PR #347, opened 2019-02-19
Fix formatting and improve docstrings for get_type_shapes.

bdice:
This PR corrects the formatting for `get_type_shapes` docstrings in `hpmc.integrate`.

The current formatting can be seen here: https://hoomd-blue.readthedocs.io/en/stable/module-hpmc-integrate.html#hoomd.hpmc.integrate.convex_polygon.get_type_shapes

### DIFF
--- a/hoomd/hpmc/integrate.py
+++ b/hoomd/hpmc/integrate.py
@@ -371,11 +371,16 @@ class mode_hpmc(_integrator):
             pos.set_def(type_list[i], shapedef + ' ' + color)
 
     def get_type_shapes(self):
-        """ Get all the types of shapes in the current simulation
+        """Get all the types of shapes in the current simulation.
 
-        Since this behaves differently for different types of shapes, the default behavior just raises an exception. Subclasses can override this to properly return
+        Since this behaves differently for different types of shapes, the
+        default behavior just raises an exception. Subclasses can override this
+        to properly return.
         """
-        raise NotImplementedError("You are using a shape type that is not implemented! If you want it, please modify the hoomd.hpmc.integrate.mode_hpmc.get_type_shapes function")
+        raise NotImplementedError(
+            "You are using a shape type that is not implemented! "
+            "If you want it, please modify the "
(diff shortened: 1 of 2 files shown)


hoomd/hpmc/integrate.py
```diff
@@ -957,9 +964,12 @@ def format_param_pos(self, param):
         return shape_def
 
     def get_type_shapes(self):
-        """ Get all the types of shapes in the current simulation
+        """Get all the types of shapes in the current simulation.
+
+        Currently assumes that all 3D shapes are convex.
```
joaander: Are these notices necessary for 2D shape classes?

hoomd/hpmc/integrate.py
```diff
@@ -851,9 +856,12 @@ def format_param_pos(self, param):
...
+
         Returns:
-            A list of dictionaries, one for each particle type in the system. Currently assumes that all 3D shapes are convex.
+            A list of dictionaries, one for each particle type in the system.
```
joaander: Would it be helpful for the method to document the contents of the dictionary?

bdice: Perhaps the right thing to do here is to document a well-defined schema for the `type_shapes` JSON format. It is currently used as a de-facto standard by a few different tools in our group.

hoomd/hpmc/integrate.py
bdice: Nope - removing now.

bdice: For now, I removed all the notices about 3D shapes and assumed convexity in HPMC. The `get_type_shapes` function is only currently implemented for convex shapes, which ensures this is always true. In `dem/pair.py`, I added the warning about assumed convexity because DEM specifies both vertex positions and face indices in its shape parameters (though I'm unsure if it actually works with non-convex shapes).

bdice:
@joaander There were some strange segfaults in the tests earlier that seem unrelated to this PR. Just wanted you to be aware in case these are sporadic failures. This PR now has proper examples and is ready for review.

Issue #351, opened 2019-02-27
Uninitialized m_hoomd_world with slurm


mphoward:
`m_hoomd_world` is used before it is initialized in `ExecutionConfiguration::guessLocalRank` when the `SLURM_LOCALID` environment variable is set. The offending lines are here:

https://github.com/glotzerlab/hoomd-blue/blob/ce50f6ab5a181a0d2221b5e3cefbc46d3c503070/hoomd/ExecutionConfiguration.cc#L752-L753

It looks like an inconsistency was introduced in this method by whoever added `hoomd_world` as an argument to the constructor, since it uses a mix of `m_hoomd_world` and `MPI_COMM_WORLD`. My guess is that lines 126-130 should probably be moved up before the GPU selection, and `MPI_COMM_WORLD` replaced by `m_hoomd_world`. @jglaser and @joaander should double check.

Reported by @AdamSimpson

bdice:
This is probably a bug introduced by my addition of SSAGES support. SSAGES manages the world communicator and necessitated this change.

bdice:
@mphoward @AdamSimpson Would one of you be able to test #352 to see if this solves the problem? I think the remaining instances of `MPI_COMM_WORLD` are correct. The case I have in mind is this: if SSAGES were to launch multiple multi-rank simulations on the same node, I think the GPU selection must use the rank in the true `MPI_COMM_WORLD`, not the MPI communicator that the individual simulation is using. In most cases, `m_hoomd_world` is the same as `MPI_COMM_WORLD`. If anyone has other thoughts, we can re-consider that behavior.

mphoward:
If that's the case, do you really want to use `m_hoomd_world` in this special check? It seems like if you are using `MPI_COMM_WORLD` to select the GPU, then you should also use `MPI_COMM_WORLD` here.

As a potential corner case, imagine that someone has 2 nodes with 2 GPUs each. They partition the `MPI_COMM_WORLD` stupidly into different HOOMD worlds so that the first world gets the first GPU on each node (`world0 = node0-gpu0 node1-gpu0`, `world1 = node0-gpu1 node1-gpu1`). The local ranks for `world0` are all 0s and the local ranks for `world1` are all 1s, but the check will raise an error in `world0` because it finds all the ranks are 0 and assumes SLURM hasn't set them. This would be avoided if the sum were taken over the full `MPI_COMM_WORLD`, since then the nonzero ranks are included.

bdice:
[...]

PR #352, opened 2019-02-27
Move communicator setup before guessLocalRank.

bdice:
This PR aims to fix the problem in #351. It seems like @mphoward's suggestion may fix the problem.


mphoward:
Great, thanks! I don't have a system with an appropriate slurm scheduler to test on, but these changes seems sensible if someone can confirm that they work.

bdice:
@adamsimpson reported that this PR resolves #351.

PR #359, opened 2019-03-04
jit module bug fixes for 2.5.1

### LINKED ISSUE #358: JIT is broken
Recent changes regarding external field support in JIT introduced linker-level dependencies between the _jit and the _hpmc targets.

This simple script fails:
```
import hoomd
from hoomd import jit
from hoomd import hpmc
```

with the error message:
```
  File "twf.py", line 2, in <module>
    from hoomd import jit
  File "/nfs/turbo/bi-vislab/jsglaser/hoomd-install-dbl/hoomd/jit/__init__.py", line 17, in <module>
    from hoomd.jit import patch
  File "/nfs/turbo/bi-vislab/jsglaser/hoomd-install-dbl/hoomd/jit/patch.py", line 5, in <module>
    from hoomd.jit import _jit
ImportError: generic_type: type "ExternalFieldJITSphere" referenced unknown base type "hpmc::ExternalFieldMono<hpmc::ShapeSphere>"
```

Obviously, it can be fixed by loading hpmc first, and then jit. But this is not really a sustainable solution (and it would be a hack if applied inside jit.py). The real culprit is the unwanted dependency between CMake subdirectories. I don't think CMake supports linking between sibling subdirectories, at least not without extra effort. I believe there should not be a cross-depency in the first place, which means that the current design of using HPMC classes inside the jit plugin is flawed and needs to be revisited.

Any takers, @vyasr, @joaander  ?

joaander:
Always import `_hmpc` in `jit` to fix  #358.

Add support for LLVM 9 using legacy interfaces.

### DIFF
--- a/hoomd/jit/KaleidoscopeJIT.h
+++ b/hoomd/jit/KaleidoscopeJIT.h
@@ -37,6 +37,21 @@
 
 #include <memory>
 
+// use legacy interfaces in LLVM 9
+#if defined LLVM_VERSION_MAJOR && LLVM_VERSION_MAJOR >= 9
+
+#define RTDYLDOBJECTLINKINGLAYER LegacyRTDyldObjectLinkingLayer
+#define IRCOMPILELAYER LegacyIRCompileLayer
+#define LOCALCXXRUNTIMEOVERRIDES LegacyLocalCXXRuntimeOverrides
+
+#else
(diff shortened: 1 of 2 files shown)


joaander:
@jglaser does this the issue for you?

Issue #365, opened 2019-03-20
Configurable streaming geometries in MPCD


mphoward:
## Description

MPCD should be made to work with streaming geometries (slit, tube, etc.) other than a bulk fluid. This is useful for simulating flows in channels or around obstacles. The general idea is to bounce particles off surfaces using specular reflection rules.

## Proposed solution

The user API is already more or less in place like `mpcd.stream.bulk`. Under the hood, a few things need to be done. The streaming method will be templated on a geometry that specifies the bounce back rules for the geometry. Each geometry will tell the integrator when a particle goes "out of bounds", puts it back on the boundary, and reflects the velocity using the local surface normal. It is too difficult to do this using the existing wall API as the user can specify pretty arbitrary setups that may be hard to detect collisions with (although eventually this could be done using BVH trees). So, each geometry will be hard-coded using the template and given a corresponding `mpcd.stream` class.

Additionally, these types of boundaries usually require so-called "virtual particles", which are particles that get inserted near the boundary right before the MPCD collision step and are removed afterwards. They are not the same as MD ghost particles, as they do not migrate between MPI ranks. Instead, they exist to give the fluid the right properties near the surfaces. Methods will be added to the MPCD particle data to add / remove these virtual particles. As a consequence, some of underlying implementation will need to change to ensure calls are made in the right order.

## Additional context

None.

## Developer

@mphoward as I already have most of this implemented and just need to clean it up. :-)

Issue #366, opened 2019-03-20
External forces in MPCD


mphoward:
## Description

MPCD doesn't use pair potentials for efficiency, but it can use external forces. MPCD should allow the user to put the MPCD particles in some sort of external force field, e.g., a constant body force to drive flow. 

## Proposed solution

The kernel that does the "streaming" (integration) of the MPCD particles is highly memory bound, and so it is not a good idea to split it into two Verlet-like kernels and load forces from memory. Rather, the external forces should just be computed on-the-fly in the kernel, since it is looking for more flops and usually external forces are cheap to evaluate anyway.

The most natural way to do this would be with some runtime solution, as I don't really want to have to compile the kernel for all possible streaming geometries (see #365) and external forces. Runtime compilation would be a possibility, but that seems like a bit of a hassle to get working. Using runtime polymorphism would be ideal, but CUDA does not allow passing objects with vtables between host and device. You can, however, have a polymorphic object in a kernel if it is constructed on the device.

I think that the most rapid development path would be to write a thin (templated) wrapper class that makes a mirror of a simple (likely const) object on both the host and the device. The device memory will be allocated on the host by the wrapper and a kernel will be called to construct the object in that memory using placement new. The wrapper can then return the appropriate pointer to host or device memory, which should be able to be passed to a kernel and dereferenced. The wrapper destructor will take care of freeing the memory. (It would be nice if there were a managed memory solution to do this more cleanly, but I haven't thought about / experimented with this.)

Once the wrapper is in place, it should be simple enough to implement an `mpcd.external` module with various forces and interface this with the MPCD integrator.

## Additional context

To be implemented after #365 is resolved.

## Developer

Assuming I can get the virtual wrapper working right, yes.

mphoward:
I've implemented a preliminary version of the polymorphic wrapper on the `mpcd-force` branch. But, I ran into a really weird issue, and so I would appreciated some feedback from @joaander and @jglaser before I proceed down that route further.

[...]

PR #369, opened 2019-03-21
Configurable streaming geometries in MPCD

mphoward:
## Description

Implements the confined streaming geometries described in #365. 

## Motivation and Context

This PR expands the functionality of the MPCD module to include confining geometries. It implements confining geometries using a templated integration method. It also adds a flexible virtual particle filling framework to help enforce the boundary condition at the surfaces. Since virtual particles are only needed for the MPCD collision, they are simply appended to the end of the MPCD ParticleData lists and have been included as needed in the collision functions (CellList, CollisionMethod, etc.).

In order to support these changes, some refactoring of the MPCD classes was needed under the hood. Basically, the virtual particles exposed a circular dependence between the CellList and the Communicator that could not be fixed with the old version of the integrator. This dependence has been broken by adding a migration signal hook to the communicator that the integrator can subscribe to. The consequence of that change, though, was that the CellList cannot be built at arbitrary times in the algorithm--the particles need to be migrated to the right place first by the integrator, as is implicitly assumed in MD.

The main culprits violating this assumption were the Sorter and the CellThermoCompute. The Sorter was being triggered early as an Updater; it is now a standalone class that is called at the proper place in the integration loop. The CellThermoCompute was only being called early when it was asked to report logged values. These log quantities were never documented for the user, and so these have simply been removed. The CellThermoCompute now allows subscribers to determine what is evaluated (with flags), as is done by ComputeThermo.

You might spot that there is a method called `set_field` included in the base streaming method. This method works and has been tested, as it was a crude pass to address #366 for a specific (constant force) external field and to test the boundary conditions for the slit. Since this code works, I have left it in but (I hope) undocumented for the time being, as the user-facing API is likely to change as the field is generalized.

This PR also resolves a miscellaneous (but related) issue by replacing `DeviceSelect::If` with `thrust::copy_if`, as I found at least one buggy case of `DeviceSelect::If` in the MPCD code when compiling with debug flags to check this PR and PR #370. There were no problems when in release mode and this may be isolated to that use case though, so I'm not recommending eliminating DeviceSelect if used elsewhere (and, thrust is likely calling it under the hood anyway.)

Resolves: #365 

## How Has This Been Tested?

There are extensive unit tests in both C++ and python for the confining geometries and virtual particle filling. We have also used the code to simulate basic flows and get the expected results. The API has been tested for a few scientific applications.

## Change log

```
* `mpcd.stream.slit` allows for simulations in parallel-plate channels. Users can implement other geometries as a plugin.
* MPCD supports virtual particle filling in bounded geometries through the `set_filler` method of `mpcd.stream` classes.
```

## Checklist:


mphoward:
@joaander One of the unit tests randomly choked on Jenkins initializing the upp11 submodule and then it looks like the other one also died prematurely. Not sure what's up with that.

mphoward:
Something is not right with the GPU sorter, possibly an issue introduced by merging up. I am looking into this.

mphoward:
> Something is not right with the GPU sorter, possibly an issue introduced by merging up. I am looking into this.

This has been resolved by 5b5921d.

mphoward:
Jenkins choked on this branch but it looks like everything passed on `mpcd-force`, so this failure can either be replayed or ignored.

PR #370, opened 2019-03-28
MPCD external force

mphoward:
## Description

Implements the external forces on MPCD particles described in #366. Currently, a constant force and a sinusoidal force are implemented, but more can be included via the polymorphic framework.

This branch is targeted onto master, but should be merged after PR #369.

## Motivation and Context

External forces are implemented within the MPCD streaming (integration) kernel using runtime polymorphism to evaluate the force in the velocity Verlet algorithm. This avoids the need to split the kernel to evaluate the forces in between or to load particle data from memory multiple times, as would be done by the MD Integrator / ForceCompute setup. We are working now to test this in an example and assess the performance, which might be useful if such polymorphism can be used elsewhere.

I implemented the polymorphism in a flexible way through the `GPUPolymorph` class, which acts as a factory / wrapper for polymorphic objects. The class is templated on the polymorphic `Base`, and then supplies factory methods to initialize ("reset") the `Derived` object on the host / device by an appropriate constructor. Because of the way CUDA works, these need to be separate objects since you can't pass vtables between the host and device address spaces. The caller must acquire the pointer in the appropriate space.

I ran into a strange issue with the polymorphic objects, where I was getting hardware stack errors (or other misc. failures) when calling a kernel that took a polymorphic object as an argument, but the polymorphic object did not have its `device_construct` kernel templated in the same file. I have a workaround in place to explicit template these, but this is not ideal as I suspect things may go wrong if people try to extend the streaming geometries or forces with a plugin. I'm currently working with a dev tech to try to figure this out with a minimum reproducer. If we can't solve it, we might need to put caveats about this in the documentation, as all testing suggests the workaround is functioning as intended.

This PR also resolves a miscellaneous (but related) issue by replacing `DeviceSelect::If` with `thrust::copy_if`, as I found at least one buggy case of `DeviceSelect::If` in the MPCD code when compiling with debug flags to check this PR. There were no problems when in release mode and this may be isolated to that use case though, so I'm not recommending eliminating DeviceSelect if used elsewhere (and, thrust is likely calling it under the hood anyway.)

Resolves: #366 

## How Has This Been Tested?

There are unit tests in place for the external forces in both C++ and python. We are starting to use this branch in a scientific application, and so this will serve as an additional check.

Additional feedback is needed from NVIDIA related to the polymorphism issue, which is suspected to be either a compiler bug or some tricky violation of the CUDA rules.

## Change log

```
* `mpcd.stream` includes an external `mpcd.force` acting on the MPCD particles. A constant force and a sine force are implemented.
```

## Checklist:


mphoward:
Separable CUDA compilation for MPCD seems to be (more or less) working other than two things:

1. One of the Jenkins build configurations (gcc54-py27-cuda8) failed with some weird linking error. I have CUDA 8 + gcc 4.8 + cmake 3.7 on my machine, and I don't get this error so it would be helpful if someone with access to the Jenkins server can investigate. Is it perhaps that this container is using a very old CMake that isn't behaving nicely?

2. I had to disable the compiled `external_field` test because of linking issues with `_mpcd` and `mpcdcuda`. I don't really understand what's wrong with that (might be related to needing to reinclude certain files in that compilation). I decided to just disable it since the code is working in the python unit tests.

In order to get things to compile and link, I needed to separate the CUDA files for MPCD into their own shared library with `CUDA_ADD_LIBRARY`, which is then linked to `_mpcd`. I couldn't get the separable compilation to work using only `CUDA_COMPILE`... it had what looked to me like linking issues with `cudadevrt` because of missing symbols, even though I was explicitly linking in that library. You also need to manually force in `fPIC` flag, even though the library is SHARED. Maybe some of this is related to 2 indirectly?

This solution is not the cleanest thing, but it does seem to work other than still needing to address 1 above. (I'm OK with punting on 2.) The ideal solution is probably to use first-class CUDA support (see #381), but this requires much more work.

@joaander @jglaser Could you help resolve the failing unit test? Any comments on the changes would also be appreciated.

mphoward:
After digging into this a little bit more, there are few things that need to be done differently (at least in connection to point 2):

1. `mpcdcuda` should be compiled as a static library (not a shared one), and there cannot be any device-side linking.
2. `_mpcd` should link to `mpcdcuda`. CMake is supposed to then trigger device-side linking of `mpcdcuda` for packaging into `_mpcd`.

By leaving `mpcdcuda` without device-side linking, it exposes its symbols to other linking libraries (e.g., the external field unit test or a plugin).

Some relevant discussion is in these issue reports:
https://gitlab.kitware.com/cmake/cmake/issues/15157
https://gitlab.kitware.com/cmake/cmake/issues/17520

This at least seems like it would work in modern CMake, but I'm not sure with FindCUDA. I will investigate.

mphoward:
> This at least seems like it would work in modern CMake, but I'm not sure with FindCUDA. I will investigate.

After some more experiments, I was unable to get the static library to work right with FindCUDA in my version of CMake (3.7.0). It seems that the static library was always getting device linked at the time it was linked into `_mpcd`, and I couldn't figure out a way to subsequently relink it into `external_field_test.cu` short of just duplicating the required `*.cu` files.

I tried some fixes for the the failing build, but I suspect it is due to an incompatibility/bug in CMake / FindCUDA / CUDA that has been resolved in subsequent versions of CMake. If this is the case (and the version of CMake in the failing container is in fact old), I'm not inclined to spend any more time trying to fix that if it can be resolved by requiring CMake 3.0 or something like that, or maybe by pasting in the current FindCUDA.cmake.

The implication of all this (as the code stands right now) is that any plugins trying to extend `ExternalField` probably will not work right, as the original `ConfinedStreamingGeometryGPU` did not get compiled with those objects. And, anyone extending `ConfinedStreamingGeometryGPU` would need to add in `ExternalField.cu` to get their new classes to work with the existing forces. This is obviously not ideal, as it is pretty limiting and counter to the philosophy elsewhere in HOOMD.

There is a chance this could be resolved using a modern CMake with first-class CUDA support if the device objects needing separable compilation were exported & installed as a static library that could then be imported into the plugin project. Then, the static library would be device-linked with the plugin CUDA files, and so all symbols should resolve properly then. But, this means bumping the minimum version of CMake to 3.8 or 3.9, and then updating HOOMD's build system.

The other option would be to drop the polymorphism in the kernel and sacrifice some performance to store/load forces in memory. This is obviously much easier to implement and could be done in a way where it only affected the performance of simulations with forces, but I'm not sure what the total performance penalty would be.

What do you all think?

mphoward:
One final set of comments, and then I think this PR is almost finalized (other than squashing some commits and tidying up docs).

I tried using CMake 3.14 with built-in CUDA support to separate the compilation. Although I could get things to compile, I ended up with segfaults using the final linked libraries / executables. I'm guessing that this is probably due to incompatibilities between the compilers / linkers & their flags that could eventually be resolved as part of a larger overhaul of the build system, but I don't have time to sort that out now. This can be revisited as part of #381.

I fixed the compilation error for the gcc54-py27-cuda8 container in commit 97afc68. The issue was that `FindCUDA.cmake` has been updated quite a bit over time (it's actually rather difficult to find a copy of it for CMake < 3.2, the oldest tag on github), and whatever CMake is used in that container does not define `CUDA_cudadevrt_LIBRARY`. This missing static library was silently ignored (and no linking took place), eventually producing the linker error shown. The solution was to find `cudadevrt` during CMake configuration if it is not already set. Also discovered in the process: separable CUDA compilation requires CMake >= 2.8.10.1, so this is now the minimum requirement (rather than just 2.8).

I thought a little more about the separable compilation and realized that it does not allow new `ExternalField` objects to be derived in a plugin library, since there is no way for the plugin to link back into the `_mpcd` library. This seems to be a fundamental limitation of the way nvcc resolves device-side symbols (and not of our design). Fortunately, the current non-virtual design of `GPUPolymorph ` actually enforces this through the explicit `reset` interface that gets exported to python, rather than making `GPUPolymorph` itself also polymorphic and letting classes derive from it (the other interface I considered). I have added some notes explaining to the user that new ExternalFields need to be implemented in core HOOMD for now.

I think it should still be possible (but haven't tried) to make a plugin that defines a new `ConfinedStreamingGeometryGPU`, but the plugin probably would need to compile with its own copy of `ExternalField.cu` to get all the necessary symbols in. This is where having a static library with unlinked device symbols would really help, as the plugin would just be able to link that in without recompiling. However, I don't anticipate this being a significant use case for the time being, and I think we can try to fix it later if it does cause problems for someone. I have documented this point.

Anyway, this is not as clean as I would like, but overall I think the use of the polymorphism is probably worth it to avoid all of the extra global loads / stores of the forces.

mphoward:
This PR should be good to go once all tests pass.

Issue #372, opened 2019-04-01
MPCD AT gives NaN particle positions


mphoward:
## Description

The MPCD example script crashes with NaN particle positions in single-precision builds on both the CPU & GPU when SRD is replaced by AT. The error is not present in double-precision builds or when the seed is changed to a different value.

I suspect the root cause is in the Box-Muller transformation used to generate normal random variates:

https://github.com/glotzerlab/hoomd-blue/blob/707e5163b0e15e5d2bae1f992f985c7d97ebef32/hoomd/mpcd/RandomNumbers.h#L93-L94

Likely, for some combination of RNG hash parameters, either `-2*log(u1)` is negative or `u1` is exactly zero, giving a NaN from that math function call. If this is true, an appropriate fix would be to (1) replace `u1` by `1.0f-u1` so that the variate is (0,1] rather than [0,1), and then (2) repeat the draw in a do loop until `-2*log(u1)` is nonnegative (see lines 50-54 for sphere point picking). The other caused could be another bug in Saru, so fingers crossed that it is the former rather than the latter, as this was very hard to track down last time.

## Script

```python
import hoomd
hoomd.context.initialize()
from hoomd import mpcd

box = hoomd.data.boxdim(L=10.)
hoomd.init.read_snapshot(hoomd.data.make_snapshot(N=0, box=box))

s = mpcd.init.make_random(N=int(5*box.get_volume()), kT=1.0, seed=42)

mpcd.integrator(dt=0.1)
mpcd.stream.bulk(period=1)
# crashes
mpcd.collide.at(seed=42, period=1, kT=1.0)
# doesn't crash
#mpcd.collide.srd(seed=42, angle=130, period=1, kT=1.0)

hoomd.run(2000)
```

## Output

```
**ERROR**: MPCD particle 3692 has position NaN
Traceback (most recent call last):
  File "test_at.py", line 17, in <module>
    hoomd.run(2000)
  File "/home/mphoward/packages/glotzerlab/hoomd-blue/v2.5.1/nompi/single/hoomd/__init__.py", line 201, in run
    context.current.system.run(int(tsteps), callback_period, callback, limit_hours, int(limit_multiple));
RuntimeError: Error computing cell list
```

## Expected output

It should have run to completion.

## Configuration

Platform:
- Linux
- GPU
- CPU

Installation method:
- Compiled from source

Versions:
- Python version: 3.4.3
- **HOOMD-blue** version: 2.5.1

## Developer

[...]

Issue #261, opened 2017-07-10
Replace Saru with random123


joaander:
**[Original report](https://bitbucket.org/glotzer/hoomd-blue/issue/261) by me.**

----------------------------------------

HOOMD's use of the Saru RNG is sporadic and inconsistent. Pull request #359 goes a long way toward making it better, but it doesn't solve the fundamental issue that Saru is not absolutely reliable and that it only has three seeds.

We should rewrite using random123 which is much more thoroughly tested and reliable. The philox 4x32 generator takes in 6 seeds. This should more than serve our needs in HOOMD without needing to resort to combining seeds (i.e. timestep+seed).

1. rng id
2. tag
3. step
4. user seed
5. rank
6. counter

tag, step, and rank may vary based on specific uses in the hoomd code (not all uses have all of these).

We should give different RNGs (i.e. langevin, HPMC grid shift, etc...) constants collected in a  header file so that it is straightforward to assign a new unique constant to a RNG.

One issue with the random123 API is that it is very low level and not trivially expandable to float and double. Since HOOMD already uses the Saru API, I propose that we re-implement  the parts of the Saru API that we use as a thin wrapper around random123. This API would maintain an internal counter to get different random numbers following Saru's stream concept - and it can generate as many as it needs. 

Since gaussian random values are commonly needed, we can also add an API for requesting these. Vectors are also a common requirement. An API to return a random vector is useful for performance as well, because random123 philox 4x32 generates 4 32 bit values with one call to the RNG.

mphoward:
I found another possible bug related to Saru. Using the three-seed hash `(309, 1313, 42)` to draw a float returns a value that is machine-identical to 1.0f:
```
hoomd::detail::Saru rng(309, 1313, 42);
rng.f() == 1.0f
```
This conflicts with the documentation that advertises a value lying in [0,1) and causes issues when the value is assumed to lie in a certain range (e.g., passing 1.0f - rng.f() to a log).

Rather than try to fix this, I am going to reject these endpoint values in my code. I hope that this revamp can maybe get some traction to fix possible issues like this, as these sorts of bugs are exceedingly difficult to track down. When I was looking at the Saru code, I was also reminded that the uniform double is drawn in a pretty lazy way (not using two true random ints), and so this is also suboptimal. Plus, it is very difficult to properly unit test the generator, and so it would be better to be using something that can really be trusted.

PR #374, opened 2019-04-01
Fix MPCD random number generators

mphoward:
## Description

MPCD uses a rejection-free Box-Muller transformation for efficiency on the GPU. But, there was no check in place to ensure that the uniform random variate passed to the logarithm was positive, and this gave inf values when the uniform variate was exactly 0. The error cropped up in a single-precision build, probably because it is easier to hit exactly 0 when you have fewer bits of mantissa. There are also underlying issues with Saru, which still does not respect the right ranges for drawing values (see comment on #261). This may be due to double-to-float precision issues.

The current solution for the MPCD code is to (1) make sure the drawn uniform variate lies in (0,1] to keep the logarithm defined and (2) keep drawing uniform values until the square-root argument is also non-negative. Most of the time, these loops will only be executed once and the overhead should be small.

The gamma distribution could in principle encounter the same issue, so I also added similar guard logic there.

## Motivation and Context

Incorrect normal variable draws in the AT collision rule manifested as NaN particle positions when the integrator operated on the velocity. This branch seems to fix the issue.

Resolves: #372 

## How Has This Been Tested?

I found two corner cases where the particle velocity had a non-finite value and then used those hashes to develop simple unit tests for the random number generators. I also reran the original simulation that crashed.

## Change log

```
* Fix a bug in the MPCD Gaussian random number generator that could lead to NaN values for the particle positions and velocities.
```

## Checklist:


mphoward:
What do you want to do with this PR given that #378 will likely also resolve this issue? This is targeted onto maint, while the other is onto master.

mphoward:
> Thanks, Mike!
> Going to merge this for 2.5.2. I will solve the conflicts with #378 when merging `maint` into `master`

OK, great! Arash and Antonia have both tested this branch, and it works for them. I think the Jenkins tests just choked on something random if you want to replay them.

joaander:
> OK, great! Arash and Antonia have both tested this branch, and it works for them. I think the Jenkins tests just choked on something random if you want to replay them.

Hopefully #378 also works :) The replay is queued in Jenkins.

PR #377, opened 2019-04-02
Multi gpu table potential

jglaser:
## Description

Enables unified memory and multi-GPU execution in TablePotentialGPU.

## Motivation and Context

I needed this to run simulations with `pair.table()` fast.

## How Has This Been Tested?

Currently only single GPU execution is unit-tested with the `test-py/test_pair_table.py` and `test/test_table_potential` unit test. The latter is only run with `-D HOOMD_SKIP_LONG_TESTS=OFF` which is not part of any build configuration. I fixed a compiler error for this configuration. Should we add this option to one of the unit tests or rather implement the same test as a python unit test?

## Change log

```
- pair.table() now takes advantage of multi-GPU execution with managed memory.
```

## Checklist:

### DIFF
--- a/hoomd/md/TablePotentialGPU.cu
+++ b/hoomd/md/TablePotentialGPU.cu
@@ -305,7 +305,7 @@ cudaError_t gpu_compute_table_forces(Scalar4* d_force,
             unsigned int run_block_size = min(block_size, max_block_size);
 
             // setup the grid to run the kernel
-            dim3 grid( N / run_block_size + 1, 1, 1);
+            dim3 grid( (range.second-range.first) / run_block_size + 1, 1, 1);
             dim3 threads(run_block_size, 1, 1);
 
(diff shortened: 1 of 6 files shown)


hoomd/md/TablePotentialGPU.cu
```diff
@@ -254,97 +260,105 @@ cudaError_t gpu_compute_table_forces(Scalar4* d_force,
...
+                {
+                cudaFuncAttributes attr;
+                cudaFuncGetAttributes(&attr, gpu_compute_table_forces_kernel<1>);
+                max_block_size = attr.maxThreadsPerBlock;
+                }
+
+            unsigned int run_block_size = min(block_size, max_block_size);
+
+            // setup the grid to run the kernel
+            dim3 grid( N / run_block_size + 1, 1, 1);
```
jglaser: instead of `N` it should be `range.second - range.first`

joaander: should we fix this before merging?

jglaser: done

PR #378, opened 2019-04-03
Replace Saru with random123

joaander:
## Description

Replace the **Saru** random number generator with `Philox4x32` from **random123**.

## Motivation and Context

**Saru** has proved to have a number of failings over the years. **random123** is better tested and should provide higher quality random numbers. `Philox4x32` also accepts 6 seeds (`Saru` took only 3). So in this PR, I take places where we mixed seeds together and provide them separately. With the additional seeds available, I give each use of the RNG in HOOMD a unique value for the first seed - these values are stored in one place (`RNGIdentifiers.h`) to avoid duplicates. This ensures that RNG uses in different portions of the code are not correlated.

To limit the amount of code changes, I call `Philox4x32` from within the existing `Saru` API. One API change I do make is to replace `gaussian_rng` with `normal` as a new member function of `Saru`. This method uses the `boxmuller` method as implemented in the **random123** examples.

Resolves: #261

## How Has This Been Tested?

All existing unit tests that extensively exercise the RNG pass (e.g. HPMC validation tests, `test_randomize_velocity`, etc...). I added new unit tests for `Saru::normal` in `mpcd/random_numbers_test.cc`.

However, not all classes changes are extensively tested (e.g. ActiveForceCompute). I ask that reviewers carefully double check places where I changed the seeds given in the `Saru` constructor to make sure we are seeding each one correctly after the change.

TODO: 
- [x] Performance test HPMC and DPD to evaluate if this costs us performance.

## Change log

```
* Use `random123` for random number generation.
```

## Checklist:


hoomd/mpcd/ATCollisionMethod.cc
```diff
@@ -103,7 +104,7 @@ void mpcd::ATCollisionMethod::drawVelocities(unsigned int timestep)
             }
 
         // draw random velocities from normal distribution
-        hoomd::detail::Saru rng(tag, timestep, m_seed);
+        hoomd::detail::Saru rng(hoomd::RNGIdentifier::ATCollisionMethod, m_seed, tag, timestep);
         mpcd::detail::NormalGenerator<Scalar,true> gen;
```
mphoward: I am drawing normal variates here and elsewhere using a wrapper I wrote for MPCD, but I can update that separately to use the new Saru interface.

joaander:
Benchmarks (from the `hoomd-benchmarks` repo).

GPU:
 - hexagon (philox): 1.22106e+08 trial moves / second
 - hexagon (Saru): 1.24958e+08 trial moves / second
 - microsphere (philox): 115.779 TPS
 - microsphere (Saru): 126.451 TPS
 
CPU (8 cores):
 - hexagon (philox): 8.60593e+06 trial moves / second
 - hexagon (Saru): 8.84309e+06 trial moves / second
 - microsphere (philox): 3.91976 TPS
 - microsphere (Saru): 3.93549 TPS

Philox performs roughtly as fast as Saru on the CPU. On the GPU, there is an unfortunate 9% performance degradation for a test DPD simulation. I think the benefits of Philox are worth keeping it. What do you think? 
 - We could implement a vector output API at the cost of making the Saru API more complex.
 - We could implement a caching mechanism at the cost of making the Saru implementation more complex (and possibly increasing register usage).

mphoward:
> * We could implement a caching mechanism at the cost of making the Saru implementation more complex (and possibly increasing register usage).

How many values does philox draw at once? 4 (so would need to cache 3 things)? Most use cases (based on the diff) involve 3 calls to the rng, so it would probably be worth it to cache the "free" values internally unless the register pressure is really bad. As a compromise, you could also cache only 1 value (pretty lightweight) to at least cut the number of draws in half.

hoomd/Saru.h
```diff
@@ -597,53 +418,41 @@ HOSTDEVICE inline double Saru::s(double a, double b)
...
     {
-    const unsigned int fullDelta=CTmultmod<delta, steps%modulus, modulus>::value;
-    /* the runtime code boils down to this single constant-filled line. */
-    return x+((x-offset>modulus-fullDelta) ? fullDelta-modulus : fullDelta);
+    r123::Philox4x32 rng;
+    r123::Philox4x32::ctr_type u = rng(m_ctr, m_key);
+    m_ctr[0] += 1;
+    float2 n = r123::boxmuller(u[0], u[1]);
+    return n.x * sigma + mu;
+    // note: If there is a need, we could add an API that returns two normally distributed numbers for no extra cost
```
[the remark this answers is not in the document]
joaander: How should that API look: `float2 Saru::normal(float2 sigma, float2 mu)`?

joaander:
I forgot to mention: I did not update HPMCMonoImplicit and related methods to avoid conflicts with https://bitbucket.org/glotzer/hoomd-blue/pull-requests/549 . We can update those later after one of these branches is merged.

jglaser:
> I forgot to mention: I did not update HPMCMonoImplicit and related methods to avoid conflicts with https://bitbucket.org/glotzer/hoomd-blue/pull-requests/549 . We can update those later after one of these branches is merged.

The GPU implicit integrator (new version -- work in progress) is somewhat heavy on random numbers, but mostly the overlap checks are the performance bottleneck. As long as the API changes are minimal you could merge this PR first, and I will adapt.

joaander:
> > * We could implement a caching mechanism at the cost of making the Saru implementation more complex (and possibly increasing register usage).
> 
> How many values does philox draw at once? 4 (so would need to cache 3 things)? Most use cases (based on the diff) involve 3 calls to the rng, so it would probably be worth it to cache the "free" values internally unless the register pressure is really bad. As a compromise, you could also cache only 1 value (pretty lightweight) to at least cut the number of draws in half.

It generates 4 32-bit values per call. 2 of those are consumed in producing a float/double output (I just pushed a commit that uses 64-bits for float to get a smaller value on the bottom of the range using u01). I will try caching the unused two and see how it goes. In the end, that is less code to change than to update everywhere else to use a vector API.

hoomd/Saru.h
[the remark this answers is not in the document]
mphoward: That would work but it is a little clunky if you want two values with the same `sigma` and `mu`, which is the common use case (will need `sigma*make_scalar2(1,1)` everywhere). This is also not ideal, but would still work with overloading:
```
void Saru::normal(float sigma, float mu, float2& vals);
void Saru::normal(double sigma, double mu, double2& vals);
```
Not sure if one is better than the other from a usage / performance perspective. Otherwise, templates would be needed.

mphoward:
Gotcha, that sounds good. It will be interesting to see if that changes the DPD performance at all.

One other (silly) question while you are updating things: do you want to keep calling this class `Saru`, or do you want to name it something more generic like `UniformGenerator` or `RandomGenerator` to avoid any future confusion about which RNG algorithm is being used?

joaander:
New benchmarks with a bit more averaging to resolve small differences:

Microsphere: 
- philox + cache: 117.806
- philox: 118.961
- Saru: 126.442

Hexagon:
- philox + cache: 1.21235e+08
- philox: 1.2128e+08
- Saru: 1.24322e+08

It turns out that DPD only needs a single random number per call, so the cache cannot help. The fact that it slows performance slightly suggests the extra registers may cause occupancy issues (I haven't investigated this with nvprof). On the bright side, with better averaging, philox is only 6% slower.

I'm going to revert the cache change for now. We have the commit in case we find cases in the future where we think it may help more than it hurts.

> One other (silly) question while you are updating things: do you want to keep calling this class `Saru`, or do you want to name it something more generic like `UniformGenerator` or `RandomGenerator` to avoid any future confusion about which RNG algorithm is being used?

Good idea. I can typedef Saru for now to provide backwards compatibility with portions of the code that haven't been updated.

mphoward:
> It turns out that DPD only needs a single random number per call, so the cache cannot help. The fact that it slows performance slightly suggests the extra registers may cause occupancy issues (I haven't investigated this with nvprof). On the bright side, with better averaging, philox is only 6% slower.
>
> I'm going to revert the cache change for now. We have the commit in case we find cases in the future where we think it may help more than it hurts.

Ah, yes right, DPD is pairwise so you only need one number there, and the cache just wastes registers. To me, 6% is worth it to have a more robust RNG.
 
> Good idea. I can typedef Saru for now to provide backwards compatibility with portions of the code that haven't been updated.

Sounds good--this typedef can be deprecated and removed later when everything is cleaned up.

Along these same lines (and just thinking out loud), do you think it's better to have the RNG be able to draw normal variates (not just uniform numbers), or could you keep that functionality separate (like I've done in `mpcd/RandomNumbers.h`)? Passing a RNG to the `NormalGenerator` (or whatever distribution you need) is consistent with the C++11 API and why I chose to do it that way in MPCD. You could actually probably just move `RandomNumbers.h` and the associated tests out of `mpcd` and into `hoomd`, and paste the guts of the Box-Muller transformation into the `NormalGenerator`.

joaander:
> Along these same lines (and just thinking out loud), do you think it's better to have the RNG be able to draw normal variates (not just uniform numbers), or could you keep that functionality separate (like I've done in `mpcd/RandomNumbers.h`)? Passing a RNG to the `NormalGenerator` (or whatever distribution you need) is consistent with the C++11 API and why I chose to do it that way in MPCD. You could actually probably just move `RandomNumbers.h` and the associated tests out of `mpcd` and into `hoomd`, and paste the guts of the Box-Muller transformation into the `NormalGenerator`.

That makes sense. There are other distributions, like `rand_select` that could be standardized to this location as well. Should we also separate out the canonical (0,1), uniform, and integer output as C++11 does?

mphoward:
If you are up for the more intensive refactoring, I think this might be a good idea. The API could closely mimic the C++11 one. Maybe something like this:

1. Current `Saru` (to be renamed `RandomEngine` or `RandomGenerator`) would act as a wrapper around philox and its state. It can supply an `operator()` that works like `Saru::u32()`. The operator could return the full philox result so that the caller can decide how many of the integers to use from the pull.

2. The canonical (0,1), uniform float, and uniform int distributions could each be implemented individually. These would be configured in their constructors and then take the `RandomEngine` as an argument to their own `operator()`. The distributions can be templated on the `Real` type for float vs. double (with template specialization) to keep the API clean and consistent with C++11. The canonical distribution would just implement `Saru::f()` & `Saru::d()`, and then the uniform float could internally make use of that. For performance, it probably makes sense to separate `Saru::f()` from `Saru::f(a,b)` so that you save a few multiplies and adds when they aren't needed.

Hopefully this would be easier to maintain / test each component moving forward.

hoomd/RandomNumbers.h
```diff
@@ -0,0 +1,513 @@
...
+class SpherePointGenerator
+    {
+    public:
+        DEVICE explicit SpherePointGenerator() {}
+
+        template<typename GeneratorType, typename Real3>
+        DEVICE inline void operator()(GeneratorType& rng, Real3& point)
+            {
+            // draw a random angle
+            const Real theta = rng.s(Real(0), Real(MPCD_2PI));
```
mphoward: `theta` should be drawn using the UniformGenerator. And then Saru can be replaced by the RandomGenerator in the MPCD code. Let me know if you want me to handle these changes though, since I am more familiar with that code.

hoomd/RandomNumbers.h
```diff
@@ -0,0 +1,513 @@
...
+        */
+        template<typename RNG>
+        DEVICE inline void operator()(Real& out1, Real& out2, RNG& rng)
+            {
+            uint64_t u0, u1;
+            generate_2u64(u0, u1, rng);
+
+            // from random123/examples/boxmuller.hpp
+            Real x, y;
+            fast::sincospi(r123::uneg11<float>(u0), x, y);
```
mphoward: Same as above.

mphoward:
Just FYI @joaander, I was actually looking at the result of one of my Jenkins builds when things glitched. It looked like the server rebooted itself and totally borked the running tests, so maybe that (+ any update it applied during the reboot) was connected to the weird behavior you started seeing.

joaander:
> Just FYI @joaander, I was actually looking at the result of one of my Jenkins builds when things glitched. It looked like the server rebooted itself and totally borked the running tests, so maybe that (+ any update it applied during the reboot) was connected to the weird behavior you started seeing.

I've seen the behavior before where some parallel Jenkins steps appear to run with parameters from others. Hopefully with this change things will be more stable.

joaander:
This is done and ready for a final review.

hoomd/RandomNumbers.h
```diff
@@ -0,0 +1,609 @@
...
+    };
+
+//! Generate random points on the surface of a sphere
+template<typename Real>
+class SpherePointGenerator
+    {
+    public:
+        DEVICE explicit SpherePointGenerator() {}
+
+        template<typename GeneratorType, typename Real3>
```
mphoward: Minor: template this on RNG for consistency?

hoomd/RandomNumbers.h
```diff
@@ -0,0 +1,609 @@
...
+            do
+                {
+                u = UniformDistribution<Real>(Real(-1.0), Real(1.0))(rng);
+                one_minus_u2 = 1.0f-u*u;
+                }
+            while (one_minus_u2 < Real(0.0));
+
+            // project onto the sphere surface
+            const Real sqrtu = fast::sqrt(one_minus_u2);
+            point.x = sqrtu * fast::cos(theta);
```
mphoward: Change this to use the new `fast::sincos` now that it exists.
```
fast::sincos(theta, point.y, point.x);
point.x *= sqrtu;
point.y *= sqrtu;
```

hoomd/RandomNumbers.h
```diff
@@ -0,0 +1,609 @@
...
+         * \returns A gamma distributed random variate
+         *
+         * The implementation of this method is inspired by that of the GSL,
+         * and also as discussed online:
+         *
+         *      http://www.hongliangjie.com/2012/12/19/how-to-generate-gamma-random-variables/
+         *
+         * The squeeze test is performed to bypass some transcendental calls.
+         */
+        template<typename GeneratorType>
```
mphoward: Minor: Use `RNG` for consistency?

hoomd/hpmc/Moves.h
```diff
@@ -180,14 +149,14 @@ template<class RNG>
 inline vec3<Scalar> generatePositionInSphere(RNG& rng, vec3<Scalar> pos_sphere, Scalar R)
     {
     // draw a random vector in the excluded volume sphere of the colloid
```
mphoward: Not positive, but this file looks like it has some bits doing something similar to `SpherePointGenerator` if you can simplify the code that way.

hoomd/md/test/test_MolecularForceCompute.cc
```diff
@@ -157,7 +157,7 @@ void comparison_test(std::shared_ptr<ExecutionConfiguration> exec_conf_cpu, std:
         for (unsigned int j = 0; j < nptl; ++j)
             {
             // choose a molecule tag 0<=mol_tag< nptl
-            unsigned int t = saru.f()*(nptl+1);
+            unsigned int t = hoomd::detail::UniformIntDistribution(nptl+1)(rng);
```
mphoward: Based on the bounds in the comment above, should this be `nptl-1`? But, I also don't understand how the original code matches the comment either.

jglaser: yeah, I believe the comment is wrong, though `nptl+1` seems correct (see line 161).

hoomd/hpmc/Moves.h
jglaser: indeed. And `SpherePointGenerator` seems more robust by one `epsilon` :)

hoomd/hpmc/UpdaterBoxMC.cc
```diff
@@ -450,7 +451,7 @@ void UpdaterBoxMC::update(unsigned int timestep)
         m_exec_conf->msg->notice(8) << "Box shear move performed at step " << timestep << std::endl;
         update_shear(timestep, rng);
         }
-    else if (move_type_select < m_Volume_weight + m_lnVolume_weight + m_Length_weight + m_Shear_weight + m_Aspect_weight)
+    else if (move_type_select <= m_Volume_weight + m_lnVolume_weight + m_Length_weight + m_Shear_weight + m_Aspect_weight)
```
jglaser: is there any consequence (i.e. bug resolved) due to this change?

hoomd/RandomNumbers.h
```diff
@@ -0,0 +1,609 @@
...
+#define DEVICE __device__
+#else
+#define DEVICE
+#endif // NVCC
+
+#define MPCD_2PI 6.283185307179586
+
+namespace hoomd
+{
+namespace detail
```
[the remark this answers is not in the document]
joaander: good suggestion. `detail` should be reserved for classes/methods that implementation details for specific methods.

hoomd/md/test/test_MolecularForceCompute.cc
joaander: Based on the code, it seems that this should be correct.
```
            // choose a molecule tag 0 <= mol_tag <= nptl
            unsigned int t = hoomd::detail::UniformIntDistribution(nptl)(rng);
```
It would be an error to assign a molecule tag of `nptl + 1` and `UniformIntDistribution` returns values inclusive of the maximum value given.

jglaser: > Based on the code, it seems that this should be correct.
> 
> ```
>             // choose a molecule tag 0 <= mol_tag <= nptl
>             unsigned int t = hoomd::detail::UniformIntDistribution(nptl)(rng);
> ```
> It would be an error to assign a molecule tag of `nptl + 1` and `UniformIntDistribution` returns values inclusive of the maximum value given.

I overlooked it was inclusive. In this case, yes, I agree.

hoomd/hpmc/UpdaterBoxMC.cc
joaander: I don't think so. Assuming for the moment that `Saru` was capable of producing float values of exactly 1.0 (which is not clear given the bugs @mphoward has found), then there was a very small chance that `move_type_select` was exactly equal to the sum of all the weights and the warning would be produced and the box move skipped.

The new random generator does produce 1.0 outputs, so I changed this to `<=`. The code would probably be cleaner and more robust using integers to select move categories, but I'm not going to take the time to rewrite it now.

joaander:
Thanks, @mphoward! I have implemented all of your excellent suggestions.

@jglaser Any idea why the validation test `lj_spheres.py-cpu-2` consistently fails on this branch? -0, -1, and -3 all pass. -2 runs with `use_clusters=0`, and `use_depletants=1`.

```
notice(2): -- HPMC stats:
notice(2): Average translate acceptance: 0.401267
notice(2): Trial moves per second:        39768.6
notice(2): Overlap checks per second:     3.89822e+07
notice(2): Overlap checks per trial move: 980.226
notice(2): Number of overlap errors:      0
notice(2): -- Implicit depletants stats:
notice(2): Depletant insertions per second:          0

** run complete **
rho_star = 0.600, U = -4.19436+-0.00183

FAIL
======================================================================
FAIL: test_statepoint (__main__.nvt_lj_sphere_energy)
----------------------------------------------------------------------

  File "/scratch/jenkins/cpu/workspace/_hoomd-blue_add-random123-master/code/hoomd/hpmc/validation/lj_spheres.py", line 131, in test_statepoint

    self.assertLessEqual(math.fabs(energy_avg-U_ref),ci*(math.fabs(U_ref)*U_ref_star_rel_err+energy_err))

AssertionError: 0.028640273437498642 not less than or equal to 0.028410479486215601
```

PR #379, opened 2019-04-03
fix bounding volume calculation for faceted ellipsoid

jglaser:
## Description

This PR fixes the calculation of the bounding volumes for `ShapeFacetedEllipsoid`. The asymmetry of the shape with facets wasn't correctly taken into account.

## Motivation and Context

I noticed that unions of faceted ellipsoids, which use OBBs internally, were overlapping, but no overlaps were reported. The tricky thing was they did this only when the leaf node capacity was small, so N^2 overlap checks were not active (the union was small enough that this was not visible with the default capacity setting). Also depletants in the `quermass_ellipsoid` branch used OBBs, so those were wrong, too.

## How Has This Been Tested?

Currently only offline.

## Change log

Since the changes have only made it into master thus far, no change log required.
## Checklist:

### DIFF
--- a/hoomd/hpmc/ShapeFacetedEllipsoid.h
+++ b/hoomd/hpmc/ShapeFacetedEllipsoid.h
@@ -284,14 +284,16 @@ struct ShapeFacetedEllipsoid
         vec3<OverlapReal> e_y(0,1,0);
         vec3<OverlapReal> e_z(0,0,1);
         quat<OverlapReal> q(orientation);
-        vec3<OverlapReal> s_x = rotate(q, sfunc(rotate(conj(q),e_x))+vec3<OverlapReal>(params.origin));
-        vec3<OverlapReal> s_y = rotate(q, sfunc(rotate(conj(q),e_y))+vec3<OverlapReal>(params.origin));
-        vec3<OverlapReal> s_z = rotate(q, sfunc(rotate(conj(q),e_z))+vec3<OverlapReal>(params.origin));
+        vec3<OverlapReal> s_x_plus = rotate(q, sfunc(rotate(conj(q),e_x))+vec3<OverlapReal>(params.origin));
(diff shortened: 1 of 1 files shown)


hoomd/hpmc/ShapeFacetedEllipsoid.h
```diff
@@ -284,14 +285,16 @@ struct ShapeFacetedEllipsoid
...
         quat<OverlapReal> q(orientation);
-        vec3<OverlapReal> s_x = rotate(q, sfunc(rotate(conj(q),e_x))+vec3<OverlapReal>(params.origin));
-        vec3<OverlapReal> s_y = rotate(q, sfunc(rotate(conj(q),e_y))+vec3<OverlapReal>(params.origin));
-        vec3<OverlapReal> s_z = rotate(q, sfunc(rotate(conj(q),e_z))+vec3<OverlapReal>(params.origin));
+        vec3<OverlapReal> s_x_plus = rotate(q, sfunc(rotate(conj(q),e_x))+vec3<OverlapReal>(params.origin));
+        vec3<OverlapReal> s_y_plus = rotate(q, sfunc(rotate(conj(q),e_y))+vec3<OverlapReal>(params.origin));
+        vec3<OverlapReal> s_z_plus = rotate(q, sfunc(rotate(conj(q),e_z))+vec3<OverlapReal>(params.origin));
+        vec3<OverlapReal> s_x_minus = rotate(q, sfunc(rotate(conj(q),-e_x))+vec3<OverlapReal>(params.origin));
+        vec3<OverlapReal> s_y_minus = rotate(q, sfunc(rotate(conj(q),-e_y))+vec3<OverlapReal>(params.origin));
+        vec3<OverlapReal> s_z_minus = rotate(q, sfunc(rotate(conj(q),-e_z))+vec3<OverlapReal>(params.origin));
```
vyasr: Can ellipsoids be asymmetric? I don't understand the reason for this change.

vyasr: Never mind, the issue arises for ellipsoid unions. That makes sense, I retract the comment.

Issue #380, opened 2019-04-04
log quantities for rigid bodies are counted twice at every `hoomd.run()`


jglaser:
## Description

I have a simulation where the potential energy output and pressure are twice their average values upon continuing the simulation with another `run()` command.

## Script

Will put script when I have a minimal reproducer, or rather a fix when I get to it. I am using `constrain.rigid()`

## Output

from the log file (second column is potential energy)
```
12107000        -43055.0381     252626.3176     0.9995534741    1.007912167     3643.560359     -47360.74145    -67610.84039    -39194.97335    -7599.75516     0.09583487076   662.1429961
12108000        -42380.78337    252626.3176     0.995785596     0.9971386409    3677.838922     -46717.38475    -66048.4744     -38837.14428    -7911.053831    0.1179516599    658.7624572
12109000        -42279.00796    252626.3176     1.005412997     1.001266048     3552.856449     -46487.30489    -67082.90642    -38130.16779    -7659.28852     0.087924868     655.4404818
12110000        -84912.95179    252626.3176     0.9963222491    0.9903331221    3646.433876     -46759.24366    -67910.25487    -38415.68488    -7338.471344    0.03190151481   656.3338954
12111000        -42611.88882    252626.3176     1.00525456      1.006248694     3536.357946     -46805.07181    -67778.57348    -38142.74926    -7733.290109    0.07279189382   656.8250417
12112000        -42536.18579    252626.3176     0.993495672     0.9939772456    3506.817905     -46697.72258
```

I issued a run command at time step 12110000, which is where the potential energy spikes.

**EDIT** or rather, it looks like this is the last time step of the previous run.
```
Time 00:00:26 | Step 12101701 / 12110000 | TPS 129.736 | ETA 00:01:03
Time 00:00:36 | Step 12103118 / 12110000 | TPS 141.668 | ETA 00:00:48
Time 00:00:46 | Step 12104582 / 12110000 | TPS 146.377 | ETA 00:00:37
Time 00:00:56 | Step 12106132 / 12110000 | TPS 154.681 | ETA 00:00:25
Time 00:01:06 | Step 12107419 / 12110000 | TPS 128.675 | ETA 00:00:20
Time 00:01:16 | Step 12109008 / 12110000 | TPS 158.854 | ETA 00:00:06
Time 00:01:21 | Step 12110000 / 12110000 | TPS 175.439 | ETA 00:00:00
Average TPS: 132.062
[...]

PR #390, opened 2019-04-18
Fix double summation

jglaser:
## Description

Recompute all `Compute` values upon every run() command, so that they have current values when `prepRun()` is called in `System.cc`.

## Motivation and Context

With the current behavior, the following would happen with rigid bodies. When a run ends and another one is started, at the next `run()` command the log values of the potential energy and the computed pressure value would double, only at the initial time step. This introduces artifacts in the dynamics of constant pressure simulations with rigid bodies.

The underlying reason was that `ForceComposite` zeroes the net force and virial of constituent particles upon summing their forces and virials, but (of course) not the arrays of the individual `ForceCompute`s that contribute to the net force. `ForceComposite` caches the central particle forces and virials (it doesn't  recompute and thus zero the constituent particle net forces if the time step doesn't change), and when the net force gets summed upon `prepRun()`, the constraint force AND the constituent particle forces are added together. That way the central particle force and virial is counted twice.

A simple solution is to force reevaluation of **all** `Compute`s after a `run()` command, including the `ForceComposite` one.

Resolves: #380 

Note: I initially thought I wanted to avoid conflicts with changes in the `multi_gpu_npt` branch (#389), which is why I based the changes off of that branch. I can rebase if necessary, or we can simple merge the other branch first.

## How Has This Been Tested?

Added a unit test.

## Change log

```
Fix a bug where rigid body energy and pressure deviated on the first time step after `run()`.
Fix a bug which could lead to invalid MPI simulations with nlist.cell() and nlist.stencil()
```

## Checklist:


jglaser:
I also fixed a bug in the calculation in the ghost layer width of cell lists for MPI simulations, which looks like a serious oversight.

@joaander Please see 8693cd943fbd443472e0cd8483822f11622c2028

mphoward:
The change you've made in 8693cd9 looks correct to me--I'm actually surprised nobody ever noticed this before.

What are the implications of the original code? You would have fewer cells covering the box than should be there based on `rlist` and the nominal cell width, but the rest of the cell list code also just takes this number of cells and runs with it (true cell width is computed separately). This means each cell is effectively larger than its nominal size, so you probably search a larger volume than required--this can happen anyway if the nominal width doesn't match the box. I don't think you would miss any ghost particles as these are stuffed into cells by the fractional coordinates & the number of cells rather than the cell size. So, if I haven't missed anything, this commit seems to fix a suboptimal cell list construction but there are probably not numerically incorrect results.

jglaser:
> The change you've made in [8693cd9](https://github.com/glotzerlab/hoomd-blue/commit/8693cd943fbd443472e0cd8483822f11622c2028) looks correct to me--I'm actually surprised nobody ever noticed this before.
> 
> What are the implications of the original code? You would have fewer cells covering the box than should be there based on `rlist` and the nominal cell width, but the rest of the cell list code also just takes this number of cells and runs with it (true cell width is computed separately). This means each cell is effectively larger than its nominal size, so you probably search a larger volume than required--this can happen anyway if the nominal width doesn't match the box. I don't think you would miss any ghost particles as these are stuffed into cells by the fractional coordinates & the number of cells rather than the cell size. So, if I haven't missed anything, this commit seems to fix a suboptimal cell list construction but there are probably not numerically incorrect results.

Yes, that makes sense. However I noticed the `nlist.stencil()` unit test was reporting illegal memory accesses. The stencil assumes that there is at least one layer of ghost cells on every side.

joaander:
Running `forceCompute` on all registered computes seems heavy handed. Could we introduce an alternate compute method or use a signal to only recompute only those that need to be?

jglaser:
> Running `forceCompute` on all registered computes seems heavy handed. Could we introduce an alternate compute method or use a signal to only recompute only those that need to be?

Note that `forceCompute` typically only sets a flag to lazily force recomputation when the next `compute()` occurs. But agreed, we could introduce an additional method for classes that modify the netforce.

*Or*: we could wrap `Integrator::computeNetForce[GPU]()` inside a `compute`-like API, so it, too, does not get called twice when the forces are not updated. This might be solution for optimal performance with frequent `run()` commands.

jglaser:
> > Running `forceCompute` on all registered computes seems heavy handed. Could we introduce an alternate compute method or use a signal to only recompute only those that need to be?
> 
> Note that `forceCompute` typically only sets a flag to lazily force recomputation when the next `compute()` occurs. But agreed, we could introduce an additional method for classes that modify the netforce.
> 
> _Or_: we could wrap `Integrator::computeNetForce[GPU]()` inside a `compute`-like API, so it, too, does not get called twice when the forces are not updated. This might be solution for optimal performance with frequent `run()` commands.

as we discussed, the best solution to this problem may be ignoring constituent particles in ComputeThermo

PR #392, opened 2019-05-02
fix handling of non-MPI_COMM_WORLD communicator in Messenger

jglaser:
## Description

Fix the usage of `MPI_COMM_WORLD` in Messenger.cc by passing it the `hoomd_world` communicator.

**EDIT** I refactored the MPI configuration as we discussed. There is now an `MPIConfiguration` class that handles the communicator handle and splits it as necessary. The three important data structures are created inside `context.py` in this order:

1. MPIConfiguration (MPI related tasks)
2. Messenger (Messaging I/O)
3. ExecutionConfiguration (per-rank parallel execution configuration)

Some MPI related routines that pertain specifically to GPUs (`guessLocalRank()`) were left in `ExecutionConfiguration`

## Motivation and Context

In `mpi4py` jobs with custom communicator arguments to `context.initialize()` we need to honor these options, otherwise the MPI RMA calls in `Messenger` will cause hangs.

## How Has This Been Tested?

Standard unit tests. I confirmed that `--msg-file` and `--shared-msg-file` work as advertised. Setting up unit tests for command line options would be good, but I'd like to defer that to the future. Also, possibly a mpi4py unit test would be good.

## Change log
```
- Fix a problem where HOOMD could not be imported in `mpi4py` jobs
- Refactor handling of MPI_Comm inside library (for developers)
```

## Checklist:

### DIFF
--- a/hoomd/context.py
+++ b/hoomd/context.py
@@ -23,16 +23,7 @@ TIME_START = time.time()
 CLOCK_START = time.clock()
 
 ## Global Messenger
-# \note This is initialized to a default messenger on load so that python code may have a unified path for sending
-# messages
-msg = _hoomd.Messenger();
-
-# only use python stdout/stderr in non-mpi runs
-if not (  'OMPI_COMM_WORLD_RANK' in os.environ
-        or 'MV2_COMM_WORLD_LOCAL_RANK' in os.environ
-        or 'PMI_RANK' in os.environ
-        or 'ALPS_APP_PE' in os.environ):
-    msg.openPython();
+msg = None;
 
 ## Global bibliography
 bib = None;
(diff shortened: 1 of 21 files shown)


hoomd/context.py
```diff
@@ -303,21 +294,30 @@ def _create_exec_conf(mpi_comm):
...
         if not handled:
             msg.error("unknown mpi_comm object: {}.\n".format(mpi_comm));
             raise RuntimeError("Invalid mpi_comm object");
 
+    # only use python stdout/stderr in non-mpi runs
+    if not (  'OMPI_COMM_WORLD_RANK' in os.environ
+            or 'MV2_COMM_WORLD_LOCAL_RANK' in os.environ
+            or 'PMI_RANK' in os.environ
+            or 'ALPS_APP_PE' in os.environ):
+        msg.openPython();
```
joaander: `openPython` should be called right after the Messenger is constructed

jglaser:
> Thanks!

We also need to support the `--shared-msg-file`,  `--msg-file` options, which this patch breaks. Working on it.

jglaser:
> > Thanks!
> 
> We also need to support the `--shared-msg-file`, `--msg-file` options, which this patch breaks. Working on it.

Should be resolved by the refactoring.

jglaser:
the `spheres_implicit` validation test failure on `gcc8` with 3 TBB threads looks unrelated and could be related to either `random123` thread safety (or lack thereof) or a problem with lambda captures in `gcc 8` (just guessing). Would be interesting to see if it is present in `next`, too.

joaander:
random123 is thread safe: the output bits are only a function of input parameters to the function. That failure did not repeat after multiple reruns of all tests.

Issue #398, opened 2019-05-09
Property API


joaander:
## Description

Parameters to HOOMD script command objects should be accessible by property.

## Proposed solution

Properties are more pythonic than setters. They also provide a more natural way for users to read current values. Examples:
```
lj.mode = 'shift'
integrator.dt = 0.001
nvt.kT = 0.5
nvt.tau = 1.0
mc.d = 0.2
```

## Additional context

Some properties in HOOMD are double buffered: stored in both python and C++. The python value is authoritative and copied to C++ at the start of a run. We should clearly document which properties are buffered. When refactoring, we should remove buffering and access the C++ values directly where possible.

## Questions

How should we map potential parameters to properties? They currently allow multiple named values to be set in a single call:
```
lj.pair_coeff.set('A', 'B', epsilon=2.0, sigma=1.0, alpha=0.5, r_cut=3.0, r_on=2.0);
```
They also allow the same parameters to be set for multiple type pairs:
```
lj.pair_coeff.set(['A', 'B'], ['C', 'D'], epsilon=1.5, sigma=2.0)
```

mphoward:
I am working on another project that needs to handle pair parameters, and I settled on making an object that thinly wraps a dict of dicts. The first dict is keyed on all the `(i,j)` pairs for types in the system, while the other dicts are keyed on some preset list of parameters (although it is not necessary to initialize either of them first, you could do this lazily and defer until later).

I use `__setitem__` to process the key, and then loop through all of the parameters passed in as a dict:
```
coeff['A','A'] = {'epsilon': 1.0, 'sigma': 1.0}
```
This makes it safe to call `__setitem__` multiple times on the same pair.

To support setting multiple types simultaneously, you could detect if either `i` or `j` in the key passed to `__setitem__` is iterable.

To access the values, you could use `__getitem__`. To be safe for this call, the `(i,j)` pairs need to be stored upper triangular internally so that someone doesn't do `coeff['A','B']['epsilon'] = 2.0`, which would fail to set `coeff['B','A']['epsilon']` if `(i,j)` and `(j,i)` were stored separately.

So, the API would look something like:
```
lj.coeff['A','A'] = {'epsilon': 1.0, 'sigma': 1.0, 'rcut': 3.0}
lj.coeff[('A',B'),'B'] = {'epsilon': 2.0, 'sigma': 1.0, 'rcut': 2.**(1./6.)}
```

jglaser:
[...]

PR #403, opened 2019-05-11
Unify HPMC integrators

### LINKED ISSUE #382: Update HPMCMonoImplicit to new RNG API
TODO: After branches are merged, we need to update the HPMC Implicit code to use the new RNG API.

jglaser:
## Description

We unify the depletants and base class integrators into a single HPMC integrator class, avoiding a lot of code duplication. The only difference between the two integration loops is that `checkDepletantOverlaps()` is called at acceptance. When the depletant fugacities are zero, it exits early.

## Motivation and Context

This simplifies maintenance and will allow putting in place the new HPMC GPU integrators later.
All 2D shapes are currently enabled, however correct 2d implicit depletant integration will have to be supported for those (separate future PR).

Needs to be merged after #402 

Resolves #382.

## How Has This Been Tested?

Covered by existing unit tests.

## Change log

- The option `implicit=True` to `hpmc.integrate.*` is no longer available. Depletants are now supported by any HPMC integrator through `mc.set_fugacity('my_type', fugacity)`
- Type swap moves in `hpmc.update.muvt()` are no longer supported (`transfer_ratio` option to `muvt.set_params()`)

## Checklist:


hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -140,6 +145,65 @@ class IntegratorHPMCMono : public IntegratorHPMC
...
+            {
+            return m_quermass;
+            }
+
+        //! Set up the additional sweep radius around every shape
+        void setSweepRadius(Scalar sweep_radius)
+            {
+            // check if supported
+            if (sweep_radius != 0.0 && !Shape::supportsSweepRadius())
+                throw std::runtime_error("This shape doesn's support setting a sweep radius to extend the surface out.\n");
```
vyasr: typo "doesn't"

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -1804,6 +2087,907 @@ bool IntegratorHPMCMono<Shape>::py_test_overlap(unsigned int type_i, unsigned in
...
+                // intersectionAABB volume
+                Scalar V =  (intersect_upper.x-intersect_lower.x)*(intersect_upper.y-intersect_lower.y)*(intersect_upper.z-intersect_lower.z);
+
+                // chooose the number of depletants in the intersection volume
+                hoomd::PoissonDistribution<Scalar> poisson(m_fugacity[type]*V);
+                #ifdef ENABLE_TBB
+                hoomd::RandomGenerator& my_rng = rng_depletants_parallel.local();
+                #else
+                hoomd::RandomGenerator& my_rng = rng_depletants;
+                #endif
```
vyasr: Can we simplify this to
```
#ifndef ENABLE_TBB
hoomd::RandomGenerator& my_rng = rng_depletants;
#endif
```

and then remove the `#else` component of the `ifdef` in the loop below? Basically, initialize the rng outside the loop if we're not using TBB, otherwise initialize it inside?

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -1804,6 +2087,907 @@ bool IntegratorHPMCMono<Shape>::py_test_overlap(unsigned int type_i, unsigned in
...
+                // intersection AABB volume
+                Scalar V = (intersect_upper.x-intersect_lower.x)*(intersect_upper.y-intersect_lower.y)*(intersect_upper.z-intersect_lower.z);
+
+                // chooose the number of depletants in the intersection volume
+                hoomd::PoissonDistribution<Scalar> poisson(-m_fugacity[type]*V);
+                #ifdef ENABLE_TBB
+                hoomd::RandomGenerator& my_rng = rng_depletants_parallel.local();
+                #else
+                hoomd::RandomGenerator& my_rng = rng_depletants;
+                #endif
```
vyasr: Same as above, can we simplify the logic in these conditional compilation steps.

hoomd/hpmc/UpdaterMuVT.h
```diff
@@ -223,6 +210,84 @@ class UpdaterMuVT : public Updater
...
+        /*
+         *! Depletant related methods
+         */
+
+        /*! Try inserting depletants into space created by changing a particle type
+         * \param timestep  time step
+         * \param n_insert Number of depletants to insert
+         * \param delta Sphere diameter
+         * \param tag Particle that is replaced
+         * \param new_type New type of particle (ignored, if ignore==True)
```
vyasr: This comment is confusing, what is the ignore parameter?

hoomd/hpmc/UpdaterMuVT.h
```diff
@@ -223,6 +210,84 @@ class UpdaterMuVT : public Updater
...
+         * \param n_trial Number of insertion trials per depletant
+         * \param lnboltzmann Log of Boltzmann factor for insertion (return value)
+         * \param type_d Depletant type
+         * \returns True if Boltzmann factor is non-zero
+         */
+        bool moveDepletantsInUpdatedRegion(unsigned int timestep, unsigned int n_insert, Scalar delta,
+            unsigned int tag, unsigned int new_type, unsigned int n_trial, Scalar &lnboltzmann,
+            unsigned int type_d);
+
+        /*! Insert depletants into such that they overlap with a particle of given tag
```
vyasr: Remove "into"

hoomd/hpmc/UpdaterMuVT.h
```diff
@@ -482,6 +556,270 @@ bool UpdaterMuVT<Shape>::boxResizeAndScale(unsigned int timestep, const BoxDim o
...
+
+                    if (overlap)
+                        break;
+
+                    } // end loop over images
+
+                if (!overlap)
+                    {
+                    // depletant in free volume
+                    extra_ndof++;
```
vyasr: Why does a depletant only add 1 DOF, not 3? What happens in quermass mode when the depletants are anisotropic and have rotational degrees of freedom? I don't think I'm quite understanding what this number of measuring, I see that it's used in the Boltzmann factor calculation in the `update` method but I don't quite follow the logic.

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -401,8 +491,59 @@ IntegratorHPMCMono<Shape>::IntegratorHPMCMono(std::shared_ptr<SystemDefinition>
...
+    ArrayHandle<hpmc_implicit_counters_t> h_counters(m_implicit_count, access_location::host, access_mode::read);
+    std::vector<hpmc_implicit_counters_t> result(this->m_pdata->getNTypes());
+
+    std::copy(h_counters.data, h_counters.data + this->m_pdata->getNTypes(), result.begin());
+    if (mode == 1)
+        {
+        for (unsigned int i = 0; i < this->m_pdata->getNTypes(); ++i)
+            result[i] = result[i] - m_implicit_count_run_start[i];
+        }
+    else
```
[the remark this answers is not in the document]
jglaser: yes. fixed.

hoomd/hpmc/IntegratorHPMCMono.h
jglaser: I don't see how that would simplify the code ...

hoomd/hpmc/UpdaterMuVT.h
jglaser: It's used for the box resize acceptance criterion, therefore 1 depletant == 1 particle degree of freedom. We should revisit this because it assumes grand-canonical statistics, with depletants being true (but noninteracting) particles that count towards the pressure. I believe we should introduce an option to make depletants only provide the multi body pair potential but not a pressure contribution. This would then also enable NPT depletants in a single box.

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -1804,6 +2088,907 @@ bool IntegratorHPMCMono<Shape>::py_test_overlap(unsigned int type_i, unsigned in
...
+#else
+template<class Shape>
+inline bool IntegratorHPMCMono<Shape>::checkDepletantOverlap(unsigned int i, vec3<Scalar> pos_i,
+    Shape shape_i, unsigned int typ_i, Scalar4 *h_postype, Scalar4 *h_orientation, unsigned int *h_overlaps,
+    hpmc_counters_t& counters, hpmc_implicit_counters_t *implicit_counters,
+    tbb::enumerable_thread_specific< hoomd::RandomGenerator >& rng_depletants_parallel)
+#endif
+    {
+    bool accept = true;
+
```
joaander: Should we have an early exit when all fugacities are 0?

hoomd/hpmc/validation/lj_spheres.py
```diff
@@ -53,7 +53,7 @@ def run_statepoint(self, Tstar, rho_star, mean_Uref, sigma_Uref, use_clusters, u
         N = len(system.particles);
 
         if use_depletants:
-            mc = hpmc.integrate.sphere(d=0.3,seed=321,implicit=True);
+            mc = hpmc.integrate.sphere(d=0.3,seed=321);
```
joaander: Is the `use_depletants=True` test still relevant here?

jglaser:
> Questions:
> 
> * Should we make TBB required in 3.0 to simplify the multiple code paths and tests?

On HPC systems, a hard dependency on TBB would be an extra compilation hurdle, so I would not recommend that.
> * Should we convert HPMC integrators over to properties now (e.g. `mc.fugacity['B'] = 0.3`) or save that for an independent PR?
That's a good idea. I will resolve this in an independent branch and put up an issue as a reminder.

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -1804,6 +2088,907 @@ bool IntegratorHPMCMono<Shape>::py_test_overlap(unsigned int type_i, unsigned in
...
+    detail::AABB aabb_i_local = shape_i.getAABB(vec3<Scalar>(0,0,0));
+    detail::AABB aabb_i_local_old = shape_old.getAABB(vec3<Scalar>(0,0,0));
+
+    #ifdef ENABLE_TBB
+    std::vector< tbb::enumerable_thread_specific<hpmc_implicit_counters_t> > thread_implicit_counters(this->m_pdata->getNTypes());
+    tbb::enumerable_thread_specific<hpmc_counters_t> thread_counters;
+    #endif
+
+    #ifdef ENABLE_TBB
+    try {
```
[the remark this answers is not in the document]
jglaser: `try`/`throw`/`catch` Is a supported way to break out of TBB for's.

hoomd/hpmc/IntegratorHPMCMono.h
jglaser: it sort of already is an early exit. The only way we could 'speed this up' for many particle types is to cache the decision.

jglaser: however, we might then also skip the reduction of the counters. Not sure if that will be measurably faster, though.

hoomd/hpmc/validation/lj_spheres.py
jglaser: no. removed. Thanks

hoomd/hpmc/UpdaterMuVT.h
vyasr: Got it, I'll follow up on the issue that you made.

vyasr:
I concur on the TBB front, at least unless and until we refactor more of the base code paths to make use of TBB-parallelism.

Regarding properties, I was thinking that would make a good hackathon project, assuming we will have one this summer. We should spend some time on #398 figuring out what good APIs would be, but the actual task seems appropriate for a less experienced HOOMD developer.

hoomd/hpmc/IntegratorHPMCMono.h
joaander: It will not likely improve performance, but it will make the code more readable.

hoomd/hpmc/validation/lj_spheres.py
joaander: there are still two tests that call `run_statepoint` with `use_depletants=True` in `lj_spheres.py`

jglaser: removed. thanks.

hoomd/hpmc/IntegratorHPMCMono.h
jglaser: now fixed.

PR #416, opened 2019-05-22
Feature/init metadata

vyasr:
## Description

This PR implements our proposed new metadata format that provides a systematic description of a complete simulation. It's still in draft mode, because there are a number of changes that we still need to make before it can be merged, but it's in a state where feedback would be very useful.

Outstanding to-do items (please add to this list as other thoughts come up):
- [ ] Implement tests for other modules
- [ ] Implement `__eq__` for HOOMD classes so that objects created from metadata can be directly checked against the original objects.

## Motivation and Context

This feature will enable almost complete reproduction of a simulation script using the JSON metadata output. For users who still prefer to work directly with a Python script, this will allow a complete capturing of the metadata associated with a simulation in a more complete manner than what is currently provided.

## How Has This Been Tested?

I've added a suite of tests that create an object, generate its metadata, then test that the object can be reloaded from metadata. However, I haven't added tests for every object in HOOMD yet, and I would like to add explicit object equality tests to make sure that nothing is lost in the metadata dump.

## Change log

```
Added ability to dump well-formed metadata about all objects in a simulation script into a format that can be directly reloaded and run in a separate script.
```

## Checklist:


vyasr:
@joaander can we add Mike to the collaborators for HOOMD so that he can also be a reviewer on this PR?

joaander:
How would the semantics of `__eq__` be meaningful outside the context of the unit tests?

vyasr:
I think they can be meaningful, in the sense that they would ensure that two objects are identical up to whatever quantities define their effects on a simulation. However, I don't think they would be useful outside of the unit tests, so I'm open to alternative suggestions on how best to test this. I plan to add a few tests that actually run a simulation, which will ensure that the reload works, but it will be difficult to test this more broadly without a better way of defining whether to objects are equivalent. How would you recommend doing that?

ahy3nz:
Summarizing some discussion with @mikemhenry and @vyasr , one bug seems to be [here](https://github.com/glotzerlab/hoomd-blue/blob/936a8ee91fa8be83ecc44d6d240f61baf7372352/hoomd/md/pair.py#L460-L464) where we are both looping, popping, and adding new key-value pairs to the metadata dictionary parameters
```
dict_keys([('A', 'A'), ('A', 'B'), ('A', 'C'), ('B', 'B'), ('B', 'C'), ('C', 'C')])
dict_keys([('A', 'B'), ('A', 'C'), ('B', 'B'), ('B', 'C'), ('C', 'C'), 'A,A'])
dict_keys([('A', 'C'), ('B', 'B'), ('B', 'C'), ('C', 'C'), 'A,A', 'A,B'])
dict_keys([('B', 'B'), ('B', 'C'), ('C', 'C'), 'A,A', 'A,B', 'A,C'])
dict_keys([('B', 'C'), ('C', 'C'), 'A,A', 'A,B', 'A,C', 'B,B'])
dict_keys([('C', 'C'), 'A,A', 'A,B', 'A,C', 'B,B', 'B,C'])
dict_keys([('C', 'C'), 'A,A', 'A,B', 'A,C', 'B,B', 'B,,,C'])
dict_keys([('C', 'C'), 'A,A', 'A,B', 'A,C', 'B,B', 'B,,,,,,,C'])
dict_keys([('C', 'C'), 'A,A', 'A,B', 'A,C', 'B,B', 'B,,,,,,,,,,,,,,,C'])
```
and ad infinitum until there's no more space to store the very-long string

mikemhenry:
That should be fixed in commit 0eee308 let me know if its still not working

ahy3nz:
> That should be fixed in commit [0eee308](https://github.com/glotzerlab/hoomd-blue/commit/0eee308980dc7dd78c7824ed3a948aac9fe52f99) let me know if its still not working

Looks good to me, thanks!

mikemhenry:
What's the current status of this PR? I think this is a pretty useful feature to add into hoomd :)

vyasr:
I think this is waiting for HOOMD 3.0 changes, because some of the proposals for 3.0 make these changes much more transparent (more object-oriented Python API for everything, property-based access to everything, etc).

vyasr:
Closing until the HOOMD v3 API is closer to ready and we can redesign this to work with that API.

mikemhenry:
Now that v3 API is closer to being ready, can we get this PR re-opened? Perhaps as a draft and I'm not sure if someone else wants to work on it, but it would be nice to get this effort going again.

joaander:
Much of the functionality designed in this PR (namely, the ability to recreate an operation from a given state dictionary) is already implemented as a core part of the v3 API. For example:
```
lj = hoomd.md.pair.LJ.from_state(state_dict)
```

The part we haven't done yet is a single command to initialize many operations.

PR #450, opened 2019-08-23
Access net virial

InnocentBug:
## Description

I added a function to the ForceCompute class that sums the virial of a particle group analog to calcForceGroup or calcEnergyGroup.

## Motivation and Context

I want to access the virial of certain forces separately. I got the idea of accessing the virials separately when working with external walls.

## How Has This Been Tested?

If there is only a single force acting in the system and the particles have zero velocity the virial components of this force should be equal to the "pressure_ij" component as logged via analyze.log(). A python test script has been added as well, which checks this condition.

## Change log

```
Enable access to net virial of particle groups for separate ForceComputes
```

## Checklist:

- [ ] I am a member of the [hoomd-contributors team](https://github.com/orgs/glotzerlab/teams/hoomd-contributors/members).

### DIFF
--- a/hoomd/md/force.py
+++ b/hoomd/md/force.py
@@ -186,7 +186,7 @@ class _force(hoomd.meta._metadata):
             g = group.all()
             virial = force.get_net_virial(g)
         """
-        return self.cpp_force.calcVirialGroup(group.cpp_group)
+        return np.asarray(self.cpp_force.calcVirialGroup(group.cpp_group))
 
 
 
--- a/hoomd/md/force.py
+++ b/hoomd/md/force.py
@@ -172,6 +172,24 @@ class _force(hoomd.meta._metadata):
 
         return (self.cpp_force.calcForceGroup(group.cpp_group).x, self.cpp_force.calcForceGroup(group.cpp_group).y, self.cpp_force.calcForceGroup(group.cpp_group).z)
 
+    def get_net_virial(self,group):
+        R""" Get the virial of a particle group.
+
+        Args:
+            group (:py:mod:`hoomd.group`): The particle group to query the virial for.
+
+        Returns:
+            The last computed virial for the members in the group.
+
+        Examples:
+
+            g = group.all()
+            virial = force.get_net_virial(g)
+        """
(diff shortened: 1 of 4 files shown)


InnocentBug:
For the future, do you prefer an issue before a PR?

InnocentBug:
It would be nice to make the loggable for the analyze.log()
Something like: log.query(ForceComputeName+"pressure_ij"). Not sure of how to put this in, though.

hoomd/md/force.py
```diff
@@ -172,6 +172,24 @@ def get_net_force(self,group):
...
+
+        Returns:
+            The last computed virial for the members in the group.
+
+        Examples:
+
+            g = group.all()
+            virial = force.get_net_virial(g)
+        """
+        return self.cpp_force.calcVirialGroup(group.cpp_group)
```
joaander: Convert return value a numpy array.

Generally user-facing API should not return C++ objects.

joaander:
> It would be nice to make the loggable for the analyze.log()
> Something like: log.query(ForceComputeName+"pressure_ij"). Not sure of how to put this in, though.

User scripts could use a log callback.

In HOOMD v3.0, we plan to completely refactor how `analyze.log` functions to more easily enable use-cases like this.

hoomd/md/force.py
InnocentBug: I am betting here on the auto conversion of std::vector to numpy array.
Do you want an explicit conversion?

InnocentBug:
> > It would be nice to make the loggable for the analyze.log()
> > Something like: log.query(ForceComputeName+"pressure_ij"). Not sure of how to put this in, though.
> 
> User scripts could use a log callback.
> 
> In HOOMD v3.0, we plan to completely refactor how `analyze.log` functions to more easily enable use-cases like this.

Yes, I am using a callback to log the data as a matrix quantity to hdf5. That works smooth.

I hope HOOMD v3.0 doesn't lose the HDF5 logging option. I use that quite extensively.
Maybe we can fit that into the new scheme as well?

Issue #442, opened 2019-08-14
Singularity Pull Installation Instructions Out of Date


rcplane:
## Description

The documented Singularity pull command now fails.
https://hoomd-blue.readthedocs.io/en/stable/installation.html

Is the new singularity image collection csadorf/software:latest ?
https://singularity-hub.org/collections/1027

## Script

singularity pull --name "software.simg" shub://glotzerlab/software

## Output

FATAL:   Failed to get manifest from Shub: The requested manifest was not found in singularity hub

## Expected output

Successfully pull image from Singularity Hub.

## Configuration

Platform:
Linux CentOS 7

Installation method:
- glotzerlab-software container

Versions:
- Python version: 2.7.5
- **HOOMD-blue** version: latest

## Developer

Documentation revision suggestion above.

csadorf:
@rcplane Hi! @joaander will need to advise on the necessary correction of the instructions in the documentation, however the containers hosted under my account (csadorf) are definitely out of date and I only updated those for my personal use. Just thought I would let you know right away.

joaander:
Singularity-hub now imposes severe limits on how many times an image may be pulled. We now host the containers on our website.
```
curl -o software.simg https://glotzerlab.engin.umich.edu/downloads/glotzerlab/software-nompi.simg
```
Full documentation: https://glotzerlab-software.readthedocs.io/en/latest/

HOOMD-blue v2.7 will fix this in the HOOMD documentation.

Issue #439, opened 2019-08-12
Possible documentation typos in md.pair.buckingham


sottcap:
## Description

I guess Buckingham potential has the form of A*exp(-r/rho)-C/r^6 but the document says it is A*exp(-r/rho)-C/r instead of r^6. Is this just a typo or the potential is implemented as -C/r? Either way I think it should be corrected.
## Script

```python
# Include a minimal script that reproduces the problem

```

## Configuration

Platform:
- Mac
- Linux
- GPU
- CPU

Installation method:
- Conda package
- Compiled from source
- glotzerlab-software container

Versions:
- Python version: ?.?.?
- **HOOMD-blue** version: ?.?.?

mphoward:
This is only a typo in the documentation, not an bug in the code.

https://github.com/glotzerlab/hoomd-blue/blob/1743c67f00821e0a44de31e137a128f84d114079/hoomd/md/EvaluatorPairBuckingham.h#L110-L112

Please feel free to submit a PR correcting the docs.

joaander:
Fixed in #459

PR #459, opened 2019-09-11
Fix misc documentation issues

joaander:
## Description

Fix documentation issues reported by users

## Motivation and Context

Resolves: #442, #439

## How Has This Been Tested?

Local sphinx documentation build

## Change log

```
* Clarify singularity image download documentation
* Correctly document the functional form of the Buckingham pair potential
* Correct typos in HPMC example snippets
```

## Checklist:

- [ ] (Glotzer group members): I am a member of the [hoomd-contributors team](https://github.com/orgs/glotzerlab/teams/hoomd-contributors/members).


lyrivera:
@joaander I recently noticed that the docstrings of the [hpmc.polyhedron](https://github.com/glotzerlab/hoomd-blue/blob/41d4845cb6273f2f1f56c71e587157cd8da850af/hoomd/hpmc/integrate.py#L1217) integrator says that only triangle meshes are supported but all the examples provide rectangular meshes. You may want to include that in this PR.

j-proc:
This looks good to me provided the tests all pass. I suspect if they don't it will be another issue since it would not pertain to documentation. Re: example code `param` vs `params`, I know most of this will likely go away if proposed API changes are made in HOOMD v3. However, is there a way we can pull all the examples from the documentation and check for syntax errors without any actual HOOMD runs? I'm not aware of such a trick, so asking the more experienced. If useful, I'll make a proposal issue and of course accept these changes.

lyrivera:
> This looks good to me provided the tests all pass. I suspect if they don't it will be another issue since it would not pertain to documentation. Re: example code `param` vs `params`, I know most of this will likely go away if proposed API changes are made in HOOMD v3. However, is there a way we can pull all the examples from the documentation and check for syntax errors without any actual HOOMD runs? I'm not aware of such a trick, so asking the more experienced. If useful, I'll make a proposal issue and of course accept these changes.

I've encountered this issue in other places before, so it would be really nice if it could be automated.

joaander:
> > However, is there a way we can pull all the examples from the documentation and check for syntax errors without any actual HOOMD runs? I'm not aware of such a trick, so asking the more experienced. If useful, I'll make a proposal issue and of course accept these changes.
> 
> I've encountered this issue in other places before, so it would be really nice if it could be automated.

Yes, [there is a way](https://ipython.readthedocs.io/en/stable/sphinxext.html). I use it in GSD. from my experiences there, I don't think it would not be practical for HOOMD. 1) It would require that readthedocs build HOOMD from source which would require requesting a time limit extension from readthedocs. 2) Every example snippet would need to be a fully functioning hoomd simulation. 3) GPU examples would not be possible because readthedocs runs on CPUs in the cloud.

For v3, my current plan is to remove all of these error-prone example snippets and instead include at least one full jupyter notebook sample script for each class. Class documentation can link to the notebook example. See [fresnel's documentation](https://fresnel.readthedocs.io/en/stable/module-geometry.html#fresnel.geometry.ConvexPolyhedron) for an example of this pattern.

Issue #286, opened 2017-11-29
Improve performance of tree neighbor lists


joaander:
**[Original report](https://bitbucket.org/glotzer/hoomd-blue/issue/286) by Michael Howard (Bitbucket: [mphoward](https://bitbucket.org/mphoward), GitHub: [mphoward](https://github.com/mphoward)).**

----------------------------------------

`nlist.tree` is slower than `nlist.cell` for certain simulations, especially standard things like the Lennard-Jones fluid. We are trying to optimize tree building / traversal and particle distance checks to boost the performance of `nlist.tree`.

1. Pull the LBVH tree out of `NeighborListGPUTree` to compartmentalize some of the tree operations and to make optimization easier.

2. Improve the way that particles are grouped into leaf nodes in the tree. Reduce the number of required traversals.

3. Vectorize the particle distance checks from the leaf nodes, and try to use shared memory when possible.

I have an undergrad student working on this. I am planning to polish off the code depending on how successful the above optimizations are and how far he gets. Tentatively, this could be completed by Spring 2018, and so I'm targeting for version 2.4 now.

joaander:
@mphoward What release should we target for improving the tree neighbor lists?

mphoward:
@joaander Do you have tentative release dates for 2.7 and 3.0? I am working on a project now where I would like to use them, so I think I will have time for it this month, next at latest.

Issue #162, opened 2016-07-01
Handle ghost particles more robustly in GPU LBVH


joaander:
**[Original report](https://bitbucket.org/glotzer/hoomd-blue/issue/162) by Michael Howard (Bitbucket: [mphoward](https://bitbucket.org/mphoward), GitHub: [mphoward](https://github.com/mphoward)).**

----------------------------------------

Right now ghost particles that lie outside of the non-bonded ghost layer are treated in an ad-hoc way in the LBVH GPU code that may cause difficulties or inefficiencies with composite bodies. (They are not ignored from the trees when they should be.)

One efficient way to fix this would be to filter out these particles when assigning the Morton codes, and then only operate on the particles that are actually inside the non-bonded ghost layer (rather than all possible ghosts). The changes that are needed are mostly to `gpu_nlist_morton_types_kernel`, but will probably propagate throughout:

* Count the ghosts that fall outside the non-bonded layer needed by the neighbor list per type
* Assign these particles 64-bits set to 1 in the first kernel to put them at the end of the sorted Morton code list
* Adjust the number of particles per type
* Adjust the number of threads / allocations needed elsewhere
* Only build the trees on these particles that are within the correct width

I will look into this when I get some time. @jglaser has another fix in place on his composite bodies MPI branch that should be pretty good for the time being.

joaander:
Duplicate of #286

PR #462, opened 2019-09-11
Improve LBVH performance

mphoward:
## Description

This PR interfaces hoomd with the LBVH from the [neighbor](https://github.com/mphoward/neighbor) library described in this [paper](https://doi.org/10.1016/j.commatsci.2019.04.004). The new LBVH uses a compressed internal representation and also queries with spherical bounding volumes, which should especially improve performance of the LBVH in double precision simulations (#286). One cost of using the compressed LBVH is that a few extra neighbors are identified. This means that the neighbor list is still overall correct (no neighbors are missed), and the extra particles don't seem to cause performance issues because the pair force calculations are pretty greedy for arithmetic.

To interface with the library, the particles are first sorted by type. During this stage, ghost particles that lie outside the box + ghost width (e.g., those in long bonds) are filtered out (#162). The sorting is a little expensive, but (1) I only sort the number of bits required by the number of types, and (2) I skip the sorting if there is only 1 type and there are no ghosts. An LBVH can then be built on each set of particles segmented by type.

Traversal is done between the segmented types and LBVHs using CUDA streams. The streams were advantageous compared to launching one kernel that traversed all particles against an LBVH at a time because the cutoff radius, etc. could be passed by value through constant memory, rather than needing to be stored in thread-local memory. (I'm guessing this reduces the overall register pressure.) The streams need some more benchmarking using multiple particle types to ensure performance is acceptable. In particular, the LBVH builds currently do not have stream concurrency because there is a sorting step that passes a few values through host memory. There might be a way to fix this using a different CUB API, at the expense of allocating extra temporary memory.

I've also experimented with the idea of staging 4 neighbors into thread-local memory, which can then be coalesced into one global write transaction. In my benchmarks with 128K LJ particles on a Quadro V100, this seemed to improve performance by a bit, but it could be sensitive to the number of registers in the GPU. (It is easy to revert if needed.) In keeping with this change, I have now aligned the neighbor list to multiples of 4 particles. It was previously aligned to 8 particles, but as far as I can tell, this was an arbitrary choice I made when working on the first LBVH implementation, and this shouldn't break anything.

Here is the rough performance data from a few benchmarks I ran on Quadro V100 with HOOMD v2.6-double as a baseline (median TPS from 4 runs). lj-liquid is HOOMD's benchmark. lj-0.2 and lj-0.8 are 128K LJ particles at T = 1.5 and density 0.2 or 0.8, using the same nlist parameters as lj-liquid and a langevin integrator (gamma = 0.1).

| algorithm | lj-0.2   |  lj-0.8  | lj-liquid | 
|--------------|----------|----------|-----------|
| cell-v2.6.0  |   2768 |  1312  | 3880 |
| tree-v2.6.0 |   1830  |  861   | 2560 |
|tree-917555|  2655 | 1200 | 3384 |

The new LBVH still can't beat the cell list for the higher density systems because of the aggressive optimizations in the cell list (non-determinism, etc.), but the new performance is still pretty good and is definitely an improvement over the original algorithm. If anyone has time to do more profiling, I'm happy to hear suggestions to get back the 10%. :-)

Switching to the library also helps simplify the neighbor list code dramatically and should make it more maintainable going forward.

I am opening this PR as a draft now to discuss a few points (@joaander @jglaser):

1. Currently, I pasted in the neighbor library, but this is already proving a bit annoying as I identified a few changes that I want to port back. What do you think about including neighbor as a submodule that is compiled into _hoomd, similarly to libgetar? I think it would be good if we could also find a way to compile and run the lbvh unit test that is bundled with the library, as this will give hoomd better coverage of this part of the code. If we go with a submodule, how do you want to manage it (glotzerlab fork, etc.)? I will try to tag a new version of neighbor with the changes I've made here.

2. I think it would be good to run performance benchmarks on a few other systems. The ones I had in mind were a WCA fluid (short cutoff), a DPD fluid (short cutoff, fast diffusion), and an LJ fluid that is synthetically given multiple types to check stream performance. I might need some help running these.

Things still to do:

- [ ] Figure out what to do with the neighbor library
- [ ] Resolve the stream concurrency issues for the LBVH builds?
- [ ] Document all of the code, and add citation to new paper
- [ ] Revise the neighbor list concepts page to include some of this discussion and benchmarking

## Motivation and Context

Resolves: #286 

## How Has This Been Tested?

The existing unit tests pass, but it would be helpful if we could run some more validation tests on established systems.

## Change log

```
* The performance of `nlist.tree` has been drastically improved for a variety of systems.
```

## Checklist:

- [ ] (Glotzer group members): I am a member of the [hoomd-contributors team](https://github.com/orgs/glotzerlab/teams/hoomd-contributors/members).


joaander:
Excellent, thanks! Maybe @jglaser has some ideas how to get that 10% extra performance.

1. Use mphoward/neighbor as a submodule. In CMake: `add_subdirectory(neighbor EXCLUDE_FROM_ALL)` and link to the `neighbor` target. See `PYBIND11_MASTER_PROJECT` in [`pybind/CMakeLists.txt`](https://github.com/pybind/pybind11/blob/master/CMakeLists.txt) for an example of a `CMakeLists.txt` that can be used to install a project or be used as a subproject. When used as a subproject, hoomd will take on the responsibility of installing the library to the correct location, *neighbor* will just build the library.

    The `CMakeLists.txt` in `neighbor` will need some updates on a separate branch to work with `next`.

2. Additional benchmarks sound like a great idea! Should we also test a MPI simulation with moderate-sized rigid bodies to test the performance considerations of the sorting and filtering process?

For validation testing, we could run [lj_spheres](https://github.com/glotzerlab/hoomd-blue/blob/master/hoomd/hpmc/validation/lj_spheres.py) or other systems from the [NIST reference](https://www.nist.gov/programs-projects/nist-standard-reference-simulation-website) as MD validation tests with multiple types of neighbor lists.

mphoward:
Thanks! I'm work on updating neighbor and adding it as a submodule. Good point about the CMake changes, I'll keep that in mind when we get closer to `next`.

Rigid bodies + MPI is probably a very good test case for the filtering. Can the DPD microsphere benchmark fit on 1 or 2 GPUs? This would probably work well for a system with multiple types in it.

I like the idea of using the LJ spheres as validation with the different neighbor lists.

mphoward:
Sorry, got held up on this. The main blocker right now is to test & fix CUDA stream concurrency. I started doing this, but I think something was not right with the test, as I had zero concurrency when there should have been at least some. (I think it relates to the tuners and not disabling them.)

If we want to launch everything in the default stream even when there are many types, this PR is almost good to go after I fix the neighbor submodule & compilation. If we need the streams to work, this will take a bit more effort. In that case, I think your target 3.0 date is most realistic for finishing.

mphoward:
I think I've added this in correctly as a submodule (with a temporary development branch of neighbor), although the way I've done it in the neighbor library is a little hacky. (I check if the current project is HOOMD, and then do the right things for that.) But, I think this solution should work for the time being.

After talking with Jens, I did a little profiling on the multiple-types systems. It turns out that it's pretty important to use CUDA streams to get good performance. Right now, things work pretty well, although there is not complete parallelism. I will try to make a few fixes to improve this a bit more, and then I think this will be ready to go.

mphoward:
OK, I think this is almost ready. Assuming all of the unit tests pass, I think that I'm done refactoring the neighbor library for now (tagged at version 0.2.0). I did my best to make use of the CUDA streams in the traversers to help systems with multiple particle types. A synthetic test (running the lj-liquid benchmark with increased numbers of types, equal # of each type) showed that the new BVH has performance efficiency (TPS with n types)/(TPS with 1 type) that is roughly the same rate as the old one, which had tightly optimized kernels. The efficiency was much worse when I put everything in the default stream. The cell list efficiency is nearly independent of number of types, which is expected since it does not have to traverse multiple structures.

I think the only outstanding items are whether or not we want to run any additional benchmarks or tests. I probably don't have enough time to do this right now, so I would appreciate help on that if it's needed.

This branch should be squashed to avoid introducing (and removing) the neighbor library to HOOMD. Should this be targeted onto master (for 2.9) or next (for 3.0)? I can take care of this once I know the right target branch.

Either way, for 3.0, I can rework the neighbor CMake to be compatible with the changes made in HOOMD. I deferred this for the time being.

mphoward:
- [ ] Update sphinx documentation for neighbor lists with new usage recommendations
- [ ] Consider adding new performance results
- [ ] Add citation to new paper to the citation generator / website
- [ ] Either merge now and update neighbor for 3.0 later, or rebase and update for next.

mphoward:
@joaander @jglaser I think this is ready, other than the decision about whether this should go onto master or next (and any associated squashing & rebasing). The only other thing would be if we want to do any more benchmarking or testing.

joaander:
I will include this in a 2.9 release. Jens will merge it up into *next* and then *hip*.

mphoward:
Great, I'm not sure whether the new code will be amenable to a HIP implementation (I chatted with Jens about this at AIChE), but I would certainly entertain PRs for *neighbor* that enable this. I'll also open an issue here in case Jens wants to try multi-GPU BVHs in HOOMD. :-)

PR #468, opened 2019-09-30
Bump embedded `gsd.c` to v1.9.2

joaander:
## Description

Use `gsd.c` from gsd v1.9.2

## Motivation and Context

This fixes a bug where GSD was unable to write or read data chunks larger than 2 GiB.

## How Has This Been Tested?

Extensive CI tests in the GSD package.

## Change log

No new change log entry is required.

## Checklist:

### DIFF
--- a/hoomd/GSDDumpWriter.cc
+++ b/hoomd/GSDDumpWriter.cc
@@ -497,7 +497,7 @@ void GSDDumpWriter::writeAttributes(const SnapshotParticleData<float>& snapshot,
         }
 
         {
-        std::vector<float> data(N*3);
+        std::vector<float> data(uint64_t(N)*3);
         data.reserve(1); //! make sure we allocate
         bool all_default = true;
 
@@ -543,7 +543,7 @@ void GSDDumpWriter::writeProperties(const SnapshotParticleData<float>& snapshot,
     uint64_t nframes = gsd_get_nframes(&m_handle);
 
         {
-        std::vector<float> data(N*3);
+        std::vector<float> data(uint64_t(N)*3);
         data.reserve(1); //! make sure we allocate
 
         for (unsigned int group_idx = 0; group_idx < N; group_idx++)
@@ -565,7 +565,7 @@ void GSDDumpWriter::writeProperties(const SnapshotParticleData<float>& snapshot,
         }
 
         {
-        std::vector<float> data(N*4);
+        std::vector<float> data(uint64_t(N)*4);
         data.reserve(1); //! make sure we allocate
         bool all_default = true;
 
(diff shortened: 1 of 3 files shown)


InnocentBug:
I cannot find a method to comment a file not touched by this PR.
But the hoomd/GSDReader.cc should be changed in order to read large files.
```
-- a/hoomd/GSDReader.cc
+++ b/hoomd/GSDReader.cc
@@ -271,7 +271,7 @@ void GSDReader::readHeader()
 */
 void GSDReader::readParticles()
     {
-    unsigned int N = m_snapshot->particle_data.size;
+    size_t N = m_snapshot->particle_data.size;
     m_snapshot->particle_data.type_mapping = readTypes(m_frame, "particles/types");
 
     // the snapshot already has default values, if a chunk is not found, the value
```
The reason is that later `N` is multiplied with integer literals (result is a integer, which may overflow for large `N`).
The overflowing argument is than passed to a function that expects `size_t`.

This is the only position I catched before my tests die (memory exhausted). Might be similar occurrences at other places (also in the Writer?).

joaander:
> I cannot find a method to comment a file not touched by this PR.
> But the hoomd/GSDReader.cc should be changed in order to read large files.

Thanks, I fixed this and similar issues in the writer. I don't have any way of testing that these fixes will work correctly for large N, however.

InnocentBug:
> > I cannot find a method to comment a file not touched by this PR.
> > But the hoomd/GSDReader.cc should be changed in order to read large files.
> 
> Thanks, I fixed this and similar issues in the writer. I don't have any way of testing that these fixes will work correctly for large N, however.

I got access to a 1TB memory machine. A test system with 385 mio particles seems to run fine.
At least the configuration is read in and written back to disk and nothing dies. (No actual integration is performed. No potentials but 1 step with an NVE integrator. See script at the bottom.)
The simulation with a single MPI rank and no GPU consumed about 400GB memory.
```python
#!/usr/bin/env python3
import time
import os

import numpy

import hoomd
hoomd.context.initialize()
import hoomd.md as md

def main():
    context = hoomd.context.SimulationContext()
    with context:
        # 385 mio particles, 19GB disk space
        system = hoomd.init.read_gsd("coordS.gsd",time_step=0)

        md.integrate.mode_standard(dt=1e-8)
        integrate_AB = md.integrate.nve(group=hoomd.group.all())

        hoomd.run(1)

        final = hoomd.dump.gsd("coordSF.gsd",group=hoomd.group.all(),period=None,truncate=True,overwrite=True)


if __name__ == "__main__":
    main()
```

joaander:
Thanks for testing!

PR #475, opened 2019-10-09
add missing GSD spec writer header to CMake

InnocentBug:
## Description

Compilation of MD potentials from external plugins fail because the GSD Spec Writer header is missing in the CMakeList.txt
Fails with a missing file error.

This is based on master and intended for master, eventhough it's a bugfix. I think the GSD Spec Writer feature is not yet available on maint.
With release of the new version, you warn plugin users that the potentials need to implement the spec-writer function, won't you?

## Motivation and Context

The introduction of GSD Spec Writer was missing adding the header to the CMakeList.txt.

No issue submited first

## How Has This Been Tested?

I tested with my own external plugins and they compile and run now.

## Change log

None, it's only a minor bugfix

## Checklist:

### DIFF
--- a/hoomd/CMakeLists.txt
+++ b/hoomd/CMakeLists.txt
@@ -107,6 +107,7 @@ set(_hoomd_headers
     GPUVector.h
     GSDDumpWriter.h
     GSDReader.h
+    GSDShapeSpecWriter.h
     HalfStepHook.h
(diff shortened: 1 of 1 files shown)


joaander:
Thanks for the fix. This should go on maint.

PR #489, opened 2019-10-24
Testing Azure Pipelines

joaander:
This PR exists to test if Azure Pipelines is able to properly trigger pull request builds.


joaander:
Azure Pipelines seems to be working now.

Issue #491, opened 2019-10-24
Periodic Torsions


ahy3nz:
## Description
Incorporate phase offsets for charrm/periodic dihedrals similar to [lammps](https://lammps.sandia.gov/doc/dihedral_charmm.html)

## Proposed solution
Add an additional argument to `hoomd.md.dihedral.harmonic` coefficients for phase offsets. The compute calculation will change slightly to account for the phase offset

## Additional context

See [thread](https://groups.google.com/forum/#!msg/hoomd-users/S4OtayowxIM/BU1yZVUyCAAJ)

## Developer

I will give this a shot

PR #492, opened 2019-10-24
Feature/update harmonic dihedral

ahy3nz:
## Description

Implementing [charmm (periodic) dihedrals](https://lammps.sandia.gov/doc/dihedral_charmm.html)
## Motivation and Context
For implementing some atomistic force fields, these types of dihedrals are needed

Resolves: #491 

## How Has This Been Tested?
Updating some of the existing `hoomd.md.dihedral.harmonic` tests to account for the additional phase offset coefficient

## Change log

```
* The `hoomd.md.dihedral.harmonic` now accepts phase offsets, p0, for CHARMM-style periodic dihedrals
```

## Checklist:

### DIFF
--- a/hoomd/md/HarmonicDihedralForceCompute.cc
+++ b/hoomd/md/HarmonicDihedralForceCompute.cc
@@ -90,8 +90,8 @@ void HarmonicDihedralForceCompute::setParams(unsigned int type, Scalar K, int si
         m_exec_conf->msg->warning() << "dihedral.harmonic: specified K <= 0" << endl;
     if (sign != 1 && sign != -1)
         m_exec_conf->msg->warning() << "dihedral.harmonic: a non unitary sign was specified" << endl;
-    if (phi_0 <= 0)
-        m_exec_conf->msg->warning() << "dihedral.harmonic: specified phi_0 <= 0" << endl;
+    if (phi_0 < 0 || phi_0 >= 2*M_PI)
+        m_exec_conf->msg->warning() << "dihedral.harmonic: specified phi_0 outside [0, 2pi)" << endl;
     }
 
 /*! DihedralForceCompute provides
@@ -257,8 +257,9 @@ void HarmonicDihedralForceCompute::computeForces(unsigned int timestep)
 
         Scalar sign = m_sign[dihedral_type];
         Scalar phi_0 = m_phi_0[dihedral_type];
-        p = p*sign + dfab*sin(phi_0);
-        dfab = dfab*sign - ddfab*sin(phi_0);
+        Scalar sin_phi_0 = fast::sin(phi_0)
+        p = p*sign + dfab*sin_phi_0;
+        dfab = dfab*sign - ddfab*sin_phi_0;
         dfab *= (Scalar)-multi;
         p += Scalar(1.0);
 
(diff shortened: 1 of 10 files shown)


ahy3nz:
`ctest` was failing locally, but I wanted to draft a PR to get comments

mphoward:
Thanks, this looks simpler to implement than I thought and is nice new functionality! The only comment I have (other than ensuring the implementation of the physics is correct and all tests pass) is to suggest two changes to the python API:

1. `p0` does work as a parameter name, but I might prefer a more descriptive name like `shift` or `phi` or `phi0`.
2. For backwards compatibility, `p0` should be made a default coefficient (with default value 0), and this should be documented. I think the dihedral coefficient object allows you to set a default--the pair coefficients definitely do, and you can look in `md.pair.lj` for an example.

ahy3nz:
Thanks for the comments! I've done a refactor into `phi_0` and allowed for default arguments to be assumed for `phi_0` if not provided.

I noticed there was a lot of similarity with [lammps charmm dihedral](https://github.com/lammps/lammps/blob/master/src/MOLECULE/dihedral_charmm.cpp) and [lammps harmonic dihedral](https://github.com/lammps/lammps/blob/master/src/MOLECULE/dihedral_harmonic.cpp), so it was a lot of translating to hoomd syntax

hoomd/md/HarmonicDihedralForceCompute.cc
```diff
@@ -246,11 +252,13 @@ void HarmonicDihedralForceCompute::computeForces(unsigned int timestep)
...
 /////////////////////////
 // FROM LAMMPS: sin_shift is always 0... so dropping all sin_shift terms!!!!
+// Adding charmm dihedral functionality, sin_shift not always 0 
 /////////////////////////
 
         Scalar sign = m_sign[dihedral_type];
-        p = p*sign;
-        dfab = dfab*sign;
+        Scalar phi_0 = m_phi_0[dihedral_type];
+        p = p*sign + dfab*sin(phi_0);
```
mphoward: You should use `fast::sin` here, as in the GPU code, and also store the value of the `sin` to reuse explicitly.

hoomd/md/test/test_harmonic_dihedral_force.cc
```diff
@@ -52,7 +52,7 @@ void dihedral_force_basic_tests(dihedralforce_creator tf_creator, std::shared_pt
```
mphoward: You should add a unit test either here (in the C++ code) or at the python level where you check that the forces are computed correctly if `phi_0` is not zero. Right now, I don't think there are any checks for this, so we can't be sure that the implementation is right.

ahy3nz: Okay see `dihedral_force_phase_shift`, sort of a contrived but simple example, I think

hoomd/md/HarmonicDihedralForceGPU.cu
```diff
@@ -179,8 +180,14 @@ void gpu_compute_harmonic_dihedral_forces_kernel(Scalar4* d_force,
 
 /////////////////////////
 // FROM LAMMPS: sin_shift is always 0... so dropping all sin_shift terms!!!!
+// Adding charmm dihedral functionality, sin_shift not always 0,
+// cos_shift not always 1
 /////////////////////////
+        Scalar sin_phi_0 = fast::sin(phi_0);
+        Scalar cos_phi_0 = fast::cos(phi_0);
+        p = p*cos_phi_0 + dfab*sin_phi_0
```
joaander: Azure found a compile error:
```suggestion
        p = p*cos_phi_0 + dfab*sin_phi_0;
```

Issue #373, opened 2019-04-01
MPCD validation tests & examples


mphoward:
## Description

MPCD needs a few validation tests to ensure everything is fully integrated at the python level. These will probably be medium-length (< 2 minute) tests that ensure momentum is globally conserved, as this is the easiest thing to check, and then maybe also that the total temperature is approximately right.

## Proposed solution

Good candidate cases would be:
1. Bulk SRD fluid.
2. Bulk AT fluid.
3. Bulk SRD fluid + embedded particles.
4. Bulk AT fluid + embedded particles.

These can be borrowed from the documentation. The validation tests may also be able to be pushed as additional hoomd examples for people looking to run MPCD simulations.

## Developer

I will implement these.

PR #496, opened 2019-10-28
MPCD validation tests

mphoward:
## Description

Adds validation tests for basic MPCD SRD and AT examples. These validation tests check for conservation of total momentum in (1) a pure solvent and (2) a solvent + some embedded particles having different mass from the solvent during a short (100 collision) run. In case (1), the momentum of each group should remain zero, while in case (2) the momentum of each group is typically nonzero (confirmed for these random seeds), but the total momentum should still be zero. These validations will most importantly error out if something chokes early with a NaN, etc. These can run pretty quickly on both CPU and GPU in both MPI & non-MPI, so I included all configurations.

This also fixes a bug where some embedded particles might have failed to migrate in MPI + GPU configurations, which would trigger an error in a production run. The validation tests caught this issue!

## Motivation and Context

Some issues are hard to catch without running the full simulation. Eventually these tests could be expanded to check other properties (e.g., temperature) or simulation geometries.

Resolves: #373 

Note: this is based on PR #495 and should be merged after, as these validation tests are meant to catch NaNs like those currently present on maint for AT + embedded particles.

## How Has This Been Tested?

The new tests pass.

## Change log

```
* Fix an error triggering migration of embedded particles for MPCD with MPI + GPU configurations.
```

## Checklist:


mphoward:
We should make sure we check the azure output for the new validation tests being run.

joaander:
@mphoward 
One of the tests is deadlocking:
```
$ ctest -V -R mpcd-validate_srd-mpi-gpu
...
530: validate_srd_test.py:064  |          slv = self.solvent.take_snapshot()
530: validate_srd_test.py:065  |          slt = self.solute.take_snapshot()
530: validate_srd_test.py:072  |          hoomd.run(100)
530: ** starting run **
530: ERROR
530: test_solvent (__main__.mpcd_srd_validation)
530: Test momentum conservation for SRD solvent. ... validate_srd_test.py:022  |          self.solute = hoomd.init.create_lattice(hoomd.lattice.sc(a=1.0), L)
530: validate_srd_test.py:022  |          self.solute = hoomd.init.create_lattice(hoomd.lattice.sc(a=1.0), L)
```
I do not know what is causing the `ERROR` output, but somehow the run does not finish (there is no `** run complete **` output and yet rank 0 continues on to the next test, causing a deadlock.

`mpcd-validate_srd-mpi-cpu` passes.

hoomd/mpcd/validation/validate_at_test.py
```diff
@@ -0,0 +1,88 @@
...
+    def test_solvent(self):
+        """Test momentum conservation for AT solvent."""
+
+        # initial momentum of both should be zero
+        slv = self.solvent.take_snapshot()
+        slt = self.solute.take_snapshot()
+        if hoomd.comm.get_rank() == 0:
+            slv_p0 = np.sum(slv.particles.velocity, axis=0)
+            slt_p0 = self.density*np.sum(slt.particles.velocity, axis=0)
+            np.testing.assert_array_almost_equal(slv_p0, [0,0,0])
```
joaander: The numpy documentation recommends `assert_allclose` over `assert_array_almost_equal`. 

To compare to 0, use `assert_allclose(a, b, atol=1e-6)` throughout.

hoomd/mpcd/validation/validate_at_test.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # each group should not have zero momentum, but total momentum should be zero
+        slv = self.solvent.take_snapshot()
+        slt = self.solute.take_snapshot()
+        if hoomd.comm.get_rank() == 0:
+            slv_p1 = np.sum(slv.particles.velocity, axis=0)
+            slt_p1 = self.density*np.sum(slt.particles.velocity, axis=0)
+            self.assertFalse(np.allclose(slv_p1, [0,0,0]))
+            self.assertFalse(np.allclose(slt_p1, [0,0,0]))
+            np.testing.assert_array_almost_equal(slv_p1 + slt_p1, [0,0,0], decimal=3)
```
joaander: This test would use `atol=1e-3`.

mphoward:
Thanks, I'll look into the deadlocked test (only ran without MPI locally) and make the suggested changes.

mphoward:
I fixed the deadlock, which had two causes: (1) there were a few missed `error() -> errorAllRanks()` conversions, so the code wasn't properly reporting that an MD particle was out-of-bounds in the MPCD cell list, and (2) there was a casting error going from an int to a char in the GPU cell list code, which meant that the MD particles weren't getting migrated properly. (This wouldn't normally be an issue if there were actually a pair potential with a reasonable `r_buff` that also triggered the migration.) This should be resolved now. I also updated the array testing.

PR #536, opened 2019-11-22
update the hpmc wall updater to 3.0 api

### LINKED ISSUE #513: Move HPMC wall updater to v.3.0 API
## Description
Refactor the HPMC wall updater to the HOOMD v.3.0 API. The template can be found in the `feature/new-object-API` branch's `Clusters` updater in `hoomd/hpmc/update.py`.

## Proposed solution
The solution involves following the template mentioned above.

## Additional context
Any pull request should merge into `feature/new-object-API` **not** `master`.

j-proc:
This is ready to go aside from the following todo's:
1. A parameter name issue. The variable "move_ratio" is actually a probability. Other updater use "move_probability" which seems much more clear but could be considered a breaking change. 
2. Verify that 3.0 won't invalidate the logged quantity in the documentation.
3. I think the examples might need a slight tune up but I'm not familiar with the rest of the API changes. These updated examples could then be moved to the notebook and linked.

fixes #513

### DIFF
--- a/hoomd/hpmc/update.py
+++ b/hoomd/hpmc/update.py
@@ -400,6 +400,7 @@ class boxmc(_updater):
 
 # TODO: consider updating the naming of move_ratio to move_probabilty
 # TODO: find out if logging documentation is still valid
+# TODO replace (periodic)trigger with :py:class:`hoomd.Triggers` once doc'ed
 class Wall(_updater):
     R""" Apply wall updates with a user-provided python callback.
 
@@ -409,16 +410,19 @@ class Wall(_updater):
         py_updater (`callable`): the python callback that performs the update
                                  moves. This must be a python method that is a
                                  function of the timestep of the simulation.
-               It must actually update the :py:class:`hoomd.hpmc.field.wall`)
-               managed object.
+                                 It must actually update the 
+                                 :py:class:`hoomd.hpmc.field.wall`) managed 
+                                 object.
         move_ratio (float): the probability with which an update move is
                             attempted.
         seed (int): the seed of the pseudo-random number generator that
                     determines whether or not an update move is attempted
-        trigger (int): the number of timesteps between update move attempt
-                       attempts.
+        trigger (int or `Trigger`): the trigger object which causes
+            the updater to be called. If an interger is given, default is 1, a
+            PeriodicTrigger with phase of 0 will be used.
+                    
 
-    Every *trigger* steps, a walls update move is tried with probability
+    Every *trigger* step, a walls update move is tried with probability
     *move_ratio*. This update move is provided by the *py_updater* callback.
     Then, update.wall only accepts an update move provided by the python
     callback if it maintains confinement conditions associated with all walls.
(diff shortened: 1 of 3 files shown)


hoomd/hpmc/update.py
```diff
@@ -398,114 +398,145 @@ def enable(self):
...
+                                   origin = [0, 0, 0], inside = True);
+        wall_updater = hpmc.update.wall(mc, ext_wall, perturb,
+                                        move_ratio = 0.5, seed = 27,
+                                        trigger = 50);
+        log = analyze.log(quantities=['hpmc_wall_acceptance_ratio'],
+                          trigger=100, filename='log.dat', overwrite=True);
 
     Example::
 
         mc = hpmc.integrate.sphere(seed = 415236);
```
b-butler: At some point we will have to make this have a deferred initialization for the walls. Since we want to be able to create walls without having a mc integrator instance at the time of creation. They need the same attach concept as well.

hoomd/hpmc/update.py
```diff
@@ -398,114 +398,145 @@ def enable(self):
...
+        py_updater (`callable`): the python callback that performs the update
+                                 moves. This must be a python method that is a
+                                 function of the timestep of the simulation.
+               It must actually update the :py:class:`hoomd.hpmc.field.wall`)
+               managed object.
+        move_ratio (float): the probability with which an update move is
+                            attempted.
+        seed (int): the seed of the pseudo-random number generator that
+                    determines whether or not an update move is attempted
+        trigger (int): the number of timesteps between update move attempt
```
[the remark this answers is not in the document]
j-proc: Added documentation but can't link it to the hoomd Trigger object, which would likely have this documentation you're referring to. Added a TODO noting that it should be changed after that is done.

hoomd/hpmc/update.py
j-proc: Opened new issue since I think it's outside the scope for these specific changes. https://github.com/glotzerlab/hoomd-blue/issues/551

hoomd/hpmc/update.py
```diff
@@ -444,7 +448,7 @@ def perturb(timestep):
           r = np.sqrt(ext_wall.get_sphere_wall_param(index = 0, param = "rsq"));
           ext_wall.set_sphere_wall(index = 0, radius = 1.5*r,
                                    origin = [0, 0, 0], inside = True);
-        wall_updater = hpmc.update.wall(mc, ext_wall, perturb,
+        wall_updater = hpmc.update.Wall(mc, ext_wall, perturb,
```
b-butler: These examples have the `HPMCIntegrator` object as an argument for the constructor. However, they are not required by `__init__`.

j-proc:
Those need to be rewritten after the API changes. It makes sense to defer
that until the documentation and example notebooks are ready does it not?

b-butler:
Hmm, it does make sense to wait until `Walls` are done on the python side too. I just don't want to merge this and forget that we needed to change documentation. Could this go in the issue you created @j-proc?

j-proc:
Sure thing.

joaander:
We still haven't implemented the walls infrastructure in v3 yet. Will evaluate work on the wall updater after we get that done.

PR #537, opened 2019-11-22
Hpmc integrator api

tommy-waltmann:
## Description

We have begun updating the api for the integrators in HPMC

## Motivation and Context

We are moving towards hoomd v3 and we need to update the api so it is in a more object oriented style, so we have to do this for hpmc submodule

Resolves: #518-520

## How Has This Been Tested?

We have added tests in files that are named corresponding to the integrators, in hoomd.hpmc.pytest

None of the documentaion has been updated yet

## Change log

Updated HPMC integrator api

## Checklist:


joaander:
I added a set of example tests for errors in `test_sphere.py`

hoomd/conftest.py
```diff
@@ -52,7 +52,55 @@ def make_simulation(particle_types=['A']):
         sim.create_state_from_snapshot(s)
         return sim
     return make_simulation
+    
+@pytest.fixture(scope='session')
```
joaander: See comments below regarding changes needed to these fixtures.

hoomd/conftest.py
```diff
@@ -52,7 +52,55 @@ def make_simulation(particle_types=['A']):
...
+            s.configuration.box = [20, 20, 20, 0, 0, 0] #make 20x20x20 cubic box
+            s.particles.N = N
+            s.particles.position[0] = [0, 0, 0]
+            s.particles.position[1] = [0, 0.25, 0]
+            s.particles.types = particle_types
 
+            #hoomd.context.current.system_definition = hoomd._hoomd.SystemDefinition(s, hoomd.context.current.device.cpp_exec_conf);
+
+        # initialize the system
+        #hoomd.context.current.system = hoomd._hoomd.System(hoomd.context.current.system_definition, 0);
```
joaander: Remove commented out code.

hoomd/hpmc/pytest/test_convex_polygon.py
```diff
@@ -1,25 +1,132 @@
...
+
+
+def test_overlaps(device, dummy_simulation_check_overlaps):
+    mc = hoomd.hpmc.integrate.ConvexPolygon(23456, d=0, a=0)
+    mc.shape['A'] = dict(vertices=[(0, (0.75**0.5) / 2),
+                                   (-0.5, -(0.75**0.5) / 2),
+                                   (0.5, -(0.75**0.5) / 2)])
+    sim = dummy_simulation_check_overlaps()
+    sim.operations.add(mc)
+    # gsd_dumper = hoomd.dump.GSD(filename='/Users/danevans/hoomd/
```
joaander: Remove commented out test code.

hoomd/hpmc/pytest/test_convex_polygon.py
```diff
@@ -1,25 +1,132 @@
...
+    # test_dump_convex_polygon.gsd', trigger=1, overwrite=True)
+    # gsd_logger = hoomd.logger.Logger()
+    # gsd_logger += mc
+    # gsd_dumper.log = gsd_logger
+    # sim.operations.add(gsd_dumper)
+    sim.operations.schedule()
+    sim.run(1)
```
joaander: Remove run(1) calls. There is no need to run steps before checking overlaps.

hoomd/hpmc/pytest/test_convex_polygon.py
```diff
@@ -1,25 +1,132 @@
...
+    overlaps = sim.operations.integrator.overlaps
```
joaander: This would be clearer as: `assert mc.overlaps > 0)`

hoomd/hpmc/pytest/test_convex_polygon.py
```diff
@@ -1,25 +1,132 @@
...
+    assert overlaps > 0
+
+    s = sim.state.snapshot
+    s.particles.position[0] = (0, 0, 0)
+    s.particles.position[1] = (0, 8, 0)
+    sim.state.snapshot = s
+    sim.operations.add(mc)
```
joaander: There is no need to re-add and re-schedule the simulation after changing the snapshot.

hoomd/hpmc/pytest/test_convex_polygon.py
```diff
@@ -1,25 +1,132 @@
...
+    overlaps = sim.operations.integrator.overlaps
+    assert overlaps > 0
+
+
+def test_shape_moves(device, dummy_simulation_check_moves):
+    mc = hoomd.hpmc.integrate.ConvexPolygon(23456)
+    mc.shape['A'] = dict(vertices=[(0, (0.75**0.5) / 2),
+                                   (-0.5, -(0.75**0.5) / 2),
+                                   (0.5, -(0.75**0.5) / 2)])
+    sim = dummy_simulation_check_moves()
```
joaander: I would suggest several changes to the simulation factories.
1) Remove `dummy_` from the prefix. These are not dummy simulations, but simulations for a specific purpose.
2) 2D shapes (like convex polygon) need particles placed in only in the z=0 plan and the dimensionality of the simulation should be set to 2. You will need to add a `dimensions` argument to the factory function to enable this.
3) Let's generalize the fixtures sufficiently so that they may be used by MD and other simulations. 
  * `dummy_simulation_check_moves` should be `lattice_simulation_factory(particle_types=['A'], dimensions=3, L=20, n=7, a=3)` - where L is the box length, dimension sets the number of dimensions, n sets the number of particles placed on each side of the cube, and a is the spacing between the particles.
  * `dummy_simulation_check_overlap` can actually use the new `lattice_simulation_factory` if you allow `n` to also be a tuple - you would place `n=(2,1)` particles with `a=0.25` to get the behavior you have now.

hoomd/hpmc/pytest/test_convex_polyhedron.py
```diff
@@ -1,32 +1,146 @@
...
+def test_overlaps(device, dummy_simulation_check_overlaps):
+
+    mc = hoomd.hpmc.integrate.ConvexPolyhedron(23456, d=0, a=0)
+    mc.shape['A'] = dict(vertices=[(0, (0.75**0.5) / 2, -0.5),
+                                   (-0.5, -(0.75**0.5) / 2, -0.5),
+                                   (0.5, -(0.75**0.5) / 2, -0.5),
+                                   (0, 0, 0.5)])
+    sim = dummy_simulation_check_overlaps()
+    sim.operations.add(mc)
+    # gsd_dumper = hoomd.dump.GSD(filename='/Users/danevans/hoomd/
```
joaander: The above comments to `test_convex_polygon.py` apply to all these tests, except that 3D shapes should of course initialize 3-d systems for testing.

joaander:
The sphinx docs fail to build due to changes made on the target branch.

hoomd/hpmc/pytest/test_convex_spheropolygon.py
```diff
@@ -0,0 +1,142 @@
...
+    poly.shape['G'] = dict(vertices=verts5, sweep_radius=4)
+
+    assert poly.shape['C']['vertices'] == verts1
+    assert poly.shape['D']['vertices'] == verts2
+    assert poly.shape['E']['vertices'] == verts3
+    assert poly.shape['F']['vertices'] == verts4
+    assert poly.shape['G']['vertices'] == verts5
+
+
+def test_convex_poly_after_attaching(device, dummy_simulation_factory):
```
b-butler: This name has `convex_poly` in it, but it is a `ConvexSpheropolygon`.

hoomd/hpmc/pytest/test_convex_spheropolygon.py
```diff
@@ -0,0 +1,142 @@
...
+    assert poly.shape['D']['ignore_statistics']
+    assert not poly.shape['E']['ignore_statistics']
+
+    np.testing.assert_allclose(poly.shape['A']['vertices'], verts1)
+    np.testing.assert_allclose(poly.shape['B']['vertices'], verts2)
+    np.testing.assert_allclose(poly.shape['C']['vertices'], verts3)
+    np.testing.assert_allclose(poly.shape['D']['vertices'], verts4)
+    np.testing.assert_allclose(poly.shape['E']['vertices'], verts5)
+
+    # check for errors on invalid input
```
b-butler: Error checking for invalid inputs should not be in a test for making sure attaching works.

hoomd/hpmc/pytest/test_convex_spheropolygon.py
```diff
@@ -0,0 +1,142 @@
...
+
+    s = sim.state.snapshot
+    if s.exists:
+        s.particles.position[0] = (0, 0, 0)
+        s.particles.position[1] = (0, 0.5, 0)
+    sim.state.snapshot = s
+    assert mc.overlaps > 0
+
+
+def test_shape_moves(device, lattice_simulation_factory):
```
b-butler: I think `test_moves` is fine.

hoomd/hpmc/pytest/test_faceted_ellipsoid.py
```diff
@@ -1,31 +1,314 @@
 import hoomd
 import hoomd.hpmc
 import hoomd.hpmc._hpmc as hpmc
+import pytest
+import copy
+
+args_1 = {"normals": [(0, 0, 1),
```
b-butler: The typical way in pytest to have module level variables for test like this is to use fixtures. Could we consider them here? They look more explicit in the test definitions.

hoomd/hpmc/pytest/test_faceted_ellipsoid.py
```diff
@@ -1,31 +1,314 @@
...
+    test_faceted_ellipsoid5 = hpmc.FacetedEllipsoidParams(args_5)
     test_dict5 = test_faceted_ellipsoid5.asDict()
     assert test_dict5 == args_5
 
+
+def test_shape_params():
+
+    mc = hoomd.hpmc.integrate.FacetedEllipsoid(23456)
+
+    mc.shape['A'] = dict()
```
b-butler: Notice here how you iterate through all the keys here of by just enumerating them. You could do something like this
````
for key, value in mc.shape['A']:
     if key == 'ignore_statistics':
        assert value is False
    else:
       assert value == hoomd.typeconverter.RequiredArg
````
Also, `is` should work for `hoomd.typeconverter.RequiredArg` since is is a class that should only have one instance. This kind of pattern would work for all of these test. For instance, when filling out the type's shape using `args_{1,...,6}` you could check the full setting with a similar for loop. This would cover more, and reduce code (removing the potential for bugs). This applies to all the test files.

File: hoomd/hpmc/pytest/test_faceted_ellipsoid.py [hunk 14]
```diff
@@ -1,31 +1,314 @@
...
+    assert mc.shape['A']['ignore_statistics']
+    assert not mc.shape['B']['ignore_statistics']
+    for key in args_1.keys():
+        assert mc.shape['A'][key] == args_1[key]
+        assert mc.shape['B'][key] == args_2[key]
+        assert mc.shape['C'][key] == args_3[key]
+        assert mc.shape['D'][key] == args_4[key]
+        assert mc.shape['E'][key] == args_5[key]
+
+    args_1_invalid = copy.deepcopy(args_1)
```
b-butler: If using pytest fixtures with a function scope (the default), you wouldn't need to work about deepcopying since they are regenerated after every test.

hoomd/hpmc/pytest/test_faceted_ellipsoid.py
```diff
@@ -1,31 +1,314 @@
...
+    args_4_invalid = copy.deepcopy(args_4)
+    args_5_invalid = copy.deepcopy(args_5)
+    args_1_invalid['normals'] = 'invalid'
+    args_2_invalid['normals'] = 1
+    args_3_invalid['normals'] = [1, 2, 3, 4]
+    args_4_invalid['origin'] = 1
+    args_5_invalid['origin'] = 'invalid'
+
+    # check for errors on invalid input
+    with pytest.raises(hoomd.typeconverter.TypeConversionError):
```
b-butler: Like before, error checking should not be done solely in the attaching test. These should cause errors whether attached or not.

hoomd/hpmc/pytest/test_polyhedron.py
```diff
@@ -1,32 +1,210 @@
 import hoomd
-import hoomd.hpmc
-import hoomd.hpmc._hpmc as hpmc
+from hoomd.hpmc import _hpmc
+from hoomd import hpmc
 
+
+args_1 = {'vertices': [(0, 0, 0),
```
b-butler: See previous comment on fixtures.

hoomd/hpmc/pytest/test_polyhedron.py
```diff
@@ -1,32 +1,210 @@
...
+    poly.shape['A'] = args_4
+    poly.shape['B'] = args_5
+
+    sim = dummy_simulation_factory(particle_types=['A', 'B'])
+    sim.operations.add(poly)
+    sim.operations.schedule()
+
+    assert not poly.shape['A']['ignore_statistics']
+    assert poly.shape['B']['ignore_statistics']
+    for key in args_4.keys():
```
b-butler: Notice this pattern, this could be used for all the tests like this.

hoomd/hpmc/pytest/test_polyhedron.py
```diff
@@ -1,32 +1,210 @@
...
+                                (3, 2, 0),
+                                (4, 2, 1),
+                                (4, 0, 2),
+                                (4, 1, 0)],
+                         overlap=[0, 0, 0, 0, 0, 0])
+    sim = lattice_simulation_factory()
+    sim.operations.add(mc)
+    sim.operations.schedule()
+    sim.run(100)
+    accepted_rejected_rot = sum(sim.operations.integrator.rotate_moves)
```
b-butler: It would be nice to check that `rotate_moves` and other loggable quantities properly return None when not attached.

hoomd/hpmc/pytest/test_sphere_union.py
```diff
@@ -0,0 +1,175 @@
...
+    test_sph_dict4 = test_sph_union4.asDict()
+    assert test_sph_dict4 == sph_union_args4
+
+
+def test_shape_params():
+
+    mc = hoomd.hpmc.integrate.SphereUnion(23456)
+
+    mc.shape['A'] = dict()
+    print(mc.shape['A'])
```
b-butler: Another `print` statement here. Please do a grep across the testing directory and remove these.

hoomd/hpmc/pytest/test_faceted_ellipsoid_union.py
```diff
@@ -0,0 +1,591 @@
...
+                                          (1, 0, 0)],
+                            'orientations': [(1, 0, 0, 0),
+                                             (1, 1, 0, 0),
+                                             (1, 0, 0, 1),
+                                             (1, 0, 0, 0)],
+                            'overlap': [1, 0, 0, 0],
+                            'capacity': 4,
+                            'ignore_statistics': 1}
+
+faceted_ell_union_args22 = {'shapes': [faceted_ell_args_1,
```
b-butler: Do we really need all these combinations? Some of these seem redundant. If we do, could we perhaps programmatically generate them since they seem to just be different combinations of the same building blocks?

hoomd/hpmc/pytest/test_faceted_ellipsoid_union.py
```diff
@@ -0,0 +1,591 @@
...
+                                             (0, 0, 1, 1),
+                                             (1, 0, 0, 0)],
+                            'overlap': [1, 1, 0, 2, 1],
+                            'capacity': 4,
+                            'ignore_statistics': 1}
+
+
+def test_dict_conversions():
+
+    test_facet_ell_union1 = _hpmc.mfellipsoid_params(faceted_ell_union_args1)
```
b-butler: Great place for a for loop

hoomd/hpmc/pytest/test_faceted_ellipsoid_union.py
```diff
@@ -0,0 +1,591 @@
+import hoomd
```
b-butler: This whole file could benefit from more uses of for loops. If you are going to leave the args as module level variables you can access them programmatically with `globals()[key]`. I really recommend pytest fixtures, however.

hoomd/hpmc/pytest/test_spheropolyhedron_union.py
```diff
@@ -0,0 +1,460 @@
+import hoomd
```
b-butler: Same comments as `test_sphere_union.py`

.azp/templates/build_test.yml
```diff
@@ -33,7 +33,7 @@ steps:
 - script: cat pytest.out.1
   displayName: Display rank 1 output
   workingDirectory: $(Build.BinariesDirectory)
-  condition: and(succeeded(), eq(variables['enable_mpi'], 'on'))
+  condition: and(or(succeeded(), failed()), eq(variables['enable_mpi'], 'on'))
```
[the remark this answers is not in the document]
joaander: no, there is a `cancelled()` state.

hoomd/hpmc/ShapeConvexPolyhedron.h
```diff
@@ -647,53 +684,50 @@ class ProjectionFuncConvexPolyhedron
...
         {
         // return the precomputed diameter
         return verts.diameter;
         }
 
-    //! Get the in-sphere radius
+    /// Get the in-sphere radius of the shape
     DEVICE OverlapReal getInsphereRadius() const
         {
         // not implemented
```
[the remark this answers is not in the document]
joaander: These are implemented for some of the template shapes, but not all - so the method needs be present in every shape. I believe this only used by the quermass feature when @jglaser plans to remove.

hoomd/hpmc/integrate.py
```diff
@@ -351,629 +359,759 @@ def type_shapes(self):
...
-        self.initialize_shape_params()
+        typeparam_shape = TypeParameter('shape', type_kind='particle_types',
+                                        param_dict=TypeParameterDict(
+                                            vertices=list,
+                                            faces=list,
+                                            sweep_radius=0.0,
+                                            capacity=4,
+                                            origin=(0, 0, 0),
+                                            hull_only=True,
+                                            overlap=list,
```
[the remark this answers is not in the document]
joaander: Yes, overlap should be optional and default to None.

joaander:
Regarding the `run(1)` it is currently necessary for MPI tests to pass. For reasons I don't recall, we initialize the classes without a Communicator object, so they think they are not in a domain decomposition simulation. The Communicator is passed from System to all of the owned objects at the start of the run.

hoomd/hpmc/ShapeConvexPolygon.h
```diff
@@ -45,25 +45,25 @@ const unsigned int MAX_POLY2D_VERTS=64;
 */
 struct PolygonVertices : ShapeParams
     {
-    /// X coordinates of the vertices
```
joaander: The `///` comments that precede a class/struct member variable or method should remain. These are special documentation comments that notify Doxygen (http://doxygen.nl/manual/docblocks.html) and other tools (such as IDEs) they are documentation.

I believe the comment requesting `///` being replaced with `//` was referring to a comment inside code that was not documenting a class or struct member.

hoomd/hpmc/ShapeEllipsoid.h
```diff
@@ -39,44 +32,45 @@
...
-    The parameter defining an ellipsoid is a OverlapReal4. First three components list the major axis in that direction.
-    The last component (w) is a ignore flag for overlaps. If w!=0, for both particles in overlap check, then overlaps
-    between those particles will be ignored.
-
-    \ingroup shape
+    Define an ellipsoid shape by the three principal semi-axes.
 */
+struct EllipsoidParams : ShapeParams
+    {
+    /// Principle semi-axis of the ellipsoid in the x-direction
```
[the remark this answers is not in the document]
joaander: This is a javadoc style documentation comment. Doxygen and other tools interpret this as documentation and will associate the comment with the declared variable: http://doxygen.nl/manual/docblocks.html

hoomd/hpmc/pytest/test_convex_polygon.py
joaander: This run(1) may no longer be necessary now that we are calling setCommunicator in attach. We should remove them and see if the tests pass in MPI.

hoomd/hpmc/module.cc
```diff
@@ -121,16 +118,6 @@ PYBIND11_MODULE(_hpmc, m)
...
-    m.def("make_poly3d_data", &make_poly3d_data);
-    m.def("make_poly3d_verts", &make_poly3d_verts);
-    m.def("make_ell_params", &make_ell_params);
-    m.def("make_sph_params", &make_sph_params);
-    m.def("make_faceted_ellipsoid", &make_faceted_ellipsoid);
-    m.def("make_sphinx3d_params", &make_sphinx3d_params);
-    m.def("make_convex_polyhedron_union_params", &make_union_params<ShapeSpheropolyhedron>);
-    m.def("make_faceted_ellipsoid_union_params", &make_union_params<ShapeFacetedEllipsoid>);
-    m.def("make_sphere_union_params", &make_union_params<ShapeSphere>);
     m.def("make_overlapreal3", &make_overlapreal3);
```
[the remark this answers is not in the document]
joaander: They can be removed.

hoomd/hpmc/ShapeConvexPolyhedron.h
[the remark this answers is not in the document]
joaander: These are vestigal code. Jens will remove in a different branch.

hoomd/conftest.py
```diff
@@ -132,3 +136,1518 @@ def pytest_sessionfinish(session, exitstatus):
 
     if exitstatus != 0 and hoomd._hoomd.is_MPI_available():
         atexit.register(abort, exitstatus)
+
+
+@pytest.fixture(scope="session")
```
b-butler: These should be in `hoomd/hpmc/conftest.py`. Also, they should only have a function scope.

hoomd/hpmc/integrate.py
```diff
@@ -425,9 +423,11 @@ def __init__(self, seed, d=0.1, a=0.1, move_ratio=0.5,
         typeparam_shape = TypeParameter('shape', type_kind='particle_types',
                                         param_dict=TypeParameterDict(
                                             vertices=list,
-                                            ignore_statistics=False,
-                                            sweep_radius=0,
-                                            len_keys=1)
+                                            ignore_statistics=bool,
```
b-butler: `ignore_statistic` and `sweep_radius` were fine. `explicit_defaults` is only for the case where you have a more complex validation function to specify. Otherwise `TypeParameterDict` will be able to tell that `False` is a boolean value and `0` is an integer. If you want a float `0.` will be recognized as such.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -0,0 +1,109 @@
...
+                    np.testing.assert_allclose(test_dict[key], val)
+                elif key != 'shapes':
+                    assert test_dict[key] == val
+
+
+def test_default_args(integrator_args):
+    for shape_integrator, valid_args, invalid_args in integrator_args():
+        mc = shape_integrator(23456)
+        mc.shape['A'] = dict()
+        for key, val in list(mc.shape["A"].items()):
```
b-butler: You don't need this list here.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -0,0 +1,109 @@
...
+                and key != 'shapes':
+                    np.testing.assert_allclose(test_dict[key], val)
+                elif key != 'shapes':
+                    assert test_dict[key] == val
+
+
+def test_default_args(integrator_args):
+    for shape_integrator, valid_args, invalid_args in integrator_args():
+        mc = shape_integrator(23456)
+        mc.shape['A'] = dict()
```
b-butler: This is unnecessary.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -0,0 +1,109 @@
...
+        for key, val in list(mc.shape["A"].items()):
+            assert mc.shape["A"][key] == val
+
+
+def test_shape_params(integrator_args):
+    for shape_integrator, valid_args, invalid_args in integrator_args():
+        mc = shape_integrator(23456)
+        for args in valid_args:
+            mc.shape["A"] = args
+            for key, val in list(args.items()):
```
b-butler: list call unnecessary.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -0,0 +1,109 @@
...
+                if key == "vertices":
+                    np.testing.assert_allclose(mc.shape["A"][key], val)
+                else:
+                    assert mc.shape["A"][key] == val
+        for args in invalid_args:
+            tf = False
```
b-butler: Look at `pytest.raises`

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -0,0 +1,109 @@
...
+    for shape_integrator, valid_args, invalid_args in integrator_args():
+        mc = shape_integrator(23456)
+        for args in valid_args:
+            mc.shape["A"] = args
+            sim = dummy_simulation_factory()
+            sim.operations.add(mc)
+            sim.operations.schedule()
+            for key, val in list(args.items()):
+                if isinstance(mc.shape["A"][key], list) and len(mc.shape["A"][key]) > 0 \
+                and key != 'shapes':
```
b-butler: Why don't we do any checks for `shapes`. This is for the unions right? Do we currently have a problem converting back to python?

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -0,0 +1,109 @@
...
+                s.particles.position[0] = (0, 0, 0)
+                s.particles.position[1] = (0, 0.5, 0)
+            sim.state.snapshot = s
+            assert mc.overlaps == 1
+
+
+def test_moves(device, lattice_simulation_factory, integrator_args):
+    dims = 3
+    for shape_integrator, valid_args, invalid_args in integrator_args():
+        if 'union' not in str(shape_integrator).lower():
```
b-butler: Do we do different testing for unions?

hoomd/hpmc/integrate.py
```diff
@@ -315,11 +315,11 @@ class Sphere(_HPMCIntegrator):
...
-        mc.shape_param.set('B', diameter=2.0)
-        mc.shape_param.set('C', diameter=1.0, orientable=True)
-        print('diameter = ', mc.shape_param['A'].diameter)
+        mc = hoomd.hpmc.integrate.Sphere(seed=415236, d=0.3, a=0.4)
+        mc.shape["A"] = dict(diameter=1.0)
+        mc.shape["B"] = dict(diameter=2.0)
+        mc.shape["C"] = dict(diameter=1.0, orientable=True)
+        print('diameter = ', mc.shape["A"]["diameter"])
 
     Depletants Example::
```
b-butler: The depletant example has not been changed.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -2,46 +2,44 @@
...
 
 def test_dict_conversion(shape_dict_conversion_args):
     for shape_params, args_list in shape_dict_conversion_args():
         for args in args_list:
             test_shape = shape_params(args)
             test_dict = test_shape.asDict()
-            for key, val in list(args.items()):
-                if isinstance(test_dict[key], list) and len(test_dict[key]) > 0 \
-                and key != 'shapes':
+            for key, val in args.items():
```
b-butler: The logic for checking all items of a dictionary being similar is the same here and in `test_shape_attached`. You could potentially write a non test function with the logic and just call it here and in the other function appropriately.

joaander:
I added `hpmc/pytest/conftest.py` to `CMakeLists.txt` and merged `feature/new-object-API` up into this branch. Aside from getting the tests to pass, what more needs to be done before merging this?

hoomd/conftest.py
```diff
@@ -87,3 +136,1266 @@ def pytest_sessionfinish(session, exitstatus):
...
+#                  sphere_union_valid_args()),
+#                 (convex_spheropolyhedron_union_parameters(),
+#                  convex_spheropolyhedron_union_valid_args()),
+#                 (sphinx_parameters(),
+#                  sphinx_valid_args())]
+#     return get_valid_args
+
+
+# @pytest.fixture(scope="session")
+# def integrator_args(convex_polygon_integrator,
```
joaander: Let's not add thousands of lines of commented code.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -0,0 +1,103 @@
...
+        mc = shape_integrator(23456)
+        for args in valid_args:
+            mc.shape["A"] = args
+            sim = dummy_simulation_factory()
+            sim.operations.add(mc)
+            sim.operations.schedule()
+            check_dict(mc.shape["A"], args)
+
+
+def test_overlaps(device, lattice_simulation_factory, integrator_args):
```
joaander: I do not see how this effectively tests the shape overlaps. It seems to assume that all valid shape arguments do not overlap in the (0,0,0) / (0,8,0) configuration and do in the (0,0,0) / (0, 0.5,0) configuration.

I'd prefer to see shape spedific hand-constructed tests for this that probe the details of each shape class.

hoomd/hpmc/test-py/hpmc_gsd_state.py
```diff
@@ -371,5 +371,18 @@ def test_convex_polyhedron(self):
         with self.assertRaises(RuntimeError):
             self.mc = hpmc.integrate.convex_polyhedron(seed=2234, d=0.3, a=0.4, restore_state=True);
 
+    def test_sphere_union(self):
```
joaander: What is this doing here? This test doesn't even run on this branch. GSD shape/state writing could bet tested for all valid shape args in `test_shape.py`.

joaander:
Ready for final updates and review. This round of tests should pass. Also, the `run(1)` before checking overlaps is no longer necessary.

joaander:
Let's get this done!

hoomd/hpmc/pytest/test_shape.py
b-butler: Agreed, but that requires a more thorough understanding of the geometry. It may be worth raising an issue about this and dealing with it in a separate pull request. Regardless, it needs to be fixed or documented (for later testing).

hoomd/hpmc/ShapePolyhedron.h
```diff
@@ -710,7 +710,7 @@ DEVICE inline bool test_narrow_phase_overlap( vec3<OverlapReal> dr,
         for (unsigned int j= 0; j< nb; j++)
             {
             unsigned int nverts_b, offs_b;
-            unsigned int jface = b.tree.getParticle(cur_node_b, j);
+            unsigned int jface = b.tree.getParticleByNode(cur_node_b, j);
```
b-butler: I can't seem to find the addition or change of `getParticleByNode` was `getParticle` just an alias to `getParticleByNode` and we are being more explicit now, or did I just miss this in an earlier commit?

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -3,40 +3,41 @@
 from hoomd.hpmc import _hpmc
 import numpy as np
 import pytest
+from hoomd.hpmc.pytest.conftest import *
+
+
+def check_dict(shape_dict, args):
+    for key, val in args.items():
```
b-butler: A developer docstring as to the function that `check_dict` does would be helpful.

hoomd/hpmc/ShapePolyhedron.h
jglaser: there are now two ways to access the particle list, `getParticleByNode` and `getParticleByIndex` (for faster iteration), used in `jit/EvaluatorUnionGPU.cuh`

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -3,40 +3,41 @@
...
+        if isinstance(shape_dict[key], list) and len(shape_dict[key]) > 0 \
+           and key != 'shapes':
+            np.testing.assert_allclose(shape_dict[key], val)
+        elif key == 'shapes':
+            for shape_args in shape_dict[key]:
+                for shape_key, shape_val in shape_args.items():
+                    if isinstance(shape_args[shape_key], list) \
+                       and len(shape_args[shape_key]) > 0:
+                        np.testing.assert_allclose(shape_args[shape_key],
```
b-butler: Isn't this just testing that `shape_args[key] == shape_args[key]`. At this point we aren't using the passed val from `args` at all.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -3,40 +3,41 @@
...
+                                                   shape_val)
+                    else:
+                        assert shape_args[shape_key] == shape_val
```
b-butler: Same as above comment.

hoomd/conftest.py
b-butler: Yeah, this can all be safely removed if the fixtures were added to a new file which they seem to be.

hoomd/hpmc/ShapePolyhedron.h
b-butler: Ah, thanks for the clarification, @jglaser.

joaander: This changed was introduced in #549. Due to a situation where ShapePolyhedron.h somehow got converted to 5-space indentation on `feature/new-object-API` and back to 4-space indentation in this PR, the conflicts cause this to appear as a change instead of a merge.

hoomd/hpmc/pytest/test_shape.py
joaander: I just don't see how it is possible to write a generic test for one overlapping configuration and one non-overlapping configuration that is guaranteed to always be correct for all valid shape arguments. Any test that checks for overlaps or not needs to have shape specific knowledge.

hoomd/hpmc/ShapeConvexPolygon.h
```diff
@@ -306,67 +329,48 @@ class SupportFuncConvexPolygon
...
         {
         // return the precomputed diameter
         return verts.diameter;
         }
 
-    //! Get the in-circle radius
+    /// Get the in-sphere radius of the shape
     DEVICE OverlapReal getInsphereRadius() const
         {
         // not implemented
```
b-butler: Maybe we should remove this or mark it as TODO.

hoomd/hpmc/ShapeFacetedEllipsoid.h
```diff
@@ -163,29 +160,148 @@ struct faceted_ellipsoid_params : param_base
...
+                OverlapReal lambda1 = OverlapReal(2.0)*(-b2*dot(n_p,n_p)+b*dotp)/denom;
+
+                vec3<OverlapReal> r = -(lambda0*n_p+lambda1*np2)/OverlapReal(2.0);
+
+                // if the line does not intersect the sphere, ignore
+                if (dot(r,r) > OverlapReal(1.0))
+                    continue;
+
+                // the line intersects with the sphere at two points, one of
+                // maximizes the support function
```
b-butler:
```suggestion
                // which maximizes the support function
```

hoomd/hpmc/ShapePolyhedron.h
```diff
@@ -1,1177 +1,1130 @@
...
+    // away from triangle, so exit early
+    float d = dot(qp, n);
+    if (d <= OverlapReal(0.0)) return false;
+    // Compute intersection t value of pq with plane of triangle. A ray
+    // intersects iff 0 <= t. Segment intersects iff 0 <= t <= 1. Delay
+    // dividing by d until intersection has been found to pierce triangle
+    vec3<OverlapReal> ap = p - a;
+    t = dot(ap, n);
+    if (t < OverlapReal(0.0)) return false;
+//    if (t > d) return false; // For segment; exclude this code line for a ray test
```
b-butler: Should this line just be deleted?

hoomd/hpmc/ShapeSphere.h
```diff
@@ -36,10 +29,19 @@
...
-//! Base class for parameter structure data types
-struct param_base
+/** HPMC shape parameter base class
+
+    HPMC shape parameters must be aligned on 32-byte boundaries for AVX acceleration. The ShapeParam
+    base class implements the necessary aligned memory allocation operations. It also provides
+    empty load_shared and allocated_shared implementations which enabled caching deep copied managed
+    data arrays in shared memory.
+
+    \todo Move base methods out into their own file. ShapeSphere.h will then no longer need to be
```
b-butler: Just for easier grepping since we don't build C++ documentation any more.
```suggestion
    TODO Move base methods out into their own file. ShapeSphere.h will then no longer need to be
```

hoomd/hpmc/ShapeSphere.h
```diff
@@ -142,70 +152,79 @@ struct sph_params : param_base
...
+    Shape classes define the interface used by IntegratorHPMCMono, ComputeFreeVolume, and other
+    classes to check for overlaps between shapes, find their extend in space, and other operations.
+    These classes are specified via template parameters to these classes so that the compiler may
+    fully inline all uses of the shape API.
+
+    ShapeSphere defines this API for spheres.
+
+    Some portions of the API (e.g. test_overlap) are implemented as specialized function templates.
+
+    \todo Should we remove orientation as a member variable from the shape API. It should be passed
```
b-butler:
```suggestion
    TODO Should we remove orientation as a member variable from the shape API. It should be passed
```

hoomd/hpmc/ShapeSphere.h
```diff
@@ -142,70 +152,79 @@ struct sph_params : param_base
...
+          when needed.
+    \todo Don't use specialized templates for things that should be methods (i.e. a.overlapsWith(b))
```
b-butler:
```suggestion
    TODO Don't use specialized templates for things that should be methods (i.e. a.overlapsWith(b))
```

hoomd/hpmc/ShapeSphere.h
```diff
@@ -142,70 +152,79 @@ struct sph_params : param_base
...
+    \todo add hpmc::shape namespace
```
b-butler:
```suggestion
    TODO add hpmc::shape namespace
```

hoomd/hpmc/ShapeSphinx.h
```diff
@@ -24,85 +17,105 @@
...
-/*! \ingroup hpmc_data_structs */
-struct sphinx3d_params : param_base
+    A Sphinx particle is represented by N spheres each with their own diameter.
+
+    ShapeSphinx represents the intersection of spheres. A  positive sphere is one where the volume
+    inside the sphere is considered, and a negative sphere is one where the volume outside the
+    sphere is considered. This shape is defined using a struct called SphinxParams whose detail
+    is named spheres. ShapeSphinx requires an incoming orientation vector, followed by parameters
+    for the struct SphinxParams. The parameter defining a sphinx is a structure containing the
+    circumsphere diameter of all spheres defined in the shape. This is followed the number of
```
b-butler:
```suggestion
    circumsphere diameter of all spheres defined in the shape. This is followed by the number of
```

hoomd/hpmc/ShapeUnion.h
```diff
@@ -114,64 +110,90 @@ struct union_params : param_base
...
-    union_params(unsigned int _N, bool _managed)
-        : N(_N)
+    #ifndef __HIPCC__
+
+    /** Construct with a given number of members
+
+        @note Use of this constructor with N != 0 is intended only for unit tests
+    */
+    DEVICE ShapeUnionParams(unsigned int _N)
+        : mpos(_N, false), morientation(_N, false), mparams(_N, false),
```
b-butler: I know that this from previous code, but maybe we should make a note to name this `m_pos` etc. to keep consistent styling.

hoomd/hpmc/module.cc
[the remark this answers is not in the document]
b-butler: Just making a note that this should still be removed.

hoomd/hpmc/pytest/conftest.py
```diff
@@ -0,0 +1,1271 @@
...
+
+@pytest.fixture(scope="function")
+def convex_polygon_integrator():
+    def make_shape():
+        return hoomd.hpmc.integrate.ConvexPolygon
+    return make_shape
+
+
+@pytest.fixture(scope="function")
+def convex_polygon_parameters():
```
b-butler: Maybe name this `convex_polygon_vertices_cpp

hoomd/hpmc/pytest/conftest.py
```diff
@@ -0,0 +1,1271 @@
...
+    return get_parameters
+
+
+@pytest.fixture(scope="function")
+def convex_polygon_valid_args():
+    def get_args():
+        args_list = [{'vertices': [(0, (0.75**0.5) / 2),
+                                   (-0.5, -(0.75**0.5) / 2),
+                                   (0.5, -(0.75**0.5) / 2)],
+                      'ignore_statistics': 0,
```
b-butler: Could we try one of these without `ignore_statistics` explicitly set. That should still work since it is optional. Same for `sweep_radius` I believe.

hoomd/hpmc/pytest/conftest.py
```diff
@@ -0,0 +1,1271 @@
...
+def ellipsoid_parameters():
+    def get_parameters():
+        return hoomd.hpmc._hpmc.EllipsoidParams
+    return get_parameters
+
+
+@pytest.fixture(scope="function")
+def ellipsoid_valid_args():
+    def get_args():
+        args_list = [{"a": 0.75, "b": 1, "c": 0.5, 'ignore_statistics': 0},
```
b-butler: Could we have optional parameters not set in one of the options.

hoomd/hpmc/pytest/test_shape.py
b-butler: Just a marker that this has not be addressed yet.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+def test_shape_params(integrator_args):
+    for shape_integrator, valid_args, invalid_args in integrator_args():
+        mc = shape_integrator(23456)
+        for args in valid_args:
+            mc.shape["A"] = args
+            check_dict(mc.shape["A"], args)
+        for args in invalid_args:
+            with pytest.raises(Exception):
```
b-butler: I think this is always the same type of exception isn't it? I can't remember for sure, but I believe it is a `hoomd.typeconverter.TypeConverterError`.

hoomd/hpmc/pytest/test_shape.py
danevans09: Fixed a25e2c3e

hoomd/hpmc/integrate.py
```diff
@@ -311,12 +316,13 @@ class Sphere(_HPMCIntegrator):
     Args:
         seed (int): Random number seed
         d (float): Maximum move displacement, Scalar to set for all types, or a
-            dict containing {type:size} to set by type.
+          dict containing {type:size} to set by type.
```
b-butler:
```suggestion
          dict containing {type: size} to set by type.
```

hoomd/hpmc/integrate.py
```diff
@@ -311,12 +316,13 @@ class Sphere(_HPMCIntegrator):
...
         a (float): Maximum rotation move to set for all types, or a dict
-            containing {type:size} to set by type.
+          containing {type:size} to set by type.
```
b-butler:
```suggestion
          containing {type: size} to set by type.
```

hoomd/hpmc/integrate.py
```diff
@@ -311,12 +316,13 @@ class Sphere(_HPMCIntegrator):
...
         move_ratio (float, only used with **orientable=True**): Ratio of
-            translation moves to rotation moves. (added in version 2.3)
+          translation moves to rotation moves. (added in version 2.3)
```
b-butler: I don't know if we need added in version 2.3 anymore since we are releasing this as 3.0.

hoomd/hpmc/integrate.py
b-butler: Actually this is false, and I think my bad. When set on construction, it is merely the default for every type. It can then be updated on a per type basis through `integrator.d['A'] = 1` or `integrator.d = dict(A=1, B=0.5)`.

hoomd/hpmc/integrate.py
b-butler: See above comment on `d`.

hoomd/hpmc/integrate.py
```diff
@@ -330,19 +336,18 @@ class Sphere(_HPMCIntegrator):
```
b-butler: Here is where I would include the documentation about `fugacity`, `interaction_matrix`, `a`, and `d`. Also I would change doc format to be something like
````
Sphere parameters

shape (particle type, dict): defines the shape of the object.
    Keys:
        * *diameter* (float, **required) - diameter of the sphere (distance units)
        * ...

d (particle type, float): the size of displacement trial moves

interaction_matrix ((particle type, particle type), bool): Whether to include overlaps between type 1 and type 2

...
````
Something similar would apply to each of the integrators. If we merge before doing this, we need to raise an issue since without this documentation, it will be hard to learn to use the integrators.

File: hoomd/hpmc/integrate.py [hunk 64]
```diff
@@ -369,1046 +374,1249 @@ def type_shapes(self):
...
+              {'type': 'Sphere', 'diameter': 2}]
 
         Returns:
             A list of dictionaries, one for each particle type in the system.
         """
         return super()._return_type_shapes()
 
 
-class convex_polygon(_HPMCIntegrator):
+class ConvexPolygon(_HPMCIntegrator):
```
b-butler: See comments on the Sphere docstring.

hoomd/hpmc/integrate.py
```diff
@@ -369,1046 +374,1249 @@ def type_shapes(self):
...
 
-    * *a* (**required**) - principle axis a of the ellipsoid (radius in the x direction) (distance units)
-    * *b* (**required**) - principle axis b of the ellipsoid (radius in the y direction) (distance units)
-    * *c* (**required**) - principle axis c of the ellipsoid (radius in the z direction) (distance units)
-    * *ignore_statistics* (**default: False**) - set to True to disable ignore for statistics tracking
+    * *a* (**required**) - principle axis a of the ellipsoid (radius in the x
+      direction) (distance units)
+    * *b* (**required**) - principle axis b of the ellipsoid (radius in the y
+      direction) (distance units)
+    * *c* (**required**) - principle axis c of the ellipsoid (radi
```
b-butler: I don't think we need this since this is for version 3.0.

b-butler: We don't need to specify for all particles. The default is the default.

b-butler:
Also, in a more meta vein, I would be willing to wait on the addressing of some of my comments particularly documentation as long is there is an open issue reporting the problems.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -24,14 +24,16 @@ def check_dict(shape_dict, args):
            and key != 'shapes':
             np.testing.assert_allclose(shape_dict[key], val)
         elif key == 'shapes':
-            for shape_args in shape_dict[key]:
+            for i in range(len(shape_dict[key])):
+                shape_args = shape_dict[key][i]
+                val_args = val[i]
                 for shape_key, shape_val in shape_args.items():
```
b-butler: Good change on the function in general we don't use `shape_val` now though so I suggest changing the line to this
```suggestion
                for shape_key in shape_args:
```

hoomd/hpmc/pytest/test_convex_polygon.py
danevans09: sim.run(1) no longer in test

hoomd/hpmc/integrate.py
```diff
@@ -1363,34 +1542,32 @@ class faceted_ellipsoid_union(_HPMCIntegrator):
...
-        # meta data
-        self.metadata_fields = ['capacity']
+        typeparam_shape = TypeParameter('shape', type_kind='particle_types',
+                                        param_dict=TypeParameterDict(
+                                            shapes=list,
+                                            positions=list,
+                                            orientations=list,
+                                            overlap=list,
+                                            ignore_statistics=False,
+                                            explict_defaults={'orientations': None,
```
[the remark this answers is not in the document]
danevans09: Fixed 0ff4c868ff4

hoomd/hpmc/pytest/test_faceted_ellipsoid_union.py
danevans09: Removed some combinations

hoomd/hpmc/module.cc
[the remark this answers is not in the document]
danevans09: Removed 0ff4c868ff4

hoomd/hpmc/ShapeUnion.h
danevans09: Added comment d98589281

hoomd/hpmc/pytest/conftest.py
danevans09: Renamed similar fixtures d98589281

hoomd/hpmc/pytest/conftest.py
danevans09: I am getting a KeyError when initializing all of the C++ integrators if I don't explicitly set each parameter, see d83f466fe

hoomd/hpmc/pytest/test_polyhedron.py
danevans09: Added ec1b7eb0e

hoomd/hpmc/pytest/test_shape.py
danevans09: I am currently getting a KeyError whenever I try to attach a union

hoomd/hpmc/pytest/conftest.py
danevans09: Made separate fixture for calling c++ objects directly 13e7126f9

hoomd/hpmc/pytest/test_shape.py
danevans09: Fixed fe04009df9

hoomd/hpmc/pytest/test_shape.py
danevans09: Fixed ba04b2da8

hoomd/hpmc/integrate.py
danevans09: Removed 1fc0c52f6

hoomd/hpmc/integrate.py
danevans09: I am confused what you mean by this and the above comment. Should I remove the '(distance units)'?

danevans09:
I think I have corrected most of the problems with the integrator testing.

1. The only integrator without a specific overlap test is Sphinx. I don't think OVITO can render sphinxes and I am having a hard time picturing the shape and how to test overlaps for it. I made one with a negligible indent diameter and treated it like a sphere, but this doesn't test that the indenting works.

2. Documentation. I pushed updated documentation for Sphere and ConvexPolygon following Brandon's outline. I will update the rest of the documentation and make sure all the examples work

hoomd/hpmc/integrate.py
```diff
@@ -369,1046 +380,1250 @@ def type_shapes(self):
...
+                shapes=[dict(vertices=[(float, float, float)],
+                             sweep_radius=0.0,
+                             ignore_statistics=False)],
+                positions=[(float, float, float)],
+                orientations=[(float, float, float, float)],
+                overlap=[bool],
+                ignore_statistics=False,
+                capacity=4,
+                len_keys=1,
+                explicit_defaults={'orientations': None,
```
jglaser, line 1506: this does not work. When I don't explicitly set `orientations` or `overlap`, I get
```
  File "/nfs/turbo/bi-vislab/jsglaser/miniconda3/envs/myenv/lib/python3.7/site-packages/hoomd/parameterdicts.py", line 214, in __init__
    self._trusted_setitem(key, type_param_dict[key])
  File "/nfs/turbo/bi-vislab/jsglaser/miniconda3/envs/myenv/lib/python3.7/site-packages/hoomd/parameterdicts.py", line 249, in _trusted_setitem
    getattr(self._cpp_obj, self._setter)(key, val)
TypeError: 'NoneType' object is not iterable
```

hoomd/hpmc/integrate.py
```diff
@@ -369,1046 +380,1250 @@ def type_shapes(self):
...
+                         normals=[(float, float, float)], offsets=[float],
+                         vertices=[(float, float, float)], origin=tuple,
+                         ignore_statistics=False)],
+                positions=[(float, float, float)],
+                orientations=[(float, float, float, float)],
+                overlap=[bool],
+                ignore_statistics=False,
+                capacity=4,
+                len_keys=1,
+                explicit_defaults={'orientations': None,
```
jglaser, line 1626: same problem as for the other *Union types

hoomd/hpmc/integrate.py
```diff
@@ -369,1046 +380,1250 @@ def type_shapes(self):
...
+            param_dict=TypeParameterDict(
+                shapes=[dict(diameter=float, ignore_statistics=False,
+                             orientable=False)],
+                positions=[(float, float, float)],
+                orientations=[(float, float, float, float)],
+                capacity=4,
+                overlap=[bool],
+                ignore_statistics=False,
+                len_keys=1,
+                explicit_defaults={'orientations': None,
```
jglaser, line 1399: same problem as for other *Union types

jglaser:
Is there a single document describing the intent and state of the planned API changes? I am having a hard time figuring out what has already been ported, what is not currently working, for which of the components at least API design ideas exist, and for which ones we do not (yet) any plans.

Specifically, I would like to re-enable the `jit` component, but I am having a hard time guessing what plans you may have had in mind. Plus, I would like to use this branch since the `SphereUnion` GSD shape export feature is only available in here, even though I may consider 'backporting' it to `next`.

hoomd/hpmc/integrate.py
```diff
@@ -109,17 +107,18 @@ def __init__(self, seed, d, a, move_ratio, nselect, deterministic):
     def attach(self, simulation):
         '''initialize the reflected c++ class'''
         sys_def = simulation.state._cpp_sys_def
-        if simulation.device.mode == 'GPU':
+        if simulation.device.mode == 'gpu':
             self._cpp_cell = _hoomd.CellListGPU(sys_def)
             if simulation._system_communicator is not None:
                 self._cpp_cell.setCommunicator(simulation._system_communicator)
             self._cpp_obj = getattr(_hpmc,
-                                    self._cpp_cls + 'GPU')(sys_def,
```
joaander, line 116: I appreciate the effort, but we currently don't have a 1-1 mapping between C++ exported class names and Python class names. I'm going to revert this to get tests pass again. Later we could bring this back as a simplification: #622

hoomd/hpmc/integrate.py
joaander: Thanks for reporting. `None` is supposed to expand to an N length list of (1,0,0,0) for orientation and 1's for overlap. We will correct this.

.azp/test.yml
```diff
@@ -222,29 +222,29 @@ stages:
...
-    steps:
-    - checkout: self
-      submodules: true
-    - script: sphinx-build -b html -d _build/doctrees -W --keep-going -n . _build/html
-      displayName: (HTML)
-      workingDirectory: sphinx-doc
-    - script: sphinx-build -b latex -d _build/doctrees -W --keep-going -n . _build/latex
-      displayName: (LaTeX)
-      workingDirectory: sphinx-doc
+# - stage: doc
```
b-butler: Do we plan to get the docs working later?

.azp/validate.yml
```diff
@@ -18,228 +18,244 @@ variables:
   ctest_stride: 4
   mpiexec_timeout: 21600
 
+# a temporary empty job so that CI will pass while the actual tests are disabled
```
b-butler: I assume we plan on adding back validation tests later? Similar to the docs?

hoomd/hpmc/ShapePolyhedron.h
```diff
@@ -827,7 +828,7 @@ DEVICE inline bool IntersectRayTriangle(const vec3<OverlapReal>& p, const vec3<O
     vec3<OverlapReal> ap = p - a;
     t = dot(ap, n);
     if (t < OverlapReal(0.0)) return false;
-//    if (t > d) return false; // For segment; exclude this code line for a ray test
+    // For segment; exclude this code line for a ray test
```
b-butler: Can this line just be deleted?

hoomd/md/pytest/test_flags.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.fixture(scope='session')
+def md_simulation_factory(device):
+    def make_simulation(particle_types=['A']):
+        s = hoomd.Snapshot(device.comm)
+
+        # initialize particles on a simple cubic lattice and perturb them
+        # slightly off the lattice positions
+        # TODO: When hpmc_integrator_api is merged, port this functionality
```
b-butler: Do we want to go ahead and port this now since this is on that branch?

hoomd/hpmc/integrate.py
```diff
@@ -311,28 +314,37 @@ def counters(self):
...
+    Sphere parameters:
 
-    Hard particle Monte Carlo integration method for spheres.
+    shape (particle type, dict): defines the shape of the object.
 
-    Sphere parameters:
+        Keys:
+            * *diameter* (float, **required**) - diameter of the sphere
+              (distance units)
+            * *ignore_statistics* (**default: False**) - set to True to ignore
```
b-butler, line 326: Could we have something like `* *ignore_statistics* (type, **default:** value)

hoomd/hpmc/integrate.py
```diff
@@ -311,28 +314,37 @@ def counters(self):
...
 
-    * *diameter* (**required**) - diameter of the sphere (distance units)
-    * *orientable* (**default: False**) - set to True for spheres with
-      orientation (added in version 2.3)
-    * *ignore_statistics* (**default: False**) - set to True to disable ignore
-      for statistics tracking
+    interaction_matrix ((particle type, particle type), bool): whether to
+    include overlaps between type 1 and type 2
+
+    seed (int): random number seed
```
b-butler, line 341: We should have these before the `TypeParameters`. Otherwise it is confusing which are regular arguments. Also, it may be helpful to give the two different headings.

````
Attributes
-------------

Regular attributes seed, etc.

TypeParameters
----------------------

shape, etc.
````

Comment (b-butler) [reply hoomd/hpmc/integrate.py:341]: I realize this changes what I said earlier and you can leave the TypeParameter heading `Integrator Name parameters` if you like.

File: hoomd/hpmc/integrate.py [hunk 78]
```diff
@@ -645,53 +687,54 @@ class Polyhedron(_HPMCIntegrator):
...
+            * *faces* (list, **required**) - list of vertex indices for every
+              face
+                * For visualization purposes, the faces **MUST** be defined with
+                  a counterclockwise winding order to produce an outward normal.
+            * *ignore_statistics* (**default: False**) - set to True to ignore
+              tracked statistics
+            * *sweep_radius* (**default: 0.0**) - radius of the sphere swept
+              around the edges of the polyhedron (distance units). Set a
+              non-zero sweep_radius to create a spheropolyhedron
+            * *overlap* (**default: None**) - only check overlap between
```
b-butler, line 711: Does not describe what the default means.

hoomd/hpmc/integrate.py
```diff
@@ -769,16 +812,16 @@ def __init__(self, seed, d=0.1, a=0.1, move_ratio=0.5,
 
         typeparam_shape = TypeParameter('shape', type_kind='particle_types',
                                         param_dict=TypeParameterDict(
-                                            vertices=list,
-                                            faces=list,
+                                            vertices=[(float, float, float)],
+                                            faces=[[int, int, int]],
```
b-butler, line 816: Ca
```suggestion
                                            faces=[[int, int, int]],  # TODO make faces use tuples instead of a list of exactly 3 elements.
```

hoomd/hpmc/integrate.py
```diff
@@ -1326,44 +1335,50 @@ def type_shapes(self):
...
+    shape (particle type, dict): defines the shape of the object.
+
+        Keys:
+            * *shapes* (**required**) - list of sphere objects to be included in
+              union
+            * *positions* (**required**) - list of centers of constituent
+              spheres in particle coordinates.
+            * *orientations* (**required**) - list of orientations of
+              constituent spheres.
+            * *overlap* (**default: 1**) - only check overlap
```
b-butler, line 1357: default's meaning is not described.

hoomd/hpmc/integrate.py
```diff
@@ -1430,42 +1445,47 @@ def type_shapes(self):
...
+    shape (particle type, dict): defines the shape of the object.
+
+        Keys:
+            * *shapes* (**required**) - list of polyhedron objects to be
+              included in union
+            * *positions* (**required**) - list of centers of constituent
+              polyhedra in particle coordinates.
+            * *orientations* (**required**) - list of orientations of
+              constituent polyhedra.
+            * *overlap* (**default: 1**) - only check overlap
```
b-butler, line 1462: default's meaning in this context is not described.

hoomd/hpmc/integrate.py
```diff
@@ -1499,58 +1519,70 @@ def __init__(self, seed, d=0.1, a=0.1, move_ratio=0.5,
...
+    shape (particle type, dict): defines the shape of the object.
+
+        Keys:
+            * *shapes* (**required**) - list of ellipsoid objects to be included
+              in union
+            * *positions* (**required**) - list of centers of constituent
+              ellipsoids in particle coordinates.
+            * *orientations* (**required**) - list of orientations of
+              constituent ellipsoids.
+            * *overlap* (**default: 1**) - only check overlap
```
b-butler, line 1561: See above.

.azp/test.yml
joaander: Yes, in another PR. The majority of the doc  build failures are due to namespace changes on `feature/new-object-aPI`.

.azp/validate.yml
joaander: Yes, we will convert existing validation tests over to pytest in another PR.

hoomd/hpmc/ShapePolyhedron.h
joaander: @jglaser would have to comment on that.

hoomd/md/pytest/test_flags.py
joaander: Done.

hoomd/hpmc/integrate.py
joaander: Done.

hoomd/hpmc/integrate.py
b-butler: Disregard comment. I disagree with my earlier opinion here.

hoomd/hpmc/integrate.py
joaander: fixed.

hoomd/hpmc/integrate.py
joaander: fixed.

hoomd/hpmc/integrate.py
joaander: fixed.

hoomd/hpmc/integrate.py
joaander: fixed.

hoomd/hpmc/integrate.py
joaander: fixed.

hoomd/hpmc/integrate.py
```diff
@@ -1460,6 +1463,32 @@ class SphereUnion(_HPMCIntegrator):
...
+                                **default:** None) -
+              Orientation of each sphere in the union. When not None,
+              ``orientations`` must have a length equal to that of
+              ``positions``. When None (the default), ``orientations`` is
+              initialized with all [1,0,0,0]'s.
+            * ``overlap`` (List[bool], **default:** None) - Check for overlaps
+              between constituent particles when ``overlap [i] & overlap[j]``
+              is nonzero (``&`` is the bitwise AND operator). When not None,
+              ``overlap`` must have a length equal to that of ``faces``. When
+              None (the default), ``overlap`` is initialized with all 1's.
```
b-butler, line 1480: While true on the C++ side, users should only see Boolean values, so should this be all `True`s?

joaander:
This is ready for final review. I have updated the documentation formatting for all HPMC integrators. You can preview the doc build here:
https://hoomd-blue.readthedocs.io/en/hpmc_integrator_api/module-hpmc-integrate.html

hoomd/hpmc/integrate.py
joaander: I thought I saw a commit where @jglaser changed it from bool back to int. I'm assuming that there is some reason for this (e.g. to treat these as bit fields for groups). Is that true @jglaser ? Or should `overlap` be boolean?

joaander: This is a bit mask. Perhaps we should rename it to `mask` to make this more explicit.

b-butler: what is the bit mask for?

b-butler: If it is user facing, we need to document its use and keep it as an int with a name like `overlap_mask`.

If not, does setting this to all ones by default break anything? If not we can still pretend it is a `bool` on the Python side, and know that C++ will overwrite the 1's when necessary.

jglaser: This bit mask allows the user to select certain constituent particles for overlap checking. **Example**: one union with two constituent particle types (read, patches) with `overlap=1` and `overlap=3`. Now have a second union particle (read, depletant), a single sphere with `overlap=2`. It will only interact with the second type of patch, but not with the first one. Nevertheless, no patch can cross any other. `1` is a reasonable default. It should be `int` not `bool`.

b-butler: Okay, I see we document this by describing the bit wise AND. Thanks for the explanation @jglaser.

joaander:
I am going to merge this now so that it stops holding up further development of v3. #632 documents the AMD memory errors.

PR #545, opened 2019-11-27
Merge LBVH changes into next

jglaser:
## Description

a merge commit that includes master and #544, resolving some conflicts and updating `neighbor` to a feature branch (`hoomd_next`)

## Checklist:


jglaser:
I'll also have to 

- [ ] adapt neighbor plugin to new cuda language support in CMake

jglaser:
I temporarily pointed the neighbor plugin to my private fork, https://github.com/jglaser/neighbor/tree/hoomd_next
https://github.com/mphoward/neighbor/pull/5

@mphoward good news! With nlist tree, the LJ particle benchmark (`N=1000000`) is now actually **faster** than with `nlist.cell()`. I will therefore switch over the nlist object in the benchmarks
https://github.com/glotzerlab/hoomd-benchmarks/tree/next

This branch compiles and runs, let's wait until unit tests pass.

jglaser:
regarding the failing unit test `stats_check.py-mpi-gpu` , failing in HPMC/GPU on cuda10_py37_mng_mpi, is the OpenMPI therein CUDA-aware? @joaander

jglaser:
> regarding the failing unit test `stats_check.py-mpi-gpu` , failing in HPMC/GPU on cuda10_py37_mng_mpi, is the OpenMPI therein CUDA-aware? @joaander

now fixed in #547

jglaser:
- [ ] point to `hoomd-3` branch in github.com/mphoward/neighbor

joaander:
Is `nlist.tree` now truly generally faster than `nlist.cell` for most benchmarks?

Can we get the last TODO item resolved soon so that we can get *next* up to date with everything in *maint* and *master*?

mphoward:
We would need to rerun the full suite of benchmarks to find out for sure. In my early tests, I could find cases where `nlist.tree` was still a bit slower (~10% in TPS). This might depend on `N`, though, as I think the slow part of `nlist.tree` relative to `nlist.cell` is now the build (tree needs sorting, cell is atomics), and this build can probably be amortized if you have enough particles to traverse.

`nlist.tree` is also probably slower if you have a bunch of types in the system, but all the types have similar cutoffs, because it always builds one tree per type and the overhead of traversing many trees turns out to be higher than I thought. We could consider writing an alternative version that builds only one tree and traverses all types against it, but it's not obvious how this will perform.

mphoward:
@joaander @jglaser Would you want to merge this with the submodule pointing to Jens' repo? I am almost finished with the changes needed to make `neighbor` header-only on [remove-hoomd](https://github.com/mphoward/neighbor/tree/remove-hoomd), and I will merge it into `hoomd-3` when its ready. But, this will then require me to modify `NeighborListGPUTree` to use the updated library interface.

One issue is that once some of these branches inevitably get deleted, we won't be able to roll back to these commits. Maybe this is OK for an unstable branch though.

joaander:
As long as the branches are merged (and not squash-merged) before they are deleted, git won't remove the old commits from them. We can still access them as long as we point to specific commits with our submodules.

mphoward:
OK, in that case, @jglaser should point the submodule in this PR onto the commit at the current head of `hoomd-3` in `neighbor`, which squash merged his compatibility changes. We can handle the header-only branch in a separate PR here and in neighbor. Then, HIP compatibility will be an additional PR for neighbor and should just be a submodule update for hoomd.

I will then merge commit `hoomd-3` into `master`.

jglaser:
done. Let's wait for unit test results.

PR #546, opened 2019-11-29
fix diameter shifting across pbc

jglaser:
## Description

Fixes the application of particle diameter shifts in the tree neighbor lists across periodic boundary conditions. Requires an updated version of the neighbor plugin, see
https://github.com/mphoward/neighbor/pull/8

I'll update the `neighbor` submodule in this PR once that above PR for `neighbor` has been merged into master. This is required for the checks in this PR to pass.

## Motivation and Context

The `patchy_protein` benchmark in `hoomd-benchmarks` failed with a particle out of the box error when the neighbor list is changed to `nlist.tree()`
https://github.com/glotzerlab/hoomd-benchmarks/tree/next

## How Has This Been Tested?

Unfortunately, the `test_neighbor_list` benchmark doesn't test the combination of periodic boundary conditions and particle diameter shifting.

## Checklist:

### DIFF
--- a/hoomd/md/NeighborListGPUTree.cc
+++ b/hoomd/md/NeighborListGPUTree.cc
@@ -381,6 +381,8 @@ void NeighborListGPUTree::traverseTree()
     // clear the neighbor counts
     cudaMemset(d_n_neigh.data, 0, sizeof(unsigned int)*m_pdata->getN());
 
+    const BoxDim& box = m_pdata->getBox();
+
     // traverse all pairs in (now-transposed) streams
     cudaDeviceSynchronize();
     for (unsigned int i=0; i < m_pdata->getNTypes(); ++i)
@@ -430,7 +432,8 @@ void NeighborListGPUTree::traverseTree()
                                                       Ni,
                                                       m_pdata->getN(),
                                                       rcut,
-                                                      rlist);
+                                                      rlist,
+                                                      box);
                 m_traversers[j]->traverse(nlist_op, query_op, map, *m_lbvhs[j], m_image_list, m_streams[i]);
                 }
             else if (m_filter_body && !m_diameter_shift)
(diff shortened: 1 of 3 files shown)


jglaser:
> Good catch, this definitely needs to be fixed.
> 
> However, couldn't this be done without modifying the neighbor library by adding a `BoxDim` to the `ParticleQueryOp`? You can then wrap `dr` in `refine`, as is done in the cell list version. I think this still works correctly because exactly one image will overlap in the broad phase detection, so it's OK to then wrap in narrow phase (no duplicates).
>
Yes it could be done, but at the expense of even more registers (an entire BoxDim!).

> I would prefer to do this over modifying `neighbor` because (1) I can't currently think of enough use cases for `refine` to always need the image, and (2) the box could be held in constant memory but the image has to be in local memory. 

I understand. Though constant memory is not a portable solution as I figured out with HIP. We should not use it. I had to eliminate all `__constant__`  variables and initialize them as `__shared__` inside the kernel

> This might increase register pressure unnecessarily for use cases that don't need to diameter shift. What do you think?

It's hard to predict, but it your consideration almost sounds like premature optimization, to be honest.
> 
> Should we also add a unit test for this?

Yes, I will add one.

mphoward:
The suggestion is actually the opposite of optimization—it’s the minimal change I can think of to fix the error. :-) I also benchmarked various schemes for dealing with the image when writing the traversal kernel, and if I remember right, not saving it did turn out to be a bit faster.

Diameter shifting is always going to be slower than a regular build because its refine method is pretty involved. So, I don’t mind if adding the BoxDim reduces performance a bit further. But, we should not allow the BoxDim to affect performance when diameter shifting is off. If you see this causing issues in benchmarks, I would then suggest that we write (or subclass) a different QueryOp. However, the compiler might be smart enough to optimize it out?

jglaser:
You're right, the LJ benchmark with tree and N=1,000,000 is marginally faster when storing a BoxDim and using the master branch of `neighbor` than the `hip` branch which currently uses the patched version of `neighbor`.

| neighbor branch | HOOMD branch | TPS |
| --------------------- | ---------------------- | ------ |
| add_image_to_refine | hip | 198.235 |
| master | fix_diameter_shift_pbc | 203.84 |

I pushed the fix you suggested.

jglaser:
Unit test added (I verified it fails without the fix applied).

PR #541, opened 2019-11-26
Support for AMD GPUs via HIP

jglaser:
## Description

Support AMD GPUs through the HIP portability layer. On NVIDIA platforms, portability is achieved via headers only, which we include as submodules (HIP/hipCUB), i.e. no additional shared library dependencies.

Currently tested with gcc and clang compilers, compiles against HIP in ROCm-2.10 (and against HIP master).

Also, a minor patch to hipcub has been applied in a fork
https://github.com/ROCmSoftwarePlatform/hipCUB/pull/58

We will want to think about adding continuous integration on that platform.

Minor changes:
- eliminate modernGPU, resolves #329 

## Motivation and Context

Frontier (the next OLCF machine) will feature AMD GPUs. Also MacbookPro's have AMD Radeon GPUs, which would be supported, ... *or, maybe not* [ see https://github.com/RadeonOpenCompute/ROCm/issues/262 ]

Related efforts in other GPU software:
LAMMPS https://github.com/lammps/lammps/pull/1590
Kokkos https://github.com/kokkos/kokkos/issues/2309
NAMD ??? [inside information: a "non-zero" effort]

resolves #427 

## How Has This Been Tested?

Standard unit tests. The `mpcd` component is probably not portable because of its use of virtual objects on the device, and is therefore disabled when targeting AMD.

Some dihedral unit tests are still failing with ROCm 2.10, but this is due to a bug in HIP's `hipfuncGetAttributes`:
https://github.com/ROCm-Developer-Tools/HIP/pull/1676

A temporary workaround is to use this custom HIP branch
https://github.com/glotzerlab/HIP/tree/hipfuncgetattributes_revertvectortypes

Additional work to allow compilation with hcc and gcc interop on HIP master **in progress**

https://github.com/ROCm-Developer-Tools/HIP/pull/1690
https://github.com/ROCm-Developer-Tools/HIP/pull/1693
https://github.com/ROCm-Developer-Tools/HIP/issues/1740

HOOMD-blue on AMD GPUs requires **Linux kernel 5.3** at minimum, otherwise intermittent slow performance or hangs are expected due to the following bug in the `amdgpu` driver

https://github.com/torvalds/linux/commit/70df8273ca0cebd2bba443a4444b4944a5929151#diff-5ba3a25a1286bd87904e4f44d248b97f

A minor issue in HIP exists and prevents loading shared libraries containing kernel symbols **after** the GPU context has already been initialized:

https://github.com/ROCm-Developer-Tools/HIP/issues/1741

[...]

Issue #507, opened 2019-11-13
Using JIT for MD pair potentials


vyasr:
## Description

The ability to specify new pair potentials via a JIT kernel rather than having to implement a full new Evaluator.

## Proposed solution

The API for this could be similar to the current `jit.patch`. We could also take inspiration from OpenMM, which uses Sympy to accept arbitrary mathematical expressions and JIT-compile them. The hard part will be ensuring that CUDA-JIT (and potentially HIP-JIT) plays nicely with the rest of our infrastructure.

## Additional context

Josh and I have discussed this possibility before, I'm just creating the issue now so that we can keep track of it.

lyrivera:
I'd be interested in contributing to this.

joaander:
Given the difficulties with LLVM, I'd like to investigate whether a non-LLVM implementation on the CPU could perform well.

PR #549, opened 2019-12-02
jit.patch.user() on GPU

jglaser:
## Description

Enables patch energy in GPU hpmc kernel + NVRTC jit

## Motivation and Context

I plan to use this to speed up free energy calculations of patchy particles, where MC is required because of the use of discontinuous order parameters

Supports patch unions and user-supplied arrays. I'd also like to enable this on AMD if possible.

## How Has This Been Tested?

Standard unit and validation tests.

## Change log

```
- `jit.patch.user()` and `jit.patch.user_union()` now support GPUs via NVRTC
```

## Checklist:


jglaser:
- [ ] add a validation test for `patch.user_union()`

jglaser:
This turned out to take me a little longer than I had hoped. The reason is that CUDA `__device__` functions do not work very well unless inlined. Therefore, the JIT now compiles the entire kernel along with the (inlined) `eval()` function.

vyasr:
This is great! This PR serves as a good testbed to evaluate the viability of #507 as well.

How do you want to sequence these changes relative to #541? Do you want to merge this branch into `hip` before that is merged? It seems perhaps safer to let #541 get merged first and then evaluate this so that PR doesn't get too bloated.

jglaser:
I patched CUB to compile with NVRTC. However, since I do not have write access I couldn't update the git submodule. @joaander Can we talk on how to implement those changes?

jglaser:
I optimized the HPMC acceptance loop further, using multiple threads per particle, while essentially maintaining determinism. This reminds me that there is a theoretical possibility for an infinite loop, which can occur if for some strange reason the acceptance probability changes due to round off error between iterations. I don't think the probability is greater than `machine_eps`, but we could put an infinite loop detection there, in which case all moves would be rejected without generating any harm. Whatever finite value the 'infinity' cutoff is, it will probably generate more false positives than avoid real hangs.

jglaser:
Also, the patch potential evaluation has been aggressively optimized using templated `__launch_bounds__`, which means that several versions of the JIT kernel are compiled, for a range of maximum allowed block sizes `32 <= m <= 1024`. These differ in their register usage, so that we pick the optimal number of registers for every block size (otherwise, the number of threads would be truncated at the maximum block size, rather than adjusting the registers to fit onto the GPU). This optimization adds to the inner-loop parallelism employed for the `user_union` variant of the kernel, which parallelizes over leaf node particles. The number of particles in the leaf nodes can be adjusted for optimum performance, using the `capacity` argument to `user_union.set_params`. Together,  this amounts to about `5*6=30` different kernel templates per JIT potential, all of which get compiled on the fly within a few sec. Only autotuning takes slightly longer (a few hundred time steps), in view of the increased number of possible kernel parameters. However, I find it quite rewarding to see a factor 10 increase in performance after tuning, at least for very complex union shapes.

lyrivera:
I was selected to review this PR, however I am not familiar enough with HOOMD's GPU code paths nor with NVRTC to provide a useful review.

vyasr:
@jglaser @joaander Should I go ahead and start reviewing this now?

jglaser:
I will have to merge next into it, but appreciate feedback

hoomd/AABB.h
```diff
@@ -267,7 +274,11 @@ struct __attribute__((visibility("default"))) AABB
 
         #endif
         }
-    } __attribute__((aligned(32)));
+    }
+#ifndef __HIPCC__
+__attribute__((aligned(32)))
```
joaander: I would hesitate to change ABI's inside and outside GPU compiled code. This is likely to cause memory errors when one code path or the other accesses AABB objects.

hoomd/AABB.h
```diff
@@ -324,6 +335,7 @@ DEVICE inline bool contains(const AABB& a, const AABB& b)
     }
 
 
+#ifndef __HIPCC__
```
joaander: We could use a platform agnostic `min()` so that this method is available in both code paths.

hoomd/hpmc/IntegratorHPMCMonoGPU.cu
```diff
@@ -106,9 +110,8 @@ __global__ void hpmc_shift(Scalar4 *d_postype,
     d_image[my_pidx] = image;
     }
 
-//!< One-thread kernel to accept reject
-__global__ void hpmc_accept(const unsigned int *d_ptl_by_update_order,
-                 const unsigned int *d_update_order_by_ptl,
+//!< Kernel to accept/reject
+__global__ void hpmc_accept(const unsigned int *d_update_order_by_ptl,
```
joaander: How do the changes to `hpmc_accept` change the performance of normal hard particle simulations?

hoomd/GPUFlags.h
```diff
@@ -151,7 +151,7 @@ template<class T> GPUFlags<T>::GPUFlags(const GPUFlags& from) : m_exec_conf(from
     memclear();
 
     // copy over the data to the new GPUFlags
-    this->resetFlags(from.readFlags);
+    this->resetFlags(from.readFlags());
```
vyasr: How did this compile before?

hoomd/RandomNumbers.h
```diff
@@ -191,7 +196,7 @@ DEVICE inline r123::Philox4x32::ctr_type RandomGenerator::operator()()
     {
     r123::Philox4x32 rng;
     r123::Philox4x32::ctr_type u = rng(m_ctr, m_key);
-    m_ctr[0] += 1;
+    m_ctr.v[0] += 1;
```
vyasr: Did the random123 APIs change?

hoomd/hpmc/GPUTree.h
```diff
@@ -148,7 +155,7 @@ class GPUTree
                 unsigned int left_child = getLeftChild(cur_node);
 
                 // is this node a leaf node?
-                if (left_child == OBB_INVALID_NODE)
+                if (left_child == 0xffffffff)
```
vyasr: Why did these all get changed from `OBB_INVALID_NODE`?

hoomd/hpmc/HPMCMiscFunctions.h
```diff
@@ -33,7 +37,11 @@ namespace detail
         #ifdef __HIP_DEVICE_COMPILE__
         return ::min(a,b);
         #else
-        return std::min(a,b);
+        #ifndef __HIPCC__
+        return std::max(a,b);
```
vyasr:
```suggestion
        return std::min(a, b);
```

hoomd/jit/EvalFactory.h
```diff
@@ -36,21 +36,33 @@ class EvalFactory
...
+            *m_alpha_union = h_alpha_union;
             }
 
     private:
         std::unique_ptr<llvm::orc::KaleidoscopeJIT> m_jit; //!< The persistent JIT engine
         EvalFnPtr m_eval;         //!< Function pointer to evaluator
-        float * m_alpha;         // Pointer to alpha array
-        float * m_alpha_union;   // Pointer to alpha array for union
+        float **m_alpha;         // Pointer to alpha array
+        float **m_alpha_union;   // Pointer to alpha array for union
```
vyasr: Are the extra levels of indirection necessary to support pointers on the GPU?

hoomd/jit/patch.py
```diff
@@ -134,12 +134,9 @@ def __init__(self, mc, r_cut, array_size=1, code=None, llvm_ir_file=None, clang_
         # check if initialization has occurred
         hoomd.context._verify_init()
 
-        # raise an error if this run is on the GPU
-        if hoomd.context.current.device.cpp_exec_conf.isCUDAEnabled():
-            hoomd.context.current.device.cpp_msg.error("Patch energies are not supported on the GPU\n");
-            raise RuntimeError("Error initializing patch energy");
+        self.compute_name = "patch"
```
vyasr: Do we need to make this unique if the user specifies multiple patches?

hoomd/jit/patch.py
```diff
@@ -226,6 +245,53 @@ def compile_user(self, array_size_iso, array_size_union, code, clang_exec, fn=No
...
 
+    def wrap_gpu_code(self, code, union = False):
+        R'''Helper function to compile the provided code into a device function
+
+        Args:
+            code (str): C++ code to compile
+            clang_exec (str): The Clang executable to use
+            fn (str): If provided, the code will be written to a file.
+            array_size_iso (int): Size of array with adjustable elements for the isotropic part. (added in version 2.8)
+            array_size_union (int): Size of array with adjustable elements for unions of shapes. (added in version 2.8)
```
vyasr: This docstring is from the `user` function and needs to be updated.

hoomd/jit/jitify_safe_headers.hpp
```diff
@@ -0,0 +1,817 @@
...
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * * Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ * * Redistributions in binary form must reproduce the above copyright
+ *   notice, this list of conditions and the following disclaimer in the
+ *   documentation and/or other materials provided with the distribution.
+ * * Neither the name of NVIDIA CORPORATION nor the names of its
+ *   contributors may be used to endorse or promote products derived
+ *   from this software without specific prior written permission.
```
vyasr: Should we update the credits file in the docs?

hoomd/hpmc/IntegratorHPMCMonoGPU.cu
```diff
@@ -159,16 +202,99 @@ __global__ void hpmc_accept(const unsigned int *d_ptl_by_update_order,
...
+            {
+            // iterate over overlapping neighbors in old configuration
+            float energy_old = 0.0f;
+            unsigned int nneigh = d_nneigh_patch_old[i];
+            bool evaluated = false;
+            for (unsigned int cur_neigh = offset; cur_neigh < nneigh; cur_neigh += group_size)
+                {
+                unsigned int primitive = d_nlist_patch_old[cur_neigh+maxn_patch*i];
+
+                unsigned int j = primitive;
```
vyasr: I don't think you need j, just reuse primitive (or rename it).

hoomd/AABB.h
jglaser: the `#ifdef` was unnecessary, removed

hoomd/AABB.h
jglaser: for now I'd leave it as is, as `merge` is usually called in the context of leaf node updates, which require a serial code path (or atomics) to avoid race conflicts. If we need to use `merge` on the GPU in the future, we can make that modification

hoomd/jit/patch.py
jglaser: done

hoomd/jit/jitify_safe_headers.hpp
jglaser: yes. done.

jglaser:
> * Some of the changes (e.g. in AABB) are unrelated to the JIT changes, right? I found the PR a bit harder to review because some changes seemed to be fixing issues with how these code paths compile on the GPU in general (for HIP), even code paths that were already GPU-enabled.

NVRTC has stricter requirements on what is allowed in JIT compilation and what not. Basically any header included in the JIT code paths needs to be safe, i.e. not include any other headers that are not in the default search path, which is set explicitly, i.e. `/usr/include/` is not by design. Some tweaks to existing header were therefore necessary to be able to use them in JIT code.

> * I'm a little confused by the changes in RandomNumbers. Did the random123 API change, requiring that we access the `ctr_type` through the `.v` attribute?

see pertinent comment above, changes to random123, and #591 

> * What is the point of the IntegratorHPMCMonoGPUJit.inc file? What logic went there? Some of the code seems redundant to the logic for computing energies in IntegratorHPMCMonoGPU.cu, and it's not clear to me why.

the `.inc` actually gets JIT compiled. It checks both the new and the old configuration for every neighbor (not only for the moved particle). This scheme enables parallelism at the level of the energy calculation, but it requires the final summation of the energies to be done at the acceptance stage (in the `.cu` file). The parallel acceptance kernel is novel.

joaander:
Does this branch change the performance for hard particle simulations without patches?

jglaser:
> Does this branch change the performance for hard particle simulations without patches?

I just tested. It does. The 1M hexagon benchmark performance decreases by 10x. The offending commit is bespoke micro-'optimization' using `GPUFlags` (6126c15). Still seeing 2x performance drop after autotuning, investigating. It has do with GPU hints in `ALWAYS_MANAGED` configuration.

jglaser:
The culprit was the memory hint on the expanded cell list. `excell` is currently not optimized for multi-GPU execution, as it is constructed on GPU 0 using the per-device cell lists and then read from multiple GPUs (gather-scatter). The memory hint led the CUDA driver to make wrong assumptions about the memory location. We should revisit the data access pattern, by taking advantage of memory locality and introducing an expanded cell lists that can be split like using a `GPUPartition`, i.e. per particle index. This implies that every particle would store its cell neighbors, not just every cell, at the cost of increased memory usage. We should evaluate the benefits and drawbacks of such an implementation.

jglaser:
there are some performance issues which I still need to investigate.

jglaser:
Hard particle benchmarks on one V100.

| benchmark | N | this branch | next | TPS (2.9.0) |
| -- | -- | -- | -- | -- |
| hexagon | 1,048,576 | **56.3811** | 11.0947 | 56.0481 |
| dodecahedron | 125,000 | 65.2757 | 35.3991 | **126.647** |
| dodecahedron | 4,096 | **827.902** | 502.179 | 455.103 |

and trial moves per second [*MPS* x 10e7]

| benchmark | N | this branch | next | 2.9.0 |
| -- | -- | -- | -- | -- |
| hexagon | 1,048,576 | 22.83 | 4.65 | **40.48** |
| dodecahedron | 125,000 | 3.41 | 1.772 | **8.490** |
| dodecahedron | 4,096 | **1.57** | 0.82 | 1.08 |

Relative to the `next`, all benchmarks receive quite a significant performance bump. The benchmarks with large numbers of particles are about 2x slower than in 2.9.0 / master, which is expected because of the doubled number of overlap checks. Conversely, performance in strong scaling should be much better with this branch.

PR #553, opened 2019-12-08
Fix uninitialized memory

jglaser:
## Description

This PR fixes some problems with uninitialized memory. Likely they only affect the results when GlobalArrays are not memset to zero, which is the case with `ALWAYS_USE_MANAGED_MEMORY`. I used valgrind to track them down.

## Motivation and Context

MPI simulations in HPMC simulations with MPI and `ALWAYS_USE_MANAGED_MEMORY` gave random results on the CPU.

## How Has This Been Tested?

I don't know why existing unit tests have not caught this. I have adjusted the tolerance of the `lj_spheres` validation test just in case (and check for too large fluctuations).

## Change log

```
- Fix uninitialized memory in some locations which could have led to unreproducible results with HPMC in MPI, in particular with `ALWAYS_USE_MANAGED_MEMORY=ON`
```

## Checklist:

### DIFF
--- a/hoomd/Communicator.cc
+++ b/hoomd/Communicator.cc
@@ -12,6 +12,7 @@
 
 #include "Communicator.h"
 #include "System.h"
+#include "HOOMDMPI.h"
 
 #include <algorithm>
 #include <hoomd/extern/pybind/include/pybind11/stl.h>
@@ -1186,6 +1187,32 @@ Communicator::Communicator(std::shared_ptr<SystemDefinition> sysdef,
     m_end.swap(end);
 
     initializeNeighborArrays();
+
+    /* create a type for pdata_element */
+    const int nitems=14;
+    int blocklengths[14] = {4,4,3,1,1,3,1,4,4,3,1,4,4,6};
+    MPI_Datatype types[14] = {MPI_HOOMD_SCALAR, MPI_HOOMD_SCALAR, MPI_HOOMD_SCALAR, MPI_HOOMD_SCALAR,
+        MPI_HOOMD_SCALAR, MPI_INT, MPI_UNSIGNED, MPI_HOOMD_SCALAR, MPI_HOOMD_SCALAR, MPI_HOOMD_SCALAR,
+        MPI_UNSIGNED, MPI_HOOMD_SCALAR, MPI_HOOMD_SCALAR, MPI_HOOMD_SCALAR};
+    MPI_Aint offsets[14];
+
+    offsets[0] = offsetof(pdata_element, pos);
+    offsets[1] = offsetof(pdata_element, vel);
+    offsets[2] = offsetof(pdata_element, accel);
+    offsets[3] = offsetof(pdata_element, charge);
+    offsets[4] = offsetof(pdata_element, diameter);
+    offsets[5] = offsetof(pdata_element, image);
+    offsets[6] = offsetof(pdata_element, body);
+    offsets[7] = offsetof(pdata_element, orientation);
+    offsets[8] = offsetof(pdata_element, angmom);
+    offsets[9] = offsetof(pdata_element, inertia);
+    offsets[10] = offsetof(pdata_element, tag);
+    offsets[11] = offsetof(pdata_element, net_force);
+    offsets[12] = offsetof(pdata_element, net_torque);
+    offsets[13] = offsetof(pdata_element, net_virial);
+
+    MPI_Type_create_struct(nitems, blocklengths, offsets, types, &m_mpi_pdata_element);
+    MPI_Type_commit(&m_mpi_pdata_element);
     }
 
 //! Destructor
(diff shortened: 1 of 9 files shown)


joaander:
This looks fine, but I'd prefer to keep the 0.99 confidence interval. With a 5% chance of a false-negative failure, 4 validation test runs, and say 4 PRs per week that gives a 1 - (1-0.05)**16
 = 55% chance of just the `lj_spheres` test failing at least once/week (not to mention all the other tests being checked at 95% confidence). With a 99% confidence, that drops to 14%. We may need even larger confidence intervals in the future when we add more validation tests just to get them all to pass in the same run.

joaander:
Case in point: https://dev.azure.com/glotzerlab/hoomd-blue/_build/results?buildId=985&view=logs&j=59990577-4998-5e26-ab13-e05396045cf6&t=75d9f71b-471c-5739-4b79-4fef151e8483&l=447
```
  File "/__w/1/s/hoomd/hpmc/validation/lj_spheres.py", line 152, in test_low_density_clusters_depletants
    self.run_statepoint(Tstar=8.50E-01, rho_star=9.00E-03, mean_Uref=-9.3973E-02, sigma_Uref=1.29E-04,
  File "/__w/1/s/hoomd/hpmc/validation/lj_spheres.py", line 137, in run_statepoint
    self.assertLessEqual(math.fabs(mean_U - mean_Uref), ci*sigma_diff)
AssertionError: 0.002380535694020705 not less than or equal to 0.0019545574894953115
```

hoomd/Communicator.cc
```diff
@@ -1186,6 +1187,32 @@ Communicator::Communicator(std::shared_ptr<SystemDefinition> sysdef,
     m_end.swap(end);
 
     initializeNeighborArrays();
+
```
mphoward: @jglaser Do I also need to implement these changes for the mpcd Communicator?

vyasr:
Under which circumstances are HPMC runs precisely binary reproducible, only single-core CPU runs? Would it make sense to restrict which configurations have the validation test run to avoid the false negatives, or is that too restrictive in what gets validated?

jglaser:
We can feel free to loosen the tolerance for the unit test. The main thing I added in that test was to check the error on the mean, to ensure it is small. What happened during testing was that the fluctuations were huge because of a bug I was tried to detect, but the unit test passed nevertheless since it assumed the 'error' was real and therefore the test result was within the combined confidence interval (error_ref + error_measure).

jglaser:
HPMC simulations are **exactly** reproducible on the CPU. On the GPU and in `next`, hard shape simulations are **perfectly** deterministic (no floating point operations there), and even in the JIT branch they are **essentially deterministic**, meaning that the only non-determinism arises from the order of energy summations **and** the random number generator for the acceptance criterion generating a random number that is exactly hitting a rounded-off float value (probability ~< machine epsilon). In practice, I find that the simulations are always deterministic. All of this doesn't require sorting the cell list, so we should probably eliminate the option in HPMC.

joaander:
Simulations are reproducible with respect to the same compiled binary of hoomd reproducing the same output given the same input when run on the same hardware. However, due to slight variations in the order of floating point operations caused by any change to the code, different compilers, runtime code generation for different GPU architectures, etc... different builds can produce different outputs given the same inputs. Thus one cannot pick a specific floating point value or extremely tight tolerance to compare with on all builds. It may be better to randomize seeds in the validation tests so that repeated tests runs truly sample different random states and we are not fooled by "mostly" reproducible values.

For the purposes of finding an bug that causes deviations smaller than the confidence interval, one would need to obtain tighter absolute tolerances by running longer tests. In this case we are also limited by the error interval around our reference data.

Significantly increasing test length is not really an option on CI platforms. Should we write a validation test suite that can be submitted manually to a cluster? We could keep the automated tests in CI short to check for gross violations, but the longer tests would be needed to check for subtle errors.

The new pytest platform we are implementing for the new API would allow the same test code to be used both in CI and on the cluster. We could add an argument to it to increase the length of the validation tests for use in cases like this. E.g.:
``` 
pytest --pyargs hoomd --validate   # run CI validation tests
pytest --pyargs hoomd --validate=16  # run tests 16x longer

mpirun -n 2 pytest-openmpi.sh -x --pyargs hoomd --validate=16 # run tests with MPI
```

hoomd/Communicator.cc
jglaser: This is recommended, at least if you are communicating arrays of structs, particular non-packed ones, such as combinations of Scalar3's and Scalar4's. The compiler is free to generate holes in the struct and align data types, such as at 16 or 32 byte boundaries. Trying to use gcc's `__attribute__((packed))` results in a lot of warnings. When you try to pass a non-contiguous struct using the `MPI_BYTE` data type, valgrind reports an uninitialized memory error inside the MPI library. This is the perhaps (one of the?) reasons why MPI has structured data types. Fun fact, there are also security concerns associated with communicating uninitialized memory regions over networks.

PR #556, opened 2019-12-09
Fix potential issues with custom MPI types in MPCD

mphoward:
## Description

Use custom MPI type for communication of MPCD particles, as proposed by Jens in PR #553.

The CellCommunicator also uses a custom struct for one MPI op. This was not easy to fix given the current template design, so I changed that struct to a contiguous type, which should still be OK. It adds 4B of message per cell, but if I remember right, latency is more of an issue than bandwidth anyway.

## Motivation and Context

No errors have been reported, so this is preventative maintenance.

## How Has This Been Tested?

Existing unit tests still pass, not sure of a case where it will fail. (Maybe running through valgrind?)

## Change log

```
* Fix potential memory-management issue in MPI for migrating MPCD particles and cell energy.
```

## Checklist:

### DIFF
--- a/hoomd/Communicator.cc
+++ b/hoomd/Communicator.cc
@@ -1211,8 +1211,13 @@ Communicator::Communicator(std::shared_ptr<SystemDefinition> sysdef,
     offsets[12] = offsetof(pdata_element, net_torque);
     offsets[13] = offsetof(pdata_element, net_virial);
 
-    MPI_Type_create_struct(nitems, blocklengths, offsets, types, &m_mpi_pdata_element);
+    MPI_Datatype tmp;
+    MPI_Type_create_struct(nitems, blocklengths, offsets, types, &tmp);
+    MPI_Type_commit(&tmp);
+
+    MPI_Type_create_resized(tmp, 0, sizeof(pdata_element), &m_mpi_pdata_element);
     MPI_Type_commit(&m_mpi_pdata_element);
+    MPI_Type_free(&tmp);
     }
 
 //! Destructor
--- a/hoomd/mpcd/CellThermoTypes.h
+++ b/hoomd/mpcd/CellThermoTypes.h
@@ -72,25 +72,21 @@ struct CellVelocityPackOp
 
 struct CellEnergyPackOp
     {
-    typedef struct
-        {
-        double energy;
-        unsigned int np;
-        } element;
+    typedef double2 element;
 
     DEVICE element pack(const double3& val) const
         {
         element e;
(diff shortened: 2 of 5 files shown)


mphoward:
Something is causing the validation tests to fail. I'll investigate.

mphoward:
I have no idea why this fails spectacularly on the GPU (it seems even CPU parts of the communicator unit tests crash) but Jens' branch works, as my changes were pretty much copy-paste. If I build with CUDA disabled, it also works correctly.

@jglaser would you be able to look at this changeset to help figure out what's wrong? It seems like there could be some underlying memory management issue with the send/receive buffers? (Maybe in the use of a GPUVector?)

jglaser:
Maybe we are missing the final padding? See the discussion of `MPI_UB` here
https://www.mpich.org/static/docs/v3.1.x/www3/MPI_Type_struct.html

mphoward:
> Maybe we are missing the final padding? See the discussion of `MPI_UB` here
> https://www.mpich.org/static/docs/v3.1.x/www3/MPI_Type_struct.html

Winner! This fixes the issue.

mphoward:
It appears this is not a portable solution, though, since `MPI_UB` and `MPI_Type_struct` have been deprecated:
https://www.open-mpi.org/faq/?category=mpi-removed#mpi-1-mpi-lb-ub

I'll try a workaround.

mphoward:
The failing unit test is now in the regular MPI communicator. I don't know if this a spurious failure, or if that struct -> MPI type needs the same patch as mine.

jglaser:
Yes. Can I commit to this branch?

mphoward:
Sure, feel free!

I'm not sure if both MPI types actually need to be "committed" (since I commit and then free the `tmp` one anyway), so if you want to play around with that as well, go ahead!

Also, this probably belies an ignorance of what the MPI library is doing under the hood, but are you sure that its necessary to use these custom types? If we are using `sizeof` to set the extent of the struct, how is that different from sending a certain number of bytes based on `sizeof` that we have already forced to lie in contiguous memory?

jglaser:
I just applied the same change to `Communicator.cc`. I think the change set is correct, I ran the `lj_spheres.py` test through valgrind again and did not observe any uninitialized memory warnings. I don't exactly know what the MPI library is doing internally, but it seems smart enough to be able to resize a custom defined struct data type, while not forgetting it's offsets (which are not necessarily the prefix sum of the `sizeof`'s of the individual data types)

joaander:
Thanks!

Issue #555, opened 2019-12-09
eliminate quermass changes in `next`


jglaser:
The HPMC quermass feature in `next` is incomplete and unneeded. It only leads to validation test failures and needs to be removed.

joaander:
To avoid conflicts, let's start this after merging *hip* into *next* and then *next* up into *hpmc_integrator_api* and/or *feature/new-object-API*. Then we can remove the quermass code on that branch.

PR #560, opened 2019-12-12
Use header-only neighbor library

mphoward:
## Description

I have updated neighbor as a header-only library. It should now be compiled entirely inside GPU code, so I have written some wrappers to avoid the need to explicitly template kernels, which is error prone and annoying.

neighbor is added by submodule to HOOMD. Because proper installation with CMake submodules got messy, I added a new target `HOOMD::neighbor` that is an interface library that points at the `extern` headers in the build / install locations. neighbor's primary dependencies are CUB / hipCUB, which are already supplied by hoomd.

neighbor 0.3.0 now supports AMD GPUs via hipper, which is another dependency. I have added hipper 0.1.2 to HOOMD as a submodule, which is similarly given a target `HOOMD::hipper`. neighbor will use this version of hipper because I have aliased it into the target `hipper::hipper` during setup.

To control hipper, we need to define `HIPPER_CUDA` to make a CUDA build or `HIPPER_HIP` to make a HIP build. I used the `HIP_PLATFORM` variable to select for this, unless there is a smarter way to do so. I set it the definition on the `HOOMD::hipper` target, so I think this should propagate to anybody that links against it (e.g., when we use hipper elsewhere).

## Motivation and Context

This should simplify the compilation and linking of the neighbor library. It also helps move toward using hipper throughout HOOMD by supplying the submodule.

## How Has This Been Tested?

It should be covered by the existing tests. @jglaser and I have also tested both neighbor and hipper on NVIDIA and AMD platforms.

## Change log

```
* Use neighbor 0.3.0 library for ``nlist.tree`` on GPUs. This library now supports both NVIDIA and AMD GPUs.
```

## Checklist:


hoomd/md/CMakeLists.txt
```diff
@@ -274,7 +274,7 @@ else()
     target_link_libraries(_md PUBLIC _hoomd)
 endif()
 if (ENABLE_CUDA)
-    target_link_libraries(_md PUBLIC _neighbor)
+    target_link_libraries(_md PRIVATE neighbor::neighbor)
```
mphoward: Something feels funny about this, but I couldn't get CMake to run properly without using `PRIVATE`.

joaander: With public `PUBLIC` cmake will add the neighbor include directories to any target that links `_md`. This is needed if `neighbor` headers are included in any of md's header files. 

With `PRIVATE`, it only adds the neighbor include directories to `_md`. This is fine if `neighbor` headers are only included in md's .cc files.

joaander:
But, then I think the headers are only copied into extern and neighbor also internally includes <cub/cub.cuh>, so both of these break plugins including hoomd's headers. I don't know enough about modern CMake though, so I could use some help @joaander @jglaser.

We aren't using modern CMake for CUB on next, yet. We always explicitly include it as "hoomd/extern/cub/cub/*". To support this in a modern way, we should instead implement a FindCUB that knows how to find CUB in extern/ both from source builds and from plugins. This would create a target that we can link to to provide the proper include path so that `<cub/cub.cuh>` works.

hoomd/md/CMakeLists.txt
mphoward: OK, thanks! That makes sense. I originally used `PUBLIC` because I did have some of the neighbor headers in `NeighborListGPUTree.cuh`. But, then I got a CMake error about not having the neighbor target in the export set for hoomd, so I switched to `PRIVATE`--this is wrong for that setup.

I think it makes alot of sense to hide the internals of `neighbor` from the public API--users just want the neighbor list objects and not the underlying implementation. I refactored so that `neighbor` is now only included in `NeighborListGPUTree.cu`, which means that `PRIVATE` should be correct.

mphoward:
> We aren't using modern CMake for CUB on next, yet. We always explicitly include it as "hoomd/extern/cub/cub/*". To support this in a modern way, we should instead implement a FindCUB that knows how to find CUB in extern/ both from source builds and from plugins. This would create a target that we can link to to provide the proper include path so that `<cub/cub.cuh>` works.

This sounds like a good plan long term, but now that I have hidden neighbor from the public API, it might not be required for this PR. I currently set `extern/cub` as `CUB_INCLUDE_DIR` before adding the neighbor subdirectory, so FindCUB in neighbor uses this as the location to make the target `CUB::CUB`, which is linked as an interface library.

mphoward:
- [ ] Point submodule back to `hoomd-3` branch after neighbor PR #9 is merged.
- [ ] Squash to avoid pointing at too many different versions of neighbor.

mphoward:
It looks like `spheres_implicit` validation test is failing. Is this expected on this branch? I assume that is HPMC and so not using the MD neighbor lists.

jglaser:
> It looks like `spheres_implicit` validation test is failing. Is this expected on this branch? I assume that is HPMC and so not using the MD neighbor lists.

Yes, don't worry about it (https://github.com/glotzerlab/hoomd-blue/issues/555)

mphoward:
Pending any benchmarks we would like to run, I think this branch is just about ready. All the failing validation tests are `sphere-implicit`, noted by Jens as being due to #555.

The only outstanding question is what to do about installing neighbor. I believe that its headers are currently copied into `include/hoomd/extern` by a glob. Technically, they do not need to be copied because they are only used in a `.cu` file. Do we want this to be the behavior?

joaander:
> The only outstanding question is what to do about installing neighbor. I believe that its headers are currently copied into `include/hoomd/extern` by a glob. Technically, they do not need to be copied because they are only used in a `.cu` file. Do we want this to be the behavior?

I have no strong preference on this. I trust your judgement.

Sorry for creating the conflicts, but we had too many outstanding large branches. I merged the hip branch to free up the blocking work. This PR will need to wait and get updated once we have a working HIP or `Runtime.h` version of the neighbor library.

mphoward:
No problem, we can resolve those once the hip-enabled neighbor is ready.

mphoward:
@joaander This is getting close to being ready. I need to resolve a few compilation issues on AMD, which I unfortunately only seem to be able to test through the CI platform because I get a linker error when I run the container locally. Todo:

- [ ] Fix AMD compilation.
- [ ] Squash, merge, and tag this neighbor version.
- [ ] Update submodule and squash commits so that they don't point to dead neighbor commits.

@jglaser Can you check if the CMake to include neighbor is done correctly? (I'm pretty confused by the new CMake setup & variable nomenclature to enable HIP vs. CUDA vs. either GPU.) In CUDA, `CUB_INCLUDE_DIR` seems to be set to some generator expression. The neighbor CMake takes this, which seems to work. In HIP, I tried forcing `HIPCUB_ROOT` to be `HIP_INCLUDE_DIR`, but I'm not sure that's actually detecting correctly. The CI platform seems to take the hipcub in `/opt/rocm/include`--is this right?

joaander:
Thanks, @mphoward! This is looking really good. How will we handle `hipper_runtime.h` files for **HOOMD** and **neighbor** in #587. Should we replicate the file, just have HOOMD include the one from the **neighbor** submodule, or what?

mphoward:
I think HOOMD should probably use its own copy of the runtime, either directly (since it is single-header) or as submodule. (I will stick `hipper_runtime.h` in a repo shortly.) That should be easier to read in future and keeps the implementation details of neighbor isolated from HOOMD.

joaander:
OK, should we use a different namespace to differentiate between the runtime header for the two packages then. I see potential problems with the same names defined in two different headers. At a minimum we could agree on the preprocessor defines that signal the build options for the two headers.

mphoward:
Hmm, good point. If the files are exactly the same, then if I put an include guard in (instead of the once pragma), there should be no problem having two copies, since only one file will ever be included. This is bad if they ever diverge from each other.

Alternatively, we could wrap one or both in a namespace (that's what CUB does for thrust--`CUB_NS_PREFIX` and `CUB_NS_POSTFIX`). Or, we could just use the supplied one directly (`#include "neighbor/hipper_runtime.h"`), with the only potential confusion being that this lives in the `hipper` namespace rather than `neighbor` namespace. What do you think would be best?

joaander:
If the runtime header is not going to remain fixed once we complete this, maybe it doesn't matter that we keep two copies of the same file that use inclusion guards instead of `#pragma once`. However, if it is going to see continued updates, than the logistics of updating it in two places might get annoying.

Could we solve this with a include path? If both HOOMD and neighbor used `#include <hipper_runtime.h>` then neighbor could use its copy when used by itself. When used inside HOOMD, the include path could be set such that it finds the one provided by HOOMD first.

@jglaser what do you think?

mphoward:
Yeah, I'm not planning to make many updates, but I'm nervous that I might need to do maintenance because HIP is under so much development, at least at first.

I think include path is a good solution. `neighbor` can try to find the header (which can be forced to a location by `hoomd`) and then fallback on the internal copy for its own builds / installs.

@jglaser do you think that would work?

joaander:
Let's move forward with that plan.

mphoward:
Great! I think we are good to go, and we can revisit if there's anything wrong when you get the new rocm images.

You might want to merge-squash this since some commits point to dead neighbor commits---up to you.

joaander:
Thanks for the suggestion. Squash merging is a good idea - the changes from this PR are simple enough.

Issue #572, opened 2020-01-02
Dynamic Bonding in MD


atravitz:
## Description

I'm proposing support for dynamic bonds in the md package, allowing for bonds to be broken and formed between particles at each time step. 

## Proposed solution
The dynamic bond module will be an updater within md. If particles are closer than`r_cut`, a random number is generated and there is a probability of `prob_form` that a bond will be formed between the two particles. Likewise, at every time step, there is `prob_break` probability that an existing bond will break. 

```python
    nl = md.nlist.tree()
    dybond = md.update.dynamic_bond(group.all(), nlist=nl, seed=123, period=1)                                                                                        
    dybond.set_params(r_cut=2.0, bond_type='harmonic', prob_form=0.6, prob_break=0.3)   
```
## Additional context

Currently, the probabilities `prob_break` and `prob_form` are defined as the probability that *one* bridge breaks or forms at a given time step, and is evaluated for every particle neighbor pair. 

I'd appreciate your thoughts on the following:

- How, or if, we should allow for multiple bonds breaking/forming at each time step,
- I have a branch where `prob_form` and `prob_break` can be tables, implemented similar to `md.bond.table` or `md.pair.table`. Should this be an option?
- Should we have a limit on the number of bonds per particle? This could be defined by particle type.

## Developer

I have a working version that I can make a PR with, but would like feedback on appropriate features and API. I'm tagging @joaander and @mikemhenry since we've already had conversations about this.

mikemhenry:
Also tagging @amburan

InnocentBug:
I like the discussion about dynamic bonds.
My application of dynamic bonds is probably too complicated to fit into this approach:
in short the creation/deletion of bonds depends on the surrounding (neighbour list access) the other scheme allows the bond anchor points to slide along existing bonds of other types.

However, I would really like to see something like this integrated into HOOMD-blue. I would definitely use as a reference of how to actually add and remove bonds efficiently.

InnocentBug:
I think, making it sensitive to the particles involved in the process is definitely helpful.

InnocentBug:
One of my current questions would be: how is that implemented with multi-GPU/MPI support.
The current framework requires to delete/create bonds as a collective MPI operation.
This of course can be a massive slow down and doesn't scale.
I haven't figured out a better way yet.
Any ideas would be great.

joaander:
[...]

PR #575, opened 2020-01-23
Dynamic bonding

atravitz:
## Description

I and several other users would like to be able to add and remove bonds between particles during HOOMD simulations. I've implemented a basic dynamic bonding feature that currently has only been tested for serial simulations on CPUs, with the goal of adding GPU capability. Adding dynamic bonding functionality for MD. Dynamic bonding is implemented as an updater, for example:

```python
nl = md.nlist.tree()
dy_bond = md.update.dynamic_bond(group=group.all(), nlist=nl, seed=1, period=1)
dy_bond.set_params(r_cut=2.0, bond_type='harmonic', prob_form=0.7, prob_break=0.4)
```

Other features that I have included on a separate branch (more specific for my own use case) that I'd be happy to include in this feature are:
   - passing in `prob_form` and `prob_break` as a table 
   - including the time step as a parameter
   - initializing each particle with a finite number of "possible bonds," such that the total number of possible bonds + current bonds in the system remains constant

Features that might be worth including but I have not included (open to suggestions):
   - adding more than one bond between particles per time step
   - adding multiple `dynamic_bond` updaters for different bond types. I have a single bond type implemented so I imagine this would be straightforward, but I'd need to write some tests.

Things I would like suggestions on how to implement are:
   - currently I am assuming simulations start with no bonds, I'd like to include in my "possible bonds" tracking which particles are initialized with bonds
   - `r_cut` is the range in which a particle will search for neighbors to bond to or break from, but I'm not sure if this will cause problems for the breakage calculations if a bonded particle is further away than `r_cut`
   - any ideas for more thorough testing, edge cases
   - getting this working on GPUs

## Motivation and Context

Several users have wanted this feature to simulate systems such as telechelic polymer and colloid mixtures,  polymer cross-linking, and epoxies.

Resolves: #572 

## How Has This Been Tested?

Python and C++ level tests are included but are not complete.

## Change log

`hoomd.md.dynamic_bond()` allows for bonds to be broken and formed during simulations.

## Checklist:


joaander:
@atravitz Do you have any plans to update this for v3 and finish this PR? Or should we close this PR?

PR #584, opened 2020-02-07
Support MSVC hoomd-blue

BinWang0213:
## Description

This is a trial of making hoomd-blue support in windows

Visual studio project can be found @ https://github.com/BinWang0213/hoomd-blue-MSVC

## Major issue

1. _upp11_ is not support in windows and the original author is no longer maintain it.
 workaround: using Catch2

2. _HIP_ library is not support for MSVC yet. 
workaround: stay in hoomd-blue 2.9

3. segmentation fault when release memory @ https://github.com/glotzerlab/hoomd-blue/blob/df7970121b19bc4f8674348ab3241055ac87153b/hoomd/GPUArray.h#L165
workaround: comment out this line

4. BondedGroupData.cu can not be compiled when  including moderngpu headers
workaround: Not found yet.

4.  __attribute__((visibility("default"))) issue
workaround: just remove this when compiled in MSVC
possible solution not have try it yet: 
```
#if defined(_MSC_VER)
#  define OWL_DLL_EXPORT __declspec(dllexport)
#  define OWL_DLL_IMPORT __declspec(dllimport)
#elif defined(__clang__) || defined(__GNUC__)
#  define OWL_DLL_EXPORT __attribute__((visibility("default")))
#  define OWL_DLL_IMPORT __attribute__((visibility("default")))
#else
#  define OWL_DLL_EXPORT
#  define OWL_DLL_IMPORT
#endif
```

5. Linux only headers
#include <execinfo.h>
#include <cxxabi.h>
#include <dlfcn.h>
#include <sys/time.h>
#include <unistd.h>

See commit for detailes

## How Has This Been Tested?

Basically, I started with the fundenmetal classes, such as GPUArray.h, GlobalArray.h  by defining some MSVC header or functions to make HOOMD-Blue capatiable for MSVC. Then I ran the unit test to check corrsponding corrections.

## Checklist:

### DIFF
--- a/hoomd/ClockSource.h
+++ b/hoomd/ClockSource.h
@@ -24,16 +24,87 @@
 
 #include <iostream>
 
+#ifdef __GNUC__
 #include <sys/time.h>
 #include <unistd.h>
+#endif
+#ifdef _WIN32
+#include <Windows.h>
+#include <time.h>
+#include <io.h>
+#endif
 
 #include <hoomd/extern/pybind/include/pybind11/pybind11.h>
 
+
+#ifdef _WIN32
+inline void usleep(__int64 usec)
+{
+	HANDLE timer;
+	LARGE_INTEGER ft;
+
+	ft.QuadPart = -(10 * usec); // Convert to 100 nanosecond interval, negative value indicates relative time
+
+	timer = CreateWaitableTimer(NULL, TRUE, NULL);
+	SetWaitableTimer(timer, &ft, 0, NULL, NULL, 0);
+	WaitForSingleObject(timer, INFINITE);
+	CloseHandle(timer);
+}
+
+
+#if defined(_MSC_VER) || defined(_MSC_EXTENSIONS)
+#define DELTA_EPOCH_IN_MICROSECS  11644473600000000Ui64
+#else
+#define DELTA_EPOCH_IN_MICROSECS  11644473600000000ULL
+#endif
+
+struct timezone
+{
+	int  tz_minuteswest; /* minutes W of Greenwich */
+	int  tz_dsttime;     /* type of dst correction */
+};
+
+inline int gettimeofday(struct timeval *tv, struct timezone *tz)
+{
+	FILETIME ft;
+	unsigned __int64 tmpres = 0;
+	static int tzflag = 0;
+
+	if (NULL != tv)
+	{
+		GetSystemTimeAsFileTime(&ft);
+
+		tmpres |= ft.dwHighDateTime;
+		tmpres <<= 32;
+		tmpres |= ft.dwLowDateTime;
+
+		tmpres /= 10;  /*convert into microseconds*/
+		/*converting file time to unix epoch*/
+		tmpres -= DELTA_EPOCH_IN_MICROSECS;
+		tv->tv_sec = (long)(tmpres / 1000000UL);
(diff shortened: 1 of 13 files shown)


joaander:
Thanks for starting this. Hopefully we can keep the changes to a minimum.

Regarding your initial comments:
1. I really do not want to rewrite every C++ unit test in HOOMD again. We maintain our own fork of upp11. Feel free to make changes to that to support MSVC compilation. The code is very minimal and should be easy to update.

2. It is unfortunate that HIP is not supported on MSVC. We have some plans to implement our own wrapper header around CUDA and HIP which would work around this problem.

Perhaps you could start on the next branch by porting HOOMD to MSVC with ENABLE_GPU=off and then later add GPU support once we complete the new abstraction layer?

3. This is strange. Memory must be freed to avoid memory leaks, so commenting out the free is not a solution. What conditions lead to the seg fault?

4. We no longer user modernGPU on the next branch. This was removed at the same time we added HIP.

5. I would much prefer the single ifdef solution you propose over adding ifdef's to every single class declaration.

joaander:
Adding @bdice as he has Windows experience.

hoomd/ClockSource.h
```diff
@@ -24,16 +24,87 @@
...
+#include <Windows.h>
+#include <time.h>
+#include <io.h>
+#endif
 
 #include <hoomd/extern/pybind/include/pybind11/pybind11.h>
 
+
+#ifdef _WIN32
+inline void usleep(__int64 usec)
```
joaander: A cleaner solution would be to re-implement the ClockSource.h methods using the cross-platform C++11 chrono (https://en.cppreference.com/w/cpp/chrono)

hoomd/HOOMDMath.h
```diff
@@ -22,8 +22,33 @@
 #include <hoomd/extern/pybind/include/pybind11/pybind11.h>
 #endif
 
+#ifdef WIN32
```
bdice: From earlier experience adding Windows compatibility for the GSD format, you must use `_WIN32` instead of `WIN32` for compatibility. CMake defines `WIN32` but the better way to detect is `_WIN32`. Let’s be consistent about that everywhere.
```suggestion
#ifdef _WIN32
```

hoomd/MemoryTraceback.cc
```diff
@@ -90,6 +98,7 @@ void MemoryTraceback::outputTraces(std::shared_ptr<Messenger> msg) const
     msg->notice(2) << "Actual allocation sizes may be larger by up to the OS page size due to alignment." << std::endl;
     msg->notice(2) << "List of memory allocations and last " << MAX_TRACEBACK-1 << " functions called at time of (re-)allocation" << std::endl;
 
+#ifdef __GNUC__
```
bdice: For features that specifically break on Windows, I would prefer to detect “not Windows” instead of `__GNUC__`. I do not know the full compiler and platform compatibility of `__GNUC__` but I suspect there may be compilers that work with HOOMD and do not have that definition.

joaander:
@BinWang0213 do you have plans to continue working on this?

BinWang0213:
@joaander This is a low priority task for me now. So I will not working on this recently.

joaander:
OK, no worries. I'm going to close this PR for now to keep the list of active PRs clean. Feel free to reopen if you pick this pack up.

Issue #583, opened 2020-02-05
GPU Script Fails with Convexpolyhedra Integrator


patricklawton:
## Description

When running on the GPU (issue does not occur with the CPU), HPMC simulation crashes for unknown reason with gsd dump. Tested the script without the gsd dump and it worked fine. 

## Script

```python
import hoomd
from hoomd import hpmc
from utils import shapetools
import euclid
from euclid.FreudShape import ConvexPolyhedron

c_val = (2 * ((-0.66) + 1) + 1)
particle_vertices = shapetools.unitVolumeVerts(
    a=1, b=2, c=c_val)
freud_shape = ConvexPolyhedron(points=particle_vertices)
particle_radius = freud_shape.getCircumsphereRadius()

dev = hoomd.device.GPU()
hoomd.context.initialize('', dev)
system = hoomd.init.create_lattice(hoomd.lattice.sc(2*particle_radius, type_name='A'), n=[5,5,5])
mc = hpmc.integrate.convex_polyhedron(seed=415236, d=0.3)
mc.shape_param.set('A', vertices=particle_vertices)

gsd = hoomd.dump.gsd(
    filename='test.gsd', group=hoomd.group.all(), period=1000, phase=0, overwrite=False)
gsd.dump_state(mc)

hoomd.run(1e6)
```

## Output

HOOMD-blue v2.9.0-667-g524380ccf GPU [CUDA] (10.2) DOUBLE HPMC_MIXED MPI SSE SSE2 
Compiled: 02/05/2020
Copyright (c) 2009-2019 The Regents of the University of Michigan.
HOOMD-blue is running on the following GPU(s):
 [0]          Quadro K6000  15 SM_3.5 @ 0.901 GHz, 12201 MiB DRAM, DIS
-----
You are using HOOMD-blue. Please cite the following:
* J A Anderson, C D Lorenz, and A Travesset. "General purpose molecular dynamics
  simulations fully implemented on graphics processing units", Journal of
  Computational Physics 227 (2008) 5342--5359
* J Glaser, T D Nguyen, J A Anderson, P Lui, F Spiga, J A Millan, D C Morse, and
  S C Glotzer. "Strong scaling of general-purpose molecular dynamics simulations
  on GPUs", Computer Physics Communications 192 (2015) 97--107
-----
-----
You are using HPMC. Please cite the following:
* J A Anderson, M E Irrgang, and S C Glotzer. "Scalable Metropolis Monte Carlo
[...]

PR #586, opened 2020-02-08
Reinstate streams for HPMC kernels

patricklawton:
## Description

Implements streams for HPMC kernels, to restrict concurrent access to managed memory regions (shape parameters). This reinstates an older implementation which had been inadvertently removed.

## Motivation and Context

The issue only occurred on pre-Pascal GPUs (Kepler) which do not support concurrent managed access.

Resolves: #583

## How Has This Been Tested?

offline, and standard unit tets

## Change log

N/A

## Checklist:

### DIFF
--- a/hoomd/ManagedArray.h
+++ b/hoomd/ManagedArray.h
@@ -132,11 +132,12 @@ class ManagedArray
 
         #ifdef ENABLE_HIP
         //! Attach managed memory to CUDA stream
-        void set_memory_hint() const
+        void attach_to_stream(hipStream_t stream) const
             {
             if (managed && ptr)
                 {
                 #if defined(__HIP_PLATFORM_NVCC__) && (CUDART_VERSION >= 8000)
+		cudaStreamAttachMemAsync(stream, ptr, 0, cudaMemAttachSingle);
                 cudaMemAdvise(ptr, sizeof(T)*N, cudaMemAdviseSetReadMostly, 0);
                 #endif
                 }
--- a/hoomd/hpmc/ComputeFreeVolumeGPU.h
+++ b/hoomd/hpmc/ComputeFreeVolumeGPU.h
@@ -78,6 +78,8 @@ class ComputeFreeVolumeGPU : public ComputeFreeVolume<Shape>
         std::unique_ptr<Autotuner> m_tuner_free_volume;     //!< Autotuner for the overlap/free volume counter
         std::unique_ptr<Autotuner> m_tuner_excell_block_size;  //!< Autotuner for excell block_size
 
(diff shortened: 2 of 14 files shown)


hoomd/hpmc/IntegratorHPMCMonoGPU.h
```diff
@@ -368,11 +373,15 @@ IntegratorHPMCMonoGPU< Shape >::IntegratorHPMCMonoGPU(std::shared_ptr<SystemDefi
         CHECK_CUDA_ERROR();
         }
     #endif
+
+    hipStreamCreate(&m_stream);
```
jglaser: we'll have to do this for every device in the execution configuration

hoomd/hpmc/ComputeFreeVolumeGPU.h
```diff
@@ -131,6 +133,7 @@ ComputeFreeVolumeGPU< Shape >::ComputeFreeVolumeGPU(std::shared_ptr<SystemDefini
 template<class Shape>
 ComputeFreeVolumeGPU<Shape>::~ComputeFreeVolumeGPU()
     {
+    hipStreamDestroy(m_stream);
```
jglaser: we'll need to match this with a `hipStreamCreate()` (for every device)

hoomd/ManagedArray.h
```diff
@@ -132,11 +132,12 @@ class ManagedArray
 
         #ifdef ENABLE_HIP
         //! Attach managed memory to CUDA stream
-        void set_memory_hint() const
+        void attach_to_stream(hipStream_t stream) const
             {
             if (managed && ptr)
                 {
                 #if defined(__HIP_PLATFORM_NVCC__) && (CUDART_VERSION >= 8000)
+		cudaStreamAttachMemAsync(stream, ptr, 0, cudaMemAttachSingle);
```
joaander: This line (and many othres in this PR) appears to have inconsistent formatting. Are you using hard tabs? You should indent only with spaces and match the indentation of the other lines in the block or argument list.

joaander:
Are there plans to finish this up soon or should we just drop support for Kepler?

jglaser:
I can finish this but it would be a kludgy solution. I'd rather like to use streams for performance, not correctness. I am in favor of dropping support for Kepler, especially as HPMC will run slow on that hardware anyway and it's thus not a good platform to advertise GPU MC.

joaander:
I think we can certainly drop support for Kepler in v3.0

joaander:
We will drop support for Kepler and Maxwell in 3.0: #608

PR #587, opened 2020-02-12
GPU Runtime Header + Tools

schwendp:
## Description
Added search and replace tools to change HIP and/or CUDA functions, typename. and macros into new gpu namespace APIs

## Motivation and Context

As described in this [link,](https://github.com/mphoward/neighbor/pull/11) we want to abstract the usage HIP and CUDA APIs through a single header file: `Runtime.h`. This file is yet to be created.

Provided tools allow users who are developers of plugins to automatically convert their code to new API (and hoomd itself too)

Related: Issue #560 

## How Has This Been Tested?
Currently being tested offline

## Change log

```
- Use new runtime API instead of HIP/CUDA directly
- Enable conversion of user code through automated tools
```

## Checklist:


mphoward:
I haven't had time to go through `replace.py` in detail, but this in general looks great and very helpful! I had two quick comments now for you:

1. Is `search.py` vestigial? It looks mostly commented out to me. If so, we should remove this.
2. This does not necessarily need to go in this PR, but would it be possible to have either `replace.py` or another tool generate a file listing which API commands are being found and replaced? Currently, we are just aiming to do our best to put a layer over the calls that are actually used (it is quite an effort to get all of them!), and knowing which ones we need would really help simplify trial-and-error compilation.

schwendp:
1. Originally, `search.py` was intended to find CUDA/HIP APIs at the beginning of a line. However, while working on the script, I realized that some CUDA/HIP APIs are called later in the line and lackadaisically decided to just print lines that contained CUDA or HIP APIs. I plan to fix this quickly after/while working on `Runtime.h` (the redundant comments should be removed then).
2. Hopefully, `search.py` will do that after I fix it.

mphoward:
I'm working on this now to shorten the header and hopefully simplify maintenance. I can submit a reverse PR or post a link to another repo (I would like to make this header-only for people to easily find in other projects) when I'm done.

mphoward:
Here is a draft rewrite of the runtime header:

https://github.com/mphoward/neighbor/blob/055f0b9385143e3723e328b032881fc4fb316979/include/neighbor/hipper_runtime.h

Rather than have too many `#if #elif #endif`, I defined a macro `HIPPER(Name)` that will expand `Name` into `cudaName` or `hipName` depending on whether `HIPPER_CUDA` or `HIPPER_HIP` is defined. This should help cut down on potential typos, and the if-else structure is still used in cases where there was renaming between CUDA and HIP.

A few other things I did:
- Modify the enums so that they can be cast to pass to API calls.
- Remove any methods that are CUDA-specific, as this runtime header should enable only the common APIs.
- Convert all defines to `static const int`, which compiler optimization should replace as constants anyway.
- Add `threadRank()` method to accommodate `threadIdx.x` vs. `hipThreadIdx_x` in mapping thread coordinates to a rank in a kernel. This is a template function, but only the 1D grid-1D block is currently implemented. Additional dimensionality can be added if hoomd uses other launch configurations.
- Place everything in the `hipper` namespace. Function calls should read like `hipper::malloc()`.

Since this PR is on a fork, I thought it might be easiest if you copied this file over yourself, if you would like it. I think I converted everything that was essential, but if I accidentally deleted a method you need, let me know! I plan to also put the header in a separate repo, as I mentioned above, so that it's easier for people to find if they want it.

I tested that the neighbor library can still build with this new header using CUDA (on `brett`), but I haven't tried under HIP yet. @joaander do you have any machines with ROCm / HIP that I could debug & test on?

joaander:
Thanks, @mphoward , the new header loos great. We do use 2-D (and even 3-D, it appears) blocks and grids in some places in HOOMD. I leave @jglaser and @schwendp to give more detailed comments as they are leading the conversion effort in HOOMD.

I do not have any local machines with ROCm /  HIP. @jglaser has access to several, one of which is running unit tests on Azure. You could use the same docker image to test compilation locally: `glotzerlab/ci:2020.01-rocm2.10_custom_gcc9_py38`

mphoward:
Thanks! That container should do the trick to catch typos, etc., and @jglaser and @schwendp can find any runtime errors with unit testing. I'll also grep for the kernels using the different dimensionality and implement the specializations for them, e.g., `threadRank<1,2>()` will map the rank in a 1d grid-2d block.

mphoward:
We are going to want to use `hipper` by submodule or copy, which I am going to take care of in #560 to supply it to `neighbor`. I am also willing to move this py script over to hipper and maintain it there, which might mean we can close this PR later.

joaander:
Closing this as any new work will be based on hipper.

Issue #486, opened 2019-10-21
Default restartable linear variant


mphoward:
## Description

The default `zero` parameter for `hoomd.variant.linear_interp` is `'now'`, which means that the current timestep is considered 0. Since this parameter is pretty analogous to `phase`, I would have expected the default to instead be 0, as it is elsewhere in hoomd. Without specifying `zero=0`, restartable jobs do not work properly, and the resulting output might be confusing to the user. I propose changing the default to `zero=0` rather than `zero='now'` to avoid this confusion.

## Script

```python
kT = hoomd.variant.linear_interp(((0,1),(1000,2)))
# use kT in an integrator...
hoomd.run_upto(1000)
```

## Output

If the job got restarted at step 500, the values of kT would be:
```
# timestep kT
0 1.0
500 1.5
# restart
500 1.0
1000 1.5
```

## Expected output

With `zero=0`, the values of kT should be:
```
# timestep kT
0 1.0
500 1.5
# restart
500 1.5
1000 2.0
```

## Configuration
- **HOOMD-blue** version: 2.6.0 (affects current version)

## Developer

Sure, this is a quick fix if change is wanted by others.

joaander:
Yes, we should make this change. This breaks existing API, so it goes in v3.

We may need to remove the `zero` arg entirely as in the new v3 OO API there is no notion of a global current timestep and `now` becomes ill-defined.

mphoward:
OK, sounds good. I will leave this open then, and you can ping me if you want the change made onto `next`. I honestly don't see much value in the `zero` option generally, since it is straightforward to shift all of the points in the list by an offset if that's the behavior the user wants. I would think having a prescribed time-varying quantity (with a fixed time origin) is much more common.

joaander:
Thanks, Mike. We hope to start making serious progress on the API over the next few weeks. We'll let you know when we have a solid foundation and will most certainly appreciate your help then.

joaander:
Thinking more outside the box: is `variant.linear_interp` sufficiently clean and flexible? Should we keep it? Should we replace it or add a more general interface that the user could customize along with a few simpler built-ins. Something like this:
```
# linear variation from start to end over steps and then hold
variant.Ramp(start=2.0, end=0.5, steps=1e6)
# Cycle from start to end over steps, then from end to start over steps, and repeat
variant.Cycle(start=0.4, end=0.6, steps=1e6)
# user defined
class User(hoomd.variant.Variant):
[...]

PR #589, opened 2020-02-18
Implement Ramp and Cycle variants

joaander:
## Description

Implement Ramp and Cycle variants in preparation for a box variant to support update.box_resize.

## Motivation and Context

Ramp and Cycle provide two common use cases for varying temperature, pressure, boxes, etc.. over time. These are simpler to understand and use then the old variant.linear_interp which has been removed.

Resolves: #486

## How Has This Been Tested?

Added pytest unit tests.

## Checklist:


hoomd/Variant.h
```diff
@@ -50,27 +50,330 @@ class PYBIND11_EXPORT VariantConstant : public Variant
...
+        Scalar operator()(uint64_t timestep)
+            {
+            if (timestep < m_t_start)
+                {
+                return m_A;
+                }
+            else if (timestep < m_t_start + m_t_ramp)
+                {
+                // interpolate between A and B
+                uint64_t v = timestep - m_t_start;
```
b-butler: Since we cast `v` to a `double`, is there a reason that this line is here and we don't use
````C++
auto s = double(timestep - m_t_start) / double(m_t_ramp);
````

File: hoomd/Variant.h [hunk 2]
```diff
@@ -50,27 +50,330 @@ class PYBIND11_EXPORT VariantConstant : public Variant
...
+                double s = double(v) / double(m_t_ramp);
+                return m_B * s + m_A * (1.0 - s);
```
b-butler: This may not be important, but is there a reason we don't use this formulation
````C++
return m_A + (m_B - m_A) * s;
````
This has one less multiplication and it is just as simple to understand. We could even remove the subtraction by internally storing `m_B - m_A` as `delta` or something. I know this is not performance critical code just some quick thoughts I had that wouldn't reduce code maintainability or readability.

Who knows, the compiler may already be (probably is) smarter than me here.

File: hoomd/Variant.h [hunk 3]
```diff
@@ -50,27 +50,330 @@ class PYBIND11_EXPORT VariantConstant : public Variant
...
+        uint64_t getTStart()
+            {
+            return m_t_start;
+            }
+
+        /// Set the length of the ramp.
+        void setTRamp(uint64_t t_ramp)
+            {
+            // doubles can only represent integers accuracy up to 2**53.
+            if (t_ramp >= 9007199254740992ull)
```
b-butler: Have to appreciate the caution. :joy:

hoomd/Variant.h
```diff
@@ -50,27 +50,330 @@ class PYBIND11_EXPORT VariantConstant : public Variant
...
+        /// The length of the ramp.
+        uint64_t m_t_ramp;
+    };
+
+
+/** Cycle variant
+
+    Variant that cycles linearly from A to B and back again over a given number of steps.
+*/
+class PYBIND11_EXPORT VariantCycle : public Variant
```
b-butler: Should we also have a periodic cycle where we have a phase and period and use sin or cos to cycle? Regardless, I think that this is the appropriate default cycle.

hoomd/variant.py
```diff
@@ -6,9 +6,99 @@
...
+    it ramps back from *B* to *A* over *t_BA* steps and repeats the cycle
+    starting with *t_A*. :py:class:`Cycle` repeats this cycle indefinitely.
+
+    .. image:: variant-cycle.svg
+
+    Attributes:
+        A (float): The first value.
+        B (float): The second value.
+        t_start (int): The start time step.
+        t_A (int): The hold time at the first value.
```
b-butler: I think the holding time at A is more direct. Same below for t_B.

hoomd/Variant.h
joaander: Thanks, this is a good simplification.

hoomd/Variant.h
joaander: The compiler is actually not allowed to make these kinds of rearrangements (by default). The order of operations matters in floating point, so `m_B * s + m_A * (1.0 - s)` is not always equal to `m_A + (m_B - m_A) * s`. Specifically, when `s=1`, the first expression will give `m_B + 0 == m_B` while the second expression will compute `((m_A + m_B) - m_A)` which will not necessarily equal `m_B` due to round-off error.

According to Wikipedia (https://en.wikipedia.org/wiki/Linear_interpolation) the form you suggest is fine on platforms that implement FMA instructions. I don't think we can guarantee that on all platforms we support, so I'll keep the code as is. The extra one math operation per time step is not going to change performance.

hoomd/Variant.h
joaander: Yes, I think a `variant.Sin` would be a good addition. I will leave that as an excessive for someone interested in using it. The hyperbolic tangent might also be nice to use as a ramp, but I agree that the simple linear interpolation methods are the most predictable defaults.

hoomd/Variant.h
joaander: There are lots of subtle considerations that go into enabling 64-bit timesteps.

hoomd/variant.py
joaander: updated.

PR #599, opened 2020-02-26
Provide energy and per-particle energies, forces, torques, and virials in _Force

joaander:
## Description

Implement loggable properties in `ForceCompute` and `_Force`.

## Motivation and Context

This will allow users to log per-particle energies. It also enables unit testing of MD forces.

## How Has This Been Tested?

In a local jupyter notebook.

## Checklist:

### DIFF
--- a/hoomd/System.cc
+++ b/hoomd/System.cc
@@ -590,8 +590,6 @@ PDataFlags System::determineFlags(unsigned int tstep)
     if (m_integrator)
         flags = m_integrator->getRequestedPDataFlags();
 
-    flags.set(); // hack to test flags always set
-
     for (auto &analyzer_trigger_pair: m_analyzers)
         {
         if ((*analyzer_trigger_pair.second)(tstep))
--- a/hoomd/System.cc
+++ b/hoomd/System.cc
@@ -590,6 +590,8 @@ PDataFlags System::determineFlags(unsigned int tstep)
     if (m_integrator)
         flags = m_integrator->getRequestedPDataFlags();
 
+    flags.set(); // hack to test flags always set
+
     for (auto &analyzer_trigger_pair: m_analyzers)
         {
         if ((*analyzer_trigger_pair.second)(tstep))
--- a/hoomd/ForceCompute.cc
+++ b/hoomd/ForceCompute.cc
@@ -17,6 +17,8 @@
 #include <iostream>
 using namespace std;
 
+#include <pybind11/numpy.h>
+
 namespace py = pybind11;
 
 #include <memory>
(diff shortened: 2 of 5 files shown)


hoomd/System.cc
```diff
@@ -590,6 +590,8 @@ PDataFlags System::determineFlags(unsigned int tstep)
     if (m_integrator)
         flags = m_integrator->getRequestedPDataFlags();
 
+    flags.set(); // hack to test flags always set
```
b-butler: Is this our permanent solution here. If not should we label this with a TODO or issue?

joaander:
The trouble with storing these arrays all the time is that invokes a permanent O(N) storage requirement on the root rank. At least with this implementation, the storage is temporary.

We could discuss with @jglaser alternate higher performance solutions.

I plan to add LJ and Bond unit tests before merging this.

hoomd/System.cc
joaander: This is not a permanent solution. I still need to test the performance implications of this.

joaander: Unfortunately, setting all flags on results in a performance loss of 10-20% on V100. Half of that comes from the extra cost of the virial in the pair potential. The other half comes from ComputeThermo calling 3 kernels instead of 1. We can mitigate the ComputeThermo overhead, but that will still leave us with a at least a 5-10% performance hit to leave the virial calculation on all the time.

We will need to keep the flag system, at least for virial computation and possible for energy. Perhaps we should provide an API for the user to set the flags for quantities they plan to request in the API? We can update the GSD and future logging output operations to request the flags in the steps when they are triggered.

Alternately, we could re-compute forces and virials when the user requests virials. However, this will not work for the pressure tensor computed by ComputeThermo - it relies on the net virial summed by Integrator.

b-butler: I think an integrator flag would probably be cleanest. We could document that it only overrides if `virial_flag == True` and `internal_flag == False` (i.e. NPT would still need pressure even if the flag on the python side was set to false). We just then need to document that the property is only available when the flag is set to true or the simulation requires that pressure be calculated (however that should not be relied upon).

We do need a way to say to the integrator, however, to only set the flag on certain timesteps to be able to quickly log the pressure, etc... We could make the python flag a trigger that can be called to determine when to set the flag (I know that it would have to be set one step before pressure was desired, but we could under the hood call `Trigger(timestep + 1)`).

The logger is fairly agnostic to what is doing the calculation by design and that would be difficult to change. I think the Trigger solution might be the best for flexibility, but may be overly complicated generally. We could duck type the setting to make `flag=True` mean `trigger.Periodic(1)`.

I am honestly not sure what the best solution here is.

b-butler:
Would there be a way to drop the arrays when a new timestep happens? I know that @bdice and @vyasr dealt with something similar in freud.

hoomd/System.cc
joaander: I opened #607 to remind us this needs to be done.

joaander:
> Would there be a way to drop the arrays when a new timestep happens? I know that @bdice and @vyasr dealt with something similar in freud.

Anything is possible. We would need to identify the use-cases that such a behavior would enable and decide whether it is worth the complexity.

I've changed my mind and have decided to merge this PR as is. We can add unit tests using this API in another pass. Going forward, I think many short and quick to review PRs will help us be more productive and keep the number of live branches lower.

hoomd/System.cc
joaander: The temporary hack has been removed for now.

Issue #592, opened 2020-02-23
Update CUB library upstream


jglaser:
## Description

**Task** Contribute the changes made to CUB in #549  to compile with NVRTC back to upstream.

## Script

N/A

## Output

N/A

## Expected output

N/A

## Configuration
N/A

## Developer

yes

joaander:
`jitify` removes the need to modify cub.

PR #609, opened 2020-03-12
Use jitify wrapper

jglaser:
## Description

Use the `jitify` wrapper for NVRTC to more flexibly handle various `#include`s in HOOMD's sources, making it possible to use the standard CUB includes without modifications, and ultimately, to use CUB as provided in upcoming versions of the CUDA toolkit.

## Motivation and Context

We don't want to maintain a fork of CUB, and jitify is currently the only viable option to include CUB headers with NVRTC. We have already been using some of `jitify`'s STL header replacements, but without the automatic, recursive rewriting of `#include`s to perform the source-level substitution.

Resolves: #592 

## How Has This Been Tested?

standard unit tests

## Change log

N/A

## Checklist:


hoomd/jit/GPUEvalFactory.h
```diff
@@ -101,16 +74,22 @@ class GPUEvalFactory
...
+                m_program->kernel(m_kernel_name).instantiate(eval_threads, launch_bounds));
             char *error;
             if (custatus != CUDA_SUCCESS)
                 {
                 cuGetErrorString(custatus, const_cast<const char **>(&error));
-                throw std::runtime_error("cufuncGetAttribute: "+std::string(error));
+                throw std::runtime_error("cuFuncGetAttribute: "+std::string(error));
                 }
+            cudaSetDevice(m_exec_conf->getGPUIds()[0]);
+            #endif
```
vyasr: Is there no corresponding logic for HIP? Do you not have to specify the device, or does it not support multi-gpu?

hoomd/jit/GPUEvalFactory.h
```diff
@@ -144,36 +129,44 @@ class GPUEvalFactory
...
 
-                // copy the array pointer to the device
-                char *error;
-                CUresult custatus = cuMemcpyHtoD(m_alpha_iso_device_ptr[idev], &d_alpha, sizeof(float *));
-                if (custatus != CUDA_SUCCESS)
+                for (auto e: m_eval_threads)
                     {
-                    cuGetErrorString(custatus, const_cast<const char **>(&error));
-                    throw std::runtime_error("cuMemcpyHtoD: "+std::string(error));
+                    for (auto l:  m_launch_bounds)
```
vyasr:
```suggestion
                    for (auto l: m_launch_bounds)
```

hoomd/jit/GPUEvalFactory.h
```diff
@@ -203,19 +205,28 @@ class GPUEvalFactory
...
 
-                // copy the array pointer to the device
-                char *error;
-                CUresult custatus = cuMemcpyHtoD(m_rcut_union_device_ptr[idev], &rcut, sizeof(float));
-                if (custatus != CUDA_SUCCESS)
+                for (auto e: m_eval_threads)
                     {
-                    cuGetErrorString(custatus, const_cast<const char **>(&error));
-                    throw std::runtime_error("cuMemcpyHtoD: "+std::string(error));
+                    for (auto l:  m_launch_bounds)
```
vyasr:
```suggestion
                    for (auto l: m_launch_bounds)
```

joaander:
@jglaser Let me know when you plan to address these comments. Since this is a major change, it would be good to merge before the new GPU runtime work modifies all files.

hoomd/jit/EvaluatorUnionGPU.cuh
```diff
@@ -95,10 +95,10 @@ union_params_t
 
 #ifdef __HIPCC__
 // Storage for shape parameters
-static __device__ union_params_t *d_union_params;
+__device__ union_params_t *d_union_params;
 
 //! Device storage of rcut value
-static __device__ float d_rcut_union;
```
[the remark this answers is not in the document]
jglaser: I don't know why this change was necessary. Ideally we would move non-static global variables into a file which we know can be compiled and linked exactly once, to avoid ambiguous linkage. De facto this is the case for this header, even though it might create problems if `EvaluatorUnionGPU.cuh` were included from different object files (or fat binaries?). The fact that it works even though the kernel **is** compiled multiple times seems to indicate that NVRTC is not prone to this issue.

hoomd/jit/GPUEvalFactory.h
```diff
@@ -11,6 +11,18 @@
...
 #include <nvrtc.h>
+
+#define JITIFY_PRINT_LOG 0
+#define JITIFY_PRINT_INSTANTIATION 0
+#define JITIFY_PRINT_SOURCE 0
+#define JITIFY_PRINT_LOG 0
+#define JITIFY_PRINT_PTX 0
+//#define JITIFY_PRINT_LINKER_LOG
+#define JITIFY_PRINT_LAUNCH 0
+#define JITIFY_PRINT_HEADER_PATHS 0
```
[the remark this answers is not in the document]
jglaser: I do not know if there would an advantage to that. NDEBUG is really used for drilling down on segfaults, because it enables `assert`s and debug information, I wouldn't want use it for anything else, as that would not be so obvious. You can, right now, log the JIT program source and error messages with `notice_level=5`.

hoomd/jit/GPUEvalFactory.h
jglaser: I suggest we implement RTC with HIP later, if it is at all possible.

jglaser: HIP doesn't support managed memory/multi-GPU just yet

hoomd/jit/GPUEvalFactory.h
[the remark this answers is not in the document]
vyasr: As long as there's a well-defined way to capture debugging information about failures here I think that's fine. It's not clear to me what information we're declining by setting all of these flags, so I'm just wondering if there should be a way to flip them for debugging.

hoomd/jit/EvaluatorUnionGPU.cuh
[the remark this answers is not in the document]
vyasr: Right that makes sense. I have no idea what symbols are actually given external linkage by NVRTC, but as you say it seems to work fine.

vyasr:
My change requests have been addressed.

PR #610, opened 2020-03-12
Move LoadBalancer to a Tuner

b-butler:
## Description

Moves `LoadBalancer{GPU}` to a tuner, and puts it in the `hoomd.tuner` namespace. Also updates the API.
Currently this is waiting on a decomposition object.

## How Has This Been Tested?

Hasn't been.


hoomd/tuner/balance.py
```diff
@@ -0,0 +1,97 @@
...
+            trigger=OnlyType(Trigger, preprocess=trigger_preprocessing))
+        self._param_dict.update(defaults)
+
+    def attach(self, simulation):
+        if simulation.device.mode == 'gpu':
+            cpp_cls = getattr(_hoomd, 'LoadBalancerGPU')
+        else:
+            cpp_cls = getattr(_hoomd, 'LoadBalancer')
+        self._cpp_obj = cpp_cls(simulation.state._cpp_sys_def,
+                                simulation.device.decomposition._cpp_obj,
```
joaander, line 95: We can complete this PR now before completing the domain decomposition API. Add a `getDomainDecomposition` method to `Communicator.h`. Then replace this with `simulation._cpp_sys.getCommunicator().getDomainDecomposition()`

The decomposition is passed here as a matter of plumbing to link up the decomposition with the balancer, as the decomposition is not part of the `SystemDefinition`.

hoomd/LoadBalancer.cc
```diff
@@ -29,12 +29,15 @@ namespace py = pybind11;
...
 LoadBalancer::LoadBalancer(std::shared_ptr<SystemDefinition> sysdef,
-                           std::shared_ptr<DomainDecomposition> decomposition)
-        : Updater(sysdef), m_decomposition(decomposition), m_mpi_comm(m_exec_conf->getMPICommunicator()),
-          m_max_imbalance(Scalar(1.0)), m_recompute_max_imbalance(true), m_needs_migrate(false),
-          m_needs_recount(false), m_tolerance(Scalar(1.05)), m_maxiter(1), m_max_scale(Scalar(0.05)),
-          m_N_own(m_pdata->getN()), m_max_max_imbalance(1.0), m_total_max_imbalance(0.0), m_n_calls(0),
-          m_n_iterations(0), m_n_rebalances(0)
+                           std::shared_ptr<Trigger> trigger)
+        : Tuner(sysdef, trigger), m_mpi_comm(m_exec_conf->getMPICommunicator()),
+          m_decomposition(m_mpi_comm->getDomainDecomposition()),
```
joaander, line 34: I don't think this will work here. `m_mpi_comm` is NULL until after the object is constructed and `setCommunicator` is called. Will need to move this into the `update()` method.

hoomd/LoadBalancer.cc
```diff
@@ -573,11 +576,13 @@ void export_LoadBalancer(py::module& m)
     {
     py::class_<LoadBalancer, Updater, std::shared_ptr<LoadBalancer> >(m,"LoadBalancer")
     .def(py::init< std::shared_ptr<SystemDefinition>, std::shared_ptr<DomainDecomposition> >())
```
joaander, line 578: Need to update the constructor arg list in the export.

hoomd/LoadBalancer.cc
b-butler: I just took your original suggestion, and did this in Python.

joaander:
Looks like we need to merge `feature/new-object-API` up into this branch.

PR #612, opened 2020-03-17
Feature/update filters

b-butler:
## Description
Add the `Type`, `SetDifference`, `Intersection`, and `Union` filters. Also, makes the `ParticleFilter` caching system more robust.

## How Has This Been Tested?
It has not, but exploring in a jupyter notebook suggests the filters work as expected.

## Checklist:


hoomd/filter/filter_.py
```diff
@@ -0,0 +1,14 @@
+class ParticleFilter:
+    def __hash__(self):
+        return NotImplementedError("Must implement hash for Filters.")
```
bdice:
```suggestion
        return NotImplementedError("Must implement hash for ParticleFilters.")
```

hoomd/filter/filter_.py
```diff
@@ -0,0 +1,14 @@
+class ParticleFilter:
```
bdice: Are users supposed to subclass this? If not, make it private. In either case, you may want to make it an abstract class.

hoomd/filter/set_.py
```diff
@@ -0,0 +1,35 @@
+from hoomd.filter.filter_ import ParticleFilter
+from hoomd import _hoomd
+
+
+class ParticleFilterSetOperations(ParticleFilter):
+    def __init__(self, f, g):
+        if f == g:
+            raise ValueError("Cannot use same filter for {}"
+                             "".format(self.__class__.__name__))
```
bdice: Why not? It should work, right? It just returns an empty list of tags for differences and the same tags as its inputs for intersection and union.

hoomd/filter/set_.py
```diff
@@ -0,0 +1,35 @@
...
+        else:
+            self._f = f
+            self._g = g
+        getattr(_hoomd, self._cpp_cls_name).__init__(self, f, g)
```
bdice: I see what this is supposed to do, but I would like a comment above this line explaining it.

hoomd/filter/set_.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __hash__(self):
+        return hash(hash(self._f) + hash(self._g))
+
+    def __eq__(self, other):
+        return type(self) == type(other) and \
+            self._f == other._f and \
+            self._g == other._g
```
bdice: May want to override this for symmetric cases such that `f` and `g` can be swapped. Depends on what assumptions you are making and/or the use cases for checking equality.

hoomd/filter/set_.py
```diff
@@ -0,0 +1,35 @@
+from hoomd.filter.filter_ import ParticleFilter
+from hoomd import _hoomd
+
+
+class ParticleFilterSetOperations(ParticleFilter):
```
bdice: 1. Should this be abstract? You can't instantiate it without `self._cpp_cls_name`.
2. Are users supposed to be able to subclass this? If not, make it private.
3. What about a set symmetric difference?

hoomd/filter/tags.py
```diff
@@ -0,0 +1,26 @@
+from hoomd.filter.filter_ import ParticleFilter
+from hoomd._hoomd import ParticleFilterTags
+import numpy as np
+
+
+class Tags(ParticleFilter, ParticleFilterTags):
```
bdice: Needs a docstring.

hoomd/filter/tags.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, tags):
+        if isinstance(tags, np.ndarray):
+            tags = tags.astype(np.uint32)
+        else:
+            tags = np.array(tags, dtype=np.uint32)
+        tags = np.unique(tags)
+        self._tags = tags
```
bdice: One line to rule them all. `np.unique` will be sorted and unique, `ascontiguousarray` with `dtype` will enforce all the other constraints.
```suggestion
        self._tags = np.ascontiguousarray(np.unique(tags), dtype=np.uint32)
```

hoomd/CMakeLists.txt
```diff
@@ -153,6 +154,14 @@ set(_hoomd_headers
     Variant.h
     VectorMath.h
     WarpTools.cuh
+    filter/ParticleFilter.h
+    filter/ParticleFilterAll.h
```
joaander: This will not install these headers in the proper location. The later cmake command `install(FILES ${_hoomd_headers} ...` installs all files into the same directory regardless of their absolute path, so there is no `filter` subdirectory in the installed header location.

We either need to put all these headers in the `hoomd/` directory or add a new `install` commands to install these headers in the proper location and remove them from the `_hoomd_headers` list.

hoomd/filter/ParticleFilter.h
```diff
@@ -0,0 +1,46 @@
...
+
+    In parallel simulations, getSelectedTags() should return only local tags.
+
+    The base class getSelectedTags() method will simply return an empty list.
+    selection semantics.
+*/
+class PYBIND11_EXPORT ParticleFilter
+    {
+    public:
+        /// constructs a base ParticleFilter (does nothing)
```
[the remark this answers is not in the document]
joaander: The triple slash comment flags this as documentation in Doxygen and other tools: https://hoomd-blue.readthedocs.io/en/feature-new-object-api/style.html#id2

hoomd/filter/ParticleFilterTags.h
```diff
@@ -0,0 +1,41 @@
...
+            }
+        virtual ~ParticleFilterTags() {}
+
+        /*! \param tag Tag of the particle to check
+            \returns true if \a m_tag_min <= \a tag <= \a m_tag_max
+        */
+        virtual std::vector<unsigned int> getSelectedTags(
+                std::shared_ptr<SystemDefinition> sysdef) const
+            {
+            return m_tags;
```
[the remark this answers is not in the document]
joaander: The root rank performs a set union of the filter outputs of all ranks. So it doesn't matter which ranks provide a tag, nor does it matter if multiple ranks provide the same tag.

hoomd/filter/ParticleFilter.h
```diff
@@ -0,0 +1,46 @@
...
+        ParticleFilter() {};
+        virtual ~ParticleFilter() {}
+
+        /// Test if a particle meets the selection criteria
```
joaander: Multiline documentation comments need to be wrapped in blocks starting with `\**`
```
\**
*/
```

hoomd/filter/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+################ Python only modules
+# copy python modules to the build directory to make it a working python package
+set(files __init__.py
+	  filter_.py
```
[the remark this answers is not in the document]
b-butler: Will do!

hoomd/filter/ParticleFilter.h
b-butler: Got it.

hoomd/filter/ParticleFilterTags.h
```diff
@@ -0,0 +1,41 @@
...
+    {
+    public:
+        /** Args:
+                tags: numpybind11array of tags to select
+        */
+        ParticleFilterTags(std::vector<unsigned int> tags)
+            : ParticleFilter(), m_tags(tags) {}
+
+        /** Args:
+                tags: numpybind11array of tags to select
```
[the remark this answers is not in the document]
b-butler: It isn't I tried to leave previous comments and probably mangled the original comments somehow.

hoomd/filter/all_.py
```diff
@@ -0,0 +1,13 @@
...
+
+class All(ParticleFilter, ParticleFilterAll):
+    def __init__(self):
+        ParticleFilterAll.__init__(self)
+
+    def __hash__(self):
+        return 0
+
+    def __eq__(self, other):
+        return type(self) == type(other)
```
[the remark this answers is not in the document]
b-butler: We cache all `ParticleFilter`s used in a given `Simulation`, using Python's `dict`, so for a given `ParticleFilter` class and instance there can only ever be one group with that particular combination. We could do the singleton class here, but to keep the pattern of other filters, I just implemented `__hash__` and `__eq__`.

hoomd/filter/filter_.py
b-butler: We can't make it an abstract class (I tried). The way pybind11 deals with inheritance prevents using metaclasses here. We would have to switch to a composition model for this to work.

hoomd/filter/set_.py
b-butler: 1. See above comment.
2. I do need to make it private.
3. The set symmetric difference can be achieved with `SetDifference(Union(a, b), Intersection(a, b))` though I could see an argument for performance reasons.

hoomd/filter/set_.py
b-butler: The idea was to prevent a user for shooting themselves in the foot. I can't think of a case where this would be useful, but if you feel strongly about it, it will work.

hoomd/filter/set_.py
b-butler: Can do.

hoomd/filter/set_.py
b-butler: That is a good point. I will need to account for that for the union and intersection.

hoomd/filter/tags.py
b-butler: Many of them do. I was not sure whether, I would include documentation in this pull request. We have been moving fast trying to get features ported or added rather than the documentation and testing to some extent right now.

hoomd/filter/tags.py
b-butler: Nice one-liner, and actually more clear than the 6 lines it replaces.

b-butler: However I will need to ensure that `ParticleFilterTags` is called with this array not the tags variable which has not been processed.

hoomd/filter/type.py
```diff
@@ -0,0 +1,19 @@
+from filter_ import ParticleFilter
```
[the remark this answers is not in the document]
b-butler: Will remove. I originally named them without the underscores, but decided to abide by the Python guidelines here.

hoomd/state.py
```diff
@@ -179,6 +180,10 @@ def scale_system(self):
         raise NotImplementedError
 
     def add_group(self, filter_):
-        group = self._groups[filter_] = _hoomd.ParticleGroup(
-            self._cpp_sys_def, filter_)
-        return group
+        cls = filter_.__class__
+        if filter_ in self._groups[cls]:
```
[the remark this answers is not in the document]
b-butler: And hopefully maintainable. :crossed_fingers:

b-butler:
@bdice Thanks for the useful comments. Regarding the use of the standard library more `copy_if` is not generally useful to us since we want to store the tags, but the decision for inclusion can involve any component of particle data. This is also why the set operations must happen on the completed arrays versus performing the condition check on a per particle basis.

With the comment on docstrings, I agree some of this is just porting old comments without the appropriate amount of checking for correctness and content.

With regards to `const`, we can definitely be more liberal in our use of it.

hoomd/CMakeLists.txt
b-butler: I have attempted to resolve this. i am not familiar with CMake so if this is not correct, let me know.

joaander:
> With the comment on docstrings, I agree some of this is just porting old comments without the appropriate amount of checking for correctness and content.

Indeed, the references to "simple" and "easy" were likely written by me a decade ago @bdice . My sincerest apologies for endeavoring to write code that is simple and easy to understand and committing the modern faux pas of actually stating that intent. In new code, we can avoid trivializing concepts by designing things with additional complexity that are challenging to use.

hoomd/CMakeLists.txt
joaander: You were just missing the `add_subdirectory(filter)`. I pushed a commit to the branch that does this.

b-butler: Ah that is a much simpler fix. I am sorry for making you revert my changes and perform the appropriate ones.

joaander: I didn't revert any changes. The `CMakeLists.txt` file you added to `filters` is what `add_subdirectory` calls.

joaander:
Tests are passing. Awaiting final comments before merge.

PR #613, opened 2020-03-20
Misc apis

joaander:
## Description

* Fix the system exception hook.
* Replace the CUDA profile start/stop commands with a context manager

## Motivation and Context

On `next`, the exception hook is broken. There is no longer a single global MPI communicator. The exception hook calls `abort_mpi` whenever this is an uncaught exception. In this PR, the default is to abort `MPI_COMM_WORLD`. When partitioning simulations, this may be undesirable as the user may want to only abort the partition that errors. This PR provides a `localize_abort` context manager where the user can opt in to only aborting on the current partition. It is not possible for this context manager to enforce that the user only use the partition within it, it is just an indication by the user that they will not communicate outside the partitions.

This PR also adds an `enable_profiling` context manager to the `GPU` object that allows the user to enable GPU profiling only for specific operations. This replaces `util.cuda_profile_start` and `util.cuda_profile_stop`.

## How Has This Been Tested?

I have tested the exception hook locally. It is not possible to test this with unit tests as `mpi_abort` kills the process. There is a minimal unit test for `enable_profiling`.

## Checklist:


hoomd/device.py
```diff
@@ -271,6 +272,29 @@ def is_available():
         """
         return _hoomd.isCUDAAvailable();
 
+    @contextlib.contextmanager
+    def enable_profiling(self):
+        """ Enable GPU profiling.
```
bdice:
```suggestion
        """Enable GPU profiling.
```

hoomd/device.py
```diff
@@ -271,6 +272,29 @@ def is_available():
...
+
+        When using GPU profiling tools on HOOMD, select the option to disable
+        profiling on start. Initialize and run a simulation long enough that
```
bdice: Select the option to disable profiling on start? Is there an option elsewhere that this should link to? I am not aware of other ways to enable profiling, so maybe I'm missing something.

hoomd/device.py
```diff
@@ -271,6 +272,29 @@ def is_available():
...
+        all autotuners have completed, then open :py:func:`enable_profiling()`
```
bdice: I can't remember for sure, but I don't think you need parentheses when using these kind of function references. I think it might even add them for you in the rendered docs.

hoomd/device.py
```diff
@@ -271,6 +272,29 @@ def is_available():
...
+        as a context manager and continue the simulation for a time. Profiling
+        stops when the context manager closes.
+
+        Example::
+
+            with device.enable_profiling():
+                sim.run(1000)
+        """
+
+        try:
```
bdice: Should this show a warning or other message if an exception is raised?

hoomd/comm.py
```diff
@@ -9,9 +9,10 @@
...
 from hoomd import _hoomd
-import hoomd;
+import hoomd
 
-import sys;
+import contextlib
+import sys
 
 
 class decomposition(object):
```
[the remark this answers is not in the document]
joaander: We will rewrite the decomposition object on another branch.

hoomd/device.py
joaander: The option is tool specific. For example, there is a checkbox in the nvvp project setup, there is a command line option for nvprof. I don't recall where the option is in NSight systems and I've never used the AMD profiling tools myself.

hoomd/device.py
joaander: The finally ensures that the profiling is stopped and the profiling tool can write the output to disk. I don't think any output other than the triggering exception itself is needed here. This is Python's recommended way to write context managers: https://docs.python.org/3.8/library/contextlib.html

hoomd/device.py
vyasr: Nsight has the same option, and IIRC it can be changed mid-run unlike nvvp (where it has to be specified when you create the workspace).

hoomd/device.py
vyasr: That is correct. It needs to access and inspect the object (since all Python functions are objects), so you don't need parenthesis.

joaander: Yes, sphinx adds them in the rendered docs.

hoomd/comm.py
```diff
@@ -379,3 +380,29 @@ def barrier(self):
...
+        to tell HOOMD that all operations within the context will use only
+        that MPI communicator so that an uncaught exception in one partition
+        will only abort that partition and leave the others running.
+        """
+
+        global _current_communicator
+        prev = _current_communicator
+
+        _current_communicator = self
+        yield None
```
bdice: Given the other conversation below about context managers, should this be wrapped in a `try/finally`?

joaander: I tried, but it failed tests. By the time the uncaught exception gets to the exception hook, the `finally:` block has executed and reset the previous communicator - which is exactly what we don't want to happen. In this case we actually want the `_current_communicator` to remain set and there is no need to reset it while the process is being torn down.

hoomd/comm.py
```diff
@@ -379,3 +380,29 @@ def barrier(self):
...
+        _current_communicator = prev
+
+# store the "current" communicator to be used for MPI_Abort calls. This defaults
+# to the world communicator, but users can opt in to a more specific
+# communicator using the GPU.localize_abort context manager
```
b-butler:
```suggestion
# communicator using the Device.localize_abort context manager
```

joaander:
Tests are passing, any final comments @bdice ?

hoomd/comm.py
joaander: Why Device? I thought your suggestion was to add the GPU specific method to only the GPU class.

b-butler: If the user has multiple GPU's through an MPI simulation, they can't selectively abort? I know that even when running on the GPU you require a CPU, but the user wouldn't have access to a CPU device as far as I am aware.

joaander: Nevermind, I was confusing `localize_abort` and `enable_profiling`

Issue #607, opened 2020-03-06
Provide API for user to request flags


joaander:
## Description

Users can now request per-particle virials, energies, etc.. on demand. These quantities are not always calculated for performance reasons.

## Proposed solution

The choice of API is not clear at this time. See some discussions for possibilities in #599.

## Additional context

For `_Force` objects, we could implement a re-compute step when one of these quantities is requested. However, this does not work for `ComputeThermo` for the total energy and pressure quantities - these depend on the net energy and net virial computed by the integrator.

Enabling virial computations all the time can slow simulations by 10-20% on V100 GPUs, so it is important to keep the performance when possible.

joaander:
In terms of the current different behavior on the GPU and CPU - we could normalize that behavior so that the user gets a consistent experience with the API.

joaander:
Done.

PR #615, opened 2020-03-26
Allow users to set the pressure computation flag

joaander:
## Description

* Simplify use of particle data flags in HOOMD at the C++ level
* Add `Simulation.always_compute_pressure` user API
* `dump.GSD` requests pressure computation in steps when it will write with the log writer.

## Motivation and Context

The v3 API allows users to request computed properties, such as the per particle virial and the total system pressure at any time. The virial and pressure are only computed on timesteps when it is needed because this reduces performance significantly. The user facing API `Simulation.always_compute_pressure` allows users to opt-in to this performance loss when they expect to query these properties.

Resolves: #607

## How Has This Been Tested?

Added minimal unit test

## Checklist:


hoomd/ParticleData.h
```diff
@@ -960,6 +955,12 @@ class PYBIND11_EXPORT ParticleData
         */
         void setFlags(const PDataFlags& flags) { m_flags = flags; }
 
+        /// Enable pressure computations
+        void setPressureFlag()
```
bdice: Why does this signature differ from the one in `System.h`? It doesn't look like the flag can be disabled here.

hoomd/md/TwoStepNPTMTK.h
```diff
@@ -137,7 +137,7 @@ class PYBIND11_EXPORT TwoStepNPTMTK : public IntegrationMethodTwoStep
         virtual void integrateStepTwo(unsigned int timestep);
 
         //! Get needed pdata flags
-        /*! TwoStepNPTMTK needs the pressure, so the isotropic_virial or pressure_tensor flag is set,
+        /*! TwoStepNPTMTK needs the pressure, so the pressure_tensor flag is set,
             depending on the integration mode
```
bdice: Does this depend on the integration mode anymore?
```suggestion
        /*! TwoStepNPTMTK needs the pressure, so the pressure_tensor flag is set.
```

hoomd/md/pytest/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# copy python modules to the build directory to make it a working python package
+set(files __init__.py
+	test_flags.py
```
bdice: tabs/spaces
```suggestion
    test_flags.py
```

hoomd/simulation.py
```diff
@@ -141,6 +141,37 @@ def verbose_run(self, value):
         self._verbose = bool(value)
         self._cpp_sys.enableQuietRun(not self.verbose_run)
 
+    @property
+    def always_compute_pressure(self):
+        """Always compute the virial and pressure.
+
+        By default, HOOMD only computes the virial and pressure on timesteps
+        where it is needed (when :py:class:`hoomd.dump.GSD` writes
```
bdice:
```suggestion
        where it is needed (when :py:class:`hoomd.dump.gsd` writes
```

hoomd/ForceCompute.cc
```diff
@@ -285,7 +288,7 @@ pybind11::object ForceCompute::getEnergiesPython()
     std::vector<double> energy(dims[0]);
 
     // This is slow: TODO implement a propert gather operation
-    for (size_t i = 0; i < dims[0]; i++)
+    for (size_t i = 0; i < m_pdata->getNGlobal(); i++)
```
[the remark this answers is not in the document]
joaander: No. Collective calls must be made by all ranks or else the simulation deadlocks.

hoomd/ComputeThermo.h
```diff
@@ -227,16 +223,8 @@ class PYBIND11_EXPORT ComputeThermo : public Compute
             #endif
 
             // return NaN if the flags are not valid
```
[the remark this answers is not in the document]
joaander: removed

hoomd/ParticleData.h
joaander: The `System` object resets the flags in `ParticleData` each timestep and `System::setPressureFlag` controls whether the pressure computation flag is always set or not. This will not take effect until at least one timestep has been run.

`ParticleData::setPressureFlag` takes effect immediately and exists so that users can request the pressure flag and then request the per particle virials from a force compute object without calling `System::run`. The flags are an indication of what data is valid, once the virials have been computed the user cannot remove them so there is no way to immediately remove the flag.

hoomd/simulation.py
joaander: The class is named `hoomd.dump.GSD` per PEP8 naming conventions.

hoomd/simulation.py
```diff
@@ -141,6 +141,37 @@ def verbose_run(self, value):
...
+        log data to a file or when using an NPT integrator). Set
+        ``always_compute_pressure`` to True to make the per particle virial,
+        net virial, and system pressure available to query any time by property
+        or through the :py:class:`hoomd.Logger` interface.
+
+        Enabling this flag will result in a moderate performance penalty
```
b-butler: This may look better in the documentation as a note since that will get special formatting, and draw the eye.

hoomd/simulation.py
```diff
@@ -141,6 +141,37 @@ def verbose_run(self, value):
...
+        """
+        if not hasattr(self, '_cpp_sys'):
+            return False
+        else:
+            return self._cpp_sys.getPressureFlag()
+
+    @always_compute_pressure.setter
+    def always_compute_pressure(self, value):
+        if not hasattr(self, '_cpp_sys'):
+            raise RuntimeError('Cannot set flag without state')
```
b-butler:
```suggestion
           # TODO make this work when not attached by automatically setting flag when state object is instantiated.
            raise RuntimeError('Cannot set flag without state')
```

hoomd/simulation.py
joaander: Done.

PR #619, opened 2020-03-31
Trigger compute caching and add new triggers

b-butler:
## Description

Implements caching for the last time step and result of a trigger automatically. Now `compute` is the virtual method that must be defined for all derived `Trigger`s.

Also adds the following triggers:

1. `After` - returns true after given step false otherwise
2. `Until` - returns true until given step false otherwise
3. `Not` - negates a single trigger
4. `And` - returns the logical AND of two triggers
5. `Or` - returns the logical OR of two triggers

## How Has This Been Tested?
locally in a jupyter notebook

### DIFF
--- a/hoomd/Trigger.cc
+++ b/hoomd/Trigger.cc
@@ -17,15 +17,14 @@ class TriggerPy : public Trigger
         using Trigger::Trigger;
 
         // trampoline method
-        bool operator()(uint64_t timestep) override
+        bool compute(uint64_t timestep) override
             {
-            PYBIND11_OVERLOAD_NAME(bool,         // Return type
+            PYBIND11_OVERLOAD_PURE(bool,         // Return type
                                    Trigger,      // Parent class
-                                   "__call__",   // name of function in python
-                                   operator(),   // Name of function in C++
+                                   compute,
                                    timestep      // Argument(s)
                               );
-        }
+            }
     };
 
 void export_Trigger(pybind11::module& m)
@@ -33,6 +32,7 @@ void export_Trigger(pybind11::module& m)
     pybind11::class_<Trigger, TriggerPy, std::shared_ptr<Trigger> >(m,"Trigger")
         .def(pybind11::init<>())
         .def("__call__", &Trigger::operator())
+        .def("compute", &Trigger::compute)
         ;
 
     pybind11::class_<PeriodicTrigger, Trigger, std::shared_ptr<PeriodicTrigger> >(m, "PeriodicTrigger")
--- a/hoomd/Trigger.cc
+++ b/hoomd/Trigger.cc
@@ -42,5 +42,54 @@ void export_Trigger(pybind11::module& m)
         .def_property("period", &PeriodicTrigger::getPeriod, &PeriodicTrigger::setPeriod)
         ;
 
+    pybind11::class_<UntilTrigger, Trigger, std::shared_ptr<UntilTrigger>
+                    >(m, "UntilTrigger")
+        .def(pybind11::init<uint64_t>(), pybind11::arg("until"))
+        .def_property("until", &UntilTrigger::getUntil, &UntilTrigger::setUntil)
+        ;
+
+    pybind11::class_<AfterTrigger, Trigger, std::shared_ptr<AfterTrigger>
+                    >(m, "AfterTrigger")
+        .def(pybind11::init<uint64_t>(), pybind11::arg("after"))
+        .def_property("after", &AfterTrigger::getAfter, &AfterTrigger::setAfter)
+        ;
+
+    pybind11::class_<NotTrigger, Trigger, std::shared_ptr<NotTrigger>
+                    >(m, "NotTrigger")
+        .def(pybind11::init<std::shared_ptr<Trigger> >(),
+             pybind11::arg("trigger"))
+        .def_property("trigger",
+                      &NotTrigger::getTrigger,
+                      &NotTrigger::setTrigger)
+        ;
+
+    pybind11::class_<AndTrigger, Trigger, std::shared_ptr<AndTrigger>
+                    >(m, "AndTrigger")
+        .def(pybind11::init<std::shared_ptr<Trigger>,
+                            std::shared_ptr<Trigger> >(),
+             pybind11::arg("trigger1"),
+             pybind11::arg("trigger2"))
+        .def_property("trigger1",
+                      &AndTrigger::getTrigger1,
+                      &AndTrigger::setTrigger1)
+        .def_property("trigger2",
+                      &AndTrigger::getTrigger2,
+                      &AndTrigger::setTrigger2)
+        ;
+
+    pybind11::class_<OrTrigger, Trigger, std::shared_ptr<OrTrigger>
+                    >(m, "OrTrigger")
+        .def(pybind11::init<std::shared_ptr<Trigger>,
+                            std::shared_ptr<Trigger> >(),
+             pybind11::arg("trigger1"),
+             pybind11::arg("trigger2"))
+        .def_property("trigger1",
+                      &OrTrigger::getTrigger1,
+                      &OrTrigger::setTrigger1)
+        .def_property("trigger2",
+                      &OrTrigger::getTrigger2,
+                      &OrTrigger::setTrigger2)
+        ;
+
     m.def("_test_trigger_call", &testTriggerCall);
     }
--- a/hoomd/Trigger.cc
+++ b/hoomd/Trigger.cc
@@ -2,6 +2,10 @@
 // This file is part of the HOOMD-blue project, released under the BSD 3-Clause License.
 
 #include "Trigger.h"
+#include <pybind11/stl.h>
+#include <pybind11/stl_bind.h>
+
+PYBIND11_MAKE_OPAQUE(std::vector<std::shared_ptr<Trigger> >);
 
 //* Method to enable unit testing of C++ trigger calls from pytest
 bool testTriggerCall(std::shared_ptr<Trigger> t, uint64_t step)
@@ -35,23 +39,42 @@ void export_Trigger(pybind11::module& m)
         .def("compute", &Trigger::compute)
         ;
 
-    pybind11::class_<PeriodicTrigger, Trigger, std::shared_ptr<PeriodicTrigger> >(m, "PeriodicTrigger")
-        .def(pybind11::init< uint64_t, uint64_t >(), pybind11::arg("period"), pybind11::arg("phase"))
+    pybind11::class_<PeriodicTrigger, Trigger,
+                     std::shared_ptr<PeriodicTrigger> >(m, "PeriodicTrigger")
+        .def(pybind11::init< uint64_t, uint64_t >(),
+             pybind11::arg("period"),
+             pybind11::arg("phase"))
         .def(pybind11::init< uint64_t >(), pybind11::arg("period"))
-        .def_property("phase", &PeriodicTrigger::getPhase, &PeriodicTrigger::setPhase)
-        .def_property("period", &PeriodicTrigger::getPeriod, &PeriodicTrigger::setPeriod)
+        .def_property("phase",
+                      &PeriodicTrigger::getPhase,
+                      &PeriodicTrigger::setPhase)
+        .def_property("period",
+                      &PeriodicTrigger::getPeriod,
+                      &PeriodicTrigger::setPeriod)
         ;
 
-    pybind11::class_<UntilTrigger, Trigger, std::shared_ptr<UntilTrigger>
-                    >(m, "UntilTrigger")
-        .def(pybind11::init<uint64_t>(), pybind11::arg("until"))
-        .def_property("until", &UntilTrigger::getUntil, &UntilTrigger::setUntil)
+    pybind11::class_<BeforeTrigger, Trigger, std::shared_ptr<BeforeTrigger>
+                    >(m, "BeforeTrigger")
+        .def(pybind11::init<uint64_t>(), pybind11::arg("timestep"))
+        .def_property("timestep",
+                      &BeforeTrigger::getTimestep,
+                      &BeforeTrigger::setTimestep)
+        ;
+
+    pybind11::class_<OnTrigger, Trigger, std::shared_ptr<OnTrigger>
+                    >(m, "OnTrigger")
+        .def(pybind11::init<uint64_t>(), pybind11::arg("timestep"))
+        .def_property("timestep",
+                      &OnTrigger::getTimestep,
+                      &OnTrigger::setTimestep)
         ;
 
     pybind11::class_<AfterTrigger, Trigger, std::shared_ptr<AfterTrigger>
                     >(m, "AfterTrigger")
-        .def(pybind11::init<uint64_t>(), pybind11::arg("after"))
-        .def_property("after", &AfterTrigger::getAfter, &AfterTrigger::setAfter)
+        .def(pybind11::init<uint64_t>(), pybind11::arg("timestep"))
+        .def_property("timestep",
+                      &AfterTrigger::getTimestep,
+                      &AfterTrigger::setTimestep)
         ;
 
     pybind11::class_<NotTrigger, Trigger, std::shared_ptr<NotTrigger>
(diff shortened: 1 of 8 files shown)


hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+*/
+class PYBIND11_EXPORT UntilTrigger : public Trigger
+    {
+    public:
+
+    UntilTrigger(uint64_t until) : Trigger(), m_until(until) {}
+
+    bool compute(uint64_t timestep)
+        {
+        if (timestep < m_until)
```
joaander, line 118: Should we use <= here so that or(until(step),after(step)) is always true?

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+*/
+class PYBIND11_EXPORT NotTrigger : public Trigger
+    {
+    public:
+        NotTrigger(std::shared_ptr<Trigger> trigger) :
+            Trigger(), m_trigger(trigger) {}
+
+        bool compute(uint64_t timestep)
+            {
+            return !(m_trigger->compute(timestep));
```
joaander, line 184: Consider using operator() here so that the child triggers will use cached results.

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+class PYBIND11_EXPORT AndTrigger : public Trigger
+    {
+    public:
+        AndTrigger(std::shared_ptr<Trigger> trigger1,
+                   std::shared_ptr<Trigger> trigger2) :
+            Trigger(), m_trigger1(trigger1), m_trigger2(trigger2) {}
+
+        bool compute(uint64_t timestep)
+            {
+            return m_trigger1->compute(timestep) &&
```
joaander, line 210: Consider using operator() here so that the child triggers will use cached results.

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+class PYBIND11_EXPORT OrTrigger : public Trigger
+    {
+    public:
+        OrTrigger(std::shared_ptr<Trigger> trigger1,
+                   std::shared_ptr<Trigger> trigger2) :
+            Trigger(), m_trigger1(trigger1), m_trigger2(trigger2) {}
+
+        bool compute(uint64_t timestep)
+            {
+            return m_trigger1->compute(timestep) ||
```
joaander, line 250: Consider using operator() here so that the child triggers will use cached results.

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+        std::shared_ptr<Trigger> getTrigger2() {return m_trigger2;}
+
+        /// Set the second trigger
+        void setTrigger2(std::shared_ptr<Trigger> trigger)
+            {
+            m_trigger2 = trigger;
+            }
+
+    protected:
+        std::shared_ptr<Trigger> m_trigger1;
```
joaander, line 273: Are `trigger1` and `trigger2` the most natural names for the child triggers?

hoomd/Trigger.h
bdice: Another way to do this would be to have `Before`, `On`, and `After` corresponding to `<`, `=`, and `>`. Then `or(before, on, after)` is always true. The language of "until" and "after" doesn't have a clear `=` to me.

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+    bool compute(uint64_t timestep)
+        {
+        if (timestep > m_after)
+            {
+            return true;
+            }
+        else
+            {
+            return false;
+            }
```
bdice, line 158:
```suggestion
        return timestep > m_after;
```

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+    bool compute(uint64_t timestep)
+        {
+        if (timestep < m_until)
+            {
+            return true;
+            }
+        else
+            {
+            return false;
+            }
```
bdice, line 125:
```suggestion
        return timestep < m_until;
```

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+    void setUntil(uint64_t until) {m_until = until;}
+
+    protected:
+        /// Always trigger until this timestep
+        uint64_t m_until;
+    };
+
+/** After trigger
+ *
+ *  Trigger every time step after and including `timestep > m_after`.
```
bdice, line 141: "and including"? Maybe I'm missing something. See other comments.

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+        }
+
+    /// Get the point where triggering will stop
+    uint64_t getAfter() {return m_after;}
+
+    /// Set the point where triggering will stop
+    void setAfter(uint64_t after) {m_after = after;}
+
+    protected:
+        /// Always trigger after this timestep
```
bdice, line 168: Make this comment specify "including" or not, depending on how you resolve other comments.

hoomd/Trigger.h
bdice: Agreed.
```suggestion
            return !m_trigger(timestep);
```

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+    public:
+        NotTrigger(std::shared_ptr<Trigger> trigger) :
+            Trigger(), m_trigger(trigger) {}
+
+        bool compute(uint64_t timestep)
+            {
+            return !(m_trigger->compute(timestep));
+            }
+
+        /// Get the trigger thats negated
```
bdice, line 187:
```suggestion
        /// Get the trigger that is negated
```

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+    {
+    public:
+        AndTrigger(std::shared_ptr<Trigger> trigger1,
+                   std::shared_ptr<Trigger> trigger2) :
+            Trigger(), m_trigger1(trigger1), m_trigger2(trigger2) {}
+
+        bool compute(uint64_t timestep)
+            {
+            return m_trigger1->compute(timestep) &&
+                   m_trigger2->compute(timestep);
```
bdice, line 211:
```suggestion
            return m_trigger1(timestep) && m_trigger2(timestep);
```

hoomd/Trigger.h
```diff
@@ -87,5 +103,176 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+    {
+    public:
+        OrTrigger(std::shared_ptr<Trigger> trigger1,
+                   std::shared_ptr<Trigger> trigger2) :
+            Trigger(), m_trigger1(trigger1), m_trigger2(trigger2) {}
+
+        bool compute(uint64_t timestep)
+            {
+            return m_trigger1->compute(timestep) ||
+                   m_trigger2->compute(timestep);
```
bdice, line 251:
```suggestion
            return m_trigger1(timestep) || m_trigger2(timestep);
```

hoomd/Trigger.h
bdice: I also wondered this -- I'd prefer to see a refactor for n-way triggers if it's not too tough. This should be straightforward using `<algorithm>` with `all_of` and `any_of`, assuming that you can make pybind work with variable numbers of arguments.

```
#include <algorithm>    // std::all_of
// ...
std::vector<std::shared_ptr<Trigger> > triggers;
triggers.push_back(some_trigger);
triggers.push_back(another_trigger);
triggers.push_back(third_trigger);
// ...
return std::all_of(triggers.begin(), triggers.end(), [](std::shared_ptr<Trigger> trigger){return trigger();}));

File: hoomd/trigger.py [hunk 13]
```diff
@@ -9,6 +9,37 @@ class Trigger(_hoomd.Trigger):
...
 
 
-class PeriodicTrigger(_hoomd.PeriodicTrigger, Trigger):
+class Periodic(_hoomd.PeriodicTrigger, Trigger):
     def __init__(self, period, phase=0):
         _hoomd.PeriodicTrigger.__init__(self, period, phase)
+
+
+class Until(_hoomd.UntilTrigger, Trigger):
+    def __init__(self, until):
```
bdice, line 18: Consider renaming `until` to `timestep`.

hoomd/trigger.py
```diff
@@ -9,6 +9,37 @@ class Trigger(_hoomd.Trigger):
...
+        if until < 0:
+            raise ValueError("until must be positive.")
+        else:
+            _hoomd.UntilTrigger.__init__(self, until)
+
+
+class After(_hoomd.AfterTrigger, Trigger):
+    def __init__(self, after):
```
bdice, line 26: Consider renaming `after` to `timestep`.

hoomd/Trigger.h
b-butler: We need to use `m_trigger->operator()(timestep)`.

hoomd/Trigger.h
b-butler: Agreed good call.

hoomd/Trigger.cc
```diff
@@ -42,5 +42,54 @@ void export_Trigger(pybind11::module& m)
...
+                    >(m, "NotTrigger")
+        .def(pybind11::init<std::shared_ptr<Trigger> >(),
+             pybind11::arg("trigger"))
+        .def_property("trigger",
+                      &NotTrigger::getTrigger,
+                      &NotTrigger::setTrigger)
+        ;
+
+    pybind11::class_<AndTrigger, Trigger, std::shared_ptr<AndTrigger>
+                    >(m, "AndTrigger")
```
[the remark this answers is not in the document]
b-butler: Yes it is possible and a nature extension.

hoomd/Trigger.h
```diff
@@ -27,10 +27,26 @@ class PYBIND11_EXPORT Trigger
...
+                m_last_trigger = triggered;
+                return triggered;
+                }
             }
+
+        virtual bool compute(uint64_t timestep) = 0;
+
+    private:
+            uint64_t m_last_timestep;
+            bool m_last_trigger;
```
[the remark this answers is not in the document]
b-butler: hmm, it makes sense, but I am not sure if it is in scope for now.

hoomd/Trigger.h
b-butler: I like that idea. This was just a first pass for these kinds of triggering.

hoomd/Trigger.h
b-butler: Agreed good call.

hoomd/Trigger.h
b-butler: I added the refactoring. I like the added flexibility it provides.

b-butler: I would prefer `All` and `Any`, but with the `All` filter, I think this is cleaner.

hoomd/Trigger.h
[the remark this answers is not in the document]
joaander: We could consider that for a later PR that updates all uses of `unsigned int timestep`. For now, we are implementing new functionality with `uint64_t` and there is no loss of generality because the calling functions that use `unsigned int` will have their values automatically converted up to `uint64_t` when they call a new method.

hoomd/Trigger.h
```diff
@@ -87,5 +108,185 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+*/
+class PYBIND11_EXPORT NotTrigger : public Trigger
+    {
+    public:
+        NotTrigger(std::shared_ptr<Trigger> trigger) :
+            Trigger(), m_trigger(trigger) {}
+
+        bool compute(uint64_t timestep)
+            {
+            return !(m_trigger->operator()(timestep));
```
joaander, line 201: If you think it is cleaner, you can also use:
```
(*m_trigger)(timestep)
```
I don't have a preference.

joaander:
Tests are failing:
```
        # test that the custom trigger can be called from c++
>       assert hoomd._hoomd._test_trigger_call(c, 0)
E       RuntimeError: Tried to call pure virtual function "Trigger::compute"
```

I think we need to replace `__call__` with `compute` in the CustomTrigger test.

hoomd/Trigger.h
b-butler: I don't either. Your suggestion is a bit shorter, but I think either is clear. I would vote to keep as is unless you or @bdice feel strongly about one of the styles.

hoomd/trigger.py
```diff
@@ -1,14 +1,58 @@
...
+
+
+class Not(_hoomd.NotTrigger, Trigger):
+    def __init__(self, trigger):
+        _hoomd.NotTrigger.__init__(self, trigger)
+
+
+class And(_hoomd.AndTrigger, Trigger):
+    def __init__(self, triggers):
+        if not hasattr(triggers, '__iter__') or isclass(triggers):
```
bdice, line 49: You might have a generator of triggers that is a class. I think that wouldn't work with this check in place. Consider:
```python
from collections.abc import Iterable

if isinstance(e, Iterable):
    # e is iterable
```
https://stackoverflow.com/questions/1952464/in-python-how-do-i-determine-if-an-object-is-iterable

bdice: You might try something like:
```python
class And(_hoomd.AndTrigger, Trigger):
    def __init__(self, triggers):
        triggers = list(triggers)  # resolves generators and other weird iterables into a list
        if not all(isinstance(t, Trigger) for t in triggers):  # ensure all items are triggers (skip this if the check can be done safely in C++!)
            raise ValueError("triggers must an iterable of Triggers.")
        _hoomd.AndTrigger.__init__(self, triggers)
```

hoomd/trigger.py
```diff
@@ -1,14 +1,58 @@
...
+            raise ValueError("triggers must an iterable of Triggers.")
+        _hoomd.AndTrigger.__init__(self, triggers)
+
+
+class Or(_hoomd.OrTrigger, Trigger):
+    def __init__(self, triggers):
+        if not hasattr(triggers, '__iter__') or isclass(triggers):
```
bdice, line 56: See comments on `And`.

sphinx-doc/module-hoomd-triggers.rst
```diff
@@ -6,7 +6,13 @@ hoomd.trigger
 .. autosummary::
     :nosignatures:
 
-    hoomd.trigger.PeriodicTrigger
+    hoomd.trigger.After
+    hoomd.trigger.Before
+    hoomd.trigger.And
```
bdice, line 11: Swap if this is supposed to be alphabetical.
```suggestion
    hoomd.trigger.And
    hoomd.trigger.Before
```

hoomd/trigger.py
```diff
@@ -1,14 +1,58 @@
 # Copyright (c) 2009-2019 The Regents of the University of Michigan
-# This file is part of the HOOMD-blue project, released under the BSD 3-Clause
 # License.
+# This file is part of the HOOMD-blue project, released under the BSD 3-Clause
```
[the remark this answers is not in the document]
bdice: Related: @joaander this could be less than 80 characters and fit on one line if we replace "the HOOMD-blue project" with "HOOMD-blue." I don't think there is any legal/other reason to prefer the longer form.
```python
# This file is part of HOOMD-blue, released under the BSD 3-Clause License.
```

hoomd/Trigger.h
```diff
@@ -87,5 +108,185 @@ class PYBIND11_EXPORT PeriodicTrigger : public Trigger
...
+        }
+
+    /// Get the point where triggering will stop
+    uint64_t getTimestep() {return m_timestep;}
+
+    /// Set the point where triggering will stop
+    void setTimestep(uint64_t timestep) {m_timestep = timestep;}
+
+    protected:
+        /// Always trigger timestep this timestep
```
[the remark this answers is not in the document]
b-butler: It is not, but now it is.

hoomd/trigger.py
b-butler: That seems clean.

hoomd/pytest/test_trigger.py
```diff
@@ -25,22 +25,22 @@ def test_periodic_properties():
...
+    a = hoomd.trigger.Periodic(period=456, phase=18)
 
     for i in range(10000):
         assert a(i) == ((i - 18) % 456 == 0)
 
     for i in range(10000000000, 10000010000):
         assert a(i) == ((i - 18) % 456 == 0)
 
-    b = hoomd.trigger.PeriodicTrigger(period=10000000000, phase=6000000000)
+    b = hoomd.trigger.Periodic(period=10000000000, phase=6000000000)
```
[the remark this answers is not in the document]
b-butler: Done.

PR #629, opened 2020-04-14
Feature/csv backend

b-butler:
## Description

This adds a Python implementation of a CSV (technically this is a space separated values file) logger backend. Only string and scalar values are allowed for logging.

## Motivation and Context

This provides a useful way to get a basic text based logger output for a file or `stdout`.

## How Has This Been Tested?

This has been experimented with on a jupyter notebook locally.

## Note
This must be merged after #628.

## Checklist:


hoomd/output/csv_backend.py
```diff
@@ -0,0 +1,190 @@
...
+        if new_keys != self._cur_headers.keys():
+            self._cur_headers = {key: max(len(key), self._min_width)
+                                 for key in new_keys}
+            self._write_header()
+
+        self._write_row(output_dict)
+
+    def _write_header(self):
+        headers = self._cur_headers
+        self._output.write(
```
joaander, line 142: Write output only on rank 0.

Also, when using actual file backed output, we will need to provide some mechanism or example showing how to open files only on the root rank.

hoomd/output/csv_backend.py
```diff
@@ -0,0 +1,190 @@
...
+            ' '.join((self._fmt.format_str(k, headers[k]) for k in headers))
+            )
+        self._output.write('\n')
```
joaander, line 145: Flush the output to ensure it is written.

hoomd/output/csv_backend.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def _write_row(self, data):
+        headers = self._cur_headers
+        self._output.write(' '.join((self._fmt(data[k], headers[k])
+                                     for k in headers))
```
joaander, line 150: Are all columns the same width? What sets the column width?

hoomd/output/csv_backend.py
```diff
@@ -0,0 +1,190 @@
...
+                           )
+        self._output.write('\n')
```
joaander, line 152: Flush the output.

hoomd/output/csv_backend.py
b-butler: All columns are not the same width, currently we set it to the max of the header size or the minimum allowed width.

hoomd/output/csv_backend.py
b-butler: Good observation, and yes we will. I imagine that this will have to be something we just show users. We could also duck type the output to accept a string as well which we can open internally with an optional mode argument.

hoomd/logging.py
```diff
@@ -102,9 +102,9 @@ def update_cls(self, cls):
 class Logger(SafeNamespaceDict):
     '''Logs HOOMD-blue operation data and custom quantities.'''
```
joaander, line 103: Document parameters to `__init__` and give a more detailed explanation of `Logger`.

hoomd/output/csv_backend.py
```diff
@@ -0,0 +1,295 @@
...
+        self.output.write('\n')
+
+    def act(self, timestep=None):
+        """Write row to designated output.
+
+        Will also write header when logged quantites are determined to have
+        changed.
+        """
+        if self._comm is not None and self._comm.rank == 0:
+            output_dict = self._get_log_dict()
```
joaander, line 234: This won't work for loggable properties that make MPI collective calls. Those properties need to be requested on all ranks. I think you can just move the `_get_log_dict` outside the `if`.

hoomd/output/csv_backend.py
joaander: OK, that is simple enough. IT will remain to be seen if users request something more complex.

b-butler: Yep, and luckily if it is too different, we can tell them to write their own back end now.

hoomd/logging.py
b-butler: This is covered in #679.

hoomd/output/csv_backend.py
b-butler: Good catch!

PR #634, opened 2020-04-23
Particle filter tests

danevans09:
## Description
This PR will test the particle filters. 

- Added the __call__ method to the _ParticleFilter class so the filters can be called 
- Wrote tests for type filter, tag filter, set union filter, set intersection filter, and set difference filter.

## Motivation and Context

The particle filters appear to be working in jupyter notebooks but need formal tests.

## How Has This Been Tested?

I tested that the output of the filter (particle tags) correctly matched either the input tags or the tags of particles of the input type(s)

## Change log

```

```

## Checklist:


hoomd/conftest.py
```diff
@@ -105,7 +105,7 @@ def make_snapshot(particle_types=['A'], dimensions=3, d=1, L=20):
                 box[2] = 1
             s.configuration.box = box
             s.configuration.dimensions = dimensions
-
+
```
joaander, line 108:
```suggestion

```

hoomd/filter/filter_.py
```diff
@@ -11,4 +14,5 @@ def __str__(self):
 
     def __call__(self, state):
```
joaander, line 15: Remove this `__call__` method and instead export `getSelectedTags` as `__call__` in `export_filters.cc`.

joaander: Wait, never-mind. I realize now that this needs access to the python only `state`.

joaander:
Reviewers, please address these questions in your in addition to any line-specific comments.
* **Is `filter(sim.state) -> list of tags` the right API for users to call filters?** If not, what would you suggest?
* **Do the tests provide sufficient coverage of the filters?** If not, what additional tests would you suggest?
* **Is the code readable?** Or there areas that need additional comments to explain the operation of this code to future developers?

hoomd/filter/filter_.py
```diff
@@ -1,4 +1,7 @@
-class _ParticleFilter:
+from hoomd._hoomd import ParticleFilter
+
+
+class _ParticleFilter(ParticleFilter):
```
vyasr, line 4: Import `collections.abc.Hashable` and change this to 
```suggestion
class _ParticleFilter(Hashable, ParticleFilter):
```
then you can get rid of the lines below.

hoomd/filter/filter_.py
```diff
@@ -11,4 +14,5 @@ def __str__(self):
 
     def __call__(self, state):
         """Needs to interact with state to get particles across MPI rank."""
-        raise NotImplementedError
+        return self._get_selected_tags(state._cpp_sys_def)
+        # raise NotImplementedError
```
vyasr, line 18: Remove this line.

hoomd/filter/filter_.py
danevans09: I'm getting a `TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases` if I try to inherit Hashable.

hoomd/pytest/test_filter.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def test_tags_filter(make_filter_snapshot, simulation_factory, tag_indices):
+    particle_types = ['A']
+    N = 10
+    filter_snapshot = make_filter_snapshot(n=N, particle_types=particle_types)
+    sim = simulation_factory(filter_snapshot)
+    inds = tag_indices
+    tag_filter = Tags(inds)
+    assert tag_filter(sim.state) == inds
```
danevans09, line 87: The tag filter tests are failing because the tag filter is returning an empty list. I'm not sure why

b-butler:
@vyasr, with pybind11 we can't use abstract base classes or any metaclasses with C++ objects. To use ABC we have to wrap the C++ object, which in the case of filters and triggers we don't.

hoomd/filter/filter_.py
b-butler: @danevans09 revert the change, this won't work with pybind.

vyasr: @b-butler Is this actually a pybind problem? I didn't think through that aspect, and it's probably not worth the hassle, but naively I would still expect something like this to work:
```
class _ParticleFilterMetaclass(type(Hashable), type(ParticleFilter)):
    pass

class _ParticleFilter(Hashable, ParticleFilter, metaclass=_ParticleFilterMetaclass):
...
```

Does that kind of metaclass hackery not play nicely with pybind? This is more for my edification, it's probably more complexity than is merited for this problem since `Hashable` is such a minimal `abc` anyway.

b-butler: @b-butler It may or it may not. pybind does not seemingly document this very well. They have a custom metaclass called `pybind11_builtins.pybind11_type` which is used. They don't specify if they overwrite `__new__` or `__init__` although I suspect they do for at least `__init__`. This lack of documentation makes it impossible to know how to properly create the hybrid metaclass (i.e. do we need to create our own `__new__` and `__init__` that we use to call `super().__new__` and such.

They do not document how to use metaclasses from Python with their metaclass, and I do not know if it is stable to rely on this even if it works. We could probably make this work through trial and error, but then is that something to trust will be stable.

vyasr: Lol you tagged yourself instead of me. But OK I get the picture, basically it might work or it might not, and not worth relying on it if pybind isn't making any promises. Fine by me. It was inevitable that there would be some gotchas in trying to inherit C++ classes in Python.

b-butler: Yes, yes I did...

hoomd/filter/filter_.py
```diff
@@ -11,4 +14,4 @@ def __str__(self):
 
     def __call__(self, state):
         """Needs to interact with state to get particles across MPI rank."""
-        raise NotImplementedError
+        return self._get_selected_tags(state._cpp_sys_def)
```
b-butler, line 17: this will fail on MPI, I think. The `ParticleFilters` were only designed to get the particles local to a rank. The union happens at the `ParticleGroup` level. We can change this of course. We could also just create a function in `hoomd.State` that grabs the particle group after updating as well.

Thoughts @joaander @vyasr?

vyasr: I don't think I know enough about how this part is planned to work going forward. It looks like there is now a single group class that is effectively constructed from a filter, is the idea that filters are more lightweight and easy to work with from python but groups must be constructed for practical usage with e.g. methods? I would avoid creating the kind of cyclic references it seems like would be introduced if you had to go Filter->State->Group->Filter, but I could be misunderstanding.

b-butler: The groups are required for C++. C++ groups own a filter. So the execution would go something like this Filter (Python) -> State (Python) -> Group (C++) -> Filter (C++) and then the C++ group would perform the union on all the rank specific tags from the filter.

The other option is to pass the MPI communicator with the state and make the call in Python `def __call__(self, simulation)`.

b-butler:
Currently MPI makes the test on the `Type` filter fail. This is due to it only gathering particles on a local rank basis. We will have to decide whether to utilize the groups the `State` holds or to give the ability to gather across ranks to the filters. If the second we need to make the Python signature `__call__(self, simulation)` so we can grab the communicator.

hoomd/filter/filter_.py
joaander: We can either a) document that filters are performed locally only in conjunction with the to-be-implemented local state API. Or b) Implement a (possibly alternate) code path in ParticleFilter that performs the union from the rank outputs. The communicator needed to do this is already available in `state._simulation.device`. https://github.com/glotzerlab/hoomd-blue/blob/619f8a99dcead6f0118e09669b42999d6792d97e/hoomd/ParticleGroup.cc#L374-L397

b-butler: We could do something like this,

`def __call__(self, state, collect=False)`. Then we could default to local tags only, but we could provide an option to collect tags across ranks (we could limit this collecting to rank 0 if desired). If we do this, we should remove the logic in `ParticleGroup`s though.

hoomd/pytest/test_filter.py
danevans09: Fixed in 5b522

hoomd/pytest/test_filter.py
```diff
@@ -35,7 +35,12 @@ def type_indices(request):
     return deepcopy(request.param)
 
 
-def test_type_filter(make_filter_snapshot, simulation_factory, type_indices):
+def test_type_filter(make_filter_snapshot,
```
joaander, line 38: Please use the `@pytest.mark.serial` function decorator.

danevans09: Fixed 4b2543522b5b98d76daaedabafd12a63b8401bfb

PR #635, opened 2020-04-27
Feature/autodoc extensions

vyasr:
## Description

The `_*.py` files used to mock the various extension modules have been removed and Sphinx's  `conf.py` has been updated to use autodoc for mocking instead.

## Motivation and Context

Currently hoomd maintains mocked up `.py` files for all extension modules to allow Sphinx to successfully import and build the documentation. However, Sphinx has had built-in support for this kind of mocking, so we should take advantage of this to avoid having to maintain these extra files ourselves. The only additional change that would have to be made is that `bumpversion` would need to update `hoomd/__init__.py` because we can't depend on being able to import the version from `_hoomd` in `__init__.py`, but that duplication is IMO a very small price to pay (and maybe there's a way around it, depending on how important it is to have the versions at the C++ level.

## Change log

```
Extension modules that were mocked for the purpose of building documentation have been removed and replaced with usage of autodoc.
```

## Checklist:


joaander:
I tried this in fresnel and it is working there. What more needs to be done to get this working in HOOMD.

vyasr:
@joaander I've updated this PR to remove the mocked extensions. The one outstanding item is deciding how to handle versioning. Currently, the `_hoomd.py` file contains a bogus version that makes import of hoomd work, but that is no longer possible so I added the version number to `hoomd/__init__.py`. An alternative would be to put a try/except block there so that the version is just unassigned when importing a docs-only uncompiled build. I don't think it makes a difference since the version in the docs depends only on what's in conf.py, but that's up to you.

The docs tests all fail right now, but it looks like it's just the 3.0 API changes; the build succeeds locally if you don't fail on all warnings.

hoomd/__init__.py
```diff
@@ -87,4 +87,5 @@ def _hoomd_sys_excepthook(type, value, traceback):
 
 sys.excepthook = _hoomd_sys_excepthook
 
-__version__ = "{0}.{1}.{2}".format(*_hoomd.__version__)
+#  __version__ = "{0}.{1}.{2}".format(*_hoomd.__version__)
+__version__ = "3.0.0"
```
joaander, line 91: Set this to 2.9.0 and add `__init__.py` to the bumpversion configuration in `setup.cfg`. Then bumpversion will keep it up to date with the others.

vyasr: 👍 done. I just wanted to confirm that you were fine with maintaining the version separately in Python.

joaander:
Thanks!

PR #646, opened 2020-05-08
Feature/new nvt method

tommy-waltmann:
## Description

Porting NVT integration method to the new v3 API. We haven't decided on how to handle the `randomize_velocities()` API in v3 yet, so this draft is to facilitate that discussion.

## Motivation and Context

We need to port everything to the new API, and this is one of those things

## How Has This Been Tested?

No tests or changes to documentation

## Change log

```
Moved NVT integration method to v3 API
```

## Checklist:

### DIFF
--- a/hoomd/md/methods.py
+++ b/hoomd/md/methods.py
@@ -108,14 +108,18 @@ class NVT(_Method):
         # initialize the reflected cpp class
         if not simulation.device.cpp_exec_conf.isCUDAEnabled():
             my_class = _md.TwoStepNVTMTK
+            thermo_cls = _hoomd.ComputeThermo
         else:
             my_class = _md.TwoStepNVTMTKGPU
+            thermo_cls = _hoomd.ComputeThermoGPU
 
         group = simulation.state.get_group(self.filter)
-        self._cpp_obj = my_class(simulation.state._cpp_sys_def,
+        cpp_sys_def = simulation.state._cpp_sys_def
+        self._cpp_obj = my_class(cpp_sys_def,
                                  group,
-                                 hoomd.compute.thermo(group).cpp_compute,
-                                 self.kT)
+                                 thermo_cls(cpp_sys_def, group, ""),
+                                 self.kT,
+                                 "")
         super().attach(simulation)
 
     def thermalize_velocities(self, seed):
--- a/hoomd/md/methods.py
+++ b/hoomd/md/methods.py
@@ -115,9 +115,11 @@ class NVT(_Method):
 
         group = simulation.state.get_group(self.filter)
         cpp_sys_def = simulation.state._cpp_sys_def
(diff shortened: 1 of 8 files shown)


hoomd/md/pytest/test_methods.py
```diff
@@ -0,0 +1,20 @@
+import hoomd
+import pytest
+import numpy
+import itertools
+
+def attach_and_sim(simulation_factory, two_particle_snapshot_factory, method):
```
tommy-waltmann, line 6: Can someone comment on the best way to write this test so that I can easily plug in each integration method and have the test run on it? I'm not that familiar with pytest, and while this will get the job done, I think there is probably a better way.

joaander: Use a parameterized fixture for the method: https://docs.pytest.org/en/latest/fixture.html#parametrizing-fixtures

joaander:
Thanks, Tommy!

Add
```
    #ifdef ENABLE_MPI
    .def("setCommunicator", &Compute::setCommunicator)
    #endif
```
to the pybind export in `IntegrationMethodTwoStep.cc` so the MPI tests will pass.

hoomd/md/pytest/test_methods.py
```diff
@@ -0,0 +1,23 @@
...
+import numpy
+import itertools
+from copy import deepcopy
+
+_int_methods = [
+    hoomd.md.methods.Langevin(hoomd.filter.All(), kT=1, seed=1),
+    hoomd.md.methods.NVT(hoomd.filter.All(), kT=1, tau=1),
+]
+
+@pytest.fixture(scope='function', params=_int_methods)
```
joaander, line 12: Use the `ids` parameter so the tests get a useful name in the output. Without the ids, the tests get named like this:
```
hoomd/md/pytest/test_methods.py::test_attach[CPU-int_method0] PASSED                                                    [ 25%]
hoomd/md/pytest/test_methods.py::test_attach[CPU-int_method1] PASSED                                                    [ 50%]
```
When it could say `CPU-Langevin` and `CPU-NVT`.

hoomd/md/methods.py
```diff
@@ -91,70 +91,40 @@ class nvt(_Method):
...
         # store metadata
-        self.group = group
-        self.kT = kT
-        self.tau = tau
-        self.metadata_fields = ['group', 'kT', 'tau']
+        param_dict = ParameterDict(
+            filter=OnlyType(_ParticleFilter),
+            kT=create_variant,
+            tau=float(tau),
+            explicit_defaults=dict(kT=kT, filter=filter)
```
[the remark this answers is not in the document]
b-butler: I recognize that I made that mistake in `Langevin`, I changed the behavior of `ParameterDict` some time back to enable some necessary functionality.

PR #647, opened 2020-05-11
Simulation tests

danevans09:
## Description

I made a test_simulation file to test that the simulation class runs, counts timesteps correctly, and initializes from gsd files correctly.

## Motivation and Context

There are currently no tests specifically for the simulation class in the new API.

## How Has This Been Tested?

I tested that simulations initialize correctly and throw the expected errors. I tested that the timestep counter is consistent with the `sim.run(nsteps)` calls. Lastly, I tested that all of the particle data is passed through the state_from_gsd method correctly.

## Change log

```

```

## Checklist:


hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,158 @@
...
+    d = tmp_path / "sub"
+    d.mkdir()
+    filename = d / "temporary_test_file.gsd"
+    file = gsd.hoomd.open(name=filename, mode='wb+')
+    sim = simulation_factory(get_snapshot(n=snap_params[0],
+                                          particle_types=snap_params[1]))
+
+    snap = sim.state.snapshot
+    box = sim.state.box
+    assert_equivalent_snapshots(make_gsd_snapshot(snap), snap)
```
joaander, line 135: This is effectively testing `create_state_from_snapshot` which should be in its own method.

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def test_state_from_gsd(simulation_factory, get_snapshot,
+                        device, state_args, tmp_path):
+    snap_params, nsteps = state_args
+
+    d = tmp_path / "sub"
+    d.mkdir()
+    filename = d / "temporary_test_file.gsd"
+    file = gsd.hoomd.open(name=filename, mode='wb+')
```
joaander, line 129: Use `with gsd.hoomd.open(name=filename, mode='wb+') as file:` to ensure that the file is written and closed before `Simulation` tries to read it. Perform all the file writes while the file is open and save the snapshots. Then test all of the reads after you are done writing the file.

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,158 @@
+import hoomd
+import numpy as np
+import pytest
+from copy import deepcopy
+import gsd.hoomd
```
joaander, line 5: We need to catch an import error here and skip the gsd test if the module is not available.

hoomd/pytest/test_simulation.py
danevans09: Fixed 491d64d72

hoomd/pytest/test_simulation.py
danevans09: Fixed fe9754f7e

hoomd/pytest/test_simulation.py
danevans09: Ok, should I delete this then because `create_state_from_snapshot` is being tested in _test_state.py_?

joaander:
Great, it looks like the skip is working. I'll add gsd to the test image later on `feature/new-object-API`

@b-butler any thoughts to finish this one off?

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,172 @@
...
+    return make_snapshot
+
+
+def make_gsd_snapshot(hoomd_snapshot):
+    s = gsd.hoomd.Snapshot()
+    snap_properties = []
+    all_attr = dir(hoomd_snapshot)
+    for att in all_attr:
+        if att[0] != '_' and att not in ['exists', 'replicate']:
+            snap_properties.append(att)
```
b-butler, line 31: This is fine, but it may be a good opportunity to use the list comprehension syntax. I don't care too much either way though.

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,172 @@
...
+        if s.exists:
+            s.configuration.box = [20, 20, 20, 0, 0, 0]
+            s.particles.N = n
+            s.particles.position[:] = np.random.uniform(-10, 10, size=(n, 3))
+            s.particles.types = particle_types
+        return s
+    return make_snapshot
+
+
+def make_gsd_snapshot(hoomd_snapshot):
```
b-butler, line 25: Instead of creating the list of top level properties, why don't you do something like,

````python
for att in all_attr:
    if cond:
        for prop in dir(getattr(hoomd_snapshot, att)):
            if cond:
                setattr()
````

File: hoomd/pytest/test_simulation.py (hunk 4 above)
Comment (b-butler) [reply hoomd/pytest/test_simulation.py:31]: Disregard this comment.

File: hoomd/pytest/test_simulation.py [hunk 6]
```diff
@@ -0,0 +1,172 @@
...
+                np.testing.assert_allclose(getattr(snap1_prop, nested_prop),
+                                           getattr(snap2_prop, nested_prop))
+
+
+def assert_equivalent_boxes(box1, box2):
+    metadata1 = box1.get_metadata()
+    metadata2 = box2.get_metadata()
+    for key in metadata1:
+        assert metadata1[key] == metadata2[key]
+
```
b-butler, line 91: This will break with the `feature/new-box-resize` PR #618. I would check their `L` and `tilts` directly.

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,172 @@
...
+
+def random_inds(n):
+    inds = []
+    for i in range(n):
+        if np.random.rand() > 0.5:
+            inds.append(i)
+    return inds
```
b-butler, line 98:
```suggestion
    return np.random.choose(np.arange(n),
                            size=int(n * np.random.rand()),
                            replace=False)

```

hoomd/pytest/test_simulation.py
danevans09: Fixed 68371cc68

hoomd/pytest/test_simulation.py
danevans09: Fixed 055059e48

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,156 @@
...
+                        assert getattr(getattr(snap1, attr), prop) == \
+                            getattr(getattr(snap2, attr), prop)
+                    else:
+                        np.testing.assert_allclose(getattr(getattr(snap1, attr),
+                                                           prop),
+                                                   getattr(getattr(snap2, attr),
+                                                           prop))
+
+
+def assert_equivalent_boxes(box1, box2):
```
b-butler, line 70: When the `feature/new-box-resize` gets merged we can just use `box1 == box2`. Unfortunately for now this is required though.

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,156 @@
+import hoomd
+import numpy as np
+import pytest
+from copy import deepcopy
+try:
+    import gsd.hoomd
+    skip_gsd = False
+except ImportError:
+    skip_gsd = True
+
```
b-butler, line 10:
```suggestion
    import gsd.hoomd
    skip_gsd = False
except ImportError:
    skip_gsd = True
skip_gsd = pytest.mark.skipif(skip_gsd, reason="gsd Python package was not found.")
```

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,156 @@
...
+               ((5, ['A', 'B']), 20),
+               ((50, ['A', 'B', 'C']), 4),
+               ((100, ['A', 'B']), 30)]
+
+
+@pytest.fixture(scope="function", params=_state_args)
+def state_args(request):
+    return deepcopy(request.param)
+
+
```
b-butler, line 124:
```suggestion

@skip_gsd
```

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,156 @@
...
+def test_state_from_gsd(simulation_factory, get_snapshot,
+                        device, state_args, tmp_path):
+    if skip_gsd:
+        pytest.skip('gsd module is not available')
```
b-butler, line 128:
```suggestion
```

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,156 @@
...
+
+    d = tmp_path / "sub"
+    d.mkdir()
+    filename = d / "temporary_test_file.gsd"
+    with gsd.hoomd.open(name=filename, mode='wb+') as file:
+        sim = simulation_factory(get_snapshot(n=snap_params[0],
+                                              particle_types=snap_params[1]))
+        snap = sim.state.snapshot
+        box = sim.state.box
+
```
b-butler, line 139:
```suggestion
```
You have this exact line below.

hoomd/pytest/test_simulation.py
```diff
@@ -97,11 +101,21 @@ def test_initialization(device, simulation_factory, get_snapshot):
...
-def test_run(simulation_factory, get_snapshot):
-    sim = simulation_factory(get_snapshot())
-    sim.operations.schedule()
+
+def test_run(simulation_factory, get_snapshot, device):
+    sim = hoomd.simulation.Simulation(device)
+    assert sim.timestep is None
+    sim.timestep = 10
+    sim = simulation_factory(get_snapshot())  # set state
     assert sim.timestep == 0
```
b-butler, line 114: Shouldn't this fail now? Since we are setting the timestep to 10?

danevans09: I believe `timestep` is being set to 0 once `sim.state` is set. I'm not sure if this is the expected behavior

b-butler: This is not. Must happen in the construction of the `System` object.

hoomd/pytest/test_simulation.py
joaander: yes

hoomd/pytest/test_simulation.py
joaander: The sim object is being reset to a new simulation by `simulation_factory`, overriding the previous object.
```
    sim = hoomd.simulation.Simulation(device)
    sim.timestep = 10
    sim = simulation_factory(get_snapshot())  # set state
    assert sim.timestep == 0
```
I think you mean:
```
    sim = hoomd.simulation.Simulation(device)
    sim.timestep = 10
    sim.create_from_snapshot(get_snapshot())
    assert sim.timestep == 0
```

danevans09: `sim.timestep` is still being set to 0 when I use `create_state_from_snapshot`. I believe this is because `sim.timestep` returns None until `_cpp_sys` is set, while `sim._timestep` returns what it is set to (in this case 10).

If I change:
```
def create_state_from_snapshot(self, snapshot):
        # initialize the system
        # Error checking
        if self.state is not None:
            raise RuntimeError("Cannot initialize more than once\n")

        self._state = State(self, snapshot)

        step = 0
        if self.timestep is not None:
            step = self.timestep

        # Store System and Reader for Operations
        self._cpp_sys = _hoomd.System(self.state._cpp_sys_def, step)
        self._init_communicator()
        self._cpp_sys.enableQuietRun(not self.verbose_run)
```

to
```
def create_state_from_snapshot(self, snapshot):
        # initialize the system
        # Error checking
        if self.state is not None:
            raise RuntimeError("Cannot initialize more than once\n")

        self._state = State(self, snapshot)

        step = 0
        if self._timestep is not None:
            step = self._timestep

        # Store System and Reader for Operations
        self._cpp_sys = _hoomd.System(self.state._cpp_sys_def, step)
        self._init_communicator()
        self._cpp_sys.enableQuietRun(not self.verbose_run)
```

and set `self._timestep` to None in `__init__`, the timestep gets set properly.

joaander: Then we should fix the `timestep` property. If the user sets it, it should no longer return `None`, but the value that the user set.

hoomd/simulation.py
```diff
@@ -103,8 +104,8 @@ def create_state_from_snapshot(self, snapshot):
         self._state = State(self, snapshot)
 
         step = 0
-        if self.timestep is not None:
-            step = self.timestep
+        if self._timestep is not None:
```
joaander, line 107: Fix this in the `timestep` property, as that will solve the problem for `snapshot`, `gsd`, and all future `create_state_from_*` methods.

danevans09: Fixed 7f0d69ad9

hoomd/pytest/test_simulation.py
danevans09: Fixed 7f0d69ad9

hoomd/pytest/test_simulation.py
```diff
@@ -0,0 +1,196 @@
...
+               ((100, ['A', 'B']), 30)]
+
+
+@pytest.fixture(scope="function", params=_state_args)
+def state_args(request):
+    return deepcopy(request.param)
+
+
+@skip_gsd
+def test_state_from_gsd(simulation_factory, get_snapshot,
```
joaander, line 169: This test fails in MPI: https://dev.azure.com/glotzerlab/hoomd-blue/_build/results?buildId=3558&view=logs&j=959d7a94-bee1-5f30-271f-627c279bcb55&t=e40ef08a-fd92-5fa7-27f4-e5706e667ef3&l=332

joaander:
MPI tests are still failing. You need to use `if snap.exists:` checks around any code that writes to snapshot arrays.

joaander:
Thanks!

Issue #552, opened 2019-12-06
New Logger API


joaander:
## Description

Replace `analyze.log` with a more complete and documented system. The new Logger system should be able to log vector quantities (i.e. force on each particle, 6-component pressure tensor, etc...). **ALL** loggable quantities should also be available via the user facing Python API. Thus, the natural implementation of the logger is in Python.

## Proposed solution
@b-butler and I are currently designing and implementing this. The overall structure is clear, but the specific form of the API calls is still under development.

### Properties

Loggable quantities will be implemented as properties. Examples:
```
lj.energy
lj.energies
thermo.pressure
mc.overlaps
```
Users can access these properties directly in their python script for real time access. There will be no need to create a logger with period=None and query it as in hoomd v2.

### Logger

A `Logger` object assembles a list of log quantities from any number of objects. Adding an object to a logger adds all of the loggable quantities by default:
```
Logger log()
log += mc
log += thermo
```
Users can specify only specific quantities if desired:
```
log.add_quantity(lj, lj.exports['energy'])
```

### Backends

Multiple backends may be implemented on top of `Logger`. We plan to implement a GSD backend first. This will enable logged quantities in GSD files to be read by OVITO. If the user wants a pure log output (no particles) at a higher rate that the trajectory output, they can pass the `None` particle filter to the GSD dump.

We also expect an HDF5 backend as a rewrite of the current HDF5 logger in HOOMD. Others are certainly possible, such as `pandas`.

### Data types

Supported data types include floats, ints, strings, 1- and 2- dimensional numpy arrays (and types convertible to numpy arrays). This choice is motivated by the support for data types in GSD and HDF5.

We will also support dicts and nested dicts that contain these base data types. More on this below.

### Namespaces

The new logging mechanism will maintain the namespace structure of the python library. While verbose, this makes a clear connection between a logged value in a file and where it came from in the simulation. In the current hoomd v2 system it is difficult to know what named quantity came from where, especially when many logged quantities are not even documented. At the Python level, the namespace is stored as a tuple, so that different backends may apply that in a way natural to their format: GSD will use /, HDF5 might use groups. I will use `/` here.

The namespace consists of several parts, the base, the class namespace, the private space and variables. Here are some examples:
```
hpmc/integrate/Sphere/overlap
md/pair/LJ/energy
particles/md/pair/LJ/energies
md/compute/Thermo/pressure
md/compute/Thermo_1/pressure
```
[...]

PR #652, opened 2020-05-13
WIP: add first try HDF5 logger, based on GSDLogWriter and h5py

InnocentBug:
## Description

The new logging API requires a reimplementation of the HDF5 logging.
I based my changes very closely on the GSDLogWriter.
In fact, I am inheriting from the GSDLogWriter.
(This is technically incorrect. There is not "is-a" relation. Probably both logger should inherit from a common parent that implements everything except the `write_frame` function, which both should implement on there own.)

## Motivation and Context
The new object API make the old HDF5 logging obsolete.
Hopefully, I can provide an updated implementation.
In reference to: #552 

I have to admit, that I didn't understand the GSDLoGWriter completely.
There are two notable gaps in my understanding.
1) who calls the _write_frame function and how is the GSDDumpWriter initialized for that.
(For the HDF5 logging, I expect now an HDF5 file that was opened on the root rank. It has to be an instance of hdf5.File. Similar to the old implementation.)
2) Does the old API from Updaters and Computes survives the change? And if yes, where are the getLogQuantity or getLogMatrixQuantitiy functions called?

Maybe it isn't necessary that I understand this for the implementation.
But I actually would like to understand it... 

## How Has This Been Tested?
It hasn't been tested at all so far.
If you have a test for the GDDLogWriter, please point it out to me.
I would want to base the test very closely on this test. (It essentially does the same with a different backend.)

## Change log

```
HDF5 logging for the new Logger API
```

## Checklist:

- [ x I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).

### DIFF
--- a/hoomd/hdf5.py
+++ b/hoomd/hdf5.py
@@ -12,6 +12,7 @@ you can use any command.
 from hoomd import _hoomd
 from hoomd.analyze import _analyzer
 import hoomd
+import hoomd.dump
 import numpy
 import os
 
@@ -358,3 +359,58 @@ class log(hoomd.analyze._analyzer):
                             hoomd.context.current.device.cpp_msg.error("Quantity position is not vacant.")
                             raise RuntimeError("Error updating quantities with log_hdf5.")
                     data_set.attrs[quantities[i]] = i
+
+
+# For the HDF5 logger, I inherit from the GSDLogWriter.
+# This makes the assumption that the GSDLogWriter creates with the log function
+# a dictionary that has the names as keys and np.arrays as values.
+# HDF5 has the restriction that the dimension of the matrices cannot change.
+# This translates to the restriction that the dimensions of the logged quantities cannot change during a simulation run.
+class HDF5LogWriter(hoomd.dump.GSDLogWriter):
+    def __init__(self,logger):
+        super().__init__(logger)
+
+    def _write_frame(self,h5file):
+
+        f = None
+        if hoomd.context.current.device.comm.rank == 0:
+        f = self.h5file
+
+        log_dict = self.log()
+        for key in log_dict.keys():
+            if f is not None:  # Only the root rank further process the received data
+                data = log_dict[key]
+                # Check the returned object
+                if not isinstance(data, numpy.ndarray):
+                    hoomd.context.current.device.cpp_msg.error("For quantity " + key + " no numpy array to log obtainable.")
+                    raise RuntimeError("Error writing matrix quantity " + key)
+
+                if key not in f:
+                    # Create a new container in hdf5 file, if not already existing.
+                    data_set = f.create_dataset(key, shape=(0,) + data.shape, maxshape=(None,) + data.shape)
+                else:
+                    data_set = f[key]
+
+                    # check compatibility of data in file and returned matrix
+                    if len(data.shape) + 1 != len(data_set.shape):
+                        msg = "Trying to log numpy array " + key + ", but dimensions are incompatible with "
+                        msg += "dimensions in file."
+                        hoomd.context.current.device.cpp_msg.error(msg)
+                        raise RuntimeError("Error writing matrix quantity " + key)
+
+                    for i in range(len(new_matrix.shape)):
+                        if data_set.shape[i + 1] != new_matrix.shape[i]:
+                            msg = "Trying to log numpy array " + key + ", but dimension " + str(i) + " is "
+                            msg += "incompatible with  dimension in file."
+                            hoomd.context.current.device.cpp_msg.error(msg)
+                            raise RuntimeError("Error writing matrix quantity " + key)
+
+                old_size = data_set.shape[0]
+
+                data_set.resize(old_size + 1, axis=0)
+                data_set[old_size, ] = data
+
+
+        if hoomd.context.current.device.comm.rank == 0:
+            # Flush the file after each write to maximize integrity of written data
+            f.flush()
--- a/hoomd/hdf5.py
+++ b/hoomd/hdf5.py
@@ -367,29 +367,30 @@ class log(hoomd.analyze._analyzer):
 # HDF5 has the restriction that the dimension of the matrices cannot change.
 # This translates to the restriction that the dimensions of the logged quantities cannot change during a simulation run.
 class HDF5LogWriter(hoomd.dump.GSDLogWriter):
-    def __init__(self,logger):
+    def __init__(self,logger,h5file):
         super().__init__(logger)
+        self.h5file = h5file
 
-    def _write_frame(self,h5file):
+    def log(self):
 
-        f = None
+        file_handle = None
         if hoomd.context.current.device.comm.rank == 0:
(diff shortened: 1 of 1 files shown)


InnocentBug:
If I am going the correct direction here, I can fix the inheritance issue I mention in the description.

InnocentBug:
Ok, in that case I will just wait a little bit and base my implementation on #629

hoomd/hdf5.py
```diff
@@ -358,3 +359,58 @@ def _write_header(self, f):
...
+                        msg += "dimensions in file."
+                        hoomd.context.current.device.cpp_msg.error(msg)
+                        raise RuntimeError("Error writing matrix quantity " + key)
+
+                    for i in range(len(new_matrix.shape)):
+                        if data_set.shape[i + 1] != new_matrix.shape[i]:
+                            msg = "Trying to log numpy array " + key + ", but dimension " + str(i) + " is "
+                            msg += "incompatible with  dimension in file."
+                            hoomd.context.current.device.cpp_msg.error(msg)
+                            raise RuntimeError("Error writing matrix quantity " + key)
```
b-butler, line 406: Is this a hold over from the previous HDF5 logger? The variable `new_matrix` is not defined in this function, and the error message is similar to what is above.

hoomd/hdf5.py
```diff
@@ -358,3 +359,58 @@ def _write_header(self, f):
...
+        f = None
+        if hoomd.context.current.device.comm.rank == 0:
+        f = self.h5file
+
+        log_dict = self.log()
+        for key in log_dict.keys():
+            if f is not None:  # Only the root rank further process the received data
+                data = log_dict[key]
+                # Check the returned object
+                if not isinstance(data, numpy.ndarray):
```
b-butler, line 384: There are many scalar (float, int) quantities that are loggable currently. Do you not plan to support them with this back-end?

hoomd/hdf5.py
```diff
@@ -358,3 +359,58 @@ def _write_header(self, f):
...
+        super().__init__(logger)
+
+    def _write_frame(self,h5file):
+
+        f = None
+        if hoomd.context.current.device.comm.rank == 0:
+        f = self.h5file
+
+        log_dict = self.log()
+        for key in log_dict.keys():
```
b-butler, line 380: It is significantly faster to not use keys.
```suggestion
        for key in log_dict:
```

hoomd/hdf5.py
```diff
@@ -358,3 +359,58 @@ def _write_header(self, f):
...
+# a dictionary that has the names as keys and np.arrays as values.
+# HDF5 has the restriction that the dimension of the matrices cannot change.
+# This translates to the restriction that the dimensions of the logged quantities cannot change during a simulation run.
+class HDF5LogWriter(hoomd.dump.GSDLogWriter):
+    def __init__(self,logger):
+        super().__init__(logger)
+
+    def _write_frame(self,h5file):
+
+        f = None
```
b-butler, line 375: I would prefer a more descriptive name for the file.

hoomd/hdf5.py
```diff
@@ -358,3 +359,58 @@ def _write_header(self, f):
...
+        if hoomd.context.current.device.comm.rank == 0:
+        f = self.h5file
```
b-butler, line 377: Where are we setting the `h5file` property?

b-butler: Also this should be indented.

b-butler:
Testing for the `GSDLogWriter` and `GSD` classes do not currently exist. They were tried out local and shown to work. Unit tests are coming however.

hoomd/hdf5.py
InnocentBug: Yes, you are right. I used my old implementation as a foundation and missed to change the name here.

hoomd/hdf5.py
InnocentBug: My assumption was that the dictionary contains only numpy arrays as values because the GSDDumpWriter does a similar assumption. And this was meant as a fall back test to double check is assumption.
I changed it now into a varaition, where I try to force convert the value into a numpy array and later check if it is an array of numerical data to exclude things like objects and such.

This covers now also simple float or int values as 1x1 arrays.

I excluded strings for now, because I think strings probably have a tendency to change in length which is not easy to handle for HDF5.
If we want to change that later on, it is a trivial fix.

hoomd/hdf5.py
InnocentBug: sure. thx for your suggestion. I am not that familiar with efficient python programming.

hoomd/hdf5.py
InnocentBug: I changed the name to file_handle.
(I considered a name that makes clear that is only valid on the root MPI rank, but couldn't come up with of a good precise name.)

hoomd/hdf5.py
InnocentBug: I changed it now, that the h5file file handle is initialized with the constructor.
That resulted from my confusion. The GSDLogWriter also accepted the GSDDumpWriter c++ backend in the function.

indention fixed.

InnocentBug:
> Thanks for the PR. I have a few questions. Do you want to prepend every key in the `HDFLogWriter` with 'log' because the `GSDLogWriter` does in the `log` function. Also, the conversion for strings and string quantities into the writer should work, just know that some arrays may be those kinds of quantities cast into NumPy arrays. Another point to make here is that `_write_frame` is for the `GSD` class to be use logger, for this class is there an existing front-end to plug into?
> 
> HDF5 can handle nested data such as `['md']['pair']['LJ']` where each level has either keys, data, or both associated with it (I may be wrong here). If that is the case, is there a reason you are using the `GSDLogWriter`'s flattened `dict` where the namespace is concatenated using '/'?

1) Yes, it is intentional to include the "log" as a prefix. The user may choose to write this data into an h5file that also contains other data not logged via the HOOMD-blue interface. And there is no problem with this approach.
In that case it is nice to have a group, where all the logged data is summarized into. 
I think the idea is very similar to the GSD logging as well.
2) No there is no front-end of this class, or the class is intended as the python front-end.
Everything that needs to be done is pure python.
I changed now the interface such that is now done in the `log` function. It propably fits the structure better.
However, after @joaander comment about #629, I wait now for this PR and may adjust the interface later on.
3) I intentionally used the flattened dict of the GSD dump writer. I guess you are referring to groups for the nested capabilities of HDF5.
The reason I use the `/` flattened dict is, that h5py automatically understands that `/` separates nested data and creates the necessary groups with the creation of the dataset.

I hope, you like my improved version a bit better.
For the interface, I would want to wait now for the PR.

joaander:
Ludwig, do you have plans to update this PR? Or should we close it?

InnocentBug:
I think you can close it.
I still want that functionalit, but the changes I introduce with the PR are probably not the basis for that.

PR #655, opened 2020-05-18
Feature/new pair force

tommy-waltmann:
## Description

This draft pull request is for the pair forces implemented in the v3 API. I'm opening this as a draft, because I would like some feedback on what I have done so far. Please ignore any changes to `Morse`, `SLJ`, and `Ewald` when giving comments, because I got a little ahead of myself when I started working on this PR. In other words, please only comment on things related to `Yukawa` and `Gauss`.

Also, I am aware its failing every MPI test because of the lack of `setCommunicator` method. The necessary changes are on #646 and I will merge those as soon as that pull request is resolved.

## Motivation and Context

We are porting classes to the v3 API, and there are a lot of pair forces, so before I start going down the line implementing all of those, I want to make sure I have done the first two right.

## How Has This Been Tested?

There are basic tests, which include all of the original cpp tests for each pair force, as well as new python tests for correct attachment and quick simulation runs in `pytest/test_pair.py`

## Change log

```
pair potentials implemented in v3 API
```

## Checklist:

- [x ] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x ] My name is on the [list of contributors](https://github.com/glotzerlab/hoomd-blue/blob/master/sphinx-doc/credits.rst).

### DIFF
--- a/hoomd/md/EvaluatorPairGauss.h
+++ b/hoomd/md/EvaluatorPairGauss.h
@@ -99,7 +99,7 @@ class EvaluatorPairGauss
             \param _params Per type pair parameters of this potential
         */
         DEVICE EvaluatorPairGauss(Scalar _rsq, Scalar _rcutsq, const param_type& _params)
-            : rsq(_rsq), rcutsq(_rcutsq), epsilon(_params.epsilon), sigma(_params.epsilon)
+            : rsq(_rsq), rcutsq(_rcutsq), epsilon(_params.epsilon), sigma(_params.sigma)
             {
             }
 
--- a/hoomd/md/EvaluatorPairGauss.h
+++ b/hoomd/md/EvaluatorPairGauss.h
@@ -25,6 +25,50 @@
 #define DEVICE
 #endif
 
+struct gauss_params
+    {
+    Scalar epsilon;
+    Scalar sigma;
+
+    #ifdef ENABLE_HIP
+    // set CUDA memory hints
+    void set_memory_hint() const
+        {
+        // default implementation does nothing
+        }
+    #endif
+
+    #ifndef __HIPCC__
+    gauss_params() {sigma = 0; epsilon = 0;}
+
+    gauss_params(pybind11::dict v)
+        {
+        sigma = v["sigma"].cast<Scalar>();
+        epsilon = v["epsilon"].cast<Scalar>();
+        }
+
+    // used to facilitate unit testing
+    gauss_params(Scalar eps, Scalar sig)
+        {
+        sigma = sig;
+        epsilon = eps;
+        }
+
+    pybind11::dict asDict()
+        {
+        pybind11::dict v;
+        v["sigma"] = sigma;
+        v["epsilon"] = epsilon;
+        return v;
+        }
+    #endif
+    }
+    #ifdef SINGLE_PRECISION
+    __attribute__((aligned(8)));
+    #else
+    __attribute__((aligned(16)));
+    #endif
+
 //! Class for evaluating the Gaussian pair potential
 /*! <b>General Overview</b>
 
@@ -47,7 +91,7 @@ class EvaluatorPairGauss
     {
     public:
         //! Define the parameter type used by this pair potential evaluator
-        typedef Scalar2 param_type;
+        typedef gauss_params param_type;
 
         //! Constructs the pair potential evaluator
         /*! \param _rsq Squared distance between the particles
@@ -55,7 +99,7 @@ class EvaluatorPairGauss
             \param _params Per type pair parameters of this potential
         */
         DEVICE EvaluatorPairGauss(Scalar _rsq, Scalar _rcutsq, const param_type& _params)
-            : rsq(_rsq), rcutsq(_rcutsq), epsilon(_params.x), sigma(_params.y)
+            : rsq(_rsq), rcutsq(_rcutsq), epsilon(_params.epsilon), sigma(_params.epsilon)
             {
             }
 
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -440,7 +440,7 @@ class Yukawa(_Pair):
 
     """
     _cpp_class_name = "PotentialPairYukawa"
-    def __init__(self, nlist, r_cut=None, r_on=0, mode='none'):
+    def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
         super().__init__(nlist, r_cut, r_on, mode)
         params = TypeParameter('params', 'particle_types',
                                TypeParamDict(kappa=float, epsilon=float,
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -453,7 +453,7 @@ class Yukawa(_Pair):
 
         return _hoomd.make_scalar2(epsilon, kappa);
 
-class ewald(pair):
+class Ewald(_Pair):
     R""" Ewald pair potential.
 
     :py:class:`ewald` specifies that a Ewald pair potential should be applied between every
@@ -497,39 +497,19 @@ class ewald(pair):
         :py:class:`ewald` for you.
 
     """
-    def __init__(self, r_cut, nlist, name=None):
-
-        # tell the base class how we operate
-
-        # initialize the base class
-        pair.__init__(self, r_cut, nlist, name);
-
-        # create the c++ mirror class
-        if not hoomd.context.current.device.cpp_exec_conf.isCUDAEnabled():
-            self.cpp_force = _md.PotentialPairEwald(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairEwald;
-        else:
-            self.nlist.cpp_nlist.setStorageMode(_md.NeighborList.storageMode.full);
-            self.cpp_force = _md.PotentialPairEwaldGPU(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairEwaldGPU;
-
-        hoomd.context.current.system.addCompute(self.cpp_force, self.force_name);
-
-        # setup the coefficient options
-        self.required_coeffs = ['kappa','alpha'];
-        self.pair_coeff.set_default_coeff('alpha', 0.0);
+    _cpp_class_name = "PotentialPairEwald"
+    def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
+        super().__init__(nlist, r_cut, r_on, mode)
+        params = TypeParameter('params', 'particle_types',
+                               TypeParamDict(kappa=float, alpha=0.0,
+                                             len_keys=2))
+        self.add_typeparam(params)
 
     def process_coeff(self, coeff):
         kappa = coeff['kappa'];
         alpha = coeff['alpha'];
 
         return _hoomd.make_scalar2(kappa, alpha)
-
-    def set_params(self, coeff):
-        """ :py:class:`ewald` has no energy shift modes """
-
-        raise RuntimeError('Not implemented for DPD Conservative');
-        return;
 
 def _table_eval(r, rmin, rmax, V, F, width):
     dr = (rmax - rmin) / float(width-1);
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -290,7 +290,7 @@ class Gauss(_Pair):
 
     """
     _cpp_class_name = "PotentialPairGauss"
-    def __init__(self, nlist, r_cut=None, r_on=0, mode='none'):
+    def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
         super().__init__(nlist, r_cut, r_on, mode)
         params = TypeParameter('params', 'particle_types',
                                TypeParameterDict(epsilon=float, sigma=float,
@@ -303,7 +303,7 @@ class Gauss(_Pair):
 
         return _hoomd.make_scalar2(epsilon, sigma);
 
-class slj(pair):
+class SLJ(_Pair):
     R""" Shifted Lennard-Jones pair potential.
 
     Args:
@@ -367,37 +367,23 @@ class slj(pair):
         slj.pair_coeff.set(['A', 'B'], ['C', 'D'], epsilon=2.0)
 
     """
-    def __init__(self, r_cut, nlist, d_max=None, name=None):
+    _cpp_class_name = 'PotentialPairSLJ'
+    def __init__(self, nlist, r_cut=None, r_on=0., mode='none', d_max=None):
+        super().__init__(nlist, r_cut, r_on, mode)
+        params = TypeParameter('params', 'particle_types',
+                               TypeParameterDict(epsilon=float, sigma=float,
+                                                 alpha=1.0, len_keys=2)
+                               )
+        self._add_typeparam(params)
 
-        # tell the base class how we operate
+        param_dict = ParameterDict(
+            d_max=float,
+            mode=OnlyFrom(['none', 'shifted'])  # mode not allowed to be xplor
+            explicit_defaults=dict(mode=mode, d_max=__) # largest particle diameter in the system taken from cpp
+            )
+        self._param_dict.update(param_dict)
 
-        # initialize the base class
-        pair.__init__(self, r_cut, nlist, name);
-
-        # update the neighbor list
-        if d_max is None :
-            sysdef = hoomd.context.current.system_definition;
-            d_max = sysdef.getParticleData().getMaxDiameter()
-            hoomd.context.current.device.cpp_msg.notice(2, "Notice: slj set d_max=" + str(d_max) + "\n");
-
-        # SLJ requires diameter shifting to be on
-        self.nlist.cpp_nlist.setDiameterShift(True);
-        self.nlist.cpp_nlist.setMaximumDiameter(d_max);
-
-        # create the c++ mirror class
-        if not hoomd.context.current.device.cpp_exec_conf.isCUDAEnabled():
-            self.cpp_force = _md.PotentialPairSLJ(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairSLJ;
-        else:
-            self.nlist.cpp_nlist.setStorageMode(_md.NeighborList.storageMode.full);
-            self.cpp_force = _md.PotentialPairSLJGPU(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairSLJGPU;
-
-        hoomd.context.current.system.addCompute(self.cpp_force, self.force_name);
-
-        # setup the coefficient options
-        self.required_coeffs = ['epsilon', 'sigma', 'alpha'];
-        self.pair_coeff.set_default_coeff('alpha', 1.0);
+        # NOTE need to set diameter shift and max diameter in the neighborlist once attached
 
     def process_coeff(self, coeff):
         epsilon = coeff['epsilon'];
@@ -408,21 +394,6 @@ class slj(pair):
         lj2 = alpha * 4.0 * epsilon * math.pow(sigma, 6.0);
         return _hoomd.make_scalar2(lj1, lj2);
 
-    def set_params(self, mode=None):
-        R""" Set parameters controlling the way forces are computed.
-
-        See :py:meth:`pair.set_params()`.
-
-        Note:
-            **xplor** is not a valid setting for :py:class:`slj`.
-
-        """
-
-        if mode == "xplor":
-            hoomd.context.current.device.cpp_msg.error("XPLOR is smoothing is not supported with slj\n");
-            raise RuntimeError("Error changing parameters in pair force");
-
-        pair.set_params(self, mode=mode);
 
 class yukawa(pair):
     R""" Yukawa pair potential.
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -437,15 +437,9 @@ class Yukawa(_Pair):
     def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
         super().__init__(nlist, r_cut, r_on, mode)
         params = TypeParameter('params', 'particle_types',
-                               TypeParamDict(kappa=float, epsilon=float,
-                                             len_keys=2))
-        self.add_typeparam(params)
-
-    def process_coeff(self, coeff):
-        epsilon = coeff['epsilon'];
-        kappa = coeff['kappa'];
-
-        return _hoomd.make_scalar2(epsilon, kappa);
+                               TypeParameterDict(kappa=float, epsilon=float,
+                                                 len_keys=2))
+        self._add_typeparam(params)
 
 class Ewald(_Pair):
     R""" Ewald pair potential.
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -776,7 +776,7 @@ class table(force._force):
 
         self.pair_coeff.set(a, b, func=_table_eval, rmin=rmin_table, rmax=rmax_table, coeff=dict(V=V_table, F=F_table, width=self.width))
 
-class morse(pair):
+class Morse(_Pair):
     R""" Morse pair potential.
 
     :py:class:`morse` specifies that a Morse pair potential should be applied between every
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -382,7 +382,7 @@ class SLJ(_Pair):
         # mode not allowed to be xplor, so re-do param dict without that option
         param_dict = ParameterDict(
             d_max=float,
-            mode=OnlyFrom(['none', 'shifted'])
+            mode=OnlyFrom(['none', 'shifted']),
             explicit_defaults=dict(mode=mode, d_max=__) # largest particle diameter in the system taken from cpp
             )
         self._param_dict.update(param_dict)
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -248,7 +248,7 @@ class LJ(_Pair):
                                )
         self._add_typeparam(params)
 
-class gauss(pair):
+class Gauss(_Pair):
     R""" Gaussian pair potential.
 
     Args:
@@ -289,26 +289,13 @@ class gauss(pair):
         gauss.pair_coeff.set(['A', 'B'], ['C', 'D'], epsilon=3.0, sigma=0.5)
 
     """
-    def __init__(self, r_cut, nlist, name=None):
-
-        # tell the base class how we operate
-
-        # initialize the base class
-        pair.__init__(self, r_cut, nlist, name);
-
-        # create the c++ mirror class
-        if not hoomd.context.current.device.cpp_exec_conf.isCUDAEnabled():
-            self.cpp_force = _md.PotentialPairGauss(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairGauss;
-        else:
-            self.nlist.cpp_nlist.setStorageMode(_md.NeighborList.storageMode.full);
-            self.cpp_force = _md.PotentialPairGaussGPU(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairGaussGPU;
-
-        hoomd.context.current.system.addCompute(self.cpp_force, self.force_name);
-
-        # setup the coefficient options
-        self.required_coeffs = ['epsilon', 'sigma'];
+    _cpp_class_name = "PotentialPairGauss"
+    def __init__(self, nlist, r_cut=None, r_on=0, mode='none'):
+        super().__init__(nlist, r_cut, r_on, mode)
+        params = TypeParameter('params', 'particle_types',
+                               TypeParameterDict(epsilon=float, sigma=float,
+                                                 len_keys=2))
+        self._add_typeparam(params)
 
     def process_coeff(self, coeff):
         epsilon = coeff['epsilon'];
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -369,6 +369,9 @@ class SLJ(_Pair):
     """
     _cpp_class_name = 'PotentialPairSLJ'
     def __init__(self, nlist, r_cut=None, r_on=0., mode='none', d_max=None):
+        if mode == 'xplor':
+            raise ValueError("xplor is not a valid mode for SLJ potential")
+
         super().__init__(nlist, r_cut, r_on, mode)
         params = TypeParameter('params', 'particle_types',
                                TypeParameterDict(epsilon=float, sigma=float,
@@ -376,9 +379,10 @@ class SLJ(_Pair):
                                )
         self._add_typeparam(params)
 
+        # mode not allowed to be xplor, so re-do param dict without that option
         param_dict = ParameterDict(
             d_max=float,
-            mode=OnlyFrom(['none', 'shifted'])  # mode not allowed to be xplor
+            mode=OnlyFrom(['none', 'shifted'])
             explicit_defaults=dict(mode=mode, d_max=__) # largest particle diameter in the system taken from cpp
             )
         self._param_dict.update(param_dict)
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -297,12 +297,6 @@ class Gauss(_Pair):
                                                  len_keys=2))
         self._add_typeparam(params)
 
-    def process_coeff(self, coeff):
-        epsilon = coeff['epsilon'];
-        sigma = coeff['sigma'];
-
-        return _hoomd.make_scalar2(epsilon, sigma);
-
 class SLJ(_Pair):
     R""" Shifted Lennard-Jones pair potential.
 
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -812,26 +812,13 @@ class morse(pair):
         morse.pair_coeff.set(['A', 'B'], ['C', 'D'], D0=1.0, alpha=3.0)
 
     """
-    def __init__(self, r_cut, nlist, name=None):
-
-        # tell the base class how we operate
-
-        # initialize the base class
-        pair.__init__(self, r_cut, nlist, name);
-
-        # create the c++ mirror class
-        if not hoomd.context.current.device.cpp_exec_conf.isCUDAEnabled():
-            self.cpp_force = _md.PotentialPairMorse(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairMorse;
-        else:
-            self.nlist.cpp_nlist.setStorageMode(_md.NeighborList.storageMode.full);
-            self.cpp_force = _md.PotentialPairMorseGPU(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairMorseGPU;
-
-        hoomd.context.current.system.addCompute(self.cpp_force, self.force_name);
-
-        # setup the coefficient options
-        self.required_coeffs = ['D0', 'alpha', 'r0'];
+    _cpp_class_name = "PotentialPairMorse"
+    def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
+        super().__init__(nlist, r_cut, r_on, mode)
+        params = TypeParameter('params', 'particle_types',
+                               TypeParamDict(D0=float, alpha=float, r0=float,
+                                             len_keys=2))
+        self.add_typeparam(params)
 
     def process_coeff(self, coeff):
         D0 = coeff['D0'];
--- a/hoomd/md/pair.py
+++ b/hoomd/md/pair.py
@@ -395,7 +395,7 @@ class SLJ(_Pair):
         return _hoomd.make_scalar2(lj1, lj2);
 
 
-class yukawa(pair):
+class Yukawa(_Pair):
     R""" Yukawa pair potential.
 
     Args:
@@ -435,26 +435,13 @@ class yukawa(pair):
         yukawa.pair_coeff.set(['A', 'B'], ['C', 'D'], epsilon=0.5, kappa=3.0)
 
     """
-    def __init__(self, r_cut, nlist, name=None):
-
-        # tell the base class how we operate
-
-        # initialize the base class
-        pair.__init__(self, r_cut, nlist, name);
-
-        # create the c++ mirror class
-        if not hoomd.context.current.device.cpp_exec_conf.isCUDAEnabled():
-            self.cpp_force = _md.PotentialPairYukawa(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairYukawa;
-        else:
-            self.nlist.cpp_nlist.setStorageMode(_md.NeighborList.storageMode.full);
-            self.cpp_force = _md.PotentialPairYukawaGPU(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name);
-            self.cpp_class = _md.PotentialPairYukawaGPU;
-
-        hoomd.context.current.system.addCompute(self.cpp_force, self.force_name);
-
-        # setup the coefficient options
-        self.required_coeffs = ['epsilon', 'kappa'];
+    _cpp_class_name = "PotentialPairYukawa"
+    def __init__(self, nlist, r_cut=None, r_on=0, mode='none'):
+        super().__init__(nlist, r_cut, r_on, mode)
+        params = TypeParameter('params', 'particle_types',
+                               TypeParamDict(kappa=float, epsilon=float,
+                                             len_keys=2))
+        self.add_typeparam(params)
 
     def process_coeff(self, coeff):
         epsilon = coeff['epsilon'];
--- a/hoomd/md/pytest/test_pair.py
+++ b/hoomd/md/pytest/test_pair.py
@@ -3,13 +3,17 @@ import pytest
 from copy import deepcopy
 
 _pairs = [
-    (hoomd.md.pair.Gauss(hoomd.md.nlist.Cell()), {'epsilon': 0.05, "sigma": 0.02}),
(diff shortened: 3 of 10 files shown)


hoomd/md/EvaluatorPairGauss.h
```diff
@@ -25,6 +25,50 @@
 #define DEVICE
 #endif
 
+struct gauss_params
```
mphoward, line 28: I'm not sure if this fits with the design of the rest of the API or is easily compatible with pybind11, but could you define these parameter structs within the evaluator so that you don't need to worry about polluting the namespace with parameter names (that will all look like `<potential>_params` anyway)? Something like:

```c++
class EvaluatorPairGauss
{
    typedef struct gauss_params
    {
    Scalar epsilon;
    Scalar sigma;
    } param_type;
}

EvaluatorPairGauss::param_type params(eps,sigma);
```

tommy-waltmann: This change seems to be compatible with pybind11, and I agree that it cleans things up, so I will make this change

hoomd/md/pytest/test_pair.py
```diff
@@ -218,6 +218,10 @@ def _calculate_force(sim):
 def test_force_energy_relationship(simulation_factory,
                                    two_particle_snapshot_factory,
                                    valid_params, nsteps):
+    # don't really test DPD and DPDLJ for this test
+    if valid_params[0] == "DPD" or valid_params[0] == "DPDLJ":
+        return
```
joaander, line 226: Skip the test instead of returning: https://docs.pytest.org/en/stable/skipping.html

tommy-waltmann:
We have made some great progress with this PR, and I'm now asking for comments on the API docs. I've reworked the first few potentials, and would like some feedback on them before I make the same changes for each potential in the PR.

hoomd/md/pair.py
```diff
@@ -278,54 +288,34 @@ class gauss(pair):
...
+    _cpp_class_name = "PotentialPairGauss"
+    def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
+        super().__init__(nlist, r_cut, r_on, mode)
+        params = TypeParameter('params', 'particle_types',
+                               TypeParameterDict(epsilon=float, sigma=float,
+                                                 len_keys=2))
+        self._add_typeparam(params)
 
-class slj(pair):
+class SLJ(_Pair):
```
tommy-waltmann, line 309: We've mentioned renaming this potential to something else previously because it was causing confusion, now would be a good time to discuss or throw out some ideas for new names.

hoomd/md/pair.py
```diff
@@ -349,103 +339,70 @@ class slj(pair):
       - *optional*: defaults to 1.0
     - :math:`r_{\mathrm{cut}}` - *r_cut* (in distance units)
       - *optional*: defaults to the global r_cut specified in the pair command
+    - :math:`r_{\mathrm{on}}` - *r_on* (in distance units)
+      - *optional*: defaults to the global r_on specified in the pair command
 
     .. attention::
         Due to the way that pair.slj modifies the cutoff criteria, a shift_mode of *xplor* is not supported.
```
tommy-waltmann, line 346:
```suggestion
        Due to the way that pair.SLJ modifies the cutoff criteria, a shift_mode of *xplor* is not supported.
```

hoomd/md/pair.py
```diff
@@ -349,103 +339,70 @@ class slj(pair):
...
 
     The actual cutoff radius for pair.slj is shifted by the diameter of two particles interacting.  Thus to determine
```
tommy-waltmann, line 348:
```suggestion
    The actual cutoff radius for pair.SLJ is shifted by the diameter of two particles interacting.  Thus to determine
```

hoomd/md/pair.py
```diff
@@ -349,103 +339,70 @@ class slj(pair):
...
     the maximum possible actual r_cut in simulation
-    pair.slj must know the maximum diameter of all the particles over the entire run, *d_max* .
-    This value is either determined automatically from the initialization or can be set by the user and can be
-    modified between runs with ``hoomd.md.nlist.nlist.set_params()``. In most cases, the correct value can be
-    identified automatically.
+    pair.slj must know the maximum diameter of all the particles over the entire run.
```
tommy-waltmann, line 350:
```suggestion
    pair.SLJ must know the maximum diameter of all the particles over the entire run.
```

hoomd/md/pair.py
```diff
@@ -349,103 +339,70 @@ class slj(pair):
...
+        self.mode = mode
 
-        """
+        # this potential needs diameter shifting on
+        self._nlist.diameter_shift = True
 
-        if mode == "xplor":
-            hoomd.context.current.device.cpp_msg.error("XPLOR is smoothing is not supported with slj\n");
-            raise RuntimeError("Error changing parameters in pair force");
+        # NOTE do we need something to automatically set the max_diameter correctly?
```
tommy-waltmann, line 393: I talked about some options for how to handle this with @b-butler , but we decided that its best if users know they need to set this themselves in the documentation.

hoomd/md/pair.py
```diff
@@ -1,10 +1,12 @@
 # Copyright (c) 2009-2019 The Regents of the University of Michigan
 # This file is part of the HOOMD-blue project, released under the BSD 3-Clause License.
 
+import hoomd
```
tommy-waltmann, line 4: Not sure if this is necessary or not.

hoomd/md/EvaluatorPairMoliere.h
```diff
@@ -38,15 +38,76 @@ class EvaluatorPairMoliere
...
+            {
+            // just holding on to these so if the user calls asDict,
+            // we can give them back correctly
+            unsigned int Zi;
+            unsigned int Zj;
+            Scalar e;
+            Scalar a0;
+
+            // these are the params the potential cares about
+            Scalar Zsq;
```
joaander, line 51: It is not possible to reconstruct 4 values from 2. We should refactor this potential to accept different parameters so that it does not need to store extra information in a performance critical data structure.
* qi: Zi*e
* qj: Zj*e
* aF

The computation of aF from a0 should be moved to the documentation to show users how to correctly calculate the value.

hoomd/md/EvaluatorPairMorse.h
```diff
@@ -45,15 +45,57 @@ class EvaluatorPairMorse
...
+            Scalar alpha;
+            Scalar r0;
+
+            #ifdef ENABLE_HIP
+            // CUDA memory hints
+            void set_memory_hints() const {}
+            #endif
+
+            #ifndef __HIPCC__
+            param_type() {D0=1; alpha=0; r0=0;}
```
joaander, line 60: Use 0 for all default constructed values.

hoomd/md/EvaluatorPairMorse.h
```diff
@@ -45,15 +45,57 @@ class EvaluatorPairMorse
...
+                {
+                pybind11::dict v;
+                v["D0"] = D0;
+                v["alpha"] = alpha;
+                v["r0"] = r0;
+                return v;
+                }
+            #endif
+            }
+            #ifdef SINGLE_PRECISION
```
joaander, line 86: This struct has 3 Scalars so use 16-byte alignment in both single and double precision.

hoomd/md/EvaluatorPairReactionField.h
```diff
@@ -47,15 +47,56 @@ class EvaluatorPairReactionField
...
+
+            #ifdef ENABLE_HIP
+            // set CUDA memory hints
+            void set_memory_hint() const {}
+            #endif
+
+            #ifndef __HIPCC__
+            param_type()
+                {
+                eps = 1;
```
joaander, line 64: Use 0 for default constructed values.

hoomd/md/EvaluatorPairReactionField.h
```diff
@@ -47,15 +47,56 @@ class EvaluatorPairReactionField
...
+                v["epsilon"] = eps;
+                v["eps_rf"] = eps_rf;
+                v["use_charge"] = use_charge;
+
+                return v;
+                }
+            #endif
+            }
+            #ifdef SINGLE_PRECISION
+            __attribute((aligned(8)));
```
joaander, line 88: Struct size is larger than 8 bytes, so use 16 byte alignment for both single and double precision.

hoomd/md/EvaluatorPairZBL.h
```diff
@@ -40,15 +41,76 @@ class EvaluatorPairZBL
 {
     public:
         //! Define the parameter type used by this pair potential evaluator
-        typedef Scalar2 param_type;
+        struct param_type
```
joaander, line 44: It appears that once we update Moliere to the new format this class can just reuse that type. The alternate aF computation will need to be listed in the documentation or provided as a convenience method.

hoomd/md/module-md.cc
```diff
@@ -257,26 +257,26 @@ PYBIND11_MODULE(_md, m)
...
     export_PotentialPair<PotentialPairMie>(m, "PotentialPairMie");
     export_PotentialPair<PotentialPairReactionField>(m, "PotentialPairReactionField");
     export_PotentialPair<PotentialPairDLVO>(m, "PotentialPairDLVO");
     export_PotentialPair<PotentialPairFourier>(m, "PotentialPairFourier");
     export_tersoff_params(m);
     export_revcross_params(m);
     export_pair_params(m);
-    export_AnisoPotentialPair<AnisoPotentialPairGB>(m, "AnisoPotentialPairGB");
-    export_AnisoPotentialPair<AnisoPotentialPairDipole>(m, "AnisoPotentialPairDipole");
+    // export_AnisoPotentialPair<AnisoPotentialPairGB>(m, "AnisoPotentialPairGB");
```
joaander, line 270: Why are so many of these commented out? This shouldn't be necessary any longer as all pair potentials have been updated.

hoomd/md/pytest/conftest.py
```diff
@@ -0,0 +1,533 @@
+import hoomd
+import pytest
+import numpy as np
+from copy import deepcopy
+import itertools
+
+np.random.seed(0)
```
joaander, line 7: Remove this. The main `conftest.py` file sets the numpy random seed to the same seed at the start of each test function.

hoomd/md/pytest/conftest.py
```diff
@@ -0,0 +1,533 @@
...
+
+
+def _make_valid_param_dicts(arg_dict, n_dicts):
+    # turn {'a': [0, 1], 'b':[2, 3]} into [{'a': 0, 'b': 2}, {'a': 1, 'b': 3}]
+    return [dict(zip(arg_dict, val)) for val in zip(*arg_dict.values())]
+
+
+def _valid_params(particle_types=['A', 'B']):
```
joaander, line 15: All of these test fixtures belong in `test_pair.py` as they are specific to those tests and not all MD tests in general.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,412 @@
+import hoomd
+import pytest
+import numpy as np
+import itertools
+import hoomd.md.pytest.conftest
+
+np.random.seed(0)
```
joaander, line 7: Remove this.

hoomd/md/EvaluatorPairBuckingham.h
```diff
@@ -58,16 +58,52 @@ class EvaluatorPairBuckingham
...
+                v["C"] = C;
+                return v;
+                }
+            #endif
+            }
+            #ifdef SINGLE_PRECISION
+            __attribute__((aligned(8)));
+            #else
+            __attribute__((aligned(16)));
+            #endif
```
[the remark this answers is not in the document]
tommy-waltmann: I've updated the potentials that I could figure out from your comments. The ones left are Fourier, Mie, Moliere, and ZBL

hoomd/md/pytest/conftest.py
danevans09: Resolved 92252586f

hoomd/md/pytest/conftest.py
danevans09: Resolved 92252586f

hoomd/md/pytest/test_pair.py
danevans09: Resolved 92252586f

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    dlvo_valid_dict = {'kappa': 1.0, 'Z': 0.1, 'A': 0.1}
+    dlvo_invalid_dicts = _make_invalid_param_dict(dlvo_valid_dict)
+    invalid_params_list.append(_make_invalid_params(dlvo_valid_dict,
+                                                    dlvo_invalid_dicts,
+                                                    hoomd.md.pair.DLVO,
+                                                    {}, "DLVO"))
+    return [params for param_list in invalid_params_list for params in param_list]
+
+
+@pytest.fixture(scope="function", params=_invalid_params(), ids=(lambda x: x[0]))
```
b-butler, line 320: Multiple will have the same id then, yes? Maybe we should append a number?

b-butler: Also is there a reason we go from a list of valid parameters with everything else constant, to a list of tuples with everything in them? I think it would be more consistent to have a tuple with a list of parameters (then the id doesn't matter any more). We do parameterize more by using a list of tuples, but I don't know if we get much from that.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+def invalid_params(request):
+    return deepcopy(request.param)
+
+
+def _forces_and_energies():
```
b-butler, line 325: Just mentioned that these we calculated then hard coded. Since this is so long, part of me would like to see this information brought to a new file like a json file that is read in for this test to be a bit cleaner.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+                               kT_dict,
+                               params[pair_potential][i],
+                               forces[pair_potential][i],
+                               energies[pair_potential][i]))
+    return param_list
+
+
+@pytest.fixture(scope="function",
+                params=_forces_and_energies(),
+                ids=(lambda x: x[0]))
```
b-butler, line 533: We will have 3 tests with the same name given this. I think having a number to signify which iteration would be helpful.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+def forces_and_energies(request):
+    return deepcopy(request.param)
+
+
+def _assert_equivalent_type_params(type_param1, type_param2):
+    for pair in type_param1:
```
b-butler, line 539: Are we guaranteed that all keys of `type_param1` are in `type_param2` and vice versa? If so this is fine, but if `type_param2` has additional keys we won't check that.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    sim = simulation_factory(two_particle_snapshot_factory(dimensions=3, d=.5))
+    integrator = hoomd.md.Integrator(dt=0.005)
+    integrator.forces.append(lj)
+    integrator.methods.append(hoomd.md.methods.Langevin(hoomd.filter.All(),
+                                                        kT=1, seed=1))
+    sim.operations.integrator = integrator
+
+    _assert_equivalent_type_params(lj.r_cut.to_dict(), {('A', 'A'): 2.5})
+    sim.operations.schedule()
+    sim.run(1)
```
b-butler, line 570:
```suggestion
    sim.run(1)
```
We don't need to explicitly call schedule any more.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+
+@pytest.mark.parametrize("mode", ['none', 'shifted', 'xplor'])
+def test_mode(simulation_factory, two_particle_snapshot_factory, mode):
+    cell = hoomd.md.nlist.Cell()
+    with pytest.raises(hoomd.typeconverter.TypeConversionError):
+        lj = hoomd.md.pair.LJ(nlist=cell, r_cut=2.5, mode=1)
+    with pytest.raises(hoomd.typeconverter.TypeConversionError):
+        lj = hoomd.md.pair.LJ(nlist=cell, r_cut=2.5, mode='str')
+    with pytest.raises(hoomd.typeconverter.TypeConversionError):
+        lj = hoomd.md.pair.LJ(nlist=cell, r_cut=2.5, mode=[1, 2, 3])
```
b-butler, line 582: This is getting tested for every parameterized value. It may be worth having two tests for this.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    lj = hoomd.md.pair.LJ(nlist=cell, r_cut=2.5, mode=mode)
+    lj.params[('A', 'A')] = {'sigma': 1, 'epsilon': 0.5}
+    snap = two_particle_snapshot_factory(dimensions=3, d=.5)
+    sim = simulation_factory(snap)
+    integrator = hoomd.md.Integrator(dt=0.005)
+    integrator.forces.append(lj)
+    integrator.methods.append(hoomd.md.methods.Langevin(hoomd.filter.All(),
+                                                        kT=1, seed=1))
+    sim.operations.integrator = integrator
+    sim.operations.schedule()
```
b-butler, line 593: unnecessary now.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    sim.operations.integrator = integrator
+    assert lj.r_on.to_dict() == {}
+
+    sim.operations.schedule()  # Before setting r_on
+    _assert_equivalent_type_params(lj.r_on.to_dict(), {('A', 'A'): 0.0})
+
+    lj.r_on[('A', 'A')] = 1.5
+    _assert_equivalent_type_params(lj.r_on.to_dict(), {('A', 'A'): 1.5})
+    sim.run(1)
+    _assert_equivalent_type_params(lj.r_on.to_dict(), {('A', 'A'): 1.5})
```
b-butler, line 619: Running shouldn't change anything. The schedule check should be enough.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    with pytest.raises(hoomd.typeconverter.TypeConversionError):
+        lj.r_on[('A', 'A')] = [1, 2, 3]
+
+    sim = simulation_factory(two_particle_snapshot_factory(dimensions=3, d=.5))
+    integrator = hoomd.md.Integrator(dt=0.005)
+    integrator.forces.append(lj)
+    integrator.methods.append(hoomd.md.methods.Langevin(hoomd.filter.All(),
+                                                        kT=1, seed=1))
+    sim.operations.integrator = integrator
+    assert lj.r_on.to_dict() == {}
```
b-butler, line 611: I would set `ron` after this to check that attaching works correctly on it. We can then set it to a new value if we want to test setting after attaching.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    dlvo_arg_dict = {'kappa': [1.0, 2.0, 5.0],
+                     'Z': [0.1, 0.5, 2.0],
+                     'A': [0.1, 0.5, 2.0]}
+    dlvo_valid_param_dicts = _make_valid_param_dicts(dlvo_arg_dict, len(combos))
+    valid_params_list.append(("DLVO", hoomd.md.pair.DLVO, {},
+                              dict(zip(combos, dlvo_valid_param_dicts))))
+    return valid_params_list
+
+
+@pytest.fixture(scope="function", params=_valid_params(), ids=(lambda x: x[0]))
```
b-butler, line 158: With the `ids` we could just do `x[1].__name__`

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+                         r_cut=2.5,
+                         mode='none')
+    for pair in pair_potential_dict:
+        pot.params[pair] = pair_potential_dict[pair]
+    _assert_equivalent_type_params(pair_potential_dict, pot.params.to_dict())
+
+
+def test_invalid_params(invalid_params):
+    pair_potential, pair_potential_dict, extra_args = invalid_params[1:]
+    print(pair_potential_dict)
```
b-butler, line 635: Remove print statement.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+                         r_cut=2.5, mode='none')
+    for pair in pair_potential_dict:
+        pot.params[pair] = pair_potential_dict[pair]
+
+    snap = lattice_snapshot_factory(particle_types=particle_types,
+                                    n=10, a=1.5, r=0.01)
+    if 'Ewald' in str(pair_potential) and snap.exists:
+        snap.particles.charge[:] = 1
+    elif 'SLJ' in str(pair_potential) and snap.exists:
+        snap.particles.diameter[:] = 2
```
b-butler, line 668: These can get nested in the conditional below.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    if snap.exists:
+        snap.particles.typeid[:] = np.random.randint(0,
+                                                     len(snap.particles.types),
+                                                     snap.particles.N)
+    sim = simulation_factory(snap)
+    sim.operations.integrator = hoomd.md.Integrator(dt=0.005)
+    sim.operations.integrator.forces.append(pot)
+    sim.operations.schedule()
```
b-butler, line 676: unnecessary.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    sim.run(10)
+    attached_pot = sim.operations.integrator.forces[0]
```
b-butler, line 678: No need to do this `pot` should be the same object.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    _assert_equivalent_type_params(attached_pot.params.to_dict(),
+                                   pair_potential_dict)
+
+
+@pytest.mark.parametrize("nsteps", [3, 5, 10])
```
b-butler, line 683: I don't think there is a reason to parametrize here.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    r = 1.0
+
+    lj = hoomd.md.pair.LJ(nlist=hoomd.md.nlist.Cell(), r_cut=r_cut)
+    lj.params[('A', 'A')] = {'sigma': 1, 'epsilon': 0.5}
+
+    snap = two_particle_snapshot_factory(dimensions=3, d=.5)
+    if snap.exists:
+        # avoid MPI errors by shifting positions by .1
+        snap.particles.position[0] = [0, 0, .1]
+        snap.particles.position[1] = [0, 0, r + .1]
```
b-butler, line 740: This is done in `conftest.py` now, I believe.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+                         r_cut=2.5, mode='none')
+    for pair in pair_potential_dict:
+        pot.params[pair] = pair_potential_dict[pair]
+
+    snap = two_particle_snapshot_factory(particle_types=particle_types, d=1.5)
+    if 'Ewald' in str(pair_potential) and snap.exists:
+        snap.particles.charge[:] = 1
+    elif 'SLJ' in str(pair_potential) and snap.exists:
+        snap.particles.diameter[:] = 2
+    sim = simulation_factory(snap)
```
b-butler, line 862: This setup seems common to a few tests. May be worth making a helper function.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+        snap = sim.state.snapshot
+        if snap.exists:
+            snap.particles.typeid[0] = particle_types.index(pair[0])
+            snap.particles.typeid[1] = particle_types.index(pair[1])
+        sim.state.snapshot = snap
+
+        calculated_forces = _calculate_force(sim)
+        forces = sim.operations.integrator.forces[0].forces
+        if forces is not None:
+            sim_forces = forces
```
b-butler, line 879: There is no need for this.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,939 @@
...
+    if snap.exists:
+        snap.particles.position[1] = initial_pos[1]
+    sim.state.snapshot = snap
+    if sim.state.snapshot.exists:
+        return Fa, Fb
+    else:
+        return 0, 0  # return dummy values if not on rank 1
+
+
+@pytest.mark.parametrize("nsteps", [1, 5, 10])
```
b-butler, line 840: I don't think the `nsteps` parametrizing is necessary.

hoomd/md/pytest/test_pair.py
b-butler: Also, we should really mention that this is for distances 0.75 and 1 in a comment at least.

b-butler: This is also far away from where it is used. Placing it closer to the first test it is used in would be helpful.

b-butler:
Also, in testing we should really consider the use of named tuples as it would help alievate all the random length tuple fixtures we have. I like the parametrized fixtures, but that would clear up some confusion in the future I think.

hoomd/conftest.py
```diff
@@ -106,7 +106,8 @@ def make_snapshot(particle_types=['A'], dimensions=3, d=1, L=20):
             s.configuration.box = box
             s.configuration.dimensions = dimensions
             s.particles.N = N
-            s.particles.position[:] = [[-d / 2, 0, 0], [d / 2, 0, 0]]
+            # shift particle positions slightly in z so MPI tests pass
+            s.particles.position[:] = [[-d / 2, 0, .1], [d / 2, 0, .1]]
```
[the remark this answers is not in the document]
tommy-waltmann: In some of the tests, we get energies of a system before running a simulation. When those tests are run with MPI, the domain decomposition splits the box above/below the z-axis. If the positions are left exactly on the boundary between domains (i.e. at (-d/2, 0, 0) and (d/2, 0, 0)) and we try to get energies before the simulation runs, the particles are not migrated properly and the test crashes with some kind of memory error.

hoomd/md/pair.py
```diff
@@ -532,55 +475,30 @@ class ewald(pair):
...
     Example::
 
-        nl = nlist.cell()
-        ewald = pair.ewald(r_cut=3.0, nlist=nl)
-        ewald.pair_coeff.set('A', 'A', kappa=1.0)
-        ewald.pair_coeff.set('A', 'A', kappa=1.0, alpha=1.5)
-        ewald.pair_coeff.set('A', 'B', kappa=1.0, r_cut=3.0, r_on=2.0);
+        nl = nlist.Cell()
+        ewald = pair.Ewald(r_cut=3.0, nlist=nl)
+        ewald.params[('A', 'A')]['kappa'] = 1.0
```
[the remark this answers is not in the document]
tommy-waltmann: I was not aware of the multiple keys syntax. Also, I don't think those brackets `[` and `]` match each other.

hoomd/md/pytest/test_pair.py
tommy-waltmann: Pytest will automatically append a number, we don't need to do that explicitly.

hoomd/md/pytest/test_pair.py
tommy-waltmann: A number will be append by pytest automatically

hoomd/md/pytest/test_pair.py
danevans09: `type_param1` is the dictionary used to set the arguments of the potential, whereas `type_param2` is the dictionary returned from the potential's `to_dict` method. So `type_param2` contains optional arguments while `type_param1` does not contain optional arguments unless they are explicitly set.

hoomd/md/EvaluatorPairMoliere.h
```diff
@@ -129,7 +159,8 @@ class EvaluatorPairMoliere
     protected:
         Scalar rsq;     //!< Stored rsq from the constructor
         Scalar rcutsq;  //!< Stored rcutsq from the constructor
-        Scalar Zsq;     //!< Zsq parameter extracted from the params passed to the constructor
+        Scalar qi;      //!< qi parameter extracted from the params passed to the constructor
+        Scalar qj;      //!< qj parameter extracted from the params passed to the constructor
```
b-butler, line 163: I don't think this should change. I misunderstood before. The params are stored elsewhere we only need `Zsq`.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+import numpy as np
+import itertools
+from copy import deepcopy
+import json
+from pathlib import Path
+from collections import namedtuple
+
+
+def _assert_equivalent_type_params(type_param1, type_param2):
+    for pair in type_param1:
```
b-butler, line 12: I think a docstring about what is expected for argument 1 and 2 would be helpful. I know a GitHub comment was made on this, but having developer documentation on that would be helpful.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+
+
+def test_invalid_mode():
+    cell = hoomd.md.nlist.Cell()
+    with pytest.raises(hoomd.typeconverter.TypeConversionError):
+        lj = hoomd.md.pair.LJ(nlist=cell, r_cut=2.5, mode=1)
+    with pytest.raises(hoomd.typeconverter.TypeConversionError):
+        lj = hoomd.md.pair.LJ(nlist=cell, r_cut=2.5, mode='str')
+    with pytest.raises(hoomd.typeconverter.TypeConversionError):
+        lj = hoomd.md.pair.LJ(nlist=cell, r_cut=2.5, mode=[1, 2, 3])
```
b-butler, line 53: A for loop with `for invalid_mode in [1, 'str', [1, 2, 3]]` would be a bit cleaner.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+
+    sim = simulation_factory(two_particle_snapshot_factory(dimensions=3, d=.5))
+    integrator = hoomd.md.Integrator(dt=0.005)
+    integrator.forces.append(lj)
+    integrator.methods.append(hoomd.md.methods.Langevin(hoomd.filter.All(),
+                                                        kT=1, seed=1))
+    sim.operations.integrator = integrator
+    assert lj.r_on.to_dict() == {}
+
+    lj.r_on[('A', 'A')] = 1.5
```
b-butler, line 87: I would like to see an assert statement after this.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+    params = []
+    for i in range(len(invalid_param_dicts)):
+        params.append({('A', 'A'): invalid_param_dicts[i]})
+    return [paramtuple(pot, params[i], extra_args) for i in range(N)]
+
+paramtuple = namedtuple('paramtuple',
+                        ['pair_potential',
+                         'pair_potential_params',
+                         'extra_args'])
+
```
b-butler, line 122: should be before where it is first used.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+
+def test_invalid_pair_key():
+    pot = hoomd.md.pair.LJ(nlist=hoomd.md.nlist.Cell())
+    with pytest.raises(KeyError):
+        pot.r_cut[3] = 2.5
+    with pytest.raises(KeyError):
+        pot.r_cut[[1, 2]] = 2.5
+    with pytest.raises(KeyError):
+        pot.r_cut['str'] = 2.5
```
b-butler, line 259: Use a for loop

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+def valid_params(request):
+    return deepcopy(request.param)
+
+def test_valid_params(valid_params):
+    pot = valid_params.pair_potential(**valid_params.extra_args,
+                                      nlist=hoomd.md.nlist.Cell(),
+                                      r_cut=2.5,
+                                      mode='none')
+    for pair in valid_params.pair_potential_params:
+        pot.params[pair] = valid_params.pair_potential_params[pair]
```
b-butler, line 433: You should be able to just set it to the entire dictionary assuming the dictionary is `{('a', 'a'): params},...}`. No need for the for loop.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+def test_attached_params(simulation_factory, lattice_snapshot_factory,
+                         valid_params):
+    pair_potential, pair_potential_dict, extra_args = valid_params
+    pair_keys = valid_params.pair_potential_params.keys()
+    particle_types = list(set(itertools.chain.from_iterable(pair_keys)))
+    pot = valid_params.pair_potential(**valid_params.extra_args,
+                                      nlist=hoomd.md.nlist.Cell(),
+                                      r_cut=2.5, mode='none')
+    for pair in valid_params.pair_potential_params:
+        pot.params[pair] = valid_params.pair_potential_params[pair]
```
b-butler, line 457: Should be able to just set parameters.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+
+    _update_snap(valid_params.pair_potential, snap)
+    if snap.exists:
+        snap.particles.typeid[:] = np.random.randint(0,
+                                                     len(snap.particles.types),
+                                                     snap.particles.N)
+    sim = simulation_factory(snap)
+    sim.operations.integrator = hoomd.md.Integrator(dt=0.005)
+    sim.operations.integrator.forces.append(pot)
+    sim.run(10)
```
b-butler, line 470:
```suggestion
    sim.run(1)
```
We still run but this should test a bit faster.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+    sim.operations.integrator = integrator
+    sim.operations.schedule()
+    for nsteps in [3, 5, 10]:
+        old_snap = sim.state.snapshot
+        sim.run(nsteps)
+        new_snap = sim.state.snapshot
+        if new_snap.exists:
+            with pytest.raises(AssertionError):
+                np.testing.assert_allclose(new_snap.particles.position,
+                                           old_snap.particles.position)
```
b-butler, line 504: Just assert that they are not all close. We should use `pytest.raises` for an `AssertionError` just use NumPy's `allclose` function and assert not that.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+    if energies is not None:
+        xplor_E = sum(energies)
+        assert xplor_E == E_r * S_r(r, r_cut, r_on)
+
+        lj_xplor.r_on[('A', 'A')] = 3.0
+        assert sum(energies) == E_r - E_rcut
+
+
+# This function calculates the forces in a two particle simulation frame by
+# finding the negative derivative of energy over inter particle distance
```
b-butler, line 589: This is better put in the function's docstring.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+                         'pair_potential_params',
+                         'extra_args',
+                         'forces',
+                         'energies'])
+
+
+def _forces_and_energies():
+    # Reference force and energy values were calculated using Mathematica 12.1.1
+    # and then stored in the json file below. Values were calculated at
+    # distances of 0.75 and 1.5 for each argument dictionary
```
b-butler, line 678: Use standard docstring `""" content """`

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+                                       sim_forces[1],
+                                       rtol=1e-06)
+
+FandEtuple = namedtuple('FandEtuple',
+                        ['pair_potential',
+                         'pair_potential_params',
+                         'extra_args',
+                         'forces',
+                         'energies'])
+
```
b-butler, line 673: Since we are using a dictionary now, we could just keep that representation, but up to you. I don't mind either way.

hoomd/conftest.py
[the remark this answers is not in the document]
b-butler: Ah, that is unfortunate. Maybe this is a bug we should make an issue about, though it is unlikely to get triggered in a normal simulation.

hoomd/md/pytest/test_pair.py
```diff
@@ -0,0 +1,751 @@
...
+    with pytest.raises(KeyError):
+        pot.r_cut[3] = 2.5
+    with pytest.raises(KeyError):
+        pot.r_cut[[1, 2]] = 2.5
+    with pytest.raises(KeyError):
+        pot.r_cut['str'] = 2.5
+
+
+def _make_valid_param_dicts(arg_dict):
+    # turn {'a': [0, 1], 'b':[2, 3]} into [{'a': 0, 'b': 2}, {'a': 1, 'b': 3}]
```
b-butler, line 263: make normal docstring.

hoomd/md/EvaluatorPairDLVO.h
```diff
@@ -129,7 +154,7 @@ class EvaluatorPairDLVO
 
                 Scalar engt1 = radprod * rmdsqsinv * A / Scalar(3.0);
                 Scalar engt2 = radprod * rmdsqminv * A / Scalar(3.0);
-                Scalar engt3 = LOG(rmdsqs * rmdsqminv) * A / Scalar(6.0);
+                Scalar engt3 = slow::log(rmdsqs * rmdsqminv) * A / Scalar(6.0);
```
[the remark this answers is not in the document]
tommy-waltmann: The other log is actually inaccurate for certain input values. This was a bug we found during unit testing for this potential.

hoomd/GPUArray.h
```diff
@@ -934,7 +934,6 @@ template<class T> void GPUArray<T>::allocate()
     h_data = std::unique_ptr<T, hoomd::detail::host_deleter<T> >(reinterpret_cast<T *>(host_ptr), host_deleter);
 
 #if defined (ENABLE_HIP)
-    assert(!d_data);
```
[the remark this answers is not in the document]
tommy-waltmann: Bottom line is that the assert is old and should have been removed a while ago, but was never done since everyone uses release builds of hoomd. This was caught during unit testing when I was using a debug build, which actually checks the assert statements.

hoomd/conftest.py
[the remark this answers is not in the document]
tommy-waltmann: It's definitely a corner case, but I'll make an issue about it just so it is recorded somewhere for future reference.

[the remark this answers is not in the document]
joaander: FYI, this issue will not occur during normal simulations. The run() loop triggers particle migration which resolves the error. When you request potential energies *before* the simulation run, migration has not yet occurred. We can skirt this issue by calling the migration when setting a snapshot and after loading a file. However, I'm not certain this will catch all corner cases.

hoomd/md/pair.py
joaander: We can deal with that with a different PR focused on fixing the other issues with SLJ.

hoomd/md/pair.py
```diff
@@ -994,8 +892,9 @@ class dpd(pair):
         integrate.nve(group=group.all())
 
     """
-    def __init__(self, r_cut, nlist, kT, seed, name=None):
-
+    _cpp_class_name = "PotentialPairDPDThermoDPD"
+    def __init__(self, nlist, kT, seed=3, r_cut=None, r_on=0., mode='none'):
+        """
         # register the citation
```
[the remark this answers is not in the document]
joaander: We can deal with removing these in a focused PR that deals with the citation management.

hoomd/md/pair.py
```diff
@@ -1414,72 +1176,57 @@ class moliere(pair):
...
-    See :py:class:`_Pair` for details on how forces are calculated and the available energy shifting and smoothing modes.
-    Use ``coeff.set`` to set potential coefficients.
-
-    The following coefficients must be set per unique pair of particle types:
+    Where each parameter is defined as:
 
     - :math:`Z_i` - *Z_i* - Atomic number of species i (unitless)
     - :math:`Z_j` - *Z_j* - Atomic number of species j (unitless)
     - :math:`e` - *elementary_charge* - The elementary charge (in charge units)
     - :math:`a_0` - *a_0* - The Bohr radius (in distance units)
```
joaander, line 1184: `a_0` isn't actually used in the above expression. Put the definition of `a_F` here with ending with `, where *a_0* is the Bohr radius (in distance units)`

hoomd/md/pair.py
```diff
@@ -1414,72 +1176,57 @@ class moliere(pair):
...
+
+    See :py:class:`_Pair` for details on how forces are calculated and the available energy shifting and smoothing modes.
+    Use ``params`` property to set potential coefficients.
+
+    The following coefficients must be set per unique pair of particle types:
+
+    - :math:`q_i` - *qi* - :math:`q_i = Z_i e`
```
joaander, line 1191: In HOOMD's system of units, this should read `q_i = Z_i e / (4 \pi \epsilon_0)` (in charge units) - and similarly for `q_j`.

hoomd/md/pair.py
```diff
@@ -1490,68 +1237,48 @@ class zbl(pair):
...
-    See :py:class:`_Pair` for details on how forces are calculated and the available energy shifting and smoothing modes.
-    Use ``coeff.set`` to set potential coefficients.
-
-    The following coefficients must be set per unique pair of particle types:
+    Where each parameter is defined as:
 
     - :math:`Z_i` - *Z_i* - Atomic number of species i (unitless)
     - :math:`Z_j` - *Z_j* - Atomic number of species j (unitless)
     - :math:`e` - *elementary_charge* - The elementary charge (in charge units)
     - :math:`a_0` - *a_0* - The Bohr radius (in distance units)
```
joaander, line 1245: Same comments here as above.

hoomd/md/pair.py
tommy-waltmann: Shouldn't that be `q_i = Z_i e / \sqrt{4 \pi \epsilon_0}` then?

PR #656, opened 2020-05-18
Improved Validation and Defaults

b-butler:
## Description

Make the validation specification more consistent and easier across the code base.

## Motivation and Context

I noticed that we were doing the same thing in multiple places using multiple solutions. This should
help fix that and help to prevent it in the future.

## How Has This Been Tested?

All existing tests still pass, and MD simulations still work locally.

## Checklist:


hoomd/typeconverter.py
```diff
@@ -1,82 +1,77 @@
 from numpy import array, ndarray
```
joaander, line 1: Add docstrings so that developers understand how these functions and classes should be used.

hoomd/util.py
```diff
@@ -265,44 +261,25 @@ def array_to_strings(value):
 
 
 def trigger_preprocessing(trigger):
-    if isinstance(trigger, int):
```
joaander, line 268: Why are tehse and other similar methods here at not in `typeconverter.py`?

hoomd/typeconverter.py
```diff
@@ -261,30 +248,7 @@ def toTypeConverter(value):
         return TypeConverterFixedLengthSequence(value)
```
joaander, line 248: Shouldn't this be named `to_type_converter`?

hoomd/util.py
b-butler: Moved them

hoomd/typeconverter.py
b-butler: Good idea done.

hoomd/typeconverter.py
b-butler: Yes, I was just using it as a "pretend" class. Done.

b-butler:
\azp run

PR #658, opened 2020-05-26
V3/improved docs

b-butler:
## Description

Improves/adds documentation for v3 classes. The PR fixes links to classes and other bugs currently in the documentation.

This isn't designed to *finish* the documentation for current classes, just to improve it to make adding to the documentation and checking the local building easier and happen without warnings on the v3 side.

## Motivation and Context

I wanted local documentation to work better.

## How Has This Been Tested?
Current tests are sufficient.

## Checklist:


hoomd/custom_action.py
```diff
@@ -134,3 +134,13 @@ class _InternalCustomAction(CustomAction, _HOOMDGetSetAttrBase):
     be created.
     """
     pass
+
+
+class ParticleDataFlags(IntEnum):
```
joaander, line 139: Can we make this a nested class inside `CustomAction` as it is only used there?
Also, and maybe outside the scope of this PR, can we move `_CustomOperation`, `CustomAction` and `Logger` out of the root namespace add into submodules/packages? I’d like to keep the root namespace clean with only the objects that all HOOMD simulations interact with.

sphinx-doc/module-hoomd-logger.rst
```diff
@@ -0,0 +1,16 @@
...
+.. rubric:: Overview
+
+.. autosummary::
+    :nosignatures:
+
+    hoomd.logger.LoggerQuantity
+
+.. rubric:: Details
+
+.. automodule:: hoomd.logger
```
joaander, line 13: We are documenting the `logger` module now, so users can use `Logger` from here.

hoomd/CMakeLists.txt
```diff
@@ -352,6 +350,9 @@ install(FILES ${_hoomd_headers}
 
 # subdirectories that are not components
 add_subdirectory(filter)
+add_subdirectory(custom)
```
joaander, line 353: Now CMake is giving strange errors related to these subdirectories. I'm not sure why.
```
CMake Error at hoomd/CMakeLists.txt:429 (add_subdirectory):
  The binary directory

    /__w/1/b/hoomd/analyze

  is already used to build a source directory.  It cannot be used to build
  source directory

    /__w/1/s/hoomd/analyze

  Specify a unique binary directory name.


CMake Error at hoomd/CMakeLists.txt:429 (add_subdirectory):
  The binary directory

    /__w/1/b/hoomd/custom

  is already used to build a source directory.  It cannot be used to build
  source directory

    /__w/1/s/hoomd/custom

  Specify a unique binary directory name.


CMake Error at hoomd/CMakeLists.txt:429 (add_subdirectory):
  The binary directory

    /__w/1/b/hoomd/update

  is already used to build a source directory.  It cannot be used to build
  source directory

    /__w/1/s/hoomd/update

  Specify a unique binary directory name
```

b-butler: Fixes as per our discussion

PR #666, opened 2020-06-04
Feature/new-npt

syjlee:
## Description

Porting NPT integration method to the new v3 API.
`randomize_velocities()` is not yet touched.
To leave a note about the more changes than other integrators,
1. parameter P and S are not separate anymore, combined into S. taking single variant/or float or 6 element of variant/or float.
2. parameter `couple` is directly ported to C++ and check for system dimension, and re-parameterized.
3. seven parameters related to `flags` which is box degrees of freedom is combined into one tuple-type parameter composed of 6 boolean elements. Also, directly ported to C++ through `param_dict`. Further bitwise operation is done in c++ files.
4. get, set functions are added and modified 
5. `def_property` was reset to match with how `param_dict` functions.

## Motivation and Context

NPT integrator is needed to be rephrased and ported for new v3 API.

## How Has This Been Tested?

Overall test is not done yet. Please comment if test should be done now before merging, and let's discuss how to do so without having to make major changes out of scope.
Partial test is done only on some newly added blobs on c++ in syntax wise.

## Change log
```
Moved NPT integration method to v3 API
```
## Checklist:

- [] My name is on the [list of contributors](https://github.com/glotzerlab/hoomd-blue/blob/master/sphinx-doc/credits.rst).


syjlee:
I forgot to make changes on `TwoStepNPTMTKGPU.h` and `TwoStepNPTMTKGPU.cc` . I'm making similar changes to TwoStepNPTMTK. Let me reflect other future comments/changes on both GPU and not GPU files at the same time.

syjlee:
What is the proper way to resolve conflicting files? The change I made in 'hoomd/md/methods.py' has conflicts on introducing `preprocess_stress`.

joaander:
What changes do you think you need to make to the `*GPU` classes? Those should just inherit the implementation from the base class.

To resolve conflicts, merge `feature/new-object-API` into your branch and then fix the conflicts.

hoomd/md/TwoStepNPTMTK.cc
```diff
@@ -839,10 +891,102 @@ void TwoStepNPTMTK::advanceThermostat(unsigned int timestep)
...
+    }
+*/
+
 TwoStepNPTMTK::couplingMode TwoStepNPTMTK::getRelevantCouplings()
     {
     couplingMode couple = m_couple;
 
+    // if sys_def is 2D, silently ignore any couplings that involve z
+    bool twod = m_sysdef->getNDimensions()==2;
+    if (twod)
```
joaander, line 974: I would make this check in `setCouple`, not `getRelevantCouplings`. We could also issue a warning to the user if they set a coupling that is ignored.

hoomd/md/TwoStepNPTMTK.h
```diff
@@ -55,12 +55,12 @@ class PYBIND11_EXPORT TwoStepNPTMTK : public IntegrationMethodTwoStep
          */
         enum baroFlags
             {
-            baro_x = 1,
-            baro_y = 2,
-            baro_z = 4,
-            baro_xy = 8,
-            baro_xz = 16,
-            baro_yz = 32
+            baro_x = 32,
```
joaander, line 58: Why change these?

hoomd/md/TwoStepNPTMTK.h
```diff
@@ -86,9 +86,17 @@ class PYBIND11_EXPORT TwoStepNPTMTK : public IntegrationMethodTwoStep
             m_T = T;
             }
 
+        //! Update the stress components
+        /*! \param S list of stress components: [xx, yy, zz, yz, xz, xy]
+         */
+        virtual void setS(std::vector<std::shared_ptr<Variant>> S)
```
joaander, line 92:
```suggestion
        virtual void setS(const std::vector<std::shared_ptr<Variant>>& S)
```

hoomd/md/TwoStepNPTMTK.h
```diff
@@ -86,9 +86,17 @@ class PYBIND11_EXPORT TwoStepNPTMTK : public IntegrationMethodTwoStep
...
+            {
+             m_S = S;
+            }
+
+    /*
     //! Update the stress components
-    /*! \param S list of stress components: [xx, yy, zz, yz, xz, xy]
-     */
+    //! \param S list of stress components: [xx, yy, zz, yz, xz, xy]
```
joaander, line 99: Remove code, don't comment it out.

hoomd/md/TwoStepNPTMTKGPU.cc
```diff
@@ -37,14 +37,14 @@ TwoStepNPTMTKGPU::TwoStepNPTMTKGPU(std::shared_ptr<SystemDefinition> sysdef,
...
                        std::shared_ptr<ComputeThermo> thermo_group_t,
                        Scalar tau,
-                       Scalar tauP,
+                       Scalar tauS,
                        std::shared_ptr<Variant> T,
-                       pybind11::list S,
-                       couplingMode couple,
-                       unsigned int flags,
+                       std::vector<std::shared_ptr<Variant>> S,
+                       string couple,
```
joaander, line 43:
```suggestion
                       std::string couple,
```

hoomd/md/TwoStepNPTMTKGPU.cc
```diff
@@ -383,9 +383,9 @@ void export_TwoStepNPTMTKGPU(py::module& m)
                        Scalar,
                        Scalar,
                        std::shared_ptr<Variant>,
-                       pybind11::list,
-                       TwoStepNPTMTKGPU::couplingMode,
-                       unsigned int,
+                       std::vector<std::shared_ptr<Variant>>,
+                       string,
```
joaander, line 387:
```suggestion
                       std::string,
```

hoomd/md/TwoStepNPTMTK.h
syjlee: @joaander I thought opposite sequence for baroflags is better to append in the order 'x,y,z,xy,xz,yz' in get function in cc file in line number 717-722. Will there be other built-in function to append in the opposite order?

hoomd/md/TwoStepNPTMTK.cc
syjlee: @joaander  I'm making the change. I'm moving the checkpoint to `setCouple` and inserted warning message in the case of z coupling ignored.

hoomd/md/TwoStepNPTMTK.cc
```diff
@@ -839,10 +891,102 @@ void TwoStepNPTMTK::advanceThermostat(unsigned int timestep)
...
+    else if (value == "xyz")
+        {
+        m_couple = couple_xyz
+        }
+    }
+
+std::string TwoStepNPTMTK::getCouple()
+    {
+    std::string delimiter = "_";
+    std::string split = m_couple.substr(1, s.find(delimiter));
```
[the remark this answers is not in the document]
syjlee: I changed `getCouple` function back with if statements.

hoomd/md/TwoStepNPTMTK.h
joaander: You are right that it makes the order of the bits in flags consistent with the order of bits in the `std::vector<bool>`. There may be some benefit to this, but given the way you explicitly convert the two back and forth the change is not needed.

SchoeniPhlippsn:
I have added the changes concerning the degrees of freedom from #691. I also tested both the NVT and NPT integrator and it fixed the issue of the exploding box (NPT). Other than that the integrators give consistent results now as well. Please let me know if there are more changes necessary.

hoomd/md/TwoStepNPTMTK.cc
```diff
@@ -41,41 +41,45 @@ TwoStepNPTMTK::TwoStepNPTMTK(std::shared_ptr<SystemDefinition> sysdef,
...
-        m_exec_conf->msg->warning() << "integrate.npt: tauP set less than 0.0" << endl;
+    if (m_tauS <= 0.0)
+        m_exec_conf->msg->warning() << "integrate.npt: tauS set less than 0.0" << endl;
 
-    if (flags == 0)
+    if (m_flags == 0)
         m_exec_conf->msg->warning() << "integrate.npt: No barostat couplings specified."
                                     << endl;
 
+     /*
```
joaander, line 76: Remove unused code.

hoomd/md/TwoStepNPTMTK.cc
```diff
@@ -689,6 +693,39 @@ void TwoStepNPTMTK::updatePropagator(Scalar nuxx, Scalar nuxy, Scalar nuxz, Scal
...
+    if (value[5] && !(twod)) 
+        flags |= int(baroFlags::baro_yz);
+    m_flags = flags;
+    }
+
+// Get Flags from integer flag to 6 element boolean tuple
+std::vector<bool> TwoStepNPTMTK::getFlags()
+    {
+    std::vector<bool> result;
+    result.push_back(m_flags & baro_yz);
```
joaander, line 720: Have you tested this by setting flags and ensuring that you get back the same flags you set? I am concerned that the flags appear to be pushed onto the result vector in the opposite order of setFlags. It appears that if you set the flags [1, 0, 0, 0, 0, 0], you will get back the flags [0, 0, 0, 0, 0, 1].

hoomd/md/TwoStepNPTMTK.h
```diff
@@ -86,18 +86,12 @@ class PYBIND11_EXPORT TwoStepNPTMTK : public IntegrationMethodTwoStep
...
             }
 
-    //! Update the stress components
-    /*! \param S list of stress components: [xx, yy, zz, yz, xz, xy]
-     */
-    virtual void setS(pybind11::list S)
+        //! Update the stress components
+        /*! \param S list of stress components: [xx, yy, zz, yz, xz, xy]
+         */
+        virtual void setS(const std::vector<std::shared_ptr<Variant>> S)
```
joaander, line 92:
```suggestion
        virtual void setS(const std::vector<std::shared_ptr<Variant>>& S)
```
Pass complex types by const reference to avoid copies.

hoomd/md/methods.py
```diff
@@ -225,251 +225,79 @@ class npt(_Method):
...
+            tau=float(tau),
+            S=(Variant,)*6,
+            tauS=float(tauS),
+            couple=str(couple),
+            box_dof=(bool,)*6,
+            nph=bool(False),
+            rescale_all=bool(rescale_all),
+            gamma=float(gamma)
+            )
+        #couple=none_or(OnlyFrom('xyz')),
```
joaander, line 254: Remove commented testing code.

hoomd/md/methods.py
```diff
@@ -225,251 +225,79 @@ class npt(_Method):
...
+        #S=(Variant,)*6,
+        #S=preprocess_stress,
+        #filter=OnlyType(_ParticleFilter)
+        param_dict.update(dict(filter=filter, kT=kT, S=preprocess_stress(S),
```
joaander, line 258: @b-butler Is this the right way to specify the stress preprocessing?

hoomd/md/methods.py
```diff
@@ -225,251 +225,79 @@ class npt(_Method):
...
-            group_copy = copy.copy(group)
-            group_copy.name = "__npt_all"
-            thermo_group = hoomd.compute.thermo(group_copy)
-            thermo_group.cpp_compute.setLoggingEnabled(False)
-        else:
-            thermo_group = hoomd.compute._get_unique_thermo(group=group)
+        # store metadata
+        param_dict = ParameterDict(
+            filter=_ParticleFilter,
+            kT=Variant,
```
joaander, line 244: `kT` is the right name for the parameter, but C++ still exports the member as `T`. Have you tested if you can set `kT` after attaching? I think that test will fail unless you export the `setT/getT` as `kT` from C++.

hoomd/md/methods.py
```diff
@@ -225,251 +225,79 @@ class npt(_Method):
...
-            self.S = [P,P,P,0,0,0]
-        else:
-            # S is a stress, should be [xx, yy, zz, yz, xz, xy]
-            if S is not None and len(S)==6:
-                self.S = S
-            else:
-                raise RuntimeError("Unrecognized stress tensor form")
-
-        S = [hoomd.variant._setup_variant_input(self.S[i]) for i in range(6)]
+    def __init__(self, filter, kT=None, tau=None, S=None, tauS=None, couple="xyz", box_dof=[True,True,True,False,False,False], rescale_all=False, gamma=0.0):
```
joaander, line 238: `kT`, `tau` and others should not default to `None`. I think this was here previously to support the implementation of `NPH`. We should just implement `NPH` with its own param_dict and attach method (or creatively re-use the `NPT` attach method via inheritance).

hoomd/md/pytest/test_methods.py
```diff
@@ -7,6 +7,7 @@
 _method_data = [
     ("Langevin", hoomd.md.methods.Langevin(hoomd.filter.All(), kT=1, seed=1)),
     ("NVT", hoomd.md.methods.NVT(hoomd.filter.All(), kT=1, tau=1)),
+    ("NPT", hoomd.md.methods.NPT(hoomd.filter.All(), kT=1, tau=1, S=1, tauS=1, box_dof=[True,True,True,False,False,False])),
 ]
```
joaander, line 11: Let's add some NPT specific tests that validate all the parameters that can be correctly set and query both before and after attaching.

hoomd/md/TwoStepNPTMTK.cc
```diff
@@ -839,10 +876,90 @@ void TwoStepNPTMTK::advanceThermostat(unsigned int timestep)
...
+            m_couple = couple_none;
+            m_exec_conf->msg->warning() << "z coupling is ignored, since system is two dimensional. coupling is none." << endl;
+            }
+        else if (value == "xyz")
+            {
+            m_couple = couple_xy;
+            m_exec_conf->msg->warning() << "z coupling is ignored, since system is two dimensional. coupling is xy." << endl;
+            }
+        }
+    }
```
b-butler, line 932: I am not sure I agree with silently ignoring `z` here. I think erroring makes the most sense.

hoomd/md/TwoStepNPTMTK.cc
```diff
@@ -839,10 +876,90 @@ void TwoStepNPTMTK::advanceThermostat(unsigned int timestep)
...
+        }
+    else if (m_couple == couple_yz)
+        {
+        couple = "yz";
+        }
+    else if (m_couple == couple_xyz)
+        {
+        couple = "xyz";
+        }
+    return couple;
```
b-butler, line 957: This is a stylistic choice, but I would prefer to see a switch statement here https://www.tutorialspoint.com/cplusplus/cpp_switch_statement.htm

hoomd/md/TwoStepNPTMTK.cc
```diff
@@ -1015,34 +1132,17 @@ void export_TwoStepNPTMTK(py::module& m)
...
+                       const string&,
+                       const std::vector<bool>&,
                        const bool>())
-        .def("setT", &TwoStepNPTMTK::setT)
-        .def("setS", &TwoStepNPTMTK::setS)
-        .def("setTau", &TwoStepNPTMTK::setTau)
-        .def("setTauP", &TwoStepNPTMTK::setTauP)
-        .def("setRescaleAll", &TwoStepNPTMTK::setRescaleAll)
-        .def("setGamma", &TwoStepNPTMTK::setGamma)
+        .def_property("T", &TwoStepNPTMTK::getT, &TwoStepNPTMTK::setT)
```
b-butler, line 1139: This needs to be `kT`

hoomd/md/methods.py
```diff
@@ -116,40 +127,31 @@ def attach(self, simulation):
...
 
     Args:
-        group (``hoomd.group``): Group of particles on which to apply this method.
+        filter (:py:mod:`hoomd.filter._ParticleFilter`): Subset of particles on which to apply this method.
         kT (:py:mod:`hoomd.variant` or :py:obj:`float`): Temperature set point for the thermostat, not needed if *nph=True* (in energy units).
         tau (float): Coupling constant for the thermostat, not needed if *nph=True* (in time units).
-        S (:py:class:`list` of :py:mod:`hoomd.variant` or :py:obj:`float`): Stress components set point for the barostat (in pressure units). In Voigt notation: [Sxx, Syy, Szz, Syz, Sxz, Sxy]
-        P (:py:mod:`hoomd.variant` or :py:obj:`float`): Isotropic pressure set point for the barostat (in pressure units). Overrides *S* if set.
-        tauP (float): Coupling constant for the barostat (in time units).
+        S (:py:class:`list` of :py:mod:`hoomd.variant` or :py:obj:`float`): Stress components set point for the barostat (in pressure units). In Voigt notation: [Sxx, Syy, Szz, Syz, Sxz, Sxy]. In case of isotropic pressure P, use [P,P,P,0,0,0]
```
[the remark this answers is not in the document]
b-butler: This is also technically incorrect. For an isotropic pressure `P` just set to `P`.

hoomd/md/methods.py
```diff
@@ -116,40 +127,31 @@ def attach(self, simulation):
...
-        x (bool): if True, rescale *Lx* and x component of particle coordinates and velocities
-        y (bool): if True, rescale *Ly* and y component of particle coordinates and velocities
-        z (bool): if True, rescale *Lz* and z component of particle coordinates and velocities
-        xy (bool): if True, rescale xy tilt factor and x and y components of particle coordinates and velocities
-        xz (bool): if True, rescale xz tilt factor and x and z components of particle coordinates and velocities
-        yz (bool): if True, rescale yz tilt factor and y and z components of particle coordinates and velocities
-        all (bool): if True, rescale all lengths and tilt factors and components of particle coordinates and velocities
-        nph (bool): if True, integrate without a thermostat, i.e. in the NPH ensemble
+        box_dof(tuple): Box degrees of freedom with six boolean elements corresponding to x, y, z, xy, xz, yz, each. (default: [True,True,True,False,False,False]) If turned on to True, rescale corresponding lengths or tilt factors and components of particle coordinates and velocities
```
b-butler, line 140: Wrap line please. Also, the default is not a list but a tuple `()` not `[]`

hoomd/md/methods.py
```diff
@@ -169,18 +171,16 @@ class npt(_Method):
     Degrees of freedom of the box specify which lengths and tilt factors of the box should be updated,
     and how particle coordinates and velocities should be rescaled.
 
-    Valid keywords for degrees of freedom are:
+    Valid form for elements of box_dof(box degrees of freedom) is :
```
b-butler, line 174:
```suggestion
    The `box_dof` tuple controls the way the box is rescaled and updated. The first three
    elements ``box_dof[:2]`` controls whether the x, y, and z box lengths are rescaled and
    updated, respectively. The last three entries control the rescaling or the tilt factors
    xy, xz, and yz. All options also appropriately rescale particle coordinates and velocities.
```

hoomd/md/methods.py
```diff
@@ -169,18 +171,16 @@ class npt(_Method):
...
-    - xy (the tilt factor xy is updated)
-    - xz (the tilt factor xz is updated)
-    - yz (the tilt factor yz is updated)
-    - all (all elements are updated, equivalent to x, y, z, xy, xz, and yz together)
+    - bool(box_dof[0]) (if True, rescale *Lx* and x component of particle coordinates and velocities, and the box length Lx is updated)
+    - bool(box_dof[1]) (if True, rescale *Ly* and y component of particle coordinates and velocities, and the box length Ly is updated)
+    - bool(box_dof[2]) (if True, rescale *Lz* and z component of particle coordinates and velocities, and the box length Lz is updated)
+    - bool(box_dof[3]) (if True, rescale xy tilt factor and x and y components of particle coordinates and velocities, and the tilt factor xy is updated)
+    - bool(box_dof[4]) (if True, rescale xz tilt factor and x and z components of particle coordinates and velocities, and the tilt factor xz is updated)
+    - bool(box_dof[5]) (if True, rescale yz tilt factor and y and z components of particle coordinates and velocities, and the tilt factor yz is updated)
```
b-butler, line 181: Remove

hoomd/md/methods.py
```diff
@@ -169,18 +171,16 @@ class npt(_Method):
...
 
-    Any of the six keywords can be combined together. By default, the x, y, and z degrees of freedom
-    are updated.
+    By default, the x, y, and z degrees of freedomare updated. [True,True,True,False,False,False]
```
b-butler, line 183:
```suggestion
    By default, the x, y, and z degrees of freedom are updated. [True,True,True,False,False,False]
```

hoomd/md/methods.py
```diff
@@ -192,17 +192,17 @@ class npt(_Method):
...
+    :py:class:`NPT` Can also apply a constant stress to the simulation box. To do so, specify the symmetric
     stress tensor *S* instead of an isotropic pressure *P*.
 
     Note:
-        :py:class:`npt` assumes that isotropic pressures are positive. Conventions for the stress tensor sometimes
+        :py:class:`NPT` assumes that isotropic pressures are positive. Conventions for the stress tensor sometimes
         assume negative values on the diagonal. You need to set these values negative manually in HOOMD.
 
-    :py:class:`npt` is an integration method. It must be used with ``mode_standard``.
+    :py:class:`NPT` is an integration method. It must be used with ``mode_standard``.
```
b-butler, line 202: `mode_standard` does not exist now. Remove.

hoomd/md/methods.py
```diff
@@ -225,251 +225,79 @@ class npt(_Method):
     A :py:class:`hoomd.compute.thermo` is automatically specified and associated with *group*.
 
     Examples::
-
-        integrate.npt(group=all, kT=1.0, tau=0.5, tauP=1.0, P=2.0)
-        integrator = integrate.npt(group=all, tau=1.0, kT=0.65, tauP = 1.2, P=2.0)
+        all = filter.All()
```
b-butler, line 228: Do not use variable name `all` this is bad practice. Use `all_` or just call `filter.All()` where needed.

hoomd/md/methods.py
```diff
@@ -225,251 +225,79 @@ class npt(_Method):
...
+        integrate.NPT(filter=all, kT=1.0, tau=0.5, tauS=1.0, S=2.0)
```
b-butler, line 229: This line does nothing any more. Remove.

hoomd/md/methods.py
b-butler: This still needs to be addressed.

hoomd/md/TwoStepNPTMTK.cc
joaander: The default value of the coupling is "xyz". Should users that run 2D systems need to opt out of this default to avoid the proposed error? What if we renamed this to "all"?

joaander (approved):
Thanks! Sorry for the delay getting back to this PR. I made a few changes. I think it is good to merge now.

Final documentation cleanup will be done after merging this into #749

PR #673, opened 2020-06-26
Add immediate mode APIs

joaander:
## Description

Add the immediate mode APIs `GSD.write_state` and `BoxResize.scale_state`.

## Motivation and Context

These static methods allow the user to write out a GSD file or scale the box with an API call directly in their script without needing to schedule an operation to occur at a given time step.
```
hoomd.dump.GSD.write_state(sim.state, filename, filter, log)
hoomd.update.BoxResize.scale_state(sim.state, new_box)
```

## How Has This Been Tested?

So far, I have tested it in the tutorials. Will add unit tests before marking this PR ready to review.

## Checklist:


joaander:
On second thought, there are currently no tests for BoxResize or GSD. I am deferring testing of the immediate mode APIs for now.

hoomd/update/box_resize.py
```diff
@@ -99,6 +99,19 @@ def get_box(self, timestep):
         else:
             return None
 
+    @staticmethod
+    def apply(state, box):
```
b-butler, line 103: Should we use update? We are using `write` for `Writer`s.

joaander: Yes, that makes more sense.

Issue #680, opened 2020-07-07
Compile errors in v2.9.2


joaander:
## Description

* `neighbor/UniformGrid.cu` does not compile with CUDA 11.
* `cudaThreadSynchronize` has been removed in CUDA 11.
* Other changes may be necessary

PR #681, opened 2020-07-07
Use neighbor fixed up for CUDA 11

mphoward:
## Description

CUDA 11 has CUB internal, so neighbor 0.2 needs to use this version rather than HOOMD's copy. This problem is already solved in neighbor 0.3 for HOOMD 3.x, which always uses `<cub/cub.cuh>`, but patches the issue for HOOMD 2.9. See:

https://github.com/mphowardlab/neighbor/commit/0956bd3ed4601a1793b7b2711af3b9880a88d8fc

## Motivation and Context

#680 

## How Has This Been Tested?

User confirmed that this fixed their compilation issue with CUDA 11.

## Checklist:


b-butler:
This seems to cause a bunch of tests on MacOS to fail.

mphoward:
The macOS failures occur even on CPU builds and all stem from a numpy import error. This doesn't seem like it could be caused by this commit, honestly, so could you check your CI configuration?

mphoward:
(The last two commits on maint have the same errors.)

b-butler:
@mphoward you are correct, I should have checked Azure first. In that case, I don't have any other comments given the simple nature of the PR.

joaander:
Sorry, I fixed the mac build errors on `next` but failed to backport those changes to `maint`. At this point we are only applying essential fixes to `maint` as we prepare for a 3.0 beta release soon so I'm not going to put too much effort into it.

I'm going to retarget this to a 2.9 branch so I don't need to keep dealing with merge conflicts as we merge fixes from `maint` all they way up to `feature/new-object-api`. I'll take care of duplicating any future fixes to ensure they make it into 3.0.

Issue #687, opened 2020-07-14
Automatically update the number of degrees of freedom in Group


joaander:
- Update Group to track the number of degrees of freedom it has.
- Query the ndof when needed by a ComputeThermo or an Integrator
- Write `update_group_dof` logic to update these values
- Call `update_group_dof` when needed
  - [x] On snapshot set.
  - [ ] ~~On releasing a local snapshot~~ (user must call).
  - [x] When changing integrators.
  - [x] When changing the list of integration methods.
  - [x] Particle filter updater
  - [x] When changing the aniso mode.
  - [x] When adding new groups.
  - and others?

joaander:
While testing a script, I found that `updateGroupDOF` needs to be called when new groups are added after attaching.

joaander:
Idea: Make the SystemDefinition own the group cache so it is available to C++.

joaander:
`cpu_local_snapshot` and `gpu_local_snapshot` are properties, so there is no way to provide an argument to opt-in to automatic group dof updates. Users will need to manually call `update_group_dof`.

PR #691, opened 2020-07-24
Store the number of degrees of freedom in ParticleGroup

joaander:
## Description

Move the storage of the number of degrees of freedom from `ComputeThermo` to `ParticleGroup`. Add a new method `State.update_group_dof` that queries the current integrator and updates the DOF provided to all groups in the state's group cache.

## Motivation and Context

In the HOOMD v3 API there is no longer a list of `ComputeThermo` objects and there  is no preamble to `run` to update the DOF for them. Instead, we make the group itself own the DOF count and update that when needed.

This PR also fixes:
* A bug where changing the methods applied to an Integrator results in a dt of 0.
* A bug where removed DOF were overcounted for any group that was not the all group.
* A bug where Langevin and BD integration had too many DOF removed (the center of mass is not fixed).

This starts work on #687 but does not complete it. We also need to add hooks to automatically call `update_group_dof` when needed to keep this count up to date.

## How Has This Been Tested?

I have tested it locally. Unfortunately, unit tests are not feasible until we implement the python interface to `ComputeThermo`.

The documentation for `update_group_dof` renders correctly.

## Checklist:


joaander:
@SchoeniPhlippsn @syjlee This PR should fix the issues with `ComputeThermo` computing incorrect values for `NVT` and `NPT` integration. Note that it currently only updates the DOF *when attaching the integrator*. If you change the integration methods later, you can manually call `update_group_dof` to recount them:
```
integrator.methods = [nvt]
sim.state.update_group_dof()
```

hoomd/md/IntegratorTwoStep.cc
```diff
@@ -104,9 +104,13 @@ void IntegratorTwoStep::update(unsigned int timestep)
         m_prof->push("Integrate");
 
     // perform the first step of the integration on all groups
-    std::vector< std::shared_ptr<IntegrationMethodTwoStep> >::iterator method;
-    for (method = m_methods.begin(); method != m_methods.end(); ++method)
+    for (auto method = m_methods.begin(); method != m_methods.end(); ++method)
```
b-butler, line 107: We could use a range-based for loop here.

hoomd/md/IntegratorTwoStep.cc
```diff
@@ -145,7 +149,7 @@ void IntegratorTwoStep::update(unsigned int timestep)
         m_prof->push("Integrate");
 
     // perform the second step of the integration on all groups
-    for (method = m_methods.begin(); method != m_methods.end(); ++method)
+    for (auto method = m_methods.begin(); method != m_methods.end(); ++method)
```
b-butler, line 152: Same as comment above. If you prefer this still just resolve the issues.

hoomd/md/IntegratorTwoStep.cc
```diff
@@ -264,29 +268,44 @@ void IntegratorTwoStep::initializeIntegrationMethods()
...
+    // proportionately remove n_dimensions DOF when there is only one momentum conserving
+    // integration method
+    Scalar periodic_dof_removed = 0;
+    if (group->getNumMembersGlobal() == m_pdata->getNGlobal() &&
+        m_methods.size() == 1 &&
+        m_methods[0]->isMomentumConserving())
+        {
+        periodic_dof_removed = Scalar(m_sysdef->getNDimensions()) *
+                               (Scalar(group->getNumMembersGlobal())
+                               / Scalar(m_pdata->getNGlobal()));
```
b-butler, line 292: Isn't this by definition 1?

hoomd/state.py
```diff
@@ -198,3 +198,31 @@ def get_group(self, filter_):
...
+
+        * The Integrator is attached to the simulation
+
+        Call it manually to force an update.
+        """
+        integrator = self._simulation.operations.integrator
+
+        # Do nothing if there is no integrator.
+        if integrator is None:
+            return
```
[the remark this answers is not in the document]
joaander: Assume that `update_group_dof` is the method that later gets triggered when we set a new snapshot. What if the user set's a snapshot before assigning an integrator? Should we have a separate internal method that does nothing when there is no integrator so that the user-facing method can throw an error?

hoomd/state.py
```diff
@@ -198,3 +198,31 @@ def get_group(self, filter_):
...
+
+        if not integrator.is_attached:
+            raise RuntimeError("Call update_group_dof after attaching")
+
+        for groups in self._groups.values():
+            for group in groups.values():
+                if integrator is not None:
```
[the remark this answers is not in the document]
joaander: Good catch. Perhaps the more correct behavior is to set all group DOF's to 0 when the integrator is None.

hoomd/md/IntegratorTwoStep.cc
joaander: Done.

hoomd/md/IntegratorTwoStep.cc
joaander: It's not a preference - much of this code was written before C++11 existed and long before it was widely available.

hoomd/md/IntegratorTwoStep.cc
joaander: Only when the group consists of all particles. HOOMD v2 incorrectly removes N_dimensions DOF from every group.

When the group is only a subset of all particles, this spreads the N_dimensions fixed degrees of freedom out proportionately. With this change if you compute the kinetic temperature of two halves of the system separately and then average them together you will get the same result as if you had computed the kinetic temperature on the all particles. There might be a better way to do this, but I can't think of anything else that would be self-consistent with a simulation of otherwise identical particles.

I also added a check to only remove these DOF when using a momentum conserving integrator. Now the kinetic temperature will be computed correctly when using Langevin where the center of mass of the system moves.

joaander:
> Overall, great work. I have a few comments to be addressed. Since I won't be reviewing next week. I will approve the PR rather than comment or request changes. I do think there are some dead code paths in state.py that should be removed, and one comment about dividing to get zero (which may not be the case, the code just seemed that way to me).

ComputeThermo will divide by 0 and then multiply by zero resulting in NaN (probably) for the kinetic temperature when there are no degrees of freedom in the group (or no integrator to provide them). We can solve this issue with an update to ComputeThermo - but we do need to decide on what the right behavior is and document it. If a user runs a simulation with no integrator or otherwise requests a ComputeThermo to compute the kinetic temperature of a group of particles that does not have an integration method applied (they are frozen) - what should the kinetic temperature be? 0 because the kinetic energy is 0? NaN because there are no degrees of freedom over which to distribute the kinetic energy? I opened #708 to track this issue.

PR #692, opened 2020-07-24
Feature/custom tuners

b-butler:
## Description
Adds custom tuners to the list of custom operations that can be implemented in Python.

## Motivation and Context

We plan to internally use this for HPMC move size tunes and similar quantities.

## How Has This Been Tested?

This likely should be tested with the other custom actions. I think it would be best to do this together in its own PR.

## Checklist:


hoomd/custom/custom_operation.py
```diff
@@ -85,6 +83,11 @@ def attach(self, simulation):
         super().attach(simulation)
         self._action.attach(simulation)
 
+    def detach(self):
+        """Detaching to a `hoomd.Simulation`."""
```
joaander, line 87: Detach from a ...

Issue #678, opened 2020-06-30
MoveSize tuner


joaander:
## Description

Implement the move size tuner as a `Tuner` in the v3 API.

## Proposed solution

```
tune = hoomd.hpmc.tune.TrialMoveSize(integrator=mc, tune=['a', 'd'], trigger=hoomd.trigger.Periodic(10))
sim.operations.tuners.append(tune)
```

## Additional context

We need to port the v2 tuner from hpmc.util to v3 as a `Tuner`, update the code to work with the new property based access to move counts, and clean up the interface.

## Developer

@b-butler could you implement this?

PR #704, opened 2020-07-31
Feature/move tuners

b-butler:
## Description

Adds a class for tuning HPMC integrator move sizes. This PR also provides a means for generic solving of simple equations of the form f(x) = y for a specific y value.

I implemented a definition and solver model where a definition defines the relationship and a solver iterators towards the target y value. For `MoveSize` the class that implements the move size tuning, it composes both these other objects to appropriately tune move sizes.

There are two solvers created in this PR one that scales x according to the ratio between the target y value and the current value (this solver is restricted to strictly positive quantities where the correlation between x and y is known and expected to be monotonic), and a solver that implements the secant method.

`MoveSize` is independent of the solver chosen; however, classmethods for easily creating `MoveSize` objects without dealing with solvers are provided.

## Motivation and Context

This brings move size tuning into HOOMD-blue version 3.

## How Has This Been Tested?

With two testing files for the generic and specific attribute tuning.

## Checklist:


vyasr:
@joaander since one goal with v3 is greater integration with the ecosystem, what are your thoughts on using scipy for Solvers here? I could understand not wanting to make scipy a hard dependency for HPMC (although in practice it's been a long time since I had trouble installing scipy anywhere), in which case we could implement one naive Solver (such as the secant method that @b-butler already has), but then allow users to install scipy to enable other (perhaps more efficient) solvers. I'm wary of introducing this kind of code into HOOMD when so many packages already do it well.

b-butler:
@vyasr I had thought about something like that, but I wasn't sure how to do so, since we are not iterating over the solution all at once. In general, we have to run the simulation for some time before recomputing the independent variables. I don't know if scipy has functionality that would support that model, (except maybe to for the function computing the root, running the simulation and returning those dependent variables). That might be a better solution in general than having an expansive solver infrastructure, but wouldn't help in the case of move tuning and things of that nature as such solvers would not be able to be used in `Tuner` or `Updater` objects.

vyasr:
That's a good point. Maybe it's just a naming issue then, the object is really more like a `SolverStep`. The solver doesn't actually solve an equation, it only takes a single iteration towards solving it. You have to manually loop to solve.

b-butler:
Addresses #678

hoomd/hpmc/tune/move_size.py
```diff
@@ -0,0 +1,333 @@
...
+        current_ratio = getattr(self.integrator,
+                                self._attr_acceptance[self.attr])
+        current_total = sum(current_ratio)
+        # We return the target when no moves are recorded since we don't want
+        # the move size to be updated. This is "hackish", but there is no right
+        # answer here since there is no concept of an acceptance_rate with no
+        # trial moves. We could error, but I think this make the class easier to
+        # use for users.
+        if current_total == 0:
+            return self._target
```
joaander, line 43: This hack is fine. We could alternately return `None` and handle that special case in the caller.

hoomd/hpmc/tune/move_size.py
```diff
@@ -0,0 +1,333 @@
...
+        self._attached = False
+
+    def act(self, timestep=None):
+        """Tune move sizes.
+
+        Args:
+            timestep (:obj:`int`, optional): Current simulation timestep. Is
+                currently ignored.
+        """
+        if not self.tuned and self.is_attached:
```
joaander, line 194: This appears to make the tuner stop adjusting move sizes after it achieves its target. What if the conditions of the simulation change (i.e. the box size, or particle parameters) that cause this tuning to be invalid? Wouldn't we want to keep tuning?

Alternately, if we want to enforce tuning to stop the run when complete we should tie this tolerance check in with the new completion framework on the `quick-compress` branch. This way, users need to opt in to re-tuning. However, it would be less convenient to use:
```
sim.operations.tuners.append(move_size_tuner)
sim.run(1e6)
# stops early
sim.operations.tuners.remove(move_size_tuner)
sim.run(1e6)
```

hoomd/hpmc/tune/move_size.py
```diff
@@ -0,0 +1,333 @@
...
+        moves (list[str]): A list of types of moves to tune. Available options
+            are 'a' and 'd'.
+        target (float): The acceptance rate for trial moves that is desired. The
+            value should be between 0 and 1.
+        solver (`hoomd.tune.SolverStep`): A solver that tunes move sizes to
+            reach the specified target.
+        types (list[str]): A list of string particle types to tune the move
+            size for, defaults to None which upon attaching will tune all types
+            in the system currently.
+        max_move_size (float): The max value of move size to attempt.
```
joaander, line 246: Is the maximum the same for both 'a' and 'd' when tuning both?

hoomd/hpmc/tune/move_size.py
```diff
@@ -0,0 +1,333 @@
...
+            max_move_size (`float`): The max value of move size to attempt.
+            gamma (float): The value of gamma to pass through to
+                `hoomd.tune.ScaleSolver`. Controls the size of corrections to
+                the move size (larger values increase stability while increasing
+                convergence time).
+            tol (float): The absolute tolerance to allow between the current
+                acceptance rate and the target before the move sizes are
+                considered tuned. The tolerance should not be too much lower
+                than the default of 0.01 as acceptance rates can vary
+                appreciable at typical tuning rates.
```
joaander, line 290:
```suggestion
                significantly at typical tuning rates.
```

hoomd/hpmc/tune/move_size.py
```diff
@@ -0,0 +1,333 @@
...
+                to attempt, default to no max size.
+            gamma (float): The value of gamma to pass through
+                to `hoomd.tune.SecantSolver`. Controls the size of corrections
+                to the move size (smaller values increase stability). Should be
+                between 0 and 1, defaults to 0.8.
+            tol (float): The absolute tolerance to allow between the current
+                acceptance rate and the target before the move sizes are
+                considered tuned. The tolerance should not be too much lower
+                than the default of 0.01 as acceptance rates can vary
+                appreciable at typical tuning rates.
```
joaander, line 324:
```suggestion
                significantly at typical tuning rates.
```

hoomd/tune/attr_tuner.py
```diff
@@ -0,0 +1,332 @@
...
+            bool: Whether the value is in the domain of x.
+        """
+        if self.domain is None:
+            return True
+        else:
+            lower_bound, upper_bound = self.domain
+            return ((lower_bound is None or lower_bound <= value)
+                    and (upper_bound is None or value <= upper_bound))
+
+    def wrap_into_domain(self, value):
```
joaander, line 33: `wrap` implies that values will wrap around periodic boundary conditions. Use `clamp` instead.

hoomd/tune/attr_tuner.py
```diff
@@ -0,0 +1,332 @@
...
+        self.__get_x = get_x
+        self.__set_x = set_x
+        self.__get_y = get_y
+        self._target = target
+        if domain is not None and not len(domain) == 2:
+            raise ValueError("domain must be a sequence of length two.")
+        self._domain = domain
+
+    def _get_x(self):
+        return self.__get_x()
```
joaander, line 171: Is it standard Python convention to have two methods with the same name and different numbers of leading underscores? To avoid confusion, consider renaming `__get_x` to `_user_get_x` and similarly for the other attributes.

hoomd/tune/attr_tuner.py
```diff
@@ -0,0 +1,332 @@
...
+            tunable (`hoomd.tune.ManualTuneDefinition`): A tunable object that
+                represents a relationship of f(x) = y.
+
+        Returns:
+            bool : Whether or not the tunable converged to the target.
+        """
+        pass
+
+    def solve(self, tunables):
+        """Iterates towards a solution for a list of tunables.
```
joaander, line 225: Document the return value of this function.

hoomd/hpmc/tune/move_size.py
b-butler: That is true. `tuned` is a settable property, though so it can be set to false to force retuning. We should mix this with the completion framework as well. I think the `tuned` property fits into this. It is completed when it is `tuned`, but that can be reset if desired; for instance, if the system changes.

hoomd/hpmc/tune/move_size.py
b-butler: Currently, yes. The `SolverStep` and `_TuneDefinition` are generic enough to allow for a difference, but I didn't how varied of maximum move sizes we wanted. Should we allow different maximums for each particle and move type?

hoomd/tune/attr_tuner.py
b-butler: I like that option better. I just didn't think of it when writing the code.

hoomd/hpmc/tune/move_size.py
```diff
@@ -0,0 +1,333 @@
...
+        # If no more trial moves have been recorded return previous
+        # acceptance_rate. In general, this conditional should not be true, if
+        # it is true different solvers may error, but this is the most natural
+        # solution, I could think of.
+        if (self.previous_total is not None
+                and self.previous_total == current_total):
+            return self.previous_ratio[0] / self.previous_total
+
+        if self.previous_ratio is None or self.previous_total > current_total:
+            acceptance_rate = current_ratio[0] / current_total
```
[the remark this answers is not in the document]
b-butler: We can tune here because the move counts are non-zero. On the first call, we can tune as long as there are moves that have been performed. Likewise, when the counters are lower than before, they have been reset, but the existing count does give us information about the acceptance rate. In both cases, we should tune I believe since we have information to do so.

The only case we don't tune is if there are no moves to get an acceptance rate for. If I am not understanding your point let me know.

hoomd/hpmc/tune/move_size.py
b-butler: I am not sure if I want to force the `SolverStep` objects to deal with a `y` of `None`. I am not sure how many cases of this problem there would be. If it is common enough, it may be worth baking into the `SolverStep` infrastructure, though. We could also just error and always use examples that use a `phase` with the trigger for attribute tuning operations.

b-butler:
My first thought regarding tuning to the maximum would be to tune the derivative to zero; however, that could lead to a minimum as well depending on the curvature of the function. This also has the problem in that we need small deltas in `x` to be able to determine the derivative (which is what the secant solver already does).

My second thought would be to create a `SolverStep` that specifically solves for these extrema problems. We could implement something that searches for a local optimum by _feeling_ the derivative and going in the correct direction. That solver would then tune towards an extrema and all objects that accept a `SolverStep` should work with it. In principal anyways, there may be incompatibilities introduced by implementation.

hoomd/hpmc/tune/move_size.py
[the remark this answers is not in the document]
joaander: Let me clarify with an example. Take this case:
```
sim.operations.tuners.append(MoveSizeTuner(..., trigger=And(After(1e8), Periodic(100)))
run(1e9)
```
Since the reading uses `current_total`, it will be the averaged value over the first 1e8 steps. The second reading will be computed from only the 100 steps after making the change. This is the inconsistency I am referring to. Depending on what happened before tuning started (i.e. compression), that first average reading may not actually reflect the current acceptance rate of the system.

hoomd/hpmc/tune/move_size.py
joaander: Yes, let's allow for this if it isn't too complex to provide the option. Depending on the diameter of the particles in the simulation users may need to provide very different maximum values. E.g. in the extreme case of a diameter 100 particle, the maximum move size for d might be 50 while (in 2D at least), the maximum move size for a is 2*pi.

hoomd/hpmc/tune/move_size.py
joaander: It's up to you. The other example case we have in mind for this (tuning r_buff for performance) is another example where the first value read will be undefined.

hoomd/hpmc/tune/move_size.py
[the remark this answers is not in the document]
b-butler: I see your point. We shouldn't tune on the first pass.  I think this makes a more compelling argument for allowing y to be `None` as well as a way of denoting to skip this for now.

hoomd/hpmc/tune/move_size.py
```diff
@@ -0,0 +1,390 @@
...
+            are 'a' and 'd'.
+        target (float): The acceptance rate for trial moves that is desired. The
+            value should be between 0 and 1.
+        solver (`hoomd.tune.SolverStep`): A solver that tunes move sizes to
+            reach the specified target.
+        types (list[str]): A list of string particle types to tune the move
+            size for, defaults to None which upon attaching will tune all types
+            in the system currently.
+        max_translation_move (float): The maximum value of a translational move
+            size to attempt.
```
joaander, line 281: For the `max_*_move` values, I assume that `None` implies no maximum. Document this.

joaander (approved):
Thanks!

PR #706, opened 2020-08-03
Brownian dynamics edited and ported based on Langevin

syjlee:
## Description
Porting Brownian integration method to the new v3 API.
Changes are only made in `methods.py`

## Motivation and Context
Brownian integrator is needed to be ported for new v3 API

## How Has This Been Tested?
Compilation on local mac machine.
A simple script was run 

## Change log
```
Moved Brownian integration method to v3 API
```

## Checklist:

- [x ] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x ] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).


joaander:
I pushed a commit that should result in the sphinx docs building without errors.

However, before merging we should update the documentation for `Langevin` and `Brownian` to document all param_dict entries and the `gamma` TypeParameter in an `Attributes:` section. See `hpmc/integrate.py` for an example of how to format this section of the documentation.

hoomd/md/methods.py
```diff
@@ -877,154 +875,58 @@ class brownian(_Method):
...
+        gamma = TypeParameter('gamma', type_kind='particle_types',
+                              param_dict=TypeParameterDict(1., len_keys=1)
+                              )
 
-        # create the compute thermo
-        hoomd.compute._get_unique_thermo(group=group)
+        gamma_r = TypeParameter('gamma_r', type_kind='particle_types',
+                                param_dict=TypeParameterDict((1., 1., 1.), len_keys=1)
+                                )
+        self._extend_typeparam([gamma,gamma_r])
```
b-butler, line 913: We should just call `super().__init__(filter, kT, seed, alpha)`, and remove the `tally_reservior_energy` from the `self._param_dict` afterwards. It will be shorter and reduce code duplication.

hoomd/md/integrate.py
```diff
@@ -102,8 +102,8 @@ class Integrator(_DynamicIntegrator):
     The following commands can be used to specify the integration methods used
     by integrate.mode_standard.
 
-    - :py:class:`hoomd.md.methods.brownian`
-    - :py:class:`hoomd.md.methods.Langevin`
+    - ``hoomd.md.methods.Brownian``
+    - ``hoomd.md.methods.Langevin``
```
joaander, line 106: @syjlee Note that single vs double back ticks mean very different things in Sphinx/restructured text. Single backticks make a reference that links to the named object. Double backticks mark a section of verbatim text that should be formatted as code.

PR #729, opened 2020-08-20
Fix bug in the state of hoomd.md._Force objects

b-butler:
## Description

Fixes a bug in the state of `hoomd.md._Force` objects. We previously stored the `_param_dict`
information in the state key `param`. We rename the new state key to `__state__`. Likewise for
consistency, we rename the `__default` key in a type parameters state `dict` to `__default__`.

This also addresses a bug where we convert any non-specified values (i.e. they are set to
`RequiredArg`) to `None` in the state. This can raise errors when using `from_state`. We now filter
out any `RequiredArg` from the type parameters.

## Motivation and Context

Needed to properly use state.

## How Has This Been Tested?

Should be covered in part through existing tests. The `from_state` methods of objects should be
tested more.

## Change log

```

```

## Checklist:

### DIFF
--- a/hoomd/operation.py
+++ b/hoomd/operation.py
@@ -249,7 +249,7 @@ class _Operation(_HOOMDGetSetAttrBase, metaclass=Loggable):
     def state(self):
         self._update_param_dict()
         state = self._typeparam_states()
-        state['params'] = dict(self._param_dict)
+        state['__params__'] = dict(self._param_dict)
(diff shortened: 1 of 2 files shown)


b-butler:
Just to note I also tested this local and found that it has the desired behavior.

mikemhenry:
This fixes both the example in issue #729 and also the bug I saw in https://github.com/mosdef-hub/mbuild/pull/698
Thanks!!!

PR #733, opened 2020-08-24
Feature/quasi2d

vyasr:
## Description

This is an initial draft of how we could enable quasi-2D simulations. It is currently implemented as a simple hack of the internals where translation moves are modified to only sample in-plane even in 3D simulations when a particular flag is set, but this implementation could be cleaned up.  I'd appreciate any thoughts at a high level on how this should work, once that's resolved I can work on cleaning up the code. I do not think that there's a need for in-depth code review yet, but enabling this should be pretty straightforward.

## Motivation and Context

@joaander and I discussed this briefly a few months ago, the goal is to enable simulations where particles can optionally rotate out of plane but are otherwise constrained to the plane. Currently this is enabled by an integer flag whose value (>0 vs >1) is used to indicate whether or not out of plane rotations are permitted.

## Change log

```
Enable simulations of 3D objects constrained to move in a plane, with out-of-plane rotations optionally enabled.
```

## Checklist:

### DIFF
--- a/.azp/test.yml
+++ b/.azp/test.yml
@@ -7,14 +7,6 @@ pr:
 # repository.
 trigger: none
 
-# Instead, trigger periodic builds on active branches.
-schedules:
-- cron: "0 2-20/6 * * *"
-  displayName: Periodic build
-  branches:
-    include:
-    - "*"
-
 variables:
   image_root: glotzerlab/ci:2021.03
 
--- a/.azp/validate.yml
+++ b/.azp/validate.yml
@@ -9,14 +9,6 @@ pr:
 # repository.
 trigger: none
 
-# Instead, trigger periodic builds on active branches.
-schedules:
-- cron: "0 2-20/6 * * *"
-  displayName: Periodic build
-  branches:
-    include:
-    - "*"
-
 variables:
   image_root: glotzerlab/ci:2021.03
 
--- a/hoomd/hpmc/IntegratorHPMCMono.h
+++ b/hoomd/hpmc/IntegratorHPMCMono.h
@@ -132,7 +132,7 @@ class IntegratorHPMCMono : public IntegratorHPMC
         typedef typename Shape::param_type param_type;
 
         //! Constructor
-        IntegratorHPMCMono(std::shared_ptr<SystemDefinition> sysdef);
+        IntegratorHPMCMono(std::shared_ptr<SystemDefinition> sysdef, unsigned int quasi2d=0);
 
         virtual ~IntegratorHPMCMono()
             {
@@ -466,6 +466,7 @@ class IntegratorHPMCMono : public IntegratorHPMC
 
         bool m_quermass;                                         //!< True if quermass integration mode is enabled
         Scalar m_sweep_radius;                                   //!< Radius of sphere to sweep shapes by
+        unsigned int m_quasi2d;                                  //!< Whether or not to run quasi2d simulations.
 
         //! Test whether to reject the current particle move based on depletants
         #ifndef ENABLE_TBB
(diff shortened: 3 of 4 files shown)


joaander:
This is fine in principle. I suggest we include `translate` in the parameter name - e.g. `translate_move_dimensions` to allow for the future possibility that rotation moves could be independently controlled.

vyasr:
@joaander I suppose that's valid from a code perspective, but I have a hard time understanding what physics would be valid in a system where 3D translations are possible but 3D rotations are not. Is there any physical system that could be reasonably studied in that manner? In any case, if we want to enable that we should do that within this PR since depending on the value of quasi2d out-of-plane rotations may or may not be allowed.

@b-butler I could use a hand determining the best way to set up the Python API. I've rebased on master, but the current API is nonfunctional because the parameter will never be passed to C++. In theory, this feature makes sense for any 3D HPMC integrator; should we introduce an additional class to the hierarchy (`HPMCIntegrator3D` or similar) and add `quasi2d` (or whatever we name the parameter) to its `param_dict`?

joaander:
> @joaander I suppose that's valid from a code perspective, but I have a hard time understanding what physics would be valid in a system where 3D translations are possible but 3D rotations are not. Is there any physical system that could be reasonably studied in that manner? In any case, if we want to enable that we should do that within this PR since depending on the value of quasi2d out-of-plane rotations may or may not be allowed.

I suppose not, but we should still include `translate_` in the name so that it is clear what the parameter is from its name alone.

vyasr:
Agreed, the renaming is a good idea regardless. How do you want to handle in/out-of-plane rotations, though? Whether or not out-of-plane rotations are allowed is determined by the value of the `quasi2d` flag, it's not just a boolean.

joaander:
> Agreed, the renaming is a good idea regardless. How do you want to handle in/out-of-plane rotations, though? Whether or not out-of-plane rotations are allowed is determined by the value of the `quasi2d` flag, it's not just a boolean.

I thought your previous suggestion was that the dimensionality of the rotation moves was not a user-settable parameter and would instead be set by the dimensionality of the system.

This is what I thought you were proposing:
System dimensions=2, translate_move_dimensions=2 - both translation and rotation are 2D
System dimensions=2, translate_move_dimensions=3 - error
System dimensions=3, translate_move_dimensions=2 - rotation in 3D, translation in 2D
System dimensions=3, translate_move_dimensions=3, both translation and rotation in 3D

vyasr:
Yes, I think I misunderstood. That set of options is what I'm proposing, I think that covers all the relevant use cases. I can make that change when I update the Python API as well.

joaander:
> Yes, I think I misunderstood. That set of options is what I'm proposing, I think that covers all the relevant use cases. I can make that change when I update the Python API as well.

Should we have allow a default `translate_move_dimensions=None` automatically set to the system dimensions on attaching to support the current behavior with no API changes? @b-butler

vyasr:
@joaander @bdice I think that would make sense. The ability to have that kind of automated setting was also IIRC the main new feature that needs to be added to support migrating the ALJ potential to the 3.0 API as well (one of the parameters must default to a value computed from one of the other parameters if it is not provided), so we can engineer that here.

vyasr:
@joaander that issue is a newer one that I assumed would be resolved independently of the ALJ potential since it's necessary for all `aniso` potentials. The specific issue I'm referring to is that in the ALJ potential there are both `sigma` and `contact_sigma` parameters, and if the user does not specify `contact_sigma` it defaults to some fraction of `sigma`. Porting that behavior to v3 requires some attach-time defaulting semantics, rather than a default encoded in the TypeParameter object at class definition.

vyasr:
@b-butler I accidentally tagged Bradley instead of you in my last post, but I meant this conversation for you.

joaander:
I cited the `aniso` issue merely as an example of the difficulties posed by attach-time resolved defaults. There was no implicit or explicit reference to ALJ intended in that reference.

Issue #697, opened 2020-07-28
Remove context.initialize


joaander:
## Purpose

`context.initialize`, `SimulationContext` and related methods have been replaced by the `Device` class. Remove them as they are now unused.

## Steps

- Remove `initialize`, `SimulationContext`, and related variables and attributes from `context.py`
- Remove `context` from the Sphinx documentation.

## Developer

When you start work on this task, please assign yourself and move it
to the *In progress* column in the associated project.

PR #735, opened 2020-08-24
Hackathon/remove context initialize

j-proc:
## Description

Removed context.py as the majority was used. The bib and options objects have been moved to the hoomd namespace and are hidden variables until their usage is finalized. The references to them have been changed as well.

## Motivation and Context

A large portion of the code is no longer used in v3. 
Resolves: #697 
Resolves: #744


hoomd/__init__.py
```diff
@@ -90,3 +89,53 @@ def _hoomd_sys_excepthook(type, value, traceback):
 sys.excepthook = _hoomd_sys_excepthook
 
 __version__ = "2.9.0"
+
+# TODO: Decide how these properties should be exposed.
+## \internal
+# \brief Gather context about HOOMD
```
b-butler, line 95: No need to use the old doxygen style comments for Python

hoomd/__init__.py
```diff
@@ -90,3 +89,53 @@ def _hoomd_sys_excepthook(type, value, traceback):
...
+class build_info():
+    ## \internal
+    # \brief Constructs the context object
```
b-butler, line 98: Same here regarding comments. Just use a standard docstring.

hoomd/__init__.py
```diff
@@ -90,3 +89,53 @@ def _hoomd_sys_excepthook(type, value, traceback):
 sys.excepthook = _hoomd_sys_excepthook
 
 __version__ = "2.9.0"
+
+# TODO: Decide how these properties should be exposed.
```
b-butler, line 93: Could you make an issue regarding this?

cbkerr: Done at #744

mikemhenry:
Just as an FYI, this makes hoomd barf on import:

```
2020-08-24T19:32:08.0355656Z hoomd/hpmc/__init__.py:143: in <module>
2020-08-24T19:32:08.0355963Z     if hoomd._bib is None:
2020-08-24T19:32:08.0356553Z E   AttributeError: module 'hoomd' has no attribute '_bib'
```

mikemhenry:
Also `if hoomd.context.current is None:` is a really common pattern in the code base (See line 83 for `hoomd/comm.py`). Do we just need to remove this check or re-implement it with the new hoomd v3 API?

joaander:
> Also `if hoomd.context.current is None:` is a really common pattern in the code base (See line 83 for `hoomd/comm.py`). Do we just need to remove this check or re-implement it with the new hoomd v3 API?

Code updated to v3 references the device through the `Simulation` object `simulation.device` (see sorter.py line 36 for an example). Code referencing `hoomd.context.current` should only exist in classes and methods that have not yet been updated to the v3  API. These will be updated one class/module at a time as that class/module is updated to v3.

hoomd/meta.py
```diff
@@ -93,7 +93,7 @@ def dump_metadata(filename=None,user=None,indent=4):
     st = datetime.datetime.fromtimestamp(ts).strftime('%Y-%m-%d %H:%M:%S')
     metadata['timestamp'] = st
     metadata['device'] = hoomd.context.current.device
-    metadata['hoomd'] = hoomd.context.HOOMDContext()
+    metadata['hoomd'] = hoomd.build_info()
```
joaander, line 96: You can remove this. `meta.py` is no longer used.

hoomd/__init__.py
```diff
@@ -90,3 +89,50 @@ def _hoomd_sys_excepthook(type, value, traceback):
...
+    @property
+    def compiler_version(self):
+        return _hoomd.__compiler_version__
+
+## Global bibliography
+_bib = None
+
+# TODO: orphaned from context.py. Only used in option.py
+## Global options
+_options = None
```
joaander, line 138: This is not needed. It is no longer used.

hoomd/__init__.py
```diff
@@ -90,3 +89,50 @@ def _hoomd_sys_excepthook(type, value, traceback):
...
+    def cuda_version(self):
+        return _hoomd.__cuda_version__
+
+    # \brief Return the compiler version
+    @property
+    def compiler_version(self):
+        return _hoomd.__compiler_version__
+
+## Global bibliography
+_bib = None
```
joaander, line 134: `hpmc` is imported above and it's reference to `_bib` fails. Perhaps `_bib` needs to be defined before importing modules?

hoomd/comm.py
```diff
@@ -113,18 +113,18 @@ def __init__(self, x=None, y=None, z=None, nx=None, ny=None, nz=None, linear=Fal
 
             """
             # do a one time update of the cuts to the global values if a global is set
-            if not self.x and self.nx == 0 and hoomd.context.options.nx is not None:
```
joaander, line 116: You don't need to undo these changes. However, no changes are necessary to code that has not been updated to the v3 API.

hoomd/__init__.py
j-proc: Moved to top before all imports. It's a simple None labeled as a global objects so hopefully won't cause confusion.

j-proc:
> Thanks! I have a few small change requests and comments.

I've reverted all the changes we made that you specified are not used. Hopefully won't have any merge conflicts with @mikemhenry this way. Bib has been moved as well.

hoomd/__init__.py
```diff
@@ -90,3 +92,14 @@ def _hoomd_sys_excepthook(type, value, traceback):
...
 
 __version__ = "2.9.0"
+
+__git_sha1__ = _hoomd.__git_sha1__
+
+__git_refspec__ = _hoomd.__git_refspec__
+
+__compile_flags__ = _hoomd.hoomd_compile_flags()
+
+__cuda_version__ = _hoomd.__cuda_version__
```
joaander, line 102: `get_cuda_version_tuple()` feeds into `__cuda__version__`. Should we change it to return a string like other `__version__` attributes?

Issue #700, opened 2020-07-28
Test BoxResize


joaander:
## Purpose

Unit test the `BoxResize` class.

## Steps

- Add tests to `hoomd/pytest/test_box_resize.py`
- Add `test_box_resize` to `hoomd/pytest/CMakeLists.txt`

## Developer

When you start work on this task, please assign yourself and move it
to the *In progress* column in the associated project.

PR #745, opened 2020-08-24
Added BoxResize tests

pepak13:
## Description

Added a unit test to test whether the BoxResize class is working properly. 

Current issues still in of work: 
1) Cannot properly rescale the particle positions. 
2) Using `hoomd.update.BoxResize.linear_volume` for one of the tests is starting with the end box. 

## Motivation and Context

Resolves: 

#700 

## How Has This Been Tested?

Tested changes to single box dimensions, multiple box dimenions, and all box dimensions. 

## Change log

```
Added BoxResize tests. 
```

## Checklist:


hoomd/pytest/test_box_resize.py
```diff
@@ -0,0 +1,126 @@
+from math import isclose
+import numpy as np
+from pytest import fixture
+
+import hoomd
+
+
+class SimulationSetup:
```
joaander, line 8: This would be easier to read if it was written as one test_ method using `pytest.mark.parameterize` to give it different box1 and box2 inputs. You could also parameterize with a flag to test either `BoxResize` or `BoxResize.linear_variant`.

hoomd/pytest/test_box_resize.py
```diff
@@ -0,0 +1,126 @@
...
+        trigger = hoomd.trigger.On(self.t_start)
+        box_resize = hoomd.update.BoxResize(
+            box1=self.box1, box2=self.box2,
+            variant=variant, trigger=trigger, scale_particles=scale_particles)
+        sim.operations.updaters.append(box_resize)
+        sim.run(self.t_start + self.t_ramp + 1)
+        return box_resize, sim.state.snapshot
+
+    def variant_linear(self, device, scale_particles=True):
+        sim = hoomd.Simulation(device)
```
joaander, line 45: Please use the `simulation_factory` fixture to create `Simulation` objects from snapshots.

hoomd/pytest/test_box_resize.py
```diff
@@ -0,0 +1,126 @@
...
+        return box_resize, sim.state.snapshot
+
+
+def test_resize_Lx(device):
+    box1_dim = np.array([1., 2., 3., 1., 2., 3.])
+    box2_dim = np.array([0.5, 2., 3., 1., 2., 3.])
+    ss = SimulationSetup(box1_dim=box1_dim, box2_dim=box2_dim)
+    box_resize, snap = ss.variant_ramp(device)
+
+    assert np.all(box_resize.get_box(0).matrix == ss.box1.matrix)
```
joaander, line 64: You should be able to check for equality between two box objects directly.

hoomd/pytest/test_box_resize.py
```diff
@@ -0,0 +1,126 @@
...
+    assert np.all(box_resize.get_box(ss.t_start + ss.t_ramp).matrix == ss.box2.matrix)
+    # assert np.all(np.isclose(snap.particles.position[:], ss.points2))
+
+    box_resize, snap = ss.variant_linear(device)
+
+    assert np.all(box_resize.get_box(0).matrix == ss.box1.matrix)
+    assert np.all(box_resize.get_box(ss.t_start + ss.t_ramp).matrix == ss.box2.matrix)
+    # assert np.all(np.isclose(snap.particles.position[:], ss.points2))
```
joaander, line 72: Use `numpy.testing.assert_allclose`

hoomd/pytest/test_box_resize.py
```diff
@@ -0,0 +1,128 @@
...
+import numpy as np
+import numpy.testing as npt
+import pytest
+import hoomd
+
+
+@pytest.fixture(scope="module")
+def fractional_coordinates(n=10):
+    """
+    TODO: Does `numpy_random_seed()` in conftest.py run for this function?
```
joaander, line 11: I don't know. That fixture is set with `@pytest.fixture(scope='function', autouse=True)` so it will be called before every function, but I don't know how that works with a module scope fixture.

You could make sure that it works in the correct order by making this a function scope fixture and include numpy_random_seed as an argument to ensure that that fixture is called first.

joaander (commented):
Looking good! @b-butler do you have any comments?

Do you have any additional tests or modifications you plan to make?

hoomd/pytest/test_box_resize.py
```diff
@@ -0,0 +1,128 @@
...
+
+    assert box_resize.get_box(0) == sys1[0]
+    assert box_resize.get_box(variant.t_start*3) == sys2[0]
+    assert sim.state.box == sys2[0]
+    if scale_particles:
+        npt.assert_allclose(sys2[1], sim.state.snapshot.particles.position)
+    else:
+        npt.assert_allclose(sys1[1], sim.state.snapshot.particles.position)
+    if variant == _variants[2]:
+       assert box_resize.get_box(variant.t_start*3+1) == sys1[0]
```
b-butler, line 108: I think a comment explicitly mentioning this is checking that we are correctly handling the `Cycle` variant is needed. Also, this should be in the `test_get_box` test I mention above.

hoomd/pytest/test_box_resize.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+_scale_particles = [True,False]
+
+
+@pytest.fixture(scope='function', params=_scale_particles)
+def scale_particles(request):
+    return request.param
+
+
```
b-butler, line 86: For this fixture we could just use `pytest.mark.parameterize` since it is only used for this function and only for the simple values `True` and `False`.

b-butler (changes requested):
It would be helpful to check intermediate box and particle positions as @joaander mentions. It may be worth creating a function given an initial box and particles that when passed a variant value can determine the new box and particles that are expected. An easier if less accurate approach would be to check that the box sizes are exactly the initial or final conditions at the correct time steps or strictly in between for other time steps in this approach positions would just use an equality or non equal check.

Also some more modularity in tests would be nice. Ideally a test should test one thing.

That being said great work and thanks for the effort thus far.

pepak13:
> Looking good! @b-butler do you have any comments?
> 
> Do you have any additional tests or modifications you plan to make?

hoomd/pytest/test_box_resize.py
pepak13: I want to note that the linear_volume box resize test is still failing for one of the systems. In addition, the other variant tests pass when the xy for sys21 is set to 0.9 (It fails when it is 2). I still do not know why this is happening.

vyasr: @joaander for any test requiring randomness, it might make sense to instead define a fixture
```
def rng():
    return numpy.random.default_rng(42)
```

This has the downside of adding to the signature of each test, but it has the upside of being self-contained and not having to worry about global numpy random state.

vyasr: There is `pytest-rng`, but that's still using the legacy numpy random number API (and this is a trivial fixture to add anyway).

joaander: > but it has the upside of being self-contained and not having to worry about global numpy random state.
Yes, this was my intention with the autouse fixture. We could use an explicit fixture instead.

klywang:
Some updates:
* Separated tests for `get_box` and `BoxResize` on the simulation state.
* Added assertions for intermediate particle positions and box size.
* Test 2D and 3D box resizing.
* Added tests to ensure `variant` and  `trigger`.
* Restructured tests with classes.
    - I found it easier to determine which test failed with the classes. Please let me know if I should change the structure back to functions only.

sphinx-doc/credits.rst
```diff
@@ -95,3 +95,868 @@ The following people have contributed to the to HOOMD-blue:
 * Wouter Ellenbroek, Eindhoven University of Technology
 * Yuan Zhou, University of Michigan
 * Åsmund Ervik, SINTEF
+
+=======
```
b-butler, line 99: Why are these changes here? Are they just hold outs from a removal in master?

hoomd/pytest/test_box_resize.py
```diff
@@ -0,0 +1,205 @@
...
+    def test_get_box(self, device, simulation_factory,
+                     get_snapshot, sys, box_resize):
+        sys1, make_sys_halfway, sys2 = sys
+        sys_halfway = make_sys_halfway(self._power)
+
+        sim = hoomd.Simulation(device)
+        sim.create_state_from_snapshot(get_snapshot())
+
+        sim.operations.updaters.append(box_resize)
+        sim.run(_t_start + _t_ramp)
```
b-butler, line 125: Is there any need to run? Could we just use `sim.run(0)`?

hoomd/pytest/test_box_resize.py
```diff
@@ -0,0 +1,205 @@
...
+        assert sim.state.box == sys_halfway[0]
+        npt.assert_allclose(sim.state.snapshot.particles.position, sys1[1])
+
+        # Finish run
+        sim.run(_t_mid)
+        assert sim.state.box == sys2[0]
+        npt.assert_allclose(sim.state.snapshot.particles.position, sys1[1])
+
+
+class TestLinearVolume(TestBoxResize):
```
b-butler, line 193: I think the idea here would be to test the factory method `hoomd.update.BoxResize.linear_volume()`. Basically we just want to test that the volume changes linearly over the box change.

b-butler (changes requested):
I left a few comments to address.

sphinx-doc/credits.rst
joaander: This looks like a merge conflict that was not resolved: it starts with `=======`

klywang: It is a merge conflict that I failed to resolve correctly. Do I remove everything from line 3 to "====="  on line 99?

hoomd/pytest/test_box_resize.py
klywang: I tried to test if box volume changed linearly with code like this:
```python
    def test_linear_volume(self, device, simulation_factory
        sim = hoomd.Simulation(device)
        sim.create_state_from_snapshot(get_snapshot())
        sim.operations.updaters.append(box_resize)
        sim.run(0)

        slope = (sys2[0].volume - sys1[0].volume)/_t_ramp
        for ts in range(_t_ramp + 1):
            estimated_volume = sys1[0].volume + ts*slope
            assert box_resize.get_box(ts + _t_start).volume == estimated_volume
```
and the test failed.

When I plot box volume for a 3D box, I get this:
linear_volume_3d



and when I plot box area for a 2D, I get this:
linear_volume2d



Am I testing/interpreting this correctly?

klywang: Does that mean we don't need to test things like `get_box` and `trigger` for the factory method? And what does the `variant` control? If I understand correctly, in order for `hoomd.update.BoxResize.linear_volume` is supposed to scale volume linearly, it uses it's own calculations to determine the appropriate coefficients for the `hoomd.variant.Power`.

sphinx-doc/credits.rst
joaander: I would recommend just replacing the credits.rst file with the one from master. Then add your names to it.

hoomd/pytest/test_box_resize.py
joaander: @klywang Are you test boxes cubic?

klywang: @joaander These are not.

joaander: Does the plot look right for cubic boxes? I'm thinking that the the power variant math @b-butler wrote may have am implicit assumption that the box is cubic.

b-butler: I think it does. I didn't realize it when writing it though.

b-butler: To get linear volume change we need to either change the way `BoxResize` works or have an internal flag that changes the behavior to allow for linear volume resizing.

klywang: Here are the results for `BoxResize.linear_volume` for 3D cubes:
box_resize3D

And 2D squares:
box_resize2D

joaander:
Discussed offline: boxes formed from `s*box1 + (1-s)*box2` don't follow a linear volume scaling when s follows a 1/3 power. Please remove the linear_volume factory method and any tests for it.

b-butler (approved):
LGTM

joaander (approved):
Thanks!

PR #746, opened 2020-08-24
Updated Documentation of pair potentials for hoomd 3.0

alacour:
## Description

We modified the documentation in ```hoomd/md/pair.py``` to bring it inline with the API of hoomd 3.0. We believe we did it correctly for simple two-body pair potentials. The API has not been updated for 3-body and anisotropic potentials, so our modifications for those potentials may need to be modified when their API has been finalized. The 3-body and anisotropic potentials are: revcross, Gay-Berne, dipole, and square_density. 

We did not change the Tersoff, Table, and ai_potential. We also did not change DLVO due to confusion about its form.

## Motivation and Context

The API is changing Hoomd 3.0, so we need to update the documentation of it.

Resolves: #709 

## How Has This Been Tested?

We have successfully built the documentation files with sphinx, but we have not examined the style.

## Change log
```
First draft of changes to documentation of pair potentials for Hoomd 3.0.
```

## Checklist:


hoomd/md/pair.py
```diff
@@ -764,6 +764,12 @@ def set_from_file(self, a, b, filename):
 class Morse(_Pair):
     R""" Morse pair potential.
 
+    Args:
+        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
```
joaander, line 768:
```suggestion
        nlist (hoomd.md.nlist._Nlist): Neighbor list
```
And similarly for all other potential classes.

hoomd/md/pair.py
```diff
@@ -776,7 +782,7 @@ class Morse(_Pair):
         \end{eqnarray*}
 
     See :py:class:`_Pair` for details on how forces are calculated and the available energy shifting and smoothing modes.
-    Use ``coeff.set`` to set potential coefficients.
+    Use ``params`` dictionary to set potential coefficients.
```
joaander, line 785: Should be able to reference this as
```
`params`
```
And `params`  should be documented in an `Attributes:` section, listing the keys to the dictionary and what they are. See the `shape` parameter in `hpmc/integrate.py` for examples of how to document `params`.

`r_cut` and `r_on` in a similar way, but they can be documented in the parent class `_Pair`.

hoomd/md/pair.py
```diff
@@ -877,15 +884,14 @@ class DPD(_Pair):
...
-        dpd.set_params(kT = 1.0)
-        integrate.mode_standard(dt=0.02)
-        integrate.nve(group=group.all())
+        nl = nlist.Cell()
+        dpd = pair.DPD(nlist=nl, kT=1.0, seed=0, r_cut=1.0)
+        dpd.params[('A', 'A')] = dict(A=25.0, gamma=4.5)
+        dpd.params[('A', 'B')] = dict(A=40.0, gamma=4.5)
+        dpd.params[('B', 'B')] = dict(A=25.0, gamma=4.5)
+        dpd.params[(['A', 'B'], ['C', 'D'])] = dict(A=40.0, gamma=4.5)
+        nve = methods.NVE(filter=all)
```
joaander, line 893: Remove NVE and Integrator examples from an example on pair forces.

b-butler:
@joaander, we could rename the arguments to `default_r_cut` and `default_r_on`. I think it makes sense to provide this as I think most users will just use one `r_cut` for a given potential.

tommy-waltmann:
> 
> 
> @joaander, we could rename the arguments to `default_r_cut` and `default_r_on`. I think it makes sense to provide this as I think most users will just use one `r_cut` for a given potential.

Another name option would be `global_r_cut` and `global_r_on`

mphoward:
> > @joaander, we could rename the arguments to `default_r_cut` and `default_r_on`. I think it makes sense to provide this as I think most users will just use one `r_cut` for a given potential.
> 
> Another name option would be `global_r_cut` and `global_r_on`

`global` is a little confusing if there are multiple pair potentials, as these are not shared between potentials, and also global is a python keyword.

Do these shared `r_cut` and `r_on` need to be separate attributes, or could their functionality be replaced by a convenience method that loops over all the pairs and sets them to the same value? (AFAIK, the reason for having these in the constructor is a legacy design from lammps.) This removes the ambiguity about where the parameter is defined: it is still per-pair, and there is no question about which value takes precedence.

joaander:
> Do these shared `r_cut` and `r_on` need to be separate attributes, or could their functionality be replaced by a convenience method that loops over all the pairs and sets them to the same value? (AFAIK, the reason for having these in the constructor is a legacy design from lammps.) This removes the ambiguity about where the parameter is defined: it is still per-pair, and there is no question about which value takes precedence.

`r_cut` and `r_on` exist as constructor arguments as a legacy from when HOOMD only supported one `r_cut` for all type pairs. Their functionality is available to the user. The following two code blocks produce identical results:
```
lj = pair.LJ(nlist=nl, r_cut=2.5)
```
```
lj = pairlLJ(nlist=nl)
lj.r_cut.default = 2.5
```
I am in favor of removing these arguments, there should be one place for the user to explicitly set these (explicit is better than implicit). Everywhere else in HOOMD that we accept parameters in the constructor*, we set the value of that parameter directly as a convenience. If we were to accept a value here, it would need to be a dictionary that expressed the full r_cut type parameter matrix - not just a default value.

* I realize that HPMC integrators accept d and a values. We should probably remove those as well.

b-butler:
@joaander that would be consistent in how we treat other type parameters. In that case, I am fine with removing it. The other option would be to allow users to provide a dictionary for the `r_cut` values, but I don't think this is really necessary.

b-butler:
However, we need to document well that users have to manually set `r_cut` after construction.

tommy-waltmann:
Don't worry about updating the docs for the triplet potentials here, that will be on #752

joaander:
OK, we got sidetracked on the discussion of default/global parameters. I opened #789 where we can deal with this at a later date. Let's review and complete this PR with those parameters as they are named now.

joaander:
My other review comments still apply. I'd like to get this wrapped up and merged next week (Oct 2). Let me know if you don't think you can get it done by then.

gaofy95:
We updated the documents based on the review comments. Basically, we updated the `nlist` reference, removed integrator from the examples, listed the coefficients under `Attributes` as keys. The documents build on the local machines. But we are not sure why some checks were not successful.

Please let us know if there is anything to be updated. Thanks!

hoomd/md/pair.py
```diff
@@ -2081,32 +2226,41 @@ class Buckingham(_Pair):
...
+            * ``C`` (``float``, **required**) - :math:`C` (in energy units)
+            
+            * ``r_cut`` (``float``, **optional**) - *r_cut* (in distance units) 
+              - *optional*: defaults to the global r_cut specified in the pair command
+              
+            * ``r_on`` (``float``, **optional**) - *r_on* (in distance units) 
+              - *optional*: defaults to the global r_on specified in the pair command
 
     .. versionadded:: 2.2
     .. versionchanged:: 2.2
```
tommy-waltmann, line 2252:
```suggestion

```

Remove these two lines

hoomd/md/pair.py
```diff
@@ -1864,38 +1989,42 @@ class ReactionField(_Pair):
     If *use_charge* is set to True, the following formula is evaluated instead:
     .. math::
```
tommy-waltmann, line 1990:
```suggestion

    .. math::
```
Add an extra empty line here so the documentation renders correctly.

hoomd/md/pair.py
```diff
@@ -1864,38 +1989,42 @@ class ReactionField(_Pair):
...
 
-    .. versionadded:: 2.1
+            * ``r_cut`` (``float``, **optional**) - *r_cut* (in distance units) 
+              - *optional*: defaults to the global r_cut specified in the pair command
+              
+            * ``r_on`` (``float``, **optional**) - *r_on* (in distance units) 
+              - *optional*: defaults to the global r_on specified in the pair command
 
+    .. versionadded:: 2.2
+    .. versionchanged:: 2.2
```
tommy-waltmann, line 2018:
```suggestion

```
Remove these two lines.

hoomd/md/pair.py
```diff
@@ -227,24 +227,31 @@ class LJ(_Pair):
...
-      - *optional*: defaults to the global r_on specified in the pair command
+    available energy shifting and smoothing modes.  Use `params` dictionary
+    to set potential coefficients. The coefficients must be set per 
+    unique pair of particle types.
+    
+    Attributes:
+        params (TypeParameter[``particle_types``, dict]):
+            The LJ potential parameters. The dictionary has the following keys:
+
+            * ``epsilon`` (``float``, **required**) - :math:`\varepsilon` - energy parameter (in energy units)
```
b-butler, line 238: Use single backticks for types like `float`

hoomd/md/pair.py
```diff
@@ -227,24 +227,31 @@ class LJ(_Pair):
...
+            * ``r_on`` (``float``, **optional**) - *r_on* (in distance units) 
+              - *optional*: defaults to the global r_on specified in the pair command
 
     Example::
 
         nl = nlist.Cell()
         lj = pair.LJ(nl, r_cut=3.0)
         lj.params[('A', 'A')] = {'sigma': 1.0, 'epsilon': 1.0}
         lj.params[('A', 'B')] = dict(epsilon=2.0, sigma=1.0, r_cut=3.0, r_on=2.0)
+
```
b-butler, line 254: Remove extra line.

hoomd/md/pair.py
```diff
@@ -227,24 +227,31 @@ class LJ(_Pair):
...
+            * ``epsilon`` (``float``, **required**) - :math:`\varepsilon` - energy parameter (in energy units)
+            
+            * ``sigma`` (``float``, **required**) - :math:`\sigma` - particle size (in distance units)
+            
+            * ``r_cut`` (``float``, **optional**) - *r_cut* (in distance units) 
+              - *optional*: defaults to the global r_cut specified in the pair command
+              
+            * ``r_on`` (``float``, **optional**) - *r_on* (in distance units) 
+              - *optional*: defaults to the global r_on specified in the pair command
```
b-butler, line 246: `r_cut` and `r_on` are both separate attributes from the `params` attribute. They should be listed as individual attributes. This applies to other classes as well.

hoomd/md/pair.py
```diff
@@ -227,24 +227,31 @@ class LJ(_Pair):
...
-    - :math:`r_{\mathrm{cut}}` - *r_cut* (in distance units)
-      - *optional*: defaults to the global r_cut specified in the pair command
-    - :math:`r_{\mathrm{on}}`- *r_on* (in distance units)
-      - *optional*: defaults to the global r_on specified in the pair command
+    available energy shifting and smoothing modes.  Use `params` dictionary
+    to set potential coefficients. The coefficients must be set per 
+    unique pair of particle types.
+    
+    Attributes:
+        params (TypeParameter[``particle_types``, dict]):
```
b-butler, line 235:
```suggestion
        params (TypeParameter[tuple[``particle_types``, ``particle_types``], dict]):
```
This needs to be changed for the other potentials as well.

hoomd/md/pair.py
```diff
@@ -209,7 +209,7 @@ class LJ(_Pair):
     R""" Lennard-Jones pair potential.
 
     Args:
-        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
+        nlist (:py:mod:`hoomd.md.nlist._Nlist`): Neighbor list
```
tommy-waltmann, line 212:
```suggestion
        nlist (:py:mod:`hoomd.md.nlist._NList`): Neighbor list
```

hoomd/md/pair.py
```diff
@@ -259,7 +266,7 @@ class Gauss(_Pair):
     R""" Gaussian pair potential.
 
     Args:
-        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
+        nlist (:py:mod:`hoomd.md.nlist._Nlist`): Neighbor list
```
tommy-waltmann, line 269:
```suggestion
        nlist (:py:mod:`hoomd.md.nlist._NList`): Neighbor list
```

hoomd/md/pair.py
```diff
@@ -308,7 +320,7 @@ class SLJ(_Pair):
     R""" Shifted Lennard-Jones pair potential.
 
     Args:
-        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
+        nlist (:py:mod:`hoomd.md.nlist._Nlist`): Neighbor list
```
tommy-waltmann, line 323:
```suggestion
        nlist (:py:mod:`hoomd.md.nlist._NList`): Neighbor list
```
This specific error is why the documentation test is not passing. It is in many places in this file, please find and fix them all.

hoomd/md/pair.py
```diff
@@ -2032,6 +2175,7 @@ class square_density(pair):
 
     [1] P. B. Warren, "Vapor-liquid coexistence in many-body dissipative particle dynamics"
     Phys. Rev. E. Stat. Nonlin. Soft Matter Phys., vol. 68, no. 6 Pt 2, p. 066702, 2003.
+
```
b-butler, line 2176: Please remove this line.

tommy-waltmann (changes requested):
Good work with this, we are almost there. The documentation test that is failing is because in many places there is an `_Nlist` where there needs to be an `_NList`. I also have a couple more suggestions that will make the docstrings render correctly.

hoomd/md/pair.py
```diff
@@ -2201,22 +2360,26 @@ class Fourier(_Pair):
...
 
     Example::
 
-        nl = nlist.cell()
-        fourier = pair.fourier(r_cut=3.0, nlist=nl)
-        fourier.pair_coeff.set('A', 'A', a=[a2,a3,a4], b=[b2,b3,b4])
+        nl = nlist.Cell()
+        fourier = pair.Fourier(r_cut=3.0, nlist=nl)
+        fourier.params[('A', 'A')] = dict(a=[a2,a3,a4], b=[b2,b3,b4])
+
```
b-butler, line 2377: Please remove this line.

hoomd/md/pair.py
```diff
@@ -1245,19 +1301,28 @@ class ZBL(_Pair):
...
+            * ``a_F`` (``float``, **required**) - :math:`a_F = \frac{0.8853 a_0}{ Z_i^{0.23} + Z_j^{0.23} }`
+
+            * ``r_cut`` (``float``, **optional**) - *r_cut* (in distance units) 
+              - *optional*: defaults to the global r_cut specified in the pair command
+              
+            * ``r_on`` (``float``, **optional**) - *r_on* (in distance units) 
+              - *optional*: defaults to the global r_on specified in the pair command
+
+        Example::
```
b-butler, line 1325: The example doesn't render properly because of the extra indent.

hoomd/md/pair.py
joaander: Perhaps:
```suggestion
        params (TypeParameter[tuple[``particle_type``, ``particle_type``], dict]):
```

hoomd/md/pair.py
joaander: All pair potentials have r_cut and r_on, you only need to document them once in `_Pair` under `Attributes`

These have a type of `(TypeParameter[tuple[``particle_types``, ``particle_types``], float]):`

joaander (changes requested):

hoomd/md/pair.py
gaofy95: If the r_cut and r_on are attributes, should the example be something like `lj.r_cut[('A','A')] = 1.0` instead of `lj.params[('A', 'B')] = dict(epsilon=2.0, sigma=1.0, r_cut=3.0, r_on=2.0)`? I am a little confused with the sample example.

joaander: Yes, `lj.r_cut[('A','A')] = 1.0` is correct. If there are examples that show `r_cut` and `r_on` as elements of the `param` dictionary, then they are incorrect. Likely they were not completely updated from the HOOMD v2 documentation.

joaander:
I'm making some final revisions to this branch. Unfortunately, line continuations do not work in raw strings so I have to convert all of the raw strings to strings and replace the `\` with `\\` in the latex equations. Line continuations are necessary to create google format docstrings within the line character limit.

joaander (approved):
Thanks!

I made some revisions to the formatting for consistency. There may be additional improvements and new patterns we could employ after a future documentation revision. For now, I think this is fine for the first beta release.

tommy-waltmann (approved):

gaofy95 (approved):

b-butler (approved):
LGTM. There are pydocstyle issues we will need to fix later. The warnings it is giving about using `r'''` is annoying since we need line continuations. We can choose to global ignore those at least later.

Issue #642, opened 2020-05-04
Error in Attaching with Multiple Simulations


b-butler:
## Description

Currently when attaching operations, we only check if the operation is attached, not what it is attached to. This could cause bugs with multiple `Simulations`. We need to enforce that an object can only be attached to a single simulation. The solution is to store a link to the `Simulation` object when attaching and check that the `id`s match. This does mean that everything has a reference to the simulation, so we could just store the id.

## Developer

I will deal with this, I just wanted to document the bug.

PR #751, opened 2020-08-25
Create adding as distinct from attaching

b-butler:
## Description

This PR sets up a distinction between adding operations and HOOMD-blue objects to lists in such as
`hoomd.Operations` and `hoomd.md.Integrator`. Adding ensures that all objects have the same
`simulation` reference and that objects are never in two different lists (or the same list twice).

Attach now simply creates the C++ object in the case of standard actions, or attaches the
`hoomd.custom.Action` class (i.e. the class set up that requires the simulation). Currently the
`attach` method of `hoomd.custom.Action` still takes in a simulation object. We could force the same
behavior here as in the rest of the classes, but I am not sure that that is the best course of
action for users. We could document that the `_simulation` attribute in `hoomd.custom.Action`
subclasses represents the containing simulation, and is available by the time `attach` is called. On
the other hand, we could name it something else like `set_up` and `break_down` to prevent this
confusion.

## Motivation and Context

This is necessary to prevent bugs in multiple simulation scripts and simplify code logic.

Resolves: #642

## How Has This Been Tested?

Existing tests pass a new test looking at the base `_add`, `_remove`, and `_added` behavior is
added.

## Checklist:


b-butler:
Fix some remaining errors in calling `_attach` and I think I will go ahead and remove `_BaseHOOMDObject.__del__`.

Issue #730, opened 2020-08-21
Port Triplet Potentials to v3 API


tommy-waltmann:
Tersoff, SquareDensity, and RevCross need to be ported to v3 API. The standard changes for porting potentials will need to be made in addition to creating a `md.many_body` namespace, changing the base class name from `_Pair` to `_NBody` in python, as well as changing the name of `PotentialTersoff` to `PotentialTriplet` in C++. 

Unit testing for this will just reuse the tests written for the standard pairs.

This issue should wait to be addressed until the standard pair potentials #655 are merged.

Issue #739, opened 2020-08-24
Rename PotentialTersoff to PotentialTriplet


tcmoore3:
All triplet potentials live in the PotentialTersoff namespace. It would be better to let these all live in a PotentialTriplet namespace instead.

PR #752, opened 2020-08-26
Feature/new triplet potentials

tommy-waltmann:
Porting the Triplet Potentials to v3 API

The potentials included in this pull request are Tersoff, RevCross, and SquareDensity. Each have had the standard changes made to their python layers and `Evaluator::param_type` structs. They have been moved to a new namespace, `hoomd.md.many_body` and each are derived from a base class `_NBody`, which was previously called `_Pair`. They reuse the tests for the pair potentials, since all the triplet potentials so far set pair coefficients.

Other changes I am considering (depending on whether we think we really need these for the beta release or not):
- Refactoring the `PotentialTersoff` class in C++ so that it is two classes, as there seems to be two separate implementations of `computeForces` right now, depending on which evaluator you plug in.
- Better documentation of the actual functional form of the Tersoff potential, as there doesn't seem to be any in the python docstrings right now.

Things need to be ported to the v3 API, and this will have to be done eventually.

Resolves: #730 and possibly #739 

## How Has This Been Tested?

I have added tests of the same form that were used for the pair potentials.

## Change log

```
added many_body namespace containing triplet potentials.
```

## Checklist:

- [x ] My name is on the [list of contributors](https://github.com/glotzerlab/hoomd-blue/blob/master/sphinx-doc/credits.rst).


tommy-waltmann:
One more note: we should have @danevans09 put these potentials in his mathematica notebook so we can test force energy accuracy and force energy relationship, but I'm not sure if the documentation is clear enough that it will be possible to run the values by hand. This is the reason I mentioned clarifying the docs earlier.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+        gamman = math.pow(gamma, n);
+        c2 = c * c;
+        d2 = d * d;
+        lambda3_cube = lambda3 * lambda3 * lambda3;
+
+        tersoff_coeffs = _hoomd.make_scalar2(C1, C2);
+        exp_consts = _hoomd.make_scalar2(lambda1, lambda2);
+        ang_consts = _hoomd.make_scalar3(c2, d2, m);
+
+        return _md.make_tersoff_params(cutoff_d, tersoff_coeffs, exp_consts, dimer_r, n, gamman, lambda3_cube, ang_consts, alpha);
```
b-butler, line 81: Is this still necessary?

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+            \end{cases}
+            \end{eqnarray*}
+
+    .. attention::
+
+        The RevCross potential models an asymmetric interaction between two different chemical moieties that can form a reversible bond.
+        This requires the definition of (at least) two different types of particles.
+        A reversible bond is only possible between two different species, otherwise :math:`v^{\left( 3b \right)}_{ijk}`, would prevent any bond.
+        In our example we then set the interactions for types A and B with ``potRevC.params[(['A','B'],['A','B'])] = dict(sigma=0.0,n=0,epsilon=0,lambda3=0)``
+        and the only non-zero energy only between the different types ``potRevC.params[('A','B')] = dict(sigma=1,n=100,epsilon=100,lambda3=1)``.
```
b-butler, line 145: Use snake case.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+            & - \dfrac{v_{ij}\left(\vec{r}_{ij}\right)}{\epsilon} \qquad r > r_{min}~.\\
+            \end{cases}
+            \end{eqnarray*}
+
+    .. attention::
+
+        The RevCross potential models an asymmetric interaction between two different chemical moieties that can form a reversible bond.
+        This requires the definition of (at least) two different types of particles.
+        A reversible bond is only possible between two different species, otherwise :math:`v^{\left( 3b \right)}_{ijk}`, would prevent any bond.
+        In our example we then set the interactions for types A and B with ``potRevC.params[(['A','B'],['A','B'])] = dict(sigma=0.0,n=0,epsilon=0,lambda3=0)``
```
b-butler, line 144:
```suggestion
        In our example we then set the interactions for types A and B with ``potRevC.params[[('A','A'),('B','B')]] = dict(sigma=0.0,n=0,epsilon=0,lambda3=0)``
```

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+        compensate the energy of multiple bonds, so it may cause unphysical situations.
+
+
+    Example::
+
+        nl = md.nlist.Cell()
+        potBondSwap = md.many_body.RevCross(r_cut=1.3,nlist=nl)
+        potBondSwap.params[(['A','B'],['A','B'])] = dict(sigma=0,n=0,epsilon=0,lambda3=0)
+        # a bond can be made only between A-B and not A-A or B-B
+        potBondSwap.params[('A','B')] = dict(sigma=1,n=100,epsilon=10,lambda3=1)
```
b-butler, line 166: Use snake case

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+    _cpp_class_name = "PotentialRevCross"
+    def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
+        super().__init__(nlist, r_cut, r_on, mode);
+        params = TypeParameter('params', 'particle_types',
+                               TypeParameterDict(sigma=2.0, n=1.0, epsilon=1.0,
+                                   lambda3=1.0, len_keys=2))
+        self._add_typeparam(params)
+
+    def attach(self, sim):
```
b-butler, line 176: Now `_attach`

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+        super().attach(sim)
+        self.nlist._cpp_obj.setStorageMode(
+            _md.NeighborList.storageMode.full)
+
+class SquareDensity(_NBody):
+    R""" Soft potential for simulating a van-der-Waals liquid
+
+    Args:
+        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
```
b-butler, line 185:
```suggestion
        nlist (:py:mod:`hoomd.md._Nlist`): Neighbor list
```

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+    _cpp_class_name = "PotentialSquareDensity"
+    def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
+        super().__init__(nlist, r_cut, r_on, mode);
+        params = TypeParameter(
+                'params',
+                'particle_types',
+                TypeParameterDict(A=0.0, B=float, len_keys=2))
+        self._add_typeparam(params)
+
+    def attach(self, sim):
```
b-butler, line 239:
```suggestion
    def attach(self, sim):
```
```suggestion
    def _attach(self, sim):
```

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+                    gamma=0.0,
+                    c=0.0,
+                    d=1.0,
+                    m=0.0,
+                    alpha=3.0,
+                    len_keys=2)
+                )
+        self._add_typeparam(params)
+
+    def attach(self, sim):
```
b-butler, line 52: See the updates to `_attach` in the current `feature/new-object-API` branch.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+        super().__init__(nlist, r_cut, r_on, mode);
+        params = TypeParameter('params', 'particle_types',
+                               TypeParameterDict(sigma=2.0, n=1.0, epsilon=1.0,
+                                   lambda3=1.0, len_keys=2))
+        self._add_typeparam(params)
+
+    def attach(self, sim):
+        super().attach(sim)
+        self.nlist._cpp_obj.setStorageMode(
+            _md.NeighborList.storageMode.full)
```
b-butler, line 179: We do this for all 3 body potentials. It may be worth making a `_ThreeBody` class that does this automatically.

sphinx-doc/module-md-pair.rst
```diff
@@ -56,10 +53,7 @@ md.pair
         Morse,
         Moliere,
         ReactionField,
-        revcross,
         SLJ,
-        square_density,
         table,
```
b-butler, line 57: Can we remove all pair potentials that are not currently updated? I know this is not the purpose of the PR but does need to be done.

b-butler (changes requested):
Overall looks good. I left mostly stylistic comments.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+            \end{eqnarray*}
+
+    .. attention::
+
+        The RevCross potential models an asymmetric interaction between two different chemical moieties that can form a reversible bond.
+        This requires the definition of (at least) two different types of particles.
+        A reversible bond is only possible between two different species, otherwise :math:`v^{\left( 3b \right)}_{ijk}`, would prevent any bond.
+        In our example we then set the interactions for types A and B with ``potRevC.params[(['A','B'],['A','B'])] = dict(sigma=0.0,n=0,epsilon=0,lambda3=0)``
+        and the only non-zero energy only between the different types ``potRevC.params[('A','B')] = dict(sigma=1,n=100,epsilon=100,lambda3=1)``.
+        Notice that the number of the minoritary species corresponds to the maximum number of bonds.
```
tcmoore3, line 146: Typo

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+                                   lambda3=1.0, len_keys=2))
+        self._add_typeparam(params)
+
+    def attach(self, sim):
+        super().attach(sim)
+        self.nlist._cpp_obj.setStorageMode(
+            _md.NeighborList.storageMode.full)
+
+class SquareDensity(_NBody):
+    R""" Soft potential for simulating a van-der-Waals liquid
```
tcmoore3, line 182:
```suggestion
class SquareDensity(_NBody):
    R""" Soft potential for simulating a van der Waals liquid
```

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+    - :math:`A` - *A* (in units of volume^-1) - mean density (*default*: 0)
+    - :math:`B` - *B* (in units of energy*volume^2) - coefficient of the harmonic density term
+
+    Example::
+
+        nl = nlist.Cell()
+        sqd = md.many_body.SquareDensity(r_cut=3.0, nlist=nl)
+        sqd.params[('A', 'A')] = dict(A=0.1, B=1.0)
+
+    For further details regarding this multibody potential, see
```
tcmoore3, line 220: See...?

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+
+    Warning:
+        Currently HOOMD does not support reverse force communication between MPI domains on the GPU.
+        Since reverse force communication is required for the calculation of multi-body potentials, attempting to use the
+        square_density potential on the GPU with MPI will result in an error.
```
tcmoore3, line 225: Should this warning be in the other 3-body potential docstrings?

tcmoore3 (approved):
Just a few stylistic things to address, otherwise looks good.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,244 @@
...
+    non-bonded particle pair in the simulation.  Despite the fact that the Tersoff potential accounts
+    for the effects of third bodies, the species of the third body is irrelevant. It can thus use
+    type-pair parameters similar to those of the pair potentials.
+
+    The Tersoff potential is a bond-order potential based on the Morse potential that accounts for the weakening of
+    individual bonds with increasing coordination number. It does this by computing a modifier to the
+    attractive term of the potential. The modifier contains the effects of third-bodies on the bond
+    energies. The potential also includes a smoothing function around the cutoff. The smoothing function
+    used in this work is exponential in nature as opposed to the sinusoid used by Tersoff. The exponential
+    function provides continuity up (I believe) the second derivative.
```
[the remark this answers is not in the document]
tommy-waltmann: Neither do I, I'm not really sure what to do with this.

hoomd/md/EvaluatorTersoff.h
```diff
@@ -23,50 +23,83 @@
...
+                Scalar c2 = c * c;
+                Scalar d2 = d * d;
+                ang_consts = make_scalar3(c2, d2, m);
+                }
+
+            pybind11::dict asDict()
+                {
+                pybind11::dict v;
+                v["C1"] = coeffs.x;
+                v["C2"] = coeffs.y;
```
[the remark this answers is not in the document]
tommy-waltmann: That's probably a good idea. Is there an example of how this should work on the python end as a tuple?

hoomd/md/EvaluatorRevCross.h
```diff
@@ -100,8 +110,8 @@ class EvaluatorRevCross
         //! Check whether a pair of particles is interactive
         DEVICE bool areInteractive()
             {
-            if ((rik_sq < rcutsq )&&(epsilon_dev!=Scalar(0.0)))                 
-            //if rik_sq < rcutsq                 
+            if ((rik_sq < rcutsq )&&(epsilon_dev!=Scalar(0.0)))
+            //if rik_sq < rcutsq
                 return true;
             else return false;
```
bdice, line 116:
```suggestion
            return (rik_sq < rcutsq) && (epsilon_dev != Scalar(0.0));
```

hoomd/md/pytest/test_potential.py
```diff
@@ -471,12 +542,19 @@ def test_attached_params(simulation_factory, lattice_snapshot_factory,
...
     sim = simulation_factory(snap)
+    _check_for_skip(sim, valid_params.pair_potential)
     sim.operations.integrator = hoomd.md.Integrator(dt=0.005)
     sim.operations.integrator.forces.append(pot)
     sim.run(1)
     _assert_equivalent_type_params(pot.params.to_dict(),
                                    valid_params.pair_potential_params)
 
+def _check_for_skip(sim, pair_potential):
+    """ Determines if the simulation is able to run this pair potential """
```
bdice, line 553:
```suggestion
    """Determines if the simulation is able to run this pair potential."""
```

hoomd/md/pytest/test_potential.py
```diff
@@ -471,12 +542,19 @@ def test_attached_params(simulation_factory, lattice_snapshot_factory,
...
+
+    if isinstance(sim.device, hoomd.device.GPU) and sim.device.communicator.num_ranks > 1 and \
+            str(pair_potential).startswith('<class \'hoomd.md.many_body'):
```
bdice, line 556: This is a bit of a rough hack -- we should use `isinstance(pair_potential, Triplet)` or something similar.

tommy-waltmann:
@joaander @b-butler This is ready for review again

hoomd/md/many_body.py
```diff
@@ -0,0 +1,343 @@
...
+
+
+class Tersoff(Triplet):
+    """Tersoff Potential.
+
+    Args:
+        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
+        r_cut (float): Default cutoff radius (in distance units).
+        r_on (float): Default turn-on radius (in distance units).
+        mode (str): Energy shifting mode.
```
bdice, line 43: I might be missing something but I don't see the definition of `mode` or the allowed options for `mode`. If I've missed that, you can just resolve this conversation.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,343 @@
...
+    term of the potential. The modifier contains the effects of third-bodies on
+    the bond energies. The potential also includes a smoothing function around
+    the cutoff. The smoothing function used in this work is exponential in
+    nature as opposed to the sinusoid used by
+    `J. Tersoff 1988 <https://journals.aps.org/prb/abstract/10.1103/PhysRevB.38.9902>`_.
+
+    Example::
+
+        nl = md.nlist.Cell()
+        tersoff = md.many_body.Tersoff(r_cut=1.3,nlist=nl)
```
bdice, line 129: Style suggestion: I recommend putting a space after commas. This is probably enforced by `yapf` in code, but that doesn't apply to examples in docstrings.
```suggestion
        tersoff = md.many_body.Tersoff(r_cut=1.3, nlist=nl)
```

hoomd/md/many_body.py
```diff
@@ -0,0 +1,343 @@
...
+        tersoff.params[('A','B')] = dict(magnitudes=(2.0, 1.0), lambda3=5.0)
```
bdice, line 130:
```suggestion
        tersoff.params[('A', 'B')] = dict(magnitudes=(2.0, 1.0), lambda3=5.0)
```

hoomd/md/many_body.py
```diff
@@ -0,0 +1,343 @@
...
+
+    :py:class:`RevCross` specifies that the revcross three-body potential
+    should be applied to every non-bonded particle pair in the simulation.
+    Despite the fact that the revcross potential accounts for the effects of
+    third bodies, it is actually just a combination of two body potential terms.
+    It can thus use type-pair parameters similar to those of the pair potentials.
+
+    The RevCross potential has been described in detail in
+    `S. Ciarella and W.G. Ellenbroek 2019 <https://arxiv.org/abs/1912.08569>`_.
+    It is based on a generalized-Lennard-Jones pairwise attraction to form bonds
```
bdice, line 171:
```suggestion
    It is based on a generalized Lennard-Jones pairwise attraction to form bonds
```

hoomd/md/many_body.py
```diff
@@ -0,0 +1,343 @@
...
+    between interacting particless:
```
bdice, line 172:
```suggestion
    between interacting particles:
```

hoomd/md/many_body.py
```diff
@@ -0,0 +1,343 @@
...
+        params (TypeParameter[tuple[``particle_type``, ``particle_type``], dict]):
+            The revcross potential parameters. The dictionary has the following keys:
+
+            * ``epsilon`` (`float`, **required**) - :math:`\\varepsilon` (in units of energy)
+
+            * ``sigma`` (`float`, **required**) - :math:`\\sigma` - (in distance units)
+
+            * ``n`` (`float`, **required**) - *n* - (unitless)
+
+            * ``lambda3`` (`float`, **required**) - *lambda3`* - (unitless)
```
bdice, line 256: Check how the docs are rendering. There are some other places where I'm not sure about the syntax correctness.
```suggestion
            * ``lambda3`` (`float`, **required**) - *lambda3* - (unitless)
```

bdice (commented):
Some comments below. I am not familiar enough with the new APIs to validate that part of the code, but I suggested some minor improvements. I will defer to other reviewers for approving this PR.

hoomd/md/pytest/test_potential.py
```diff
@@ -471,12 +540,19 @@ def test_attached_params(simulation_factory, lattice_snapshot_factory,
...
     sim.run(1)
     _assert_equivalent_type_params(pot.params.to_dict(),
                                    valid_params.pair_potential_params)
 
+def _check_for_skip(sim, pair_potential):
+    """Determines if the simulation is able to run this pair potential."""
+
+    if isinstance(sim.device, hoomd.device.GPU) and sim.device.communicator.num_ranks > 1 and \
+            issubclass(pair_potential, hoomd.md.many_body.Triplet):
+        pytest.skip("Cannot run triplet potentials with GPU+MPI enabled")
```
b-butler, line 555: We should probably define this before using it.

hoomd/md/many_body.py
b-butler: Shouldn't `*lambda3*` be in math notation given sigma and varepsilon?

hoomd/md/many_body.py
```diff
@@ -0,0 +1,343 @@
...
+
+    Use `params` dictionary to set potential coefficients. The coefficients must be set per unique pair of particle types.
+
+    Attributes:
+        params (TypeParameter[tuple[``particle_type``, ``particle_type``], dict]):
+            The SquareDensity potential parameters. The dictionary has the
+            following keys:
+
+            *``A`` (`float`, **required**) - :math:`A` - mean density (in units of volume^-1, *default*:0)
+            *``B`` (`float`, **required**) - :math:`B` - coefficient of the harmonic density term (in units of energy*volumne^2)
```
b-butler, line 321: These don't render well with sphinx.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,343 @@
...
+            * ``lambda3`` (`float`) - :math:`\\lambda_3` - exponential factor in :math:`\\chi_{ij}` (in units of 1/length, *default*: 0.0)
+            * ``dimer_r`` (`float`) - :math:`r_D` - length shift in attractive and repulsive terms (in units of length, *default*: 1.5)
+            * ``cutoff_thickness`` (`float`) - :math:`r_{CT}` - distance which defines the different regions of the potential (in units of length, *default*: 0.2)
+            * ``alpha`` (`float`) - :math:`\\alpha` - decay rate of the cutoff term :math:`f_C(r)` (dimensionless, *default*: 3.0)
+            * ``n`` (`float`) - :math:`n` - power in :math:`b_{ij}` (dimensionless, *default*: 0.0)
+            * ``gamma`` (`float`) - :math:`\\gamma` - coefficient in :math:`b_{ij}` (dimensionless, *default*: 0.0)
+            * ``c`` (`float`) - :math:`c` - coefficient in :math:`g(\\theta)` (dimensionless, *default*: 0.0)
+            * ``d`` (`float`) - :math:`d` - coefficient in :math:`g(\\theta)` (dimensionless, *default*: 1.0)
+            * ``m`` (`float`) - :math:`m` - coefficient in :math:`g(\\theta)` (dimensionless, *default*: 0.0)
```
b-butler, line 115: This section should be at the end of the docstring (besides any examples).

hoomd/md/many_body.py
```diff
@@ -0,0 +1,343 @@
...
+
+    .. attention::
+
+        The RevCross potential models an asymmetric interaction between two
+        different chemical moieties that can form a reversible bond. This
+        requires the definition of (at least) two different types of particles.
+        A reversible bond is only possible between two different species,
+        otherwise :math:`v^{\\left( 3b \\right)}_{ijk}`, would prevent any bond.
+        In our example we then set the interactions for types A and B with
+        ``rev_c.params[[('A','B'),('A','B')]] = dict(sigma=0.0,n=0,epsilon=0,lambda3=0)``
```
b-butler, line 224: Unfortunately this renders poorly when my browser is set to half a screen. I am not sure of the best way to change this though.

b-butler (changes requested):
I have a few documentation comments to address.

hoomd/md/many_body.py
joaander: We could break the code block into two at the `=`.

joaander (changes requested):
Thanks. I agree with the previous reviews. This branch looks good overall. To finish this, check the rendered docs (https://hoomd-blue--752.org.readthedocs.build/en/752/module-md-many_body.html) and adjust the rst formatting. I suggest building the docs with sphinx locally so you don't need to wait for readthedocs to see every small change.

hoomd/md/many_body.py
tommy-waltmann: The definition of mode is in `md.pair.Pair`. It is explained in `md.many_body.Triplet`, that the triplet potentials use many of the same arguments as the pair potentials, and the common documentation on those is in `md.pair.Pair`. Do you think this is enough, or should we be more explicit about this somehow?

joaander: Yes, we should be more explicit. We haven't officially decided on a policy for this but the practice we last discussed was to use the :inherited-members: sphinx option for classes like `Triplet` so that the relevant inherited docs are present in the local HTML page as the classes users actually use. See `module-hommd-operation.rst` for an example of now to enable this option for one class and not others.

hoomd/md/pytest/test_potential.py
```diff
@@ -424,6 +454,45 @@ def _valid_params(particle_types=['A', 'B']):
                                         dict(zip(combos,
                                                  dlvo_valid_param_dicts)),
                                         {}))
+
+    tersoff_arg_dict = {
+            'cutoff_thickness': [0.1, 0.5, 1.0],
```
joaander, line 459: Regarding a test that randomly throws particles out of the box, I suggest changing the test to just run 1 step. The purpose of a "run" test like this is to ensure that the kernel executes and does something. Ensuring a stable simulation will require a proper selection of step size, potential parameters, and the initial condition which is challenging to accomplish in this fully parameterized fashion.

If the simulation is still unstable enough to through particles out of the box in just 1 step then we either need to use more appropriate parameters for the potential or consider dropping that test entirely.

joaander (changes requested):
Thanks! The docs look much better now. We just need to document the inherited members of `Pair` better and get that test to pass.

Looking at it closer, `:inherited-members:` won't capture all of the docs from `Pair`. I think for now we should just also copy and paste the `Attributes:` section of the `Pair` docstring so the docs for `r_cut` appears in `Triplet`.

hoomd/md/many_body.py
tommy-waltmann: What we need to be showing is the documentation under the class declaration, not the inherited members, and I can't find an option for this. Any ideas?

joaander: Sphinx doesn't inherit docstrings themselves, so you'll have to copy and paste.

joaander (approved):
Thanks!

joaander:
@b-butler, any final comments before we merge this one?

b-butler (approved):
LGTM @tommy-waltmann thanks for all the work! I left one comment you can choose to address before merging. The documentation looks good as well when I pulled it up.

hoomd/md/pytest/test_potential.py
```diff
@@ -452,6 +521,14 @@ def _update_snap(pair_potential, snap):
         snap.particles.diameter[1] = 0.5
 
 
+def _check_for_skip(sim, pair_potential):
```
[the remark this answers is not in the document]
tommy-waltmann: The way this method is used in the tests means that it could be used for any reason to skip a test. Right now, it only applies to triplet potentials, but as we add more types of potentials (like anisotropic), we may want to skip some of those tests in some cases as well.

tommy-waltmann:
The fact that the managed memory problem is occurring again for a `cuda11` test is concerning. Should we alter this test as we did the `cuda10` test?

hoomd/md/pytest/test_potential.py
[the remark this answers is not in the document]
b-butler: Right but I would prefer to have a few functions that skip for different reasons than one function that skips them for multiple as the former I more clear to future developers.

b-butler:
Maybe we should spent some more time on this then. It failing on CUDA 11 Is troubling.

joaander:
@tommy-waltmann how do you plan to alter the test? The changes you made before seemed to indicate the the only failing configuration was cuda10 with always_managed_memory=on. I don't know what new changes you plan to make.

That failing test has one reporting that the box is too small, one test where the expected forces are slightly out of tolerance and another where the tersoff test sent a particle flying outside the box. The test was run on `mays`. Is there any pattern in what hosts those tests fail on? Maybe one of the GPUs is in a bad state?

hoomd/md/pytest/test_potential.py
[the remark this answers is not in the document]
tommy-waltmann: I think having one helper function looks cleaner in the calling code rather than calling each of the many helper functions. I don't think its really a big deal at this point, if we need to change it later we can.

tommy-waltmann:
I can try re-running the failing test to see if the GPU isn't in a bad state anymore. The diff on the merge commit is huge, and it may have changed something causing the test to fail, but I couldn't diagnose what the problem is.

hoomd/md/NeighborList.cc
```diff
@@ -438,22 +440,14 @@ void NeighborList::updateRList()
...
-            // copy the first matrix as a starting point
-            memcpy(h_r_cut.data, h_consumer_r_cut.data, sizeof(Scalar)*m_r_cut.getNumElements());
-            }
-        else
+        // take the maximum
+        for (unsigned int i=0; i < m_pdata->getNTypes(); ++i)
             {
-            // take the maximum
-            for (unsigned int i=0; i < m_pdata->getNTypes(); ++i)
+            for (unsigned int j=0; i < m_pdata->getNTypes(); ++i)
```
bdice, line 446: Should this be a `j`?
```suggestion
            for (unsigned int j=0; j < m_pdata->getNTypes(); ++j)
```

joaander: Good catch.

hoomd/md/pytest/test_potential.py
```diff
@@ -573,19 +651,20 @@ def test_run(simulation_factory, lattice_snapshot_factory, valid_params):
...
+    sim.operations._schedule()
+    old_snap = sim.state.snapshot
+    sim.run(2)
+    new_snap = sim.state.snapshot
+    if new_snap.exists:
+        assert not np.allclose(new_snap.particles.position,
+                               old_snap.particles.position)
 
 
 def test_energy_shifting(simulation_factory, two_particle_snapshot_factory):
```
tommy-waltmann, line 670: Why is this test being skipped now?

It mentions a bug, has the bug been fixed? If it has, we shouldn't be skipping this test anymore.

hoomd/md/NeighborListGPUStencil.cc
```diff
@@ -241,8 +241,17 @@ void NeighborListGPUStencil::buildNlist(unsigned int timestep)
...
+        oss << "nlist: Simulation box is too small! Particles would be interacting with themselves."
+            << "rmax=" << rmax << std::endl;
+
+        if (box.getPeriodic().x)
+            oss << "nearest_plane_distance.x=" << nearest_plane_distance.x << std::endl;
+        if (box.getPeriodic().y)
+            oss << "nearest_plane_distance.y=" << nearest_plane_distance.y << std::endl;
+        if (this->m_sysdef->getNDimensions() == 3 && box.getPeriodic().z)
+            oss << "nearest_plane_distance.z=" << nearest_plane_distance.z << std::endl;
+        throw std::runtime_error(oss.str());
```
tommy-waltmann, line 254: Was this only left in the code for debugging purpsoes?

hoomd/md/NeighborListStencil.cc
```diff
@@ -136,8 +136,17 @@ void NeighborListStencil::buildNlist(unsigned int timestep)
...
+        oss << "nlist: Simulation box is too small! Particles would be interacting with themselves."
+            << "rmax=" << rmax << std::endl;
+
+        if (box.getPeriodic().x)
+            oss << "nearest_plane_distance.x=" << nearest_plane_distance.x << std::endl;
+        if (box.getPeriodic().y)
+            oss << "nearest_plane_distance.y=" << nearest_plane_distance.y << std::endl;
+        if (this->m_sysdef->getNDimensions() == 3 && box.getPeriodic().z)
+            oss << "nearest_plane_distance.z=" << nearest_plane_distance.z << std::endl;
+        throw std::runtime_error(oss.str());
```
tommy-waltmann, line 149: Was this only left in the code for debugging purposes?

tommy-waltmann (commented):
I took one last look over this PR and had some questions, in addition to the one I raised above this review. @b-butler @joaander This is ready for review again.

hoomd/md/NeighborListStencil.cc
bdice: @tommy-waltmann I think this is meant to be left as it is. It raises an error if the box is too small. The new code in this diff just makes the message more verbose and useful to the reader.

hoomd/md/pytest/test_potential.py
joaander: @b-butler added this skip.

hoomd/md/NeighborListStencil.cc
joaander: Correct, the message is now more useful for users to understand why they are getting the error.

hoomd/md/PotentialTersoff.h
```diff
@@ -82,18 +82,41 @@ class PotentialTersoff : public ForceCompute
         //! Destructor
         virtual ~PotentialTersoff();
 
-        //! Set the pair parameters for a single type pair
+        //! Set and get the pair parameters for a single type pair
```
joaander, line 85:
```suggestion
        //! Set the pair parameters for a single type pair
```
This method only sets the parameters.

hoomd/md/pair/pair.py
```diff
@@ -15,10 +15,9 @@
 import math
 
 
-class pair(force._force):
```
joaander, line 18: Remove this class.

hoomd/md/pytest/test_potential.py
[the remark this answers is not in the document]
joaander: Yes, the name should be clearer for what this skips. I cannot imagine how one function could possibly handle all cases. Different tests will have different configurations that need to be skipped. The name should be explicit: `skip_if_triplet_gpu_mpi`.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,402 @@
+import hoomd
+from hoomd.md import _md
+from hoomd.md.pair import Pair
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeparam import TypeParameter
+
+
+class Triplet(Pair):
```
joaander, line 8: Why does this inherit from Pair? This isn't an is-a relationship. The only commonalities are that both use neighbor lists and have r_cut type parameters. I think this should ineherit from `force.Force`.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,402 @@
...
+    """Common three body potential documentation.
```
joaander, line 9: This docstring appears to be copied directly from `Pair`. Much (maybe all?) of this is inaccurate. Triplet potentials have a different functional form compared to pair potentials. Also, add the attribute documentation for the `r_cut` type parameter.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,402 @@
...
+        V_{ij}(r)  =  4 \\varepsilon \\left[ \\left( \\dfrac{ \\sigma}{r_{ij}} \\right)^{2n}
+        - \\left( \\dfrac{ \\sigma}{r_{ij}} \\right)^{n} \\right] \\qquad r<r_{cut}
+        \\end{eqnarray*}
+
+    with the following coefficients:
+
+    - :math:`\\varepsilon` - *epsilon* (in energy units)
+    - :math:`\\sigma` - *sigma* (in distance units)
+    - :math:`n` - *n* (unitless)
+    - :math:`m` - *m* (unitless)
```
joaander, line 247: Remove. There is no `m` parameter.

joaander: Actually, remove this whole list. Look at the `md.pair` documentation style to see how we label the math symbols for the parameters only in one place in the typeparam documentation below.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+class SquareDensity(Triplet):
+    """Soft potential for simulating a van der Waals liquid.
+
+    Args:
+        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
+        r_cut (float): Default cutoff radius (in distance units).
+        r_on (float): Default turn-on radius (in distance units).
+        mode (str): Energy shifting mode.
```
joaander, line 342:
```suggestion
```

hoomd/md/many_body.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+class Tersoff(Triplet):
+    """Tersoff Potential.
+
+    Args:
+        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
+        r_cut (float): Default cutoff radius (in distance units).
+        r_on (float): Default turn-on radius (in distance units).
+        mode (str): Energy shifting mode.
```
joaander, line 103:
```suggestion
```
These are not used. `r_on` also needs to be removed from C++.

hoomd/md/many_body.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+class RevCross(Triplet):
+    """Reversible crosslinker three-body potential to model bond swaps.
+
+    Args:
+        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
+        r_cut (float): Default cutoff radius (in distance units).
+        r_on (float): Default turn-on radius (in distance units).
+        mode (str): Energy shifting mode.
```
joaander, line 221:
```suggestion
```

hoomd/md/many_body.py
joaander: This shows up again in later comments where the inherited `r_on`, `mode` and `compute_energy` are not present in the C++ implementation of triplet potentials.

joaander (changes requested):
Thanks!
There are a few misc changes needed and then this is ready to merge.

hoomd/md/many_body.py
tommy-waltmann: I think I did this for the purpose of code re-use, back when I mistakenly thought there was an `r_on` for triplet potentials.

joaander (approved):
Thanks for the huge effort @tommy-waltmann!

Issue #717, opened 2020-08-13
Make hoomd.device.auto a function rather than class


b-butler:
## Purpose

Given there are CPU and GPU device classes, `hoomd.device.Auto` should be a function `hoomd.device.auto` that returns either a CPU or GPU object. 

## Tasks

- [ ] Remove `hoomd.device.Auto`
- [ ] Write function `hoomd.device.auto`
- [ ] Document and test function

## Pertinent Information

Since we default to GPU, we could just try to create a GPU device object and if it fails return a CPU device in Python.

joaander:
The C++ `ExecutionConfiguration` class implements the auto-selection logic. The new `auto` function can construct this object and then return a constructed `CPU` or `GPU` object based on which mode `ExecutionConfiguration` chooses.

PR #754, opened 2020-08-26
device.auto

j-proc:
## Description

The auto function now returns a CPU or GPU device object rather than existing as it's own class. 

## Motivation and Context

Resolves: #717 
Closes: #747

### DIFF
--- a/hoomd/device.py
+++ b/hoomd/device.py
@@ -359,9 +359,8 @@ class CPU(_Device):
             self.num_cpu_threads = num_cpu_threads
 
 
-def auto(communicator=None, msg_file=None, shared_msg_file=None, notice_level=2):
-    """
-    Allow simulation hardware to be chosen automatically by HOOMD-blue
+def auto_select(communicator=None, msg_file=None, shared_msg_file=None, notice_level=2):
+    """Allow simulation hardware to be chosen automatically by HOOMD-blue.
 
     Args:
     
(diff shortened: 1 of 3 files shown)


j-proc:
Was able to just do very small changes, so I based it on your refactor branch and should be trivial to merge in now.

hoomd/device.py
```diff
@@ -359,28 +359,29 @@ def __init__(self,
             self.num_cpu_threads = num_cpu_threads
 
 
-class Auto(_Device):
+def auto(communicator=None, msg_file=None, shared_msg_file=None, notice_level=2):
```
joaander, line 362: Let's make the method a verb. `auto_select`?

hoomd/device.py
```diff
@@ -359,28 +359,29 @@ def __init__(self,
...
     """
     Allow simulation hardware to be chosen automatically by HOOMD-blue
```
joaander, line 364: Fix the pydocstyle errors:
```
device.py:363 in public function `auto`:
        D212: Multi-line docstring summary should start at the first line
device.py:363 in public function `auto`:
        D415: First line should end with a period, question mark, or exclamation point (not 'e')
```

hoomd/device.py
j-proc: Done

hoomd/device.py
j-proc: Done

Issue #512, opened 2019-11-13
Move HPMC boxmc updater to v3.0 API


b-butler:
## Description
Refactor the HPMC boxmc updater to the HOOMD v.3.0 API. The template can be found in the `feature/new-object-API` branch's `Clusters` updater in `hoomd/hpmc/update.py`.

## Proposed solution
The solution involves following the template mentioned above.

## Additional context
Any pull request should merge into `feature/new-object-API` **not** `master`.

PR #758, opened 2020-08-28
Update boxmc to hoomd v3.0

lyrivera:
This PR moves the `boxmc` updater into the new v3.0 API. We took the liberty to remove some unused variable, methods and pybind11 exports as well as renaming some variables to meet `HOOMD` style standards. Docstrings have also been updated but still need to add the link to the examples notebook.

## Motivation and Context

Resolves: #512

## How Has This Been Tested?

Tested only locally for obvious issues. 

## Checklist:


hoomd/hpmc/UpdaterBoxMC.cc
```diff
@@ -16,25 +16,23 @@ namespace hpmc
 UpdaterBoxMC::UpdaterBoxMC(std::shared_ptr<SystemDefinition> sysdef,
                              std::shared_ptr<IntegratorHPMC> mc,
                              std::shared_ptr<Variant> P,
-                             const Scalar frequency,
                              const unsigned int seed)
         : Updater(sysdef),
           m_mc(mc),
           m_P(P),
```
bdice, line 22: Should this be renamed `m_beta_P` to match the other times this name is used?

hoomd/hpmc/UpdaterBoxMC.h
```diff
@@ -240,59 +394,26 @@ class UpdaterBoxMC : public Updater
...
-        Scalar m_Volume_delta;                      //!< Amount by which to change parameter during box-change
-        float m_Volume_weight;                     //!< relative weight of volume moves
-        Scalar m_lnVolume_delta;                      //!< Amount by which to change parameter during box-change
-        float m_lnVolume_weight;                   //!< relative weight of volume moves
-        Scalar m_Volume_A1;                         //!< Ratio of Lx to Ly to use in isotropic volume changes
-        Scalar m_Volume_A2;                         //!< Ratio of Lx to Lz to use in isotropic volume changes
+        Scalar m_volume_delta;                      //!< Amount by which to change parameter during box-change
+        float m_volume_weight;                     //!< relative weight of volume moves
+        Scalar m_ln_volume_delta;                      //!< Amount by which to change parameter during box-change
+        float m_ln_volume_weight;                   //!< relative weight of volume moves
```
bdice, line 404:
```suggestion
        Scalar m_volume_delta;                      //!< Amount by which to change volume during box-change
        float m_volume_weight;                      //!< relative weight of volume moves
        Scalar m_ln_volume_delta                    //!< Amount by which to change log volume during box-change
        float m_ln_volume_weight;                   //!< relative weight of log volume moves
```

hoomd/hpmc/update.py
```diff
@@ -14,388 +14,171 @@
...
-        if weight is not None:
-            self.volume_weight = float(weight)
-
-        if delta is not None:
-            self.volume_delta = float(delta)
-
-        self.cpp_updater.volume(self.volume_delta, self.volume_weight);
-        return {'delta': self.volume_delta, 'weight': self.volume_weight};
+            * ``weight`` (float) - relative weight of log(V) box moves relative to other box move types.
+            * ``delta`` (float) - maximum change of **ln(V)** (where V is box area (2D) or volume (3D)).
```
bdice, line 40: It's weird that one line says "log" and the other says "ln." This should be made consistent.

hoomd/hpmc/update.py
```diff
@@ -14,388 +14,171 @@
...
-        self.check_initialization();
+        length (dict):
+            Enable/disable isobaric box length moves and set parameters (change box lengths independently).
+            The dictionary has the following keys:
 
-        if weight is not None:
-            self.ln_volume_weight = float(weight)
+            * ``weight`` (float) - maximum change of the box thickness for each pair of parallel planes
+                     connected by the corresponding box edges. I.e. maximum change of HOOMD-blue box parameters Lx, Ly, Lz.
+            * ``delta`` (list or tuple) -  maximum change of the box tilt factor xy, xz, yz.
```
bdice, line 55: Looks like this docstring is incorrectly copied.
```suggestion
            * ``delta`` (list or tuple) -  maximum change of the box lengths Lx, Ly, Lz.
```

bdice (commented):
I would like to use the BoxMC feature in the near term. I reviewed the code and it looks good overall. Nice work, @lyrivera! I would like to suggest a different approach for the random weighted selection of the move type. Here is a code example: https://godbolt.org/z/7oYTo9 The `weight_sums` vector should be computed in advance (instead of for every move) with `std::partial_sum`, and the move can be selected using `std::lower_bound`. This should reduce the amount of computations on each step, as well as make the computation more general for different move types.

joaander:
> I would like to use the BoxMC feature in the near term. I reviewed the code and it looks good overall. Nice work, @lyrivera! I would like to suggest a different approach for the random weighted selection of the move type. Here is a code example: https://godbolt.org/z/7oYTo9 The `weight_sums` vector should be computed in advance (instead of for every move) with `std::partial_sum`, and the move can be selected using `std::lower_bound`. This should reduce the amount of computations on each step, as well as make the computation more general for different move types.

Yes, this code should be improved. Dynamic memory allocations are very expensive and should be avoided every time step. We can keep the std::vector's as member variables and reuse them. Better yet, we should implement a re-usable `WeightedSelection` class in `RandomNumbers.h` and `BoxMC` could keep one of these classes as a member variable.

lyrivera:
@bdice @joaander We tried to limit the C++ changes to simply renaming and removing unused stuff, but we'll work on the suggested improvements soon. Thanks!

bdice:
@joaander Great idea, I’d be happy to contribute to that after this is merged.
@lyrivera Totally understandable to keep the scope narrow since it affects so much code, just wanted to share the idea for an improvement while it was on my mind.

joaander:
Make an issue and we can tackle it later.

joaander:
@lyrivera can you please merge master into this branch and resolve conflicts? I will review after it is up to date.

lyrivera:
> @lyrivera can you please merge master into this branch and resolve conflicts? I will review after it is up to date.

Sure thing. Will work on it later today.

lyrivera:
> > @lyrivera can you please merge master into this branch and resolve conflicts? I will review after it is up to date.
> 
> Sure thing. Will work on it later today.

@joaander sorry for the delay. I postponed it for a day but didn't put it in my TODO list and totally forgot about it.

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,171 @@
...
+        if not self.is_attached:
+            return None
+        else:
+            return self._cpp_obj.getCounters(1)
 
-    def get_volume_acceptance(self):
-        R""" Get the average acceptance ratio for volume changing moves.
+    @log(flag="sequence")
+    def volume_moves(self):
+        R""" Get a tuple with the accepted and rejected volume and length moves.
```
joaander, line 130:
```suggestion
        """tuple[int, int]: The accepted and rejected volume and length moves.
```

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,171 @@
...
+        if counter is None:
+            return (0.0, 0.0)
+        else:
+            return counter.volume
 
-    def get_ln_volume_acceptance(self):
-        R""" Get the average acceptance ratio for log(V) changing moves.
+    @log(flag="sequence")
+    def ln_volume_moves(self):
+        R""" Get a tuple with the accepted and rejected ln(V) moves.
```
joaander, line 144:
```suggestion
        """tuple[int, int]: The accepted and rejected ln(V) moves.
```

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,171 @@
...
+        if counter is None:
+            return (0.0, 0.0)
+        else:
+            return counter.ln_volume
 
-    def get_shear_acceptance(self):
-        R"""  Get the average acceptance ratio for shear changing moves.
+    @log(flag="sequence")
+    def shear_moves(self):
+        R""" Get a tuple with the accepted and rejected shear moves.
```
joaander, line 158:
```suggestion
        """tuple[int, int]: The accepted and rejected shear moves.
```

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,171 @@
...
+        if counter is None:
+            return (0.0, 0.0)
+        else:
+            return counter.shear
 
-    def get_aspect_acceptance(self):
-        R"""  Get the average acceptance ratio for aspect changing moves.
+    @log(flag="sequence")
+    def aspect_moves(self):
+        R""" Get a tuple with the accepted and rejected aspect moves.
```
joaander, line 172:
```suggestion
        """tuple[int, int]: The accepted and rejected aspect moves.
```

joaander (changes requested):
* Please reset the `hipper` submodule to v0.2.0.
* Add unit tests to ensure that parameters can be set.

joaander:
Also, fix the base class for BoxMC so that tests will run: https://dev.azure.com/glotzerlab/hoomd-blue/_build/results?buildId=6613&view=logs&j=b9a5787c-1cca-55f6-5e30-47b4f2efc53a&t=f8bd10e5-360c-52a8-e4e0-655a006c0e98&l=35

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,154 @@
...
+        super().__init__(trigger)
 
-        if weight is not None:
-            self.shear_weight = float(weight)
+        _default_dict = dict(weight=float(0), delta=float(0))
+        param_dict = ParameterDict(seed=int(seed),
+                                   volume=_default_dict,
+                                   ln_volume=_default_dict,
+                                   aspect=_default_dict,
+                                   length=dict(weight=float(0), delta=[float(0)]*3),
```
b-butler, line 87:
```suggestion
                                   length=dict(weight=float(0), delta=(float(0),)*3),
```

b-butler (changes requested):
I think it makes sense to give `volume` another key `mode` which can be `None` or `'ln'` or something of the like (`'standard'` and `'ln'` would also work). Since `ln_volume` moves shouldn't be paired with with `volume` moves. @joaander what do you think on this?

Otherwise, LGTM. We need to get the tests passing though.

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,154 @@
...
-        Example::
-
-            mc = hpmc.integrate.shape(..);
-            mc.shape_param[name].set(....);
-            box_update = hpmc.update.boxmc(mc, betaP=10, seed=1)
-            run(100)
-            s_accept = box_update.get_shear_acceptance()
+        counter = self.counter
+        if counter is None:
+            return (0.0, 0.0)
```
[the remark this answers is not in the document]
lyrivera: Agree and fixed

hoomd/hpmc/pytest/test_boxmc.py
```diff
@@ -0,0 +1,193 @@
...
+
+    sim = simulation_factory(two_particle_snapshot_factory())
+    sim.operations.updaters.append(boxmc)
+
+    # BoxMC requires an HPMC integrator
+    mc = hoomd.hpmc.integrate.Sphere(seed=1)
+    mc.shape['A'] = dict(diameter=1)
+    sim.operations.integrator = mc
+
+    sim.operations._schedule()
```
joaander, line 76:
```suggestion
    sim.run(0)
```
Let's use public APIs in tests where possible.

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,138 @@
...
     Args:
 
-        mc (:py:mod:`hoomd.hpmc.integrate`): HPMC integrator object for system on which to apply box updates
-        betaP (:py:class:`float` or :py:mod:`hoomd.variant`): :math:`\frac{p}{k_{\mathrm{B}}T}`. (units of inverse area in 2D or
-                                                    inverse volume in 3D) Apply your chosen reduced pressure convention
-                                                    externally.
         seed (int): random number seed for MC box changes
+        betaP (float or :py:mod:`hoomd.variant`): :math:`\frac{p}{k_{\mathrm{B}}T}`. (units of inverse area in 2D or
+                                                    inverse volume in 3D) Apply your chosen reduced pressure convention externally.
+        trigger (hoomd.trigger.Trigger): Select the timesteps to perform box trial moves.
```
joaander, line 27: Let's add a more extended paragaph length description of what BoxMC is between `Args` and `Attributes`. This should include a discussion of the weights.

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,138 @@
...
+            * ``weight`` (float) - relative weight of shear box moves relative to other box move types.
+            * ``delta`` (list or tuple) -  maximum change of the box tilt factor xy, xz, yz.
+            * ``reduce`` (float) - Maximum number of lattice vectors of shear to allow before applying lattice reduction.
+                     Shear of +/- 0.5 cannot be lattice reduced, so set to a value < 0.5 to disable (default 0)
+                     Note that due to precision errors, lattice reduction may introduce small overlaps which can be
+                     resolved, but which temporarily break detailed balance.
 
-        Note:
-            When an argument is None, the value is left unchanged from its current state.
+    Note:
```
joaander, line 64: With the weights mentioned in the description (see comment above), we should remove this whole `Note`.

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,138 @@
...
 
-        Example::
+        One or more Monte Carlo move types may be applied to evolve the simulation box. By default,
+        no moves are applied (all *delta* and *weight* values for all move types default to 0).
 
-            box_update.shear(delta=(0.01, 0.00, 0.0)) # 2D box changes
-            box_update.shear(delta=(0.01, 0.01, 0.01), weight=2)
-            box_update.shear(delta=(0.10, 0.01, 0.01), weight=0.15) # sample xy more aggressively
+        Pressure inputs to update.BoxMC are defined as :math:`\beta P`. Conversions from a specific definition of reduced
```
joaander, line 69: Pressure units are described in the tutorial and do not need to be mentioned here.

joaander (changes requested):
Thanks! Almost there, this needs a few additional documentation tweaks.

bdice (approved):
It looks like a few methods could become properties and also need tests. I recommend merging this PR soon and deferring those changes for a follow-up PR. If another reviewer agrees, let's create an issue for those changes and merge this.

Also noted a couple small typos that I'm going to fix directly on this branch.

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,141 @@
...
-            box_update = hpmc.update.boxmc(mc, betaP=10, seed=1)
-            run(100)
-            s_accept = box_update.get_shear_acceptance()
+        if not self.is_attached:
+            return None
+        else:
+            return self._cpp_obj.getCounters(1)
 
+    @log(flag="sequence")
+    def volume_moves(self):
```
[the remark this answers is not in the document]
b-butler: The `log` decorator makes methods properties by default. This is to prevent in nearly all cases,
```
@property 
@log 
def foo(self):
    pass
```

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,141 @@
...
+            * ``weight`` (float) - relative weight of aspect box moves relative to other box move types.
+            * ``delta`` (float) - maximum relative change of box aspect ratio.
+
+        length (dict):
+            Enable/disable isobaric box length moves and set parameters (change box lengths independently).
+            The dictionary has the following keys:
+
+            * ``weight`` (float) - maximum change of the box thickness for each pair of parallel planes
+                    connected by the corresponding box edges. I.e. maximum change of HOOMD-blue box parameters Lx, Ly, Lz.
+            * ``delta`` (list or tuple) -  maximum change of the box lengths Lx, Ly, Lz.
```
b-butler, line 63: Should document as just list.

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,141 @@
...
+
+        shear (dict):
+            Enable/disable isobaric box shear moves and set parameters. The dictionary
+            has the following keys:
+
+            * ``weight`` (float) - relative weight of shear box moves relative to other box move types.
+            * ``delta`` (list or tuple) -  maximum change of the box tilt factor xy, xz, yz.
```
b-butler, line 70: Should be just a tuple

hoomd/hpmc/update.py
```diff
@@ -15,387 +15,141 @@
...
-        boxMC.ln_volume(delta=0.01, weight=2.0)
-        boxMC.length(delta=(0.1,0.1,0.1), weight=4.0)
-        run(30) # perform approximately 10 volume moves and 20 length moves
-
+    Attributes:
+        volume (dict):
+            Enable/disable isobaric volume moves and set parameters (scale the box lengths uniformly).
+            The dictionary has the following keys:
+
+            * ``mode`` (string, **default:** ``standard``) - choose between ``standard`` for conventional volume
```
b-butler, line 45:
```suggestion
            * ``mode`` (str, **default:** ``standard``) - choose between ``standard`` for conventional volume
```

hoomd/hpmc/UpdaterBoxMC.h
```diff
@@ -35,98 +35,164 @@ class UpdaterBoxMC : public Updater
...
-            \param weight relative likelihood of volume move
+        //! Set parameters for box volume moves from a dictionary
+        /*! dict keys:
+             mode - choose between standard and log volume moves
+             delta - maximum size of volume change
+             weight - relative likelihood of volume move.
         */
-        void ln_volume(const Scalar delta_lnV,
-                       const float weight)
+        void setVolumeParams(pybind11::dict d)
```
b-butler, line 76: This also needs to zero out the weight of the other mode.

hoomd/hpmc/UpdaterBoxMC.h
```diff
@@ -35,98 +35,164 @@ class UpdaterBoxMC : public Updater
...
+        */
+        pybind11::dict getLengthParams()
+            {
+            pybind11::dict d;
+            d["weight"] = m_length_weight;
+            pybind11::list l;
+            l.append(m_length_delta[0]);
+            l.append(m_length_delta[1]);
+            l.append(m_length_delta[2]);
+            d["delta"] = l;
```
b-butler, line 111: This should be converted to a tuple.

hoomd/hpmc/UpdaterBoxMC.h
```diff
@@ -35,98 +35,164 @@ class UpdaterBoxMC : public Updater
...
+        */
+        pybind11::dict getShearParams()
+            {
+            pybind11::dict d;
+            d["weight"] = m_shear_weight;
+            pybind11::list l;
+            l.append(m_shear_delta[0]);
+            l.append(m_shear_delta[1]);
+            l.append(m_shear_delta[2]);
+            d["delta"] = l;
```
b-butler, line 148: Should be converted to a tuple

hoomd/hpmc/UpdaterBoxMC.h
```diff
@@ -214,65 +280,33 @@ class UpdaterBoxMC : public Updater
...
+        float m_volume_weight;                    //!< relative weight of volume moves
+        Scalar m_ln_volume_delta;                 //!< Amount by which to log volume parameter during box-change
+        float m_ln_volume_weight;                 //!< relative weight of log volume moves
+        std::string m_volume_mode;                //!< volume moves mode: standard or logarithmic
+        Scalar m_volume_A1;                       //!< Ratio of Lx to Ly to use in isotropic volume changes
+        Scalar m_volume_A2;                       //!< Ratio of Lx to Lz to use in isotropic volume changes
 
-        Scalar m_Length_delta[3];                   //!< Max length change in each dimension
-        float m_Length_weight;                     //!< relative weight of length change moves
+        Scalar m_length_delta[3];                  //!< Max length change in each dimension
```
b-butler, line 295: Is there a reason we are not using `Scalar3` @joaander?

b-butler (changes requested):
Overall looks good. I have a few comments that need addressing before merging though.

hoomd/hpmc/update.py
lyrivera: I'm assuming you meant tuple here?

lyrivera:
> It looks like a few methods could become properties and also need tests. I recommend merging this PR soon and deferring those changes for a follow-up PR. If another reviewer agrees, let's create an issue for those changes and merge this.
> 
> Also noted a couple small typos that I'm going to fix directly on this branch.

I'll add the tests here since they're simple enough.

hoomd/hpmc/update.py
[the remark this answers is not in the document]
bdice: In that case, should this be `log_property` or something like that? Most of the property-like decorators I know have property in the name, such as `abstractproperty`. The word “log” alone doesn’t convey the other crucial meaning here.

hoomd/hpmc/update.py
b-butler: Yes sorry.

hoomd/hpmc/update.py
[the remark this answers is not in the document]
b-butler: We could create a separate method `log_property`. Right now `log` accepts a key-word only argument `is_property` which defaults to `True`.

hoomd/hpmc/UpdaterBoxMC.h
joaander: The original author of the code wrote it this way. I think either way is fine, but at this point changing it to `Scalar3` also requires updating the for loops over dimensions in the implementation.

b-butler: That is fine then. I was wondering if there was a non-historic reason for the choice. This does not need to be addressed here. It would be nice to be consistent though throughout the code base on this eventually, but honestly that is pretty low priority.

lyrivera:
I added unit tests for the counters and addressed all the other comments.

joaander (approved):
I fixed some minor documentation formatting issues. This looks good!

b-butler (approved):
I added some final changes and formatting. LGTM.

b-butler:
\azp validate

PR #760, opened 2020-08-31
Fix/amd issues

jglaser:
## Description

Support compilation on AMD with more recent ROCm versions (>= 3.5.0), which are based on hip-clang. Select GPU target from CMake cached variable. Support more recent CMake (>3.14.0) with AMD.

## Motivation and Context

AMD support has degraded lately and needs to be updated. We should also revisit and update the AMD CI rig (I don't know what its status is currently).

Resolves: N/A

## How Has This Been Tested?

offline, on AMD hardware

## Change log

N/A

## Checklist:


joaander (approved):
Thanks!

The AMD test rig is offline because the images don't currently build with the custom HIP. Perhaps the use of a standard HIP version will work.

Please make any changes to the test rig on `feature/new-object-API` as it has significant non-merge-able changes from `next.

PR #762, opened 2020-09-01
Enable validation tests in pytest

joaander:
## Description

Add the ability to label validation tests in pytest with `pytest.mark.validate`. These tests are not run by default, pass the `--validate` command line argument to enable:
```
pytest hoomd --validate
```
Run only validation tests with the `-m` mark filter:
```
pytest hoomd --validate -m validate
```

## Motivation and Context

Longer running tests that ensure the validity of HOOMD do not need to be run on every commit. Keeping these longer runs out of the CI system keeps it from getting too busy.

## How Has This Been Tested?

I have applied the `validate` mark to the longer running `QuickCompress` tests. If this PR is functioning correctly, these tests should not run by default on CI, only after an explicit request is made.

## Checklist:


hoomd/conftest.py
```diff
@@ -196,6 +220,9 @@ def pytest_configure(config):
     config.addinivalue_line(
         "markers",
         "cupy_optional: tests that should pass with and without CuPy.")
+    config.addinivalue_line(
+        "markers",
+        "validate: Tests that perform longer validation tests.")
```
bdice, line 225: Slightly less redundant phrasing:
```suggestion
        "validate: Tests that perform long-running validations.")
```

Issue #696, opened 2020-07-28
Implement compute.Thermo


joaander:
## Purpose

Implement `compute.Thermo` - the Python interface to `ComputeThermo`. This new class provides loggable quantities for all of the computed properties. 

## Considerations

- We should consider whether `Thermo` should be in the `md` namespace or not.
- Should we use a more explicit name: `ThermodynamicQuantities`?

## Steps

- Implement `Thermo`
- Add tests to `hoomd/pytest/[md]/test_thermo.py`
- Add `test_custom_thermo.py` to `hoomd/[md]/pytest/CMakeLists.txt`

## Developer

When you start work on this task, please assign yourself and move it
to the *In progress* column in the associated project.

joaander:
When re-implementing, make the recommended documentation change: https://groups.google.com/g/hoomd-users/c/rpomGe8CKDI

PR #767, opened 2020-09-03
Compute Thermo ported to V3 API

tommy-waltmann:
## Description

This pull request ports `compute.thermo` to the new API where it is now `compute.ThermoQuantities`. Optionally this pull request could also handle porting `compute.thermoHMA` if needed

Everything is currently implemented in the `hoomd.compute` namespace, I thought we might want to have more discussion about whether or not to move this to `hoomd.md.compute`.

## Motivation and context

This needs to be merged before the V3 beta is released. Porting `compute.thermo` is important because most validation testing must wait until after this pull request is merged.

Resolves: #696 

## How has this been tested?

I have added some very basic tests of `compute.ThermoQuantities` in `hoomd/pytest/test_thermo.py`. The tests cover adding/removing this from a simulation object (since a compute is an operation) and tests that calling the loggable properties does not fail and returns None when it should.

Further validation testing of the thermo quantities could be done, but I wasn't sure whether to put that in this PR.

Do more tests need to be added to ensure a compute being an operation is implemented correctly?

## Change log

```
compute.thermo is now compute.ThermoQuantities
```

## Checklist:


hoomd/ComputeThermo.h
```diff
@@ -206,6 +206,93 @@ class PYBIND11_EXPORT ComputeThermo : public Compute
             return h_properties.data[thermo_index::potential_energy];
             }
 
+
+        // <----------------- Components of the pressure tensor --------------
+
+        Scalar getPressureXX()
```
joaander, line 212: Use one method to return the pressure tensor. You can either write a C++ method that returns the pressure tensor as a python list [xx, xy, xz, yy, yz, zz] or you can export the `PressureTensor` struct to Python and build that list in the Python property after calling the existing `getPressureTensor` method.

hoomd/compute.py
```diff
@@ -12,139 +12,34 @@
...
-            else:
-                hoomd.context.current.device.cpp_msg.error("Restoring state from {reader_name} is not currently supported for {name}\n".format(reader_name=hoomd.context.current.state_reader.__name__, name=self.__class__.__name__));
-            raise RuntimeError("Can not restore state information!");
 
-
-# set default counter
-_compute.cur_id = 0;
-
-class thermo(_compute):
+class ThermoQuantities(_Thermo):
```
joaander, line 29: Let use full words in class names: `ThermodynamicQuantities`

hoomd/compute.py
```diff
@@ -12,139 +12,34 @@
...
     Args:
-        group (``hoomd.group``): Group to compute thermodynamic properties for.
+        filter (``hoomd.filter``): Particle filter to compute thermodynamic properties for.
 
-    :py:class:`hoomd.compute.thermo` acts on a given group of particles and calculates thermodynamic properties of those particles when
-    requested. A default :py:class:`hoomd.compute.thermo` is created that operates on the group of all particles. Integration methods
-    such as ``hoomd.md.integrate.nvt`` automatically create an internal :py:class:`hoomd.compute.thermo` for the group that they operate on.
-    If thermodynamic properties are needed on additional groups, a user can specify additional :py:class:`hoomd.compute.thermo` commands.
+    :py:class:`hoomd.compute.ThermoQuantities` acts on a given group of particles and calculates thermodynamic properties of those particles when
+    requested. A default :py:class:`hoomd.compute.ThermoQuantities` is created that operates on the group of all particles. Integration methods
```
joaander, line 36: A default compute is no longer created, remove this sentence. You also don't need to mention that integrators internally create these - that is an implementation detail the user doesn't need to know.

hoomd/compute.py
```diff
@@ -12,139 +12,34 @@
...
+    such as ``hoomd.md.integrate.NVT`` automatically create an internal :py:class:`hoomd.compute.ThermoQuantities` for the group that they operate on.
+    If thermodynamic properties are needed on additional groups, a user can specify additional :py:class:`hoomd.compute.ThermoQuantities` commands.
 
     Whether they are automatically created or created by a user, all specified thermos are available for logging via
```
joaander, line 40: Really, this whole documentation just needs to be re-written. It probably just needs a few sentences that state the class computes thermodynamic quantities and interoperates with the `Logger`.

hoomd/compute.py
```diff
@@ -191,71 +86,148 @@ class thermo(_compute):
...
+        if isinstance(self._simulation.device, hoomd.device.CPU):
+            thermo_cls = _hoomd.ComputeThermo
+        else:
+            thermo_cls = _hoomd.ComputeThermoGPU
+        group = self._simulation.state.get_group(self._filter)
+        self._cpp_obj = thermo_cls(self._simulation.state._cpp_sys_def, group, "")
+        super()._attach()
+
+    @log
+    def temperature(self):
```
joaander, line 106:
```suggestion
    def kinetic_temperature(self):
```

hoomd/compute.py
```diff
@@ -191,71 +86,148 @@ class thermo(_compute):
...
+        if self._attached:
+            self._cpp_obj.compute(self._simulation.timestep)
+            return self._cpp_obj.getPressure()
         else:
-            self.cpp_compute = _hoomd.ComputeThermoGPU(hoomd.context.current.system_definition, group.cpp_group, suffix);
+            return None
 
-        hoomd.context.current.system.addCompute(self.cpp_compute, self.compute_name);
+    @log
+    def pressureXX(self):
```
joaander, line 122: Define just one `pressure_tensor` method.

hoomd/compute.py
```diff
@@ -191,71 +86,148 @@ class thermo(_compute):
...
+    def pressureZZ(self):
+        if self._attached:
+            self._cpp_obj.compute(self._simulation.timestep)
+            return self._cpp_obj.getPressureZZ()
+        else:
+            return None
 
-        A disabled thermo compute can be re-enabled with :py:meth:`enable()`.
-        """
+    @log
```
joaander, line 169: @b-butler Are the default log parameters appropriate, or should this explicitly state flags here?

hoomd/compute.py
```diff
@@ -191,71 +86,148 @@ class thermo(_compute):
...
+    def kineticEnergy(self):
```
joaander, line 170:
```suggestion
    def kinetic_energy(self):
```
And similarly use PEP8 names throughout.

hoomd/compute.py
```diff
@@ -191,71 +86,148 @@ class thermo(_compute):
...
+    def potentialEnergy(self):
+        if self._attached:
+            self._cpp_obj.compute(self._simulation.timestep)
+            return self._cpp_obj.getPotentialEnergy()
+        else:
+            return None
 
-        Examples::
+    @log
+    def NDOF(self):
```
joaander, line 202: Let's not abbreviate: Use `degrees_of_freedom`

hoomd/pytest/test_thermo.py
```diff
@@ -0,0 +1,45 @@
...
+    for qty in _thermo_qtys:
+        calc_qty = getattr(thermo, qty)
+        assert type(calc_qty) == float
+
+    # detach from simulation and test properties again
+    sim.operations.remove(thermo)
+    assert len(sim.operations.computes) == 0
+    for qty in _thermo_qtys:
+        assert getattr(thermo, qty) == None
+
```
joaander, line 45: Additional testing would be great! Please add tests that define specific particle systems and check that the properties all return the correct values.

joaander (changes requested):
Thanks! This is a great start! I left some specific change requests and comments.

Let's keep PRs short and leave ThermoHMA for a separate PR after this one.

joaander:
Also - yes please move `ThermodynamicQuantities` to `hoomd.md.compute`. This will prevent user confusion by making it clear that this computes quantities for MD simulations and not MC.

hoomd/ComputeThermo.h
b-butler: I think returning the Python list will be cleanest.

hoomd/ComputeThermo.h
```diff
@@ -241,6 +328,28 @@ class PYBIND11_EXPORT ComputeThermo : public Compute
             return p;
             }
 
+        // <--------------- Degree of Freedom Data
+
+        Scalar getNDOF()
+            {
+            return Scalar(m_group->getTranslationalDOF() + m_group->getRotationalDOF());
+            }
```
b-butler, line 336: We could also just make this a Python property. Up to you though.

hoomd/ComputeThermo.cc
```diff
@@ -416,6 +416,16 @@ void export_ComputeThermo(py::module& m)
...
+    .def("getPressureXX", &ComputeThermo::getPressureXX)
+    .def("getPressureXY", &ComputeThermo::getPressureXY)
+    .def("getPressureXZ", &ComputeThermo::getPressureXZ)
+    .def("getPressureYY", &ComputeThermo::getPressureYY)
+    .def("getPressureYZ", &ComputeThermo::getPressureYZ)
+    .def("getPressureZZ", &ComputeThermo::getPressureZZ)
+    .def("getNDOF", &ComputeThermo::getNDOF)
+    .def("getTranslationalDOF", &ComputeThermo::getTranslationalDOF)
+    .def("getRotationalDOF", &ComputeThermo::getRotationalDOF)
+    .def("getNumParticles", &ComputeThermo::getNumParticles)
```
b-butler, line 428: We could make some of these pybind properties so we don't have to use `()` everywhere. That may be a bit cleaner as all of these are simple quantities that don't take arguments.

hoomd/operations.py
```diff
@@ -13,7 +13,7 @@ def _triggered_op_conversion(value):
 class Operations:
     def __init__(self, simulation=None):
         self._simulation = simulation
-        self._compute = list()
+        self._computes = SyncedList(OnlyType(_Compute), lambda x: x._cpp_obj)
```
b-butler, line 16: Can we move this down a line to the other synced lists.

hoomd/compute.py
b-butler: Agreed.

hoomd/compute.py
b-butler: The default is `scalar` so this should be fine.

hoomd/compute.py
```diff
@@ -191,71 +86,148 @@ class thermo(_compute):
...
+    def numParticles(self):
+        if self._attached:
+            return self._cpp_obj.getNumParticles()
+        else:
+            return None
 
-        hoomd.context.current.thermo.append(self)
 
-class thermoHMA(_compute):
+class thermoHMA(_Compute):
```
b-butler, line 230: If we are going to change the parent class for this class then we can just remove the `_compute` class.

hoomd/compute.py
b-butler: :+1:

b-butler (changes requested):
Thanks @tommy-waltmann for the code. You managed to get a negative PR in even with these changes. I agree with Josh's current comments, and made a few of my own.

hoomd/ComputeThermo.h
tommy-waltmann: Why use just one method to get pressure tensor components? Also, we still want to expose each component of the pressure tensor individually as its own loggable quantity, right?

hoomd/compute.py
tommy-waltmann: I'm not completely up to date on the newest abilities of the logger, but won't logging a list quantity be a problem?

hoomd/ComputeThermo.cc
tommy-waltmann: I don't have a strong opinion either way. Once I make the changes for the pressure tensor, there will be one quantity that is not simple anymore.

hoomd/compute.py
tommy-waltmann: I left this because there are classes in hpmc that still have `_compute` as their base class, but I can change those too.

hoomd/ComputeThermo.h
joaander: Because the pressure tensor is a single "object" that should be gettable with one name. The elements of that tensor are naturally accessed by index.

hoomd/compute.py
joaander: No, the new logger can handle arbitrary data quantities, up to and including large numpy arrays with the number of particles in the system or larger. A single 6-element tensor is trivially handled.

joaander:
You'll need to update the `ComputeThermoHMAGPU` files with the new ones from `feature/new-object-API`. There wasn't a need to move the C++ files in order to reorganized the Python namespace.

tommy-waltmann:
> 
> 
> You'll need to update the `ComputeThermoHMAGPU` files with the new ones from `feature/new-object-API`. There wasn't a need to move the C++ files in order to reorganized the Python namespace.

I may need to move those C++ files back to integrate the changes from `feature/new-object-API`, but they should end up in the md/ directory when this PR is merged, right?

joaander:
There does not need to be a 1:1 correspondence between python namespaces what directories C++ class files are stored in.

hoomd/compute.py
tommy-waltmann: I think we should keep all the documentation related to the quantities and the put them underneath their properties. The class documentation can be shortened significantly.

hoomd/md/compute.py
```diff
@@ -0,0 +1,276 @@
...
+
+        Calculated as
+
+          .. math::
+
+              W = \frac{1}{2} \sum_{i}\sum_{j \ne i} \vec{F}_{ij} \cdot \vec{r_{ij}} + \sum_{k} \vec{F}_{k} \cdot \vec{r_{k}}
+
+          where :math:`\vec{F}_{ij}` are pairwise forces between particles and :math:`\vec{F}_k` are forces due to explicit constraints, implicit rigid
+          body constraints, external walls, and fields. In 2D simulations,
+          :math:`P = (K + \frac{1}{2}\cdot W)/A` where :math:`A` is the area of the simulation box.
```
joaander, line 85: Also list the `P` expression for 3D boxes. Or better yet, list one general expression that applies to both 2D and 3D.

hoomd/md/compute.py
```diff
@@ -0,0 +1,276 @@
...
+    def pressure_tensor(self):
+        R"""
+        (:math:`P_{xx}`, :math:`P_{xy}`, :math:`P_{xz}`, :math:`P_{yy}`, :math:`P_{yz}`, :math:`P_{zz}`).
+
+        Instantaneous pressure tensor of the group (in pressure units).
+
+          .. math::
+
+              P_{ij} = \left[  \sum_{k\in[0..N)} m_k v_{k,i} v_{k,j} +
+                               \sum_{k\in[0..N)} \sum_{l > k} \frac{1}{2} \left(\vec{r}_{kl,i} \vec{F}_{kl,j} + \vec{r}_{kl,j} \vec{F}_{kl, i} \right) \right]/V
```
joaander, line 104: Add: Where `V` is the total simulation box volume (or area in 2D).

hoomd/md/compute.py
```diff
@@ -0,0 +1,276 @@
...
+        R""" :math:`K`, total kinetic energy of all particles in the group (in energy units). """
+        if self._attached:
+            self._cpp_obj.compute(self._simulation.timestep)
+            return self._cpp_obj.kinetic_energy
+        else:
+            return None
+
+    @log
+    def translational_kinetic_energy(self):
+        R""" :math:`K_{\mathrm{trans}}`, translational kinetic energy of all particles in the group (in energy units). """
```
joaander, line 123: Add a forumla to make the computation explicit.

hoomd/md/compute.py
```diff
@@ -0,0 +1,276 @@
...
+        if self._attached:
+            self._cpp_obj.compute(self._simulation.timestep)
+            return self._cpp_obj.translational_kinetic_energy
+        else:
+            return None
+
+    @log
+    def rotational_kinetic_energy(self):
+        R""" :math:`K_{\mathrm{rot}}`, rotational kinetic energy of all particles in the group (in energy units). """
```
joaander, line 132: Add a formula to make the computation explicit.

hoomd/md/compute.py
```diff
@@ -0,0 +1,276 @@
...
+        """
+        if self._attached:
+            self._cpp_obj.compute(self._simulation.timestep)
+            return self._cpp_obj.pressure_tensor
+        else:
+            return None
+
+    @log
+    def kinetic_energy(self):
+        R""" :math:`K`, total kinetic energy of all particles in the group (in energy units). """
```
joaander, line 114: Add a formula to make this computation explicit (K is the sum of K_trans and K_rot).

hoomd/md/compute.py
```diff
@@ -0,0 +1,276 @@
...
+        R""" :math:`U`, potential energy that the group contributes to the entire system (in energy units). """
+        if self._attached:
+            self._cpp_obj.compute(self._simulation.timestep)
+            return self._cpp_obj.potential_energy
+        else:
+            return None
+
+    @log
+    def degrees_of_freedom(self):
+        R""" :math:`N_{\mathrm{dof}}`, number of degrees of freedom given to the group by its integration method. """
```
joaander, line 150: Note that this is the sum of transnational and rotational dof.

hoomd/md/compute.py
```diff
@@ -0,0 +1,276 @@
...
+class ThermodynamicQuantities(_Thermo):
+    R""" Compute thermodynamic properties of a group of particles.
+
+    Args:
+        filter (``hoomd.filter``): Particle filter to compute thermodynamic properties for.
+
+    :py:class:`ThermodynamicQuantities` acts on a given group of particles and calculates thermodynamic properties of
+    those particles when requested. All specified :py:class:`ThermodynamicQuantities` objects can be added to a logger
+    for logging during a simulation, see :py:class:`hoomd.logging.Logger` for more details.
+
```
joaander, line 36: Let's add a note here about pairwise interactions:
```
Note:

    For pairwise terms U_ij and F_ij \cdot r_ij), HOOMD-blue assigns half of each term to a per particle quantity. When *filter* selects a subset of the particles in the system, `ThermoDynamicQuantities` sums these per-particle assignments over only particles selected by the filter.
```
What do people think? Is that clear enough, or should we instead provide a full LaTeX math equation for U and P that include this explicitly?

hoomd/md/compute.py
```diff
@@ -0,0 +1,276 @@
...
+    def degrees_of_freedom(self):
+        R""" :math:`N_{\mathrm{dof}}`, number of degrees of freedom given to the group by its integration method. """
+        if self._attached:
+            return self._cpp_obj.degrees_of_freedom
+        else:
+            return None
+
+    @log
+    def translational_degrees_of_freedom(self):
+        R""" :math:`N_{\mathrm{dof, trans}}`, number of translational degrees of freedom given to the group by its integration method. """
```
joaander, line 158: Add a note:
```
When using a single integration method that is momentum conserving that operations on all particles, `N_{\mathrm{dof, trans}} = DN - D - N_constraints` where D is the dimensionality of the system.

Note:
    The removal of `D` degrees of freedom accounts for the pinned center of mass in periodic boundary conditions. When the *filter* in `ThermoDynamicQuantities` selects a subset of all particles, the removed degrees of freedom are spread proportionately.

File: hoomd/md/pytest/test_thermo.py [hunk 23]
```diff
@@ -0,0 +1,140 @@
...
+    ('rotational_degrees_of_freedom', int),
+    ('num_particles', int),
+]
+
+def test_attach_detach(simulation_factory, two_particle_snapshot_factory):
+    # test before attaching to simulation
+    group = hoomd.filter.All()
+    thermo = hoomd.md.compute.ThermodynamicQuantities(group)
+    for qty, typ in _thermo_qtys:
+        assert getattr(thermo, qty) == None
```
joaander, line 24:
```suggestion
        assert getattr(thermo, qty) is None
```

hoomd/md/pytest/test_thermo.py
```diff
@@ -0,0 +1,140 @@
...
+    # make sure quantities are computable without failure
+    for qty, typ in _thermo_qtys:
+        calc_qty = getattr(thermo, qty)
+        assert type(calc_qty) == typ
+
+    # detach from simulation and test properties again
+    sim.operations.remove(thermo)
+    assert len(sim.operations.computes) == 0
+    for qty, typ in _thermo_qtys:
+        assert getattr(thermo, qty) == None
```
joaander, line 41:
```suggestion
        assert getattr(thermo, qty) is None
```

hoomd/md/pytest/test_thermo.py
```diff
@@ -0,0 +1,140 @@
...
+    sim.operations.schedule()
+    sim.run(1)
+
+    assert thermo.num_particles == 2
+    assert thermo.rotational_degrees_of_freedom == 0
+    assert thermo.translational_degrees_of_freedom == 3
+    assert thermo.degrees_of_freedom == 3
+
+    np.testing.assert_allclose(thermo.potential_energy, 0.0)
+    #np.testing.assert_allclose(thermo.rotational_kinetic_energy, 0.0, rtol=1e-5)
```
joaander, line 66: We should check that rotational_kinetic_energy is computed correctly here. I think it returns NaN now. It should return 0. We should also add a test that adds moment of inertia and angular momentum to particles to check that it computes a non-zero rotational energy.

hoomd/md/pytest/test_thermo.py
```diff
@@ -0,0 +1,140 @@
...
+    np.testing.assert_allclose(thermo.translational_kinetic_energy, 4.0, rtol=1e-5)
+    np.testing.assert_allclose(thermo.kinetic_energy, 4.0, rtol=1e-5)
+    np.testing.assert_allclose(thermo.kinetic_temperature, 2*thermo.kinetic_energy/thermo.degrees_of_freedom, rtol=1e-5)
+    np.testing.assert_allclose(thermo.pressure, 2.0/3*thermo.kinetic_energy/20**3, rtol=1e-5)
+    (pxx, pxy, pxz, pyy, pyz, pzz) = thermo.pressure_tensor
+    np.testing.assert_allclose(pxx, 8.0/20.0**3, rtol=1e-5)
```
joaander, line 72: Use one assert_allclose on the full vector tensor to make this more readable.

hoomd/mpcd/integrate.py
```diff
@@ -103,7 +103,7 @@ class slit(_bounce_back):
     This method is the MD analog of :py:class:`.stream.slit`, which documents additional details
     about the geometry.
 
-    A :py:class:`hoomd.compute.thermo` is automatically specified and associated with *group*.
+    A :py:class:`hoomd.md.compute.ThermodynamicQuantities` is automatically specified and associated with *group*.
```
joaander, line 106: Remove this. It is not true.

hoomd/mpcd/integrate.py
```diff
@@ -177,7 +177,7 @@ class slit_pore(_bounce_back):
     This method is the MD analog of :py:class:`.stream.slit_pore`, which documents additional details
     about the geometry.
 
-    A :py:class:`hoomd.compute.thermo` is automatically specified and associated with *group*.
+    A :py:class:`hoomd.md.compute.ThermodynamicQuantities` is automatically specified and associated with *group*.
```
joaander, line 180: Remove this as well.

sphinx-doc/module-md-compute.rst
```diff
@@ -1,15 +1,17 @@
...
 
-    hoomd.compute.thermo
+    ThermodynamicQuantities
 
 .. rubric:: Details
 
-.. automodule:: hoomd.compute
+.. automodule:: hoomd.md.compute
     :synopsis: Compute system properties.
     :members:
```
joaander, line 17:
```suggestion
    :members: ThermodynamicQuantities
```
This will hide the unimplemented thermoHMA.

joaander (changes requested):
Thanks, Tommy! This is in really good shape now. I have several suggested improvements to the documentation and a few additional test requests. That is all we need to finish this off.

Another additional test we need is that a Logger can grab all loggable quantities before and after attaching (use the no defaults option in the test).

hoomd/md/compute.py
tommy-waltmann: I'm trying to look through the source code to figure out the formula myself, can you clarify what exactly is happening around line 280 of `md/ComputeThermo.cc`? I don't understand what the variable `s` is or why `p` is a quaternion.

hoomd/md/compute.py
tommy-waltmann: I think the math equation would make it clearer what is happening, especially why it assigns half of each term to a per particle quantity. Do users on the email list ask about this often?

joaander: No, it is not a common question, but it is good to be explicit.

hoomd/md/compute.py
joaander: This is `1/2(L_x**2/I_x + L_y**2 / I_y+ L_z**2 / I_z)` where I is the moment of inertia and L is the angular momentum in the (diagonal) reference frame of the particle.

hoomd/md/compute.py
joaander: OK. In the pressure equations we can note that the i and j indicies are particle tags and write the sums over `k \in filter`.

For the potential energy U, we would write a similar expression but break it into components. `U=U_pair + U_bond + U_angle + ...` and then explicitly document the expressions for the terms. Something like `U_pair = sum over i \in filter 1/2 V_ij`

tommy-waltmann: I don't quite understand what you mean for the pressure (`P`) equations. Why would we write the sums over `k` (and presumably `k'`) if we want to document the usage of `i` and `j`?

tommy-waltmann: You are talking about the sums in the equation for `W`, right?

joaander: I was speaking in general terms. All equations that sum over pairs will have some indices that run over the filtered subset and others that run over all particles.

hoomd/md/compute.py
```diff
@@ -0,0 +1,344 @@
...
+        """
+        if self._attached:
+            self._cpp_obj.compute(self._simulation.timestep)
+            return self._cpp_obj.potential_energy
+        else:
+            return None
+
+    @log
+    def degrees_of_freedom(self):
+        """
```
joaander, line 206: Check this with pydocstyle. You'll find a number of errors that need addressing, like newlines and spaces immediately after """.

Please ignore any pydocstyle errors in thermoHMA - we'll fix those when we update that class.

hoomd/md/compute.py
```diff
@@ -0,0 +1,344 @@
...
+        :math:`N_{\\mathrm{dof}}`, number of degrees of freedom given to the group by its integration method.
+
+        Note:
```
joaander, line 209: For consistency with the rest of the docs here you can replace this note with an equation.

hoomd/md/compute.py
```diff
@@ -0,0 +1,344 @@
...
+    @log
+    def pressure(self):
+        """
+        :math:`P`, instantaneous pressure of the group (in pressure units).
+
+        Calculated as:
+
+        .. math::
+
+            P = \\left(\\frac{2}{D} \\cdot K_{\\mathrm{trans}} + \\frac{1}{2} \\cdot W \\right) / V,
```
joaander, line 80:
```suggestion
            P = \\left(\\frac{2}{D} \\cdot K_{\\mathrm{trans}} + \\frac{1}{D} \\cdot W \\right) / V,
```
Or factor out the 1/D.

hoomd/md/compute.py
```diff
@@ -0,0 +1,344 @@
...
+
+        where :math:`D` is the dimensionality of the system, :math:`V` is the total volume of the simulation box (or area in 2D), and :math:`W` is calculated as:
+
+        .. math::
+
+            W = \\frac{1}{2} \\sum_{i} \\sum_{j \\ne i} \\vec{F}_{ij} \\cdot \\vec{r_{ij}} + \\sum_{k} \\vec{F}_{k} \\cdot \\vec{r_{k}},
```
joaander, line 86:
```suggestion
            W = \\frac{1}{2} \\sum_{i \\in \\mathrm{filter}} \\sum_{j} \\vec{F}_{ij} \\cdot \\vec{r_{ij}} + \\sum_{k} \\vec{F}_{k} \\cdot \\vec{r_{k}},
```

hoomd/md/compute.py
```diff
@@ -0,0 +1,344 @@
...
+    @log(flag='sequence')
+    def pressure_tensor(self):
+        """
+        (:math:`P_{xx}`, :math:`P_{xy}`, :math:`P_{xz}`, :math:`P_{yy}`, :math:`P_{yz}`, :math:`P_{zz}`).
+
+        Instantaneous pressure tensor of the group (in pressure units), calculated as:
+
+          .. math::
+
+              P_{ij} = \\left[  \\sum_{k\\in[0..N)} m_k v_{k,i} v_{k,j} +
```
joaander, line 105: This also needs to be updated with the outer particle sums over particles in the filter and the inner sum running over all particles.

hoomd/md/pytest/test_thermo.py
joaander: What I mean is:
```
numpy.testing.assert_allclose(thermo.pressure_tensor, [8/20**3, 0, 0, 0, 0, 0])
```

hoomd/md/compute.py
```diff
@@ -0,0 +1,344 @@
...
+
+        Examples::
+
+            my_thermo.enable()
+
+        See ``disable``.
+        """
+        _compute.enable(self)
+
+## \internal
```
joaander, line 331: Remove `_get_unique_thermo`. It is no longer used.

joaander (changes requested):
I clarified my previous comments.

joaander (approved):
Thanks!

hoomd/md/pytest/test_thermo.py
```diff
@@ -0,0 +1,166 @@
...
+    np.testing.assert_allclose(thermo.rotational_kinetic_energy, 0.0, rtol=1e-5)
+    np.testing.assert_allclose(thermo.translational_kinetic_energy, 4.0, rtol=1e-5)
+    np.testing.assert_allclose(thermo.kinetic_energy, 4.0, rtol=1e-5)
+    np.testing.assert_allclose(thermo.kinetic_temperature, 2*thermo.kinetic_energy/thermo.degrees_of_freedom, rtol=1e-5)
+    np.testing.assert_allclose(thermo.pressure, 2.0/3*thermo.kinetic_energy/20**3, rtol=1e-5)
+    np.testing.assert_allclose(thermo.pressure_tensor, [8.0/20.0**3, 0., 0., 0., 0., 0.], rtol=1e-5)
+
+
+def test_basic_system_2d(simulation_factory, lattice_snapshot_factory):
+    group = hoomd.filter.Type(['A'])
```
b-butler, line 75: If we are going to name the other filter `groupB` this should be `groupA` although I prefer the names `filter_A` and `filter_B`.

hoomd/md/pytest/test_thermo.py
```diff
@@ -0,0 +1,166 @@
...
+    assert thermoB.rotational_degrees_of_freedom == 0
+    assert thermoB.translational_degrees_of_freedom == 4
+    assert thermoB.degrees_of_freedom == 4
+    np.testing.assert_allclose(thermoB.potential_energy, 0.0)
+    np.testing.assert_allclose(thermoB.rotational_kinetic_energy, 0.0, rtol=1e-5)
+    np.testing.assert_allclose(thermoB.translational_kinetic_energy, 4.0, rtol=1e-3)
+    np.testing.assert_allclose(thermoB.kinetic_energy, 4.0, rtol=1e-3)
+    np.testing.assert_allclose(thermoB.kinetic_temperature, 2*thermoB.kinetic_energy/thermoB.degrees_of_freedom, rtol=1e-3)
+    np.testing.assert_allclose(thermoB.pressure, thermoB.kinetic_energy/2.0**2, rtol=1e-3)
+    np.testing.assert_allclose(thermoB.pressure_tensor, (8.0/2.0**2, 0., 0., 0., 0., 0.), rtol=1e-3, atol=1e-1)
```
b-butler, line 120: It may be worth abstracting this into a function since it is the same check with just different values. I am fine either way though.

hoomd/md/pytest/test_thermo.py
```diff
@@ -0,0 +1,166 @@
...
+    sim = simulation_factory(snap)
+    sim.always_compute_pressure = True
+    sim.operations.add(thermo)
+
+    integrator = hoomd.md.Integrator(dt=0.0001)
+    integrator.aniso = True
+    integrator.methods.append(hoomd.md.methods.NVT(group, tau=1, kT=1))
+    sim.operations.integrator = integrator
+
+    sim.operations.schedule()
```
b-butler, line 148: This call is unnecessary.

hoomd/md/pytest/test_thermo.py
```diff
@@ -0,0 +1,166 @@
...
+    sim.always_compute_pressure = True
+    sim.operations.add(thermo)
+    sim.operations.add(thermoB)
+
+    integrator = hoomd.md.Integrator(dt=0.0001)
+    integrator.methods.append(hoomd.md.methods.NVT(group, tau=1, kT=1))
+    integrator.methods.append(hoomd.md.methods.Langevin(groupB, kT=1, seed=3))
+    sim.operations.integrator = integrator
+
+    sim.operations.schedule()
```
b-butler, line 93: unnecessary.

hoomd/md/pytest/test_thermo.py
```diff
@@ -0,0 +1,166 @@
...
+        snap.particles.velocity[:] = [[-2, 0, 0], [2, 0, 0]]
+    sim = simulation_factory(snap)
+    sim.always_compute_pressure = True
+    sim.operations.add(thermo)
+
+    integrator = hoomd.md.Integrator(dt=0.0001)
+    integrator.methods.append(hoomd.md.methods.NVT(group, tau=1, kT=1))
+    sim.operations.integrator = integrator
+
+    sim.operations.schedule()
```
b-butler, line 57: Method call unnecessary.

b-butler (changes requested):
There are still references to `compute.thermo` in `hoomd/md/ComputeThermo.cc` and `hoomd/mpcd/integrate.py` the later is less important.

A brief aside. It is hard to review some of the documentation as the lines are too long in GitHub to view accurately. If you could wrap lines to cap at 80 or 100 that would help reviewing this code and also just make the experience nicer on some editors like vim or emacs.

Besides these comments, and the specific ones I added, the PR looks good.

tommy-waltmann:
> 
> 
> There are still references to `compute.thermo` in `hoomd/md/ComputeThermo.cc` and `hoomd/mpcd/integrate.py` the later is less important.
> 
> A brief aside. It is hard to review some of the documentation as the lines are too long in GitHub to view accurately. If you could wrap lines to cap at 80 or 100 that would help reviewing this code and also just make the experience nicer on some editors like vim or emacs.
> 
> Besides these comments, and the specific ones I added, the PR looks good.

Can you point to the line in `ComputeThermo.cc` where `compute.thermo` is mentioned? I've been through the code and I can't see where it is.

b-butler:
Line 181 in `hoomd/md/ComputeThermo.cc` and line 55 in `hoomd/mpcd/integrate.py`.

hoomd/md/compute.py
```diff
@@ -22,11 +22,14 @@ class ThermodynamicQuantities(_Thermo):
     """Compute thermodynamic properties of a group of particles.
 
     Args:
-        filter (``hoomd.filter``): Particle filter to compute thermodynamic properties for.
+        filter (``hoomd.filter``): Particle filter to compute thermodynamic
```
b-butler, line 25:
```suggestion
        filter (`hoomd.filter._ParticleFilter`): Particle filter to compute thermodynamic
```

b-butler (approved):
LGTM

joaander:
Thanks, all!

PR #776, opened 2020-09-05
New Data Structures for HOOMD objects

b-butler:
## Description

This fleshes out HOOMD's data structures that sync with C++. It adds classes that can inform parents on modification. This allows us to do things like `lj.params[('A', 'A')].epsilon =4` and expect it to work correctly.

The main caveat here is that upon attaching the data structures may not always be up to date if C++ changes them (in general it shouldn't and this is only for complex structures like lists, dictionaries, and sets). However, the setting of these parameters using the new data structures still works as expected even if it isn't necessarily up to date.

We rely on Python's reference system to only every expose a user to the exact same data structure object to represent the same data. This may seem obvious, but it is required to allow us to break the connection of internal data structures from them parent's to prevent erroneous updates.

## Motivation and context

I was motivated by #758 and conversations had with the development team. Since the beta release will be rolling out soon, I thought a more robust system for handling HOOMD object's attributes was in order.

## How has this been tested?

Existing tests pass, but more should be added.

## Checklist:


b-butler:
I am having some documentation build warnings to due with `MutableMapping`. I am not sure how to solve them currently.

joaander (commented):
I approve of this concept in general. I leave a detailed review of how it is implemented in Python to others.

I do wonder if there are better names for HOOMDDict, HOOMDList, and HOOMDList.

joaander:
@vyasr @bdice I would appreciate your review.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+    def _update(self):
+        """Signal that the object has been updated."""
+        if self._buffered:
+            return
+        elif parent is not None:
+            self._parent._handle_update(self, self._label)
+
+
+def _get_inner_typeconverter(type_def, desired_type):
+    """Handles more complicated data structure specification.
```
bdice, line 44: Could this be made more descriptive?

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+
+    When using nested containers where an entry could be a list or None for
+    instance requires the use of something like typeconverter's OnlyIf or
+    Either. We autdetect the internal type specification for these object and if
+    they are of the correct type use these.
```
bdice, line 49: I'm having trouble parsing what is meant here. I tried writing a suggestion but I don't know how to resolve the antecedents of "them" / "these" / etc.
```suggestion
    When using nested containers, where an entry could be a list or None for
    instance requires the use of something like typeconverter's OnlyIf or
    Either. We detect the internal type specification for these objects and if
    they are of the correct type, use these.
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+
+    Args:
+        type_def (`hoomd.data.typeconverter.TypeConverter`):
+            The type converter which is desired to be introspected to find the
+            desired type.
```
bdice, line 54:
```suggestion
            Type converter which is introspected to find the desired type.
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+        desired_type (`hoomd.data.typeconverter.TypeConverter`):
+            A child class of `hoomd.data.typeconverter.TypeConverter` that
+            signals what type converter types is needed.
```
bdice, line 57: type converter types? I'm unsure what this means.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+    """
+    rtn_type_def = None
```
bdice, line 59: Use descriptive variable names like `return_type_def` -- I'm deleting this line and defining it in another suggestion below.
```suggestion
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+    # Simple case the type_def is already sufficient
```
bdice, line 60:
```suggestion
    # If the type_def is the desired type, return it immediately.
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+    if isinstance(type_def, desired_type):
+        rtn_type_def = type_def
```
bdice, line 62:
```suggestion
        return type_def
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+                if len(matches) > 1:
+                    raise ValueError(
+                        "Parameter defined with multiple valid definitions of "
+                        "the same type.")
+                elif len(matches) == 1:
+                    rtn_type_def = matches[0]
+    # When no valid type converter can be found error out.
+    if rtn_type_def is None:
+        raise ValueError("Unable to find type definition for attribute.")
+    return rtn_type_def
```
bdice, line 86: I would try to separate out the comments so that "multiple valid container representations" or "allowed to be None" align with the checks where those are implemented. Part of the challenge is that I don't know what the `OnlyIf` or `Either` classes are meant to do. Reading their docstrings didn't help, the `OnlyIf` docstring mentions "Not strictly necessary, but keeps the theme of the other classes, and allows pre/post-processing and optionally allows None." Is that class meant to act like [typing.Optional](https://docs.python.org/3/library/typing.html#typing.Optional)?
```suggestion
    # More complicated cases, deals with nested type definitions where the
    # nested containers have other requirements such as multiple valid container
    # representations or entry is allowed to be None. In these cases, we check
    # the TypeConverterValue instance's converter for the desired container type
    # converter type.
    return_type_def = None
    if isinstance(type_def, TypeConverterValue):
        if isinstance(type_def.converter, OnlyIf):
            if isinstance(type_def.converter.cond, desired_type):
                return_type_def = type_def.converter.cond  # Can this be None? If not, return here.
        elif isinstance(type_def.converter, Either):
            matches = [spec for spec in type_def.converter.specs if isinstance(spec, desired_type)]
            if len(matches) > 1:
                raise ValueError(
                    "Parameter defined with multiple valid definitions of "
                    "the same type.")
            elif len(matches) == 1:
                return_type_def = matches[0]
    # When no valid type converter can be found error out.
    if return_type_def is None:
        raise ValueError("Unable to find type definition for attribute.")
    return return_type_def
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+    elif isinstance(type_def, TypeConverterValue):
+        if isinstance(type_def.converter, OnlyIf):
+            if isinstance(type_def.converter.cond, desired_type):
+                rtn_type_def = type_def.converter.cond
+        elif isinstance(type_def.converter, Either):
+            for spec in type_def.converter.specs:
+                matches = []
+                if isinstance(spec, desired_type):
+                    matches.append(spec)
+                if len(matches) > 1:
```
bdice, line 77: This can't happen, check the logic -- `matches = []` happens inside the for loop. I attempt to resolve this in my rewrite below.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+from contextlib import contextmanager
+from copy import deepcopy
+from itertools import cycle
+
+from hoomd.data.typeconverter import (
+    TypeConversionError, TypeConverterValue, OnlyIf, Either,
+    TypeConverterMapping, TypeConverterSequence)
+
+
+class _HOOMDDataStructures(metaclass=ABCMeta):
```
bdice, line 12: I recommend avoiding HOOMD in the class name. Also this should be singular (Structure, not Structures).
```suggestion
class _SyncedDataStructure(metaclass=ABCMeta):
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+                        "the same type.")
+                elif len(matches) == 1:
+                    rtn_type_def = matches[0]
+    # When no valid type converter can be found error out.
+    if rtn_type_def is None:
+        raise ValueError("Unable to find type definition for attribute.")
+    return rtn_type_def
+
+
+def _to_hoomd_data_structure(data, type_def, parent=None, label=None):
```
bdice, line 89:
```suggestion
def _to_synced_data_structure(data, type_def, parent=None, label=None):
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+    """Convert raw data to a synced _HOOMDDataStructures object.
+
+    This does nothing if the data is not detected to be a supported HOOMD data
+    structure type. Currently we support mutable sequences, mappings, and sets.
+
+    Args:
```
bdice, line 95: Missing `type_def` from Args.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+    if rtn_type_def is None:
+        raise ValueError("Unable to find type definition for attribute.")
+    return rtn_type_def
+
+
+def _to_hoomd_data_structure(data, type_def, parent=None, label=None):
+    """Convert raw data to a synced _HOOMDDataStructures object.
+
+    This does nothing if the data is not detected to be a supported HOOMD data
+    structure type. Currently we support mutable sequences, mappings, and sets.
```
bdice, line 93:
```suggestion
    This returns ``data`` if the data is not a supported data structure (sequence, mapping, or set).
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+            object.
+        parent:
+            The parent container if any for the data.
+        label:
+            Label to indicate to the parent where the child belongs. For
+            instance, if the parent is a mapping the label would be the key for
+            this value.
+    """
+    if isinstance(data, MutableMapping):
+        typing = _get_inner_typeconverter(type_def, TypeConverterMapping)
```
bdice, line 107: Don't use `typing`, that's a standard library module name. Also occurs below.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+            Raw data to be potentially converted to a _HOOMDDataStructures
+            object.
+        parent:
+            The parent container if any for the data.
+        label:
+            Label to indicate to the parent where the child belongs. For
+            instance, if the parent is a mapping the label would be the key for
+            this value.
+    """
+    if isinstance(data, MutableMapping):
```
bdice, line 106: Should you check against `Mapping` instead? I believe the existing logic would just create a mutable copy of an immutable data structure. Same below for `MutableSequence` and `MutableSet`.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+        typing = _get_inner_typeconverter(type_def, TypeConverterMapping)
+        return HOOMDDict(typing, parent, data, label)
+    elif isinstance(data, MutableSequence):
+        typing = _get_inner_typeconverter(type_def, TypeConverterSequence)
+        return HOOMDList(typing, parent, data, label)
+    elif isinstance(data, MutableSet):
+        return HOOMDSet(typing, parent, data, label)
+    else:
+        return data
```
bdice, line 115:
```suggestion
    return data
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+
+
+class HOOMDList(MutableSequence, _HOOMDDataStructures):
+    """List with type validation.
+
+    Use `to_base` to get a plain `list`.
```
bdice, line 121: Try to use rST-compatible syntax.
```suggestion
    Use :meth:`to_base` to convert to a list.
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+
+    Warning:
+        Should not be instantiated by users.
```
bdice, line 124:
```suggestion
        Users should not need to instantiate this class.
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+    """
+    def __init__(self, type_definition, parent, initial_value=None,
+                 label=None):
+        self._type_definition = type_definition
+        self._parent = parent
+        self._list = []
```
bdice, line 130: Move this further down, next to the data initialization.

bdice: Maybe rename the internal storage to `self._data` for all these classes?
```suggestion
        self._data = []
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+
+    def __getitem__(self, index):
+        """Get the value(s) associated with entry index of the list.
+
+        Supports slicing.
+        """
+        return self._list[index]
+
+    def __setitem__(self, index, value):
+        """Set an item(s) at index to the given value.
```
bdice, line 145:
```suggestion
        """Set item(s) at index to the given value(s).
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+
+        Support assignment with slices.
```
bdice, line 147:
```suggestion
        Supports assignment with slices.
```

bdice: I don't know the exact use cases here but it's worth noting that NumPy-like data structures can also be assigned using a NumPy array of indices instead of a slice object. _If_ that is a useful feature, you could consider handling the case where the indices are a sequence (a list or array of indices).

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+                iterable_start = len(self) - 1
+            else:
+                iterable_start = 0
+        if end is None:
+            if step is not None and step < 0:
+                iterable_end = -1
+            else:
+                iterable_end = len(self)
+        step = 1 if step is None else step
+        return range(iterable_start, iterable_end, step)
```
bdice, line 184: I think this whole function is equivalent to:
```suggestion
    def _slice_to_iterable(self, slice_):
        return range(len(self))[slice_]
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+        return self._list[index]
+
+    def __setitem__(self, index, value):
+        """Set an item(s) at index to the given value.
+
+        Support assignment with slices.
+        """
+        if isinstance(index, slice):
+            with self._buffer():
+                for i, v in zip(self._slice_to_iterable(index), value):
```
bdice, line 151: If you agree with my rewrite of `_slice_to_iterable` below, then you should inline that method here and delete it below (the helper function is only called once).

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+                    self[i] = v
+        else:
+            type_def = self._get_type_def(index)
+            try:
+                validated_value = type_def(value)
+            except TypeConversionError as err:
+                raise TypeConversionError(
+                    "Error in setting item {} in list.".format(index)) from err
```
bdice, line 159:
```suggestion
                    f"Error in setting item {index} in list.") from err
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+                iterable_end = len(self)
+        step = 1 if step is None else step
+        return range(iterable_start, iterable_end, step)
+
+    def _get_type_def(self, index):
+        entry_index = index % len(self._type_definition)
+        return self._type_definition[entry_index]
+
+    def __delitem__(self, index):
+        """Delele the index-th item from the list.
```
bdice, line 191:
```suggestion
        """Delete the item at the given index.
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+                raise RuntimeError("Deleting items from list has caused an "
+                                   "invalid state.") from err
+        self._update()
+
+    def __len__(self):
+        """Get the length of the list."""
+        return len(self._list)
+
+    def insert(self, index, value):
+        """Insert a value at the index-th index of the list.
```
bdice, line 215:
```suggestion
        """Insert a value at the given index.
```

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+            except TypeConversionError as err:
+                raise RuntimeError("Deleting items from list has caused an "
+                                   "invalid state.") from err
+        self._update()
+
+    def __len__(self):
+        """Get the length of the list."""
+        return len(self._list)
+
+    def insert(self, index, value):
```
bdice, line 214: Strong recommendation: match the Python docs for all your standard container docstrings. Matching established conventions is better than inventing your own (such as "index-th"). Reference: https://docs.python.org/3/library/stdtypes.html#mutable-sequence-types

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+                The value to insert into the list.
+        """
+        if index >= len(self):
+            index = len(self)
+
+        type_def = self._get_type_def(index)
+        try:
+            validated_value = type_def(value)
+        except TypeConversionError as err:
+            raise TypeConversionError(
```
bdice, line 230: Seems like you should raise a `ValueError` from the underlying `TypeConversionError`. That indicates that it's a problem in the user-provided value (and the source exception is still provided using `from err`).

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,497 @@
...
+            if index != len(self) and len(self._type_definition) > 1:
+                try:
+                    _ = self._type_definition(self._list)
+                except TypeConversionError as err:
+                    raise TypeConversionError(
+                        "List insertion invalidated list.") from err
+        self._update()
+
+    def extend(self, other):
+        """Add a list to the end of this list.
```
bdice, line 246:
```suggestion
        """Extends this list with the contents of ``other``.
```

hoomd/data/param_dict.py
```diff
@@ -0,0 +1,144 @@
+from copy import deepcopy
```
bdice, line 1: I'd use a full name for the module like `parameter_dict.py` that matches the class name.

bdice (changes requested):
Here's a partial review -- I only got through about 30% of the code, but had enough comments to share for a first iteration.

hoomd/data/data_structures.py
b-butler: This is supposed to act like a standard list. We could support basic NumPy indexing, but I think that might be more confusing.

b-butler:
@bdice thanks for the constructive comments!

hoomd/data/data_structures.py
b-butler: `to_base` will resolve correctly with sphinx and we don't use it elsewhere in hoomd.

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -19,6 +19,10 @@ def check_dict(shape_dict, args):
     Useful for more complex nested dictionaries (like the shape key in unions)
     Used to test that the dictionary passed in is what gets passed out
     """
+    try:
+        shape_dict = shape_dict.to_base()
+    except AttributeError:
+        pass
```
bdice, line 25: What raises an `AttributeError`? I haven't looked at the test data, but perhaps there is some kind of check you can have here to make sure that those cases are reasonable and that this won't ignore real problems?

bdice (changes requested):
I completed a first pass over this code. I am not very familiar with the context for this code (which appears to be quite complex), but I did my best to critique what I saw.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,447 @@
...
+            return self[attr]
+        except KeyError:
+            raise AttributeError("{} object has not attribute {}.".format(
+                self, attr))
+
+    def __setattr__(self, attr, value):  # noqa: D105
+        if attr in self._data:
+            self[attr] = value
+        else:
+            super().__setattr__(attr, value)
```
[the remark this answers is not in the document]
b-butler: I can see some fragility in how it is defined now, but in terms of limit the assignment of "new" keys, it already prevents that. Are you suggesting we make the class behave more like it has a `__slots__` attribute?

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,447 @@
...
+        else:
+            # Disconnect child from parent to prevent child signaling update
+            if isinstance(self._data[key], _SyncedDataStructure):
+                self._data[key]._parent = None
+            self._data[key] = _to_synced_data_structure(
+                validated_value, type_def, self, key)
+        self._update()
+
+    def __delitem__(self, key):  # noqa: D105
+        raise RuntimeError("mapping does not support deleting keys.")
```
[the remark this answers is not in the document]
b-butler: The typical use case is that this dictionary will be linked to a C++ object which needs these exact keys, therefore we can't let users delete or add keys.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,447 @@
...
+        for key, entry in self.items():
+            if isinstance(entry, _SyncedDataStructure):
+                return_data[key] = entry.to_base()
+            else:
+                try:
+                    use_entry = deepcopy(entry)
+                except Exception:
+                    use_entry = entry
+                return_data[key] = use_entry
+        return return_data
```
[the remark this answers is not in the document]
b-butler: Nice refactor, and I believe they are, but i need to check the pybind11 error that gets raised.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,447 @@
...
+        return item in self._data
+
+    def __iter__(self):  # noqa: D105
+        yield from self._data
+
+    def __len__(self):  # noqa: D105
+        return len(self._data)
+
+    def add(self, item):  # noqa: D102
+        if item not in self:
```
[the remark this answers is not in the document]
b-butler: Good catch. I was doing this to prevent all the preprocessing, but due to the prepreprocessing the first check would likely fail anyways.

hoomd/data/type_param_dict.py
```diff
@@ -28,6 +30,37 @@ def proper_type_return(val):
...
+    attributes to be easily synced between C++/Python and validated in Python.
+
+    Implements hard coded key validation which expects string keys when
+    ``_len_keys == 1`` and tuples of strings when ``_len_keys > 1``. However, we
+    also provide convenience means of specifying multiple keys on a single
+    `__setitem__` or `__getitem__`. Anywhere a string is expected, a list or
+    list-like object containing strings is also accepted. For multiple string
+    keys this will result in the product of the list with the other string
+    values for instance, ``(['a', 'b'], 'a')`` will expand to ``[('a', 'a'),
+    ('a', 'b')]``. Likewise, a list of tuples of the appropriate length can also
```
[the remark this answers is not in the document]
b-butler: Can do, and to explain, we store keys using lexical ordering so we don't have to store values for both `('a', 'b')` and `('b', 'a')`. This also prevents errors as they must be the same. The other option would involve checking for the current inclusion of any permutation of a key.

hoomd/hpmc/pytest/test_shape.py
b-butler: I need to examine this code again to remember why it is okay and necessary.

hoomd/data/type_param_dict.py
```diff
@@ -116,12 +146,14 @@ def _yield_keys(self, key):
 
         The order is necessary so ('A', 'B') is equivalent to ('B', A').
         '''
+        keys = self._validate_and_split_key(key)
         if self._len_keys > 1:
-            keys = self._validate_and_split_key(key)
+            # We always set ('a', 'b') instead of ('b', 'a') to reduce storage
```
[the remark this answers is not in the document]
b-butler: These are dictionary keys so not having to store both `('a', 'b')` and `('b', 'a')` reduces storage. We allow unsorted access as well so if the user sets a key using an unsorted tuple he can access the data with the same key, we just don't store it that way.

vyasr:
I'm happy to provide a review for this in the near future. I should also say that I intend to finish up the work on glotzerlab/signac#249, and tbh I'm wondering if extracting the final version of that into a separate package and reusing it here will be a better solution prior to the final v3 release of HOOMD. Right now this is a ton of logic directly duplicated from the design in signac. I was already considering eventually extracting that into a package, but didn't want to touch on that until it was merged into signac properly; these changes to HOOMD definitely suggest that it might be beneficial to spin that off separately.

hoomd/data/type_param_dict.py
```diff
@@ -221,14 +309,23 @@ def to_dettached(self):
...
         for key in self._yield_keys(key):
-            vals[key] = getattr(self._cpp_obj, self._getter)(key)
+            cpp_val = getattr(self._cpp_obj, self._getter)(key)
+            if (key in self._dict and
+                    isinstance(self._dict[key], _SyncedDataStructure)):
+                self._dict[key]._parent = None
+            data_struct = _to_synced_data_structure(
+                cpp_val, self._type_converter, self, key)
+            self._dict[key] = data_struct
+            vals[key] = data_struct
```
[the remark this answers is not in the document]
b-butler: I added a comment to try to explain the reasoning for the logic here.

hoomd/data/data_structures.py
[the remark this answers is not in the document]
b-butler: I believe I have resolved this.

hoomd/data/data_structures.py
```diff
@@ -0,0 +1,447 @@
...
+    def __setattr__(self, attr, value):  # noqa: D105
+        if attr in self._data:
+            self[attr] = value
+        else:
+            super().__setattr__(attr, value)
+
+    def __setitem__(self, key, item):  # noqa: D105
+        if key not in self._type_definition:
+            raise KeyError(
+                "Cannot set value for non-existent key {}.".format(key))
```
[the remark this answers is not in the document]
b-butler: I have switched to format strings here.

joaander:
There are lots of good discussions on nesting, properties and user expectations in #819. I leave the decision and implementation up to you all. Python does not make this easy. In C++, you *know* if you have a reference or a copy.

vyasr:
Just an update, I haven't reviewed this yet because I've been putting my time into the signac PR that I mentioned. I _think_ that once the details are worked out there we should be able to copy a lot of it in directly and just implement a couple of small classes defining how HOOMD objects should be synchronized. Once that PR is closer to complete I'll ping @b-butler to discuss details.

joaander:
@b-butler I will let you handle the merge of master into this branch. Let me know if you have any questions.

joaander:
Is this functionality we are still planning to complete? Or should we close this PR?

b-butler:
@joaander yes and no, some of these changes were necessary for #944, but others weren't. I know @vyasr thought that using the syncing code he wrote for signac would make this simpler to implement, so I am currently waiting on #944, and then I will probably re-implement the changes using the signac code, but I guess that means I can close this to open another PR later.

Issue #777, opened 2020-09-05
Rename move_ratio in HPMC


vyasr:
## Description

The `move_ratio` property of HPMC integrators is confusingly named. It is documented as the "ratio of translation moves to rotation moves", but really it's the percentage of moves that are translational: the actual check that is performed is `bool move_type_translate = !shape_i.hasOrientation() || (move_type_select < m_move_ratio)`. If it were truly a ratio, the default value of 0.5 would imply 1 translational move for every 2 rotational moves, but it actually implies a 1:1 ratio.

## Proposed solution

We should rename this property in v3. Something like `translational_move_probability` would more accurately reflect the meaning.

## Developer

I'm happy to help with this, although it might also be natural to fold into other ongoing efforts.

joaander:
Yes, this is a good change to make. Feel free to start on this any time, it doesn't interfere with any other ongoing work.

joaander:
#779 makes some of the changes, but there are more places to fix.

PR #779, opened 2020-09-09
Feature/rename move ratio

vyasr:
## Description

This PR renames instance of `move_ratio` throughout HPMC to `*move_probability` to more accurately reflect the meaning of these parameters.

**I'm not quite sure how to handle `hpmc.update.Clusters`, which I have not yet changed.** There are multiple parameters and they each affect probabilities in different ways in different situations. I glanced at the C++ code, and offhand it seems like line reflections and pivots are applied per particle then swaps are applied to all types in a cluster, but I'll have to spend more than 5 minutes looking through that code to understand exactly how the different input probabilities factor in to the final move choice. Note that the validity of pivot moves depends on whether or not particles are anisotropic, and I don't think the type swaps are applied across all types, so all the parameters are somewhat conditional. If @jglaser or anyone else knows the answer offhand, I can apply the appropriate naming conventions.

**Update I'm going to avoid changing the cluster updater in this PR, so it should be ready for review.

## Motivation and context

Partially addresses #777 (update.Clusters will be handled separately at a later date).

## How has this been tested?

Covered by existing tests.

## Change log

None needed IMO.

## Checklist:


jglaser:
I suggest saving changes to `UpdaterClusters` for later. At present, they will only generate (even more) merge conflicts with ongoing work in #659 .

vyasr:
That's fine with me, I've updated the PR description to reflect that `update.clusters` won't be changed in this PR. In that case, this PR is ready to review and merge.

joaander (approved):
Thanks!

PR #783, opened 2020-09-18
Reorder operations in run

### LINKED ISSUE #574: Reorder actions in run()
## Description

Change the order of operations in `run()` so that analyzers run last.

## Proposed solution

`run()` should run steps in this order:
1. Updaters (at step t)
2. Integrator (move system to step t+1)
3. Analyzers (at step t+1)

## Additional context

This solves user confusion where (e.g.) a trajectory dump with period=100 and run(1000) fails to write output for step 1000. This change also enables simplified single file trajectory/restart workflows with no need for `write_restart` at the end of the script.

As a downside, output files will no longer include step 0.

### LINKED ISSUE #710: Add documentation for hoomd/simulation.py
## Purpose

This file needs to have better documentation of its class `Simulation` for public facing properties, and methods. In addition, its docstring needs to be elaborated.

## Tasks

- [ ] Expand docstring for `hoomd.Simulation`.
- [ ] Add docstrings for pubilic (not leading with an underscore) methods and properties of `hoomd.Simulation`, or after discussion explicitly mark code with the appropriate error code.
- [ ] Add file to CI pydocstyle checking.

## Pertinent Information

* Information about how the class works can be found in the new HOOMD-blue tutorial [here](https://github.com/glotzerlab/hoomd-examples/pull/12). This should provide some idea about how the class in question and many of its methods work.
````python
def foo(bar):  # noqa: D103
    pass

class Baz:  # noqa: D101
    def egg(self):  # noqa: D102
        pass
````

### PR DESCRIPTION
**Pull request by joaander**
## Description

* Place the call to analyzers at the end of the run() loop body. Resolves #574
* Also, document the Simulation object. Resolves #710

## Motivation and context

Many users expect that a periodic trigger and a run() for an integer number of periods (e.g. period=100, run(500)) will trigger the analyzers on final step computed by the run. This PR includes an optional flag to additional call analyzers before the start of the loop body for cases where users want output on step 0 or the first step in a new stage of the simulation.

## How has this been tested?

This PR adds unit tests to verify the change.

All updated sphinx documentation renders properly in a local build.

## Note

This PR is based on #767 to avoid merge conflicts. I will take it out of draft state once that PR is merged.

## Checklist:

### DIFF
--- a/hoomd/simulation.py
+++ b/hoomd/simulation.py
@@ -202,13 +202,13 @@ class Simulation(metaclass=Loggable):
             if value:
                 self._state._cpp_sys_def.getParticleData().setPressureFlag()
 
-    def run(self, steps, check_writer_triggers_on_initial_step=False):
+    def run(self, steps, write_at_start=False):
         """Advance the simulation a number of steps.
 
         Args:
             steps (int): Number of steps to advance the simulation.
 
-            check_writer_triggers_on_initial_step (bool): When True, writers
+            write_at_start (bool): When True, writers
                with triggers that evaluate True for the initial step will be
                exected before the time step loop.
 
@@ -219,7 +219,7 @@ class Simulation(metaclass=Loggable):
         state in the order: Tuners, Updaters, Integrator, then Writers following
         the logic in this pseudocode::
 
-            if check_writer_triggers_on_initial_step:
+            if write_at_start:
                 for writer in operations.writers:
                     if writer.trigger(timestep):
                         writer.write(timestep)
@@ -247,11 +247,11 @@ class Simulation(metaclass=Loggable):
         capture the final output of the last step of the run loop. For example,
         a writer with a trigger ``hoomd.trigger.Periodic(period=100, phase=0)``
         active during a ``run(500)`` would write on steps 100, 200, 300, 400,
-        and 500. Set ``check_writer_triggers_on_initial_step=True`` on the first
+        and 500. Set ``write_at_start=True`` on the first
         call to `run` to also obtain output at step 0.
 
         Warning:
-            Using ``check_writer_triggers_on_initial_step=True`` in subsequent
+            Using ``write_at_start=True`` in subsequent
             calls to `run` will result in duplicate output frames.
         """
         # check if initialization has occurred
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by joaander** [comment]
Anyone have a suggestion for a better name than `check_writer_triggers_on_initial_step` for the optional parameter?

**Comment by joaander** [comment]
> Anyone have a suggestion for a better name than `check_writer_triggers_on_initial_step` for the optional parameter?

`write_at_start`?

File: hoomd/simulation.py [hunk 1]
```diff
@@ -109,33 +148,39 @@ def create_state_from_snapshot(self, snapshot):
 
     @property
     def state(self):
+        """State: The current simulation state."""
```
b-butler, line 151:
```suggestion
        """hoomd.State: The current simulation state."""
```

hoomd/simulation.py
```diff
@@ -109,33 +148,39 @@ def create_state_from_snapshot(self, snapshot):
...
         return self._state
 
     @property
     def operations(self):
+        """Operations: The operations that apply to the state."""
```
b-butler, line 156:
```suggestion
        """hoomd.Operations: The operations that apply to the state."""
```

hoomd/simulation.py
```diff
@@ -109,33 +148,39 @@ def create_state_from_snapshot(self, snapshot):
...
-    def sanity_check(self):
-        raise NotImplementedError
-
-    def advance(self, runsteps):
-        raise NotImplementedError
-
-    def apply_operations(self, operations):
-        raise NotImplementedError
-
     @property
```
b-butler, line 159: Should we make `Simulation` loggable? We could make `TPS` loggable then. Otherwise we can use the `logger['TPS'] = (sim, 'tps', 'scalar')` syntax.

b-butler (approved):
In general, it looks good to me. I left a few suggestions regarding the type of some of `Simulation`'s properties. They render fine, but I figured we should probably include the full namespace we expect users to use. However, I don't feel strongly about it so I will approve the PR regardless.

hoomd/simulation.py
joaander: Yes, we should. I opened #797 as a reminder and also noted a few other loggable quantities we may want to provide.

hoomd/simulation.py
joaander: Done.

hoomd/simulation.py
joaander: Done.

PR #785, opened 2020-09-18
Add documentation for hoomd.State

b-butler:
## Description

Adds documentation for `hoomd.State`.

## Motivation and context

The public facing class needed documentation.

## How has this been tested?

`pydocstyle` passes for this file and sphinx builds without errors.

## Checklist:


hoomd/state.py
```diff
@@ -34,15 +40,29 @@ def _create_domain_decomposition(device, box):
...
+    properties. Accessing these quantities through a `State` object directly is
+    a bit faster than accessing them through a local snapshot and much faster
+    than accessing them through `State.snapshot`.
+
+    Note:
+        This object should never be directly instantiated by users. There is no
+        way to set a state created outside of a `hoomd.Simulation` object to a
+        simulation. See `hoomd.Simulation.create_state_from_gsd` and
+        `hoomd.Simulation.create_state_from_snapshot` for information about
+        instantiating `State` objects.
```
joaander, line 61:
```suggestion
        simulation. Use `hoomd.Simulation.create_state_from_gsd` and
        `hoomd.Simulation.create_state_from_snapshot` to instantiate a 
        `State` object.
```
Use active voice.

hoomd/state.py
```diff
@@ -34,15 +40,29 @@ def _create_domain_decomposition(device, box):
 
 
 class State:
-    R""" Simulation state.
-
-    Args:
-        snapshot
+    R"""The state of a `hoomd.Simulation` object.
```
joaander, line 43: Should we use `"""` instead of `R"""` everywhere? I recall running into problems with the `R"""` in some tools, though now I can't remember which. PEP257 recommends `r"""` when using backslashes.

hoomd/state.py
```diff
@@ -55,52 +75,54 @@ def __init__(self, simulation, snapshot):
...
+        # filters with poorly implemented __hash__ and __eq__ from causing cache
+        # errors.
         self._groups = defaultdict(dict)
 
     @property
     def snapshot(self):
-        cpp_snapshot = self._cpp_sys_def.takeSnapshot_double()
-        return Snapshot._from_cpp_snapshot(cpp_snapshot,
-                                           self._simulation.device.communicator)
+        """hoomd.Snapshot: All data of a simulation's current microstate.
```
joaander, line 91: We need to document the `hoomd.Snapshot` object in detail, though that should be the subject of another PR. If use of `Snapshot` is as discouraged as these docs suggest, should we move the `Snapshot` class into `data` in our namespace reorg?

hoomd/state.py
```diff
@@ -121,8 +143,45 @@ def snapshot(self, snapshot):
 
         self._cpp_sys_def.initializeFromSnapshot(snapshot._cpp_obj)
 
+    @property
+    def particle_types(self):
+        """list[str]: List of all particle types in the simulation."""
```
joaander, line 148:
```suggestion
        """List[str]: List of all particle types in the simulation."""
```
I thought we were using the Python typing notation. Should we mention that these are all copies and are not settable?

joaander (changes requested):
Thanks! I have a few minor change requests and a few questions below.

hoomd/state.py
b-butler: I was until I noticed that it wouldn't link to the python standard documentation. We should mention that. I think the type documentation could be a good point of discussion.

hoomd/state.py
b-butler: I didn't mean to discourage its use, I more was trying to prevent users from doing something like
````python
snapshot = sim.state.snapshot
if snapshot.exists:
    snapshot.particles.velocities[:] = [0, 0, 0]
# or
if snapshot.exists:
    snapshot.particles.position[1] = [1, 2, 3]
sim.state.snapshot = snapshot
````
Overall, I think that the plain snapshot is easiest to manipulate especially given an MPI simulation. I just wanted to inform users that if only manipulating a few pieces of data or doing some blanket change like above, it is faster to use the local snapshot API.

I am not sure what the right balance here between warning users about using a sub-optimal method (for some use cases) versus the ease in use of the snapshot.

File: hoomd/state.py (hunk 2 above)
Comment (b-butler) [reply hoomd/state.py:43]: I tried this with pydocstyle without any problems, but I can remove it or use `r"""` on your preference.

Comment (bdice) [reply hoomd/state.py:43]: I think lowercase `r"""` is preferable. As for raw vs. normal strings, the case I always run into that requires raw strings is when documenting the parameters `*args` and `**kwargs`. There is some magic in how you place the backslashes so that Sphinx doesn’t try to render the asterisks as italic/bold font. I have commented about this before and can find examples/demos if needed. Otherwise I prefer non-raw strings wherever possible.

File: hoomd/state.py (hunk 3 above)
Comment (joaander) [reply hoomd/state.py:91]: In reference docs, I would prefer to state facts and leave suggested usages for tutorials. E.g., we could state that `snapshot` is an O(N_particles + N_bonds + ...) operation that provides a copy of the data while the local snapshot methods are nominally O(1) and provides a reference.

File: hoomd/state.py (hunk 4 above)
Comment (joaander) [reply hoomd/state.py:152]: `list[str]` links only because Sphinx has hardcoded this for list, dict, and tuple: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists

We should discuss and decide on a standard to use. If links to `typing.List` are desired, then we can explicitly include those. I have no particular preference as long as we are consistent and clear.

File: hoomd/state.py [hunk 5]
```diff
@@ -34,15 +40,29 @@ def _create_domain_decomposition(device, box):
...
+    Provides access (read/write) to a `hoomd.Simulation` object's particle,
+    bond, angle, etc. data. Data access is facilitated through two complementary
+    APIs: *global* and *local* snapshots (note that global does not refer to
+    variable scope here). See `State.snapshot`, `State.cpu_local_snapshot`, and
+    `State.gpu_local_snapshot` for information about these data access patterns.
+    In addition, many commonly used smaller quantites such as the number of
+    particles in a simulation are available directly through `State` object
+    properties. Accessing these quantities through a `State` object directly is
+    a bit faster than accessing them through a local snapshot and much faster
+    than accessing them through `State.snapshot`.
```
joaander, line 54:
```suggestion
    properties.
```
Unnecessary.

hoomd/state.py
```diff
@@ -55,52 +75,58 @@ def __init__(self, simulation, snapshot):
...
+
+        `State.snapshot` should be used when all of a simulation's state
+        information is desired in a single object. When accessed, data across
+        all MPI ranks and from GPUs is gathered on the root MPI rank's memory.
+        When accessing data in MPI simulations then it is recommended to use a
+        ``if snapshot.exists:`` conditional to prevent attempting to access data
+        on a non-root rank.
+
+        This property can be set to the state using an entirely new
+        `hoomd.Snapshot` object.  Example use cases in which a simulation's
```
joaander, line 101:
```suggestion
        This property can be set to replace the system state with the given
        `hoomd.Snapshot` object.  Example use cases in which a simulation's
```
"entirely new" here is awkward. I could be an existing snapshot that has been modified.

joaander (approved):
Thanks! I think this is fine as is. I've included a few optional suggestions to improve.
Any thoughts, @bdice ?

hoomd/state.py
bdice: This is an annoying inconsistency between Sphinx and Python typing. 😕 I want to recommend `list` over `List`, based on how the output looks. The link to the Python docs for `list` is what I would expect here.

hoomd/state.py
```diff
@@ -161,33 +224,9 @@ def N_dihedrals(self):
...
-    @property
-    def special_pair_types(self):
-        return self._cpp_sys_def.getPairData().getTypes()
-
     @property
     def box(self):
-        """The state's box (a :py:class:`hoomd.Box` object).
+        """hoomd.Box: The current simulation box.
 
         Ediing the box directly is not allowed.  For example
```
bdice, line 231:
```suggestion
        Editing the box directly is not allowed. For example
```

hoomd/state.py
```diff
@@ -257,17 +293,20 @@ def update_group_dof(self):
...
-        The `hoomd.data.LocalSnapshot` object is only usable within a
-        context manager (i.e. ``with sim.state.cpu_local_snapshot as data:``)
-        The interface is similar to that of the `hoomd.Snapshot`. Data is local
-        to a given MPI rank. The returned arrays are `hoomd.array.HOOMDArray`
-        objects. Through this interface zero-copy access is available (access is
-        guarenteed to be zero-copy when running on CPU, may be zero copy if
+        Data in `State.cpu_local_snapshot` is MPI rank local, and the
+        `hoomd.data.LocalSnapshot` object is only usable within a context
+        manager (i.e. ``with sim.state.cpu_local_snapshot as data:``).  Attempts
+        to assess data outside the context manager will result in errors.  The
```
bdice, line 303: Sometimes I've noticed that vim inserts two spaces after periods when wrapping. Not sure if that's how this happened.
```suggestion
        manager (i.e. ``with sim.state.cpu_local_snapshot as data:``). Attempts
        to assess data outside the context manager will result in errors. The
```

hoomd/state.py
```diff
@@ -257,17 +293,20 @@ def update_group_dof(self):
...
+        local snapshot interface is similar to that of `hoomd.Snapshot`.
+
+        The `hoomd.data.LocalSnapshot` data access is mediated through
+        `hoomd.array.HOOMDArray` objects. This lets us ensure memory safety when
+        directly accessing HOOMD-blue's data. The interface provides zero-copy
+        access (zero-copy is guarenteed on CPU, access may be zero copy if
```
bdice, line 309:
```suggestion
        access (zero-copy is guarenteed on CPU, access may be zero-copy if
```

hoomd/state.py
```diff
@@ -292,18 +331,20 @@ def cpu_local_snapshot(self):
...
-        within a context manager (i.e. ``with sim.state.gpu_local_snapshot as
-        data:``) The interface is similar to that of the `hoomd.Snapshot`. Data
-        is local to a given MPI rank. The returned arrays are
-        `hoomd.array.HOOMDGPUArray` objects. Through this interface potential
-        zero-copy access is available (access cannot be guarenteed to be
-        zero-copy, but will be if the most recent copy of the data is on the
+        Data in `State.gpu_local_snapshot` is GPU local, and the
+        `hoomd.data.LocalSnapshotGPU` object is only usable within a context
+        manager (i.e. ``with sim.state.gpu_local_snapshot as data:``).  Attempts
+        to assess data outside the context manager will result in errors.  The
```
bdice, line 341:
```suggestion
        manager (i.e. ``with sim.state.gpu_local_snapshot as data:``). Attempts
        to assess data outside the context manager will result in errors. The
```

bdice (approved):
A few minor edits but otherwise LGTM.

joaander:
> * My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

FYI @b-butler, I added your name to `credits.rst` in `feature/new-object-API`.

joaander:
Thanks!

Issue #648, opened 2020-05-11
V3 Randomize Velocities API


tommy-waltmann:
## Description

There may be a better, more general way to expose the `randomize_velocities()` API that is currently done through the integration methods

## Proposed solution

We had talked about potentially exposing this API through the simulation state instead of through the integration method. For example, it might be used via `simulation.state.thermalize_velocities()`, instead of `nvt.randomize_velocities()` for an `NVT` integration method.

## Additional context

Another option we had considered is moving the implementation of `randomize_velocities()` to the `_Method` class if we decide not to move it to the state.

## Developer

Once we come to a decision as a group as to how to proceed, then yes, I could handle this if need be.

vyasr:
To give some context, I proposed moving `randomize_velocities` into the state for a couple of reasons. The reason this discussion came up is because the v3 API introduces additional complexity associated with the attachment process, which allows simulations to be specified almost completely without ever creating C++ objects. While this process is desirable in general because it allows composability of protocols in very natural ways by just importing different python modules that each set up e.g. a force field via a set of pair potentials, it introduces difficulties for any method that is meant to occur instantaneously rather than at some specific interval. One common example is resizing a box, and randomizing velocities is another. Since conceptually methods are meant to be `Triggered` in the v3 setup, having one execute instantaneously is an antipattern and also causes implementation problems because it then requires that the simulation state be attached to the C++ instance. Since randomization is really a property of the state, not a property of the integrator, it makes sense to move the logic there, and it conveniently resolves the difficulties mentioned above. The one downside is that initializing velocities in a way consistent with integrator parameters (i.e. with the same temperature applied to the same groups of particles) is now the user's responsibility.

jglaser:
An issue is also that internal integrator variables (like the thermostat or barostat degrees of freedom) are currently being randomized because they, too, are velocities, which is one if the reasons why the randomization is specific to the integration method being used. Differently from particle velocities, there is only a finite number of them (such as, one for every box degree of freedom). We would need to find a way how to extend the state by these variables, depending on which integration method is used.

joaander:
[...]

PR #749, opened 2020-08-25
Improve documentation on methods.py and integrate.py

syjlee:
## Description
Adding and Editing the documentation on methods.py and integrate.py.
The documentation covers public methods and properties of classes.
The private methods are commented with `# noqa` to mark as undocumented.
Mainly args and attributes are taken care of to improve.
Currently working on rephrasing descriptions on args and attributes and also on extended params on Langevin and Brownian.

## Motivation and Context
To resolve the Issue #721 .

## How Has This Been Tested?
Tested locally generated doc html through sphinx

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [ ] My name is on the [list of contributors](https://github.com/glotzerlab/hoomd-blue/blob/master/sphinx-doc/credits.rst).

Issue #229, opened 2017-02-23
Use 64-bit time step counters


joaander:
**[Original report](https://bitbucket.org/glotzer/hoomd-blue/issue/229) by me.**

----------------------------------------

* There are many "unsinged int timestep" and "unsigned int m_timestep" phrases in the code that can be bulk replaced.
* System.h and System.cc will need a read-through to find all instances where it stores a timestep like variable in an unsinged int (even temporarily in a delta).
* Special care must be taken in DPD and anywhere else that uses the current timestep in a Saru RNG - the easiest thing to do is cast out the lower 32 bits of the number and use that for the seed, though that will make RNGs repeat every ~4 billion steps. Alternately, one could mix the higher 32 bits into one of the other seed entries.
* the gsd file format already stores time step as a uint64_t, so no changes are needed there.

joaander:
We will implement this in v3.0.

Questions: The DCD file format cannot represent 64-bit timestep counters. Should we remove DCD, let the counter wrap around, or error when attempting to write an invalid step to DCD?

Aside: DCD can also only represent fixed periods and the new `Trigger` API in HOOMD will make it even easier for users to use non-fixed period outputs.

mphoward:
I think we should keep supporting DCD because users of other simulation packages casually trying or transitioning to HOOMD might prefer having this file format for their other tools.

I would error out if the timestep exceeds the DCD spec. This can be clearly documented, and the user should decide how they want to handle longer runs (reset the step, create a new file, etc.)

vyasr:
I agree with @mphoward, removing DCD would probably alienate significant portions of our user base. There's no need to improve its support, just to keep it where it is now.

PR #788, opened 2020-09-25
Implement thermalize_particle_momenta

joaander:
## Description

* Implement `State.thermalize_particle_momenta`.
* Implement `_Method.thermalize_extra_dof`.

## Motivation and context

HOOMD v2 had `ranomize_velocities` as a method of the integrator methods. This randomized particle velcoities, angular momenta, and the extra degrees of freedom of the NVT and NPT integration methods. This PR breaks those into two methods. The particle velocities and angular momenta are set by `State.thermalize_particle_momenta(filter, kT, seed)` and the integration method degrees of freedom are set by `_Method.thermalize_extra_dof(seed)`.

This is a more appropriate abstraction as discussed in #648.

Resolves: #648

## How has this been tested?

I added unit tests.

The sphinx documentation for `State.theramlize_particle_momenta` renders correctly. I did not check `_Method` - that can be handled after merging with #749 which make significant improvements to the integration methods documentation.

## Checklist:


hoomd/md/methods.py
```diff
@@ -19,8 +19,37 @@
...
+    def thermalize_extra_dof(self, seed):
+        """Set the extra degrees of freedom to random values.
+
+        Args:
+            seed (int): Random number seed
+
+        Some integration methods, such as `NVT` and `NPT` employ extra degrees
+        of freedom for the thermostat and/or barostat. `thermalize_extra_dof`
+        sets these to random values. This method performs no operation for
+        integration methods that lack internal degrees of freedom.
```
b-butler, line 41: The reason for the above comment is expressed in these lines. I worry that it may be confusing to users to have this function as a no-op for most integrators although most common use cases would indeed have this perform a meaningful action.

hoomd/md/methods.py
```diff
@@ -19,8 +19,37 @@
...
+
+        Note:
+            The seed for the pseudorandom number stream includes the
+            simulation timestep and the provided *seed*.
+        """
+
+        if not self._attached:
+            raise RuntimeError(
+                "Call Simulation.schedule before thermalize_extra_dof")
```
b-butler, line 50:
```suggestion
                "Call Simulations.operations.schedule before thermalize_extra_dof")
```

hoomd/md/pytest/test_methods.py
```diff
@@ -274,6 +274,33 @@ def test_npt_attributes_attached_3d(simulation_factory,
...
+    all_ = hoomd.filter.All()
+    constant_t = hoomd.variant.Constant(2.0)
+    constant_s = [
+        hoomd.variant.Constant(1.0),
+        hoomd.variant.Constant(2.0),
+        hoomd.variant.Constant(3.0),
+        hoomd.variant.Constant(0.125),
+        hoomd.variant.Constant(.25),
+        hoomd.variant.Constant(.5),
+    ]
```
b-butler, line 289: These should work without the explicit conversion to a constant variant right?

hoomd/md/pytest/test_methods.py
```diff
@@ -274,6 +274,33 @@ def test_npt_attributes_attached_3d(simulation_factory,
...
+                               tau=2.0,
+                               S=constant_s,
+                               tauS=2.0,
+                               couple='xyz')
+
+    sim = simulation_factory(two_particle_snapshot_factory())
+    sim.operations.integrator = hoomd.md.Integrator(0.005, methods=[npt])
+    sim.operations.schedule()
+
+    npt.thermalize_extra_dof(100)
```
b-butler, line 301: There isn't an easy way to do this, but it would be nice to check for actually randomization of the extra DOF.

b-butler (changes requested):
Good work @joaander. I have a few comments include a question regarding the placement of `thermalize_extra_dofs`.

hoomd/RNGIdentifiers.h
```diff
@@ -58,6 +58,7 @@ struct RNGIdentifier
     static const uint32_t SlitGeometryFiller = 0xdb68c12c;
     static const uint32_t SlitPoreGeometryFiller = 0xc7af9094;
     static const uint32_t UpdaterQuickCompress = 0x00981234;
+    static const uint32_t ParticleGroupThermalize = 1;
```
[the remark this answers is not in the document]
joaander: In a later PR, these will IDs will get fewer bits and they will all get renumbered sequentially (#229)

hoomd/md/methods.py
joaander: We can implement in only those methods where it does something: NVT, NPT, and NPH.

hoomd/md/pytest/test_methods.py
joaander: Yes. This was copied and pasted from a test where the explicit object was needed for `is` testing.

hoomd/md/pytest/test_methods.py
joaander: I will add the extra DOF to the param dict and then we can validate they get set in this test.

hoomd/md/methods.py
joaander: done.

hoomd/md/methods.py
```diff
@@ -132,6 +115,31 @@ def _attach(self):
...
+    def thermalize_extra_dof(self, seed):
+        """Set the extra degrees of freedom to random values.
+
+        Args:
+            seed (int): Random number seed
+
+        Some integration methods, such as `NVT` and `NPT` employ extra degrees
+        of freedom for the thermostat and/or barostat. `thermalize_extra_dof`
+        sets these to random values. This method performs no operation for
+        integration methods that lack internal degrees of freedom.
```
b-butler, line 128: Since no classes have this method where it does nothing, we no longer need this statement.

hoomd/md/methods.py
```diff
@@ -132,6 +115,31 @@ def _attach(self):
...
+
+        Note:
+            The seed for the pseudorandom number stream includes the
+            simulation timestep and the provided *seed*.
+        """
+
+        if not self._attached:
+            raise RuntimeError(
+                "Call Simulation.operations.schedule before thermalize_extra_dof")
```
b-butler, line 137: We should probably change this to `Simulation.run(0)` now.

b-butler (changes requested):
I have two small comments, but otherwise this looks good to me!

hoomd/md/methods.py
joaander: Done.

hoomd/md/methods.py
joaander: Done.

joaander:
This PR is ready for another round of reviews.

b-butler (approved):
LGTM. I think this separation of thermalizing makes a lot of intuitive sense.

hoomd/ParticleGroup.cc
```diff
@@ -789,6 +792,100 @@ unsigned int ParticleGroup::intersectionSize(std::shared_ptr<ParticleGroup> othe
     return n;
     }
 
+void ParticleGroup::thermalizeParticleMomenta(Scalar kT, unsigned int seed, unsigned int timestep)
+    {
+    unsigned int group_size = this->getNumMembers();
+
+    const unsigned int D = Scalar(m_sysdef->getNDimensions());
```
vyasr, line 799: I would prefer the use of `dimensions` instead of `D`, easy to forget the meaning of single letter variables later in the code.

hoomd/ParticleGroup.cc
```diff
@@ -789,6 +792,100 @@ unsigned int ParticleGroup::intersectionSize(std::shared_ptr<ParticleGroup> othe
...
+        {
+        unsigned int j = this->getMemberIndex(group_idx);
+        unsigned int ptag = h_tag.data[j];
+
+        // Seed the RNG
+        hoomd::RandomGenerator rng(hoomd::RNGIdentifier::ParticleGroupThermalize, seed, ptag, timestep);
+
+        // Generate a random velocity
+        Scalar mass =  h_vel.data[j].w;
+        Scalar sigma = fast::sqrt(kT / mass);
```
vyasr, line 835: I _think_ that we looked at this at some point and the fast sqrt isn't actually doing anything special right now (in theory you could do smarter algorithms than what `slow::sqrt` should do using inverses etc). In case I'm wrong or if that ever changes, though, are there any issues with using the less precise sqrt with respect to the accuracy of the resulting velocity distribution?

vyasr: Same question applies to the angular mementa below.

hoomd/ParticleGroup.cc
```diff
@@ -789,6 +792,100 @@ unsigned int ParticleGroup::intersectionSize(std::shared_ptr<ParticleGroup> othe
...
+    vec3<Scalar> tot_momentum(0, 0, 0);
+
+    // Loop over all particles in the group
+    for (unsigned int group_idx = 0; group_idx < group_size; group_idx++)
+        {
+        unsigned int j = this->getMemberIndex(group_idx);
+        unsigned int ptag = h_tag.data[j];
+
+        // Seed the RNG
+        hoomd::RandomGenerator rng(hoomd::RNGIdentifier::ParticleGroupThermalize, seed, ptag, timestep);
```
vyasr, line 831: Any reason not to just use index `j` directly in the rng and avoid accessing the tags? Saves having to map the particle tags array unless there's some reason you need the consistency of a tag rather than the index for reproducibility reasons.

hoomd/ParticleGroup.cc
```diff
@@ -789,6 +792,100 @@ unsigned int ParticleGroup::intersectionSize(std::shared_ptr<ParticleGroup> othe
...
+        // Generate a random velocity
+        Scalar mass =  h_vel.data[j].w;
+        Scalar sigma = fast::sqrt(kT / mass);
+        hoomd::NormalDistribution<Scalar> normal(sigma);
+        h_vel.data[j].x = normal(rng);
+        h_vel.data[j].y = normal(rng);
+        if (D > 2)
+            h_vel.data[j].z = normal(rng);
+        else
+            h_vel.data[j].z = 0; // For 2D systems
```
vyasr, line 842: I'm personally a fan of the ternary operator for such short snippets, especially when you're choosing not to enclose the if/else blocks in braces because they're so short (that's a clear sign for me that `? :` syntax is appropriate. Not sure if you have are stylistically opposed, though.

hoomd/ParticleGroup.cc
```diff
@@ -789,6 +792,100 @@ unsigned int ParticleGroup::intersectionSize(std::shared_ptr<ParticleGroup> othe
...
+    for (unsigned int group_idx = 0; group_idx < group_size; group_idx++)
+        {
+        unsigned int j = this->getMemberIndex(group_idx);
+        Scalar mass =  h_vel.data[j].w;
+        h_vel.data[j].x = h_vel.data[j].x - com_momentum.x / mass;
+        h_vel.data[j].y = h_vel.data[j].y - com_momentum.y / mass;
+        if (D > 2)
+            h_vel.data[j].z = h_vel.data[j].z - com_momentum.z / mass;
+        else
+            h_vel.data[j].z = 0; // For 2D systems
```
vyasr, line 886: Same ternary comment as above (either apply the change in both places or neither depending on your stylistic preference).

hoomd/ParticleGroup.h
```diff
@@ -315,6 +315,16 @@ class PYBIND11_EXPORT ParticleGroup
         */
         unsigned int intersectionSize(std::shared_ptr<ParticleGroup> other);
 
+        /** Thermalize the particle momenta
+
+            Draw gaussian random momenta for all particles in the group.
```
vyasr, line 320:
```suggestion
            Draw Gaussian random momenta for all particles in the group.
```
My personal preference, I don't know that capitalization is really standard but it's more common in my experience.

hoomd/md/TwoStepNPTMTK.cc
```diff
@@ -981,18 +981,13 @@ TwoStepNPTMTK::couplingMode TwoStepNPTMTK::getRelevantCouplings()
...
-void TwoStepNPTMTK::randomizeVelocities(unsigned int timestep)
+void TwoStepNPTMTK::thermalizeExtraDOF(unsigned int seed, unsigned int timestep)
     {
-    if (m_shouldRandomize == false)
-        {
-        return;
-        }
-
-    m_exec_conf->msg->notice(6) << "TwoStepNPTMTK randomizing velocities" << std::endl;
+    m_exec_conf->msg->notice(6) << "TwoStepNPTMTK randomizing extra DOF" << std::endl;
```
vyasr, line 986: I find the `extra` naming to be confusing, I have to read the class or the documentation in the header to recognize which DOF are under consideration. Could we instead name it to something like `thermalizeReservoirDOF`? I'm open to other names as well, I realize that my suggestion might be too broad if there are reservoirs (e.g. Gibbs ensemble simulations) where that's not quite right either, but I'd like the name to more clearly indicate what DOF are being considered here.

hoomd/md/TwoStepNVTMTK.cc
```diff
@@ -440,14 +440,9 @@ void TwoStepNVTMTK::advanceThermostat(unsigned int timestep, bool broadcast)
...
-void TwoStepNVTMTK::randomizeVelocities(unsigned int timestep)
+void TwoStepNVTMTK::thermalizeExtraDOF(unsigned int seed, unsigned int timestep)
     {
-    if (m_shouldRandomize == false)
-        {
-        return;
-        }
-
-    m_exec_conf->msg->notice(6) << "TwoStepNVTMTK randomizing velocities" << std::endl;
+    m_exec_conf->msg->notice(6) << "TwoStepNVTMTK randomizing extra DOF" << std::endl;
```
vyasr, line 445: Same naming note as in TwoStepNPTMTK

hoomd/md/methods.py
```diff
@@ -77,6 +83,14 @@ class NVT(_Method):
...
         tau (float): Coupling constant for the Nosé-Hoover thermostat. (in time
             units).
+
+        translational_thermostat_dof (tuple[float, float]): Additional degrees
+            of freedom for the translational thermostat (:math:`\xi`,
+            :math:`\eta`)
+
+        rotational_thermostat_dof (tuple[float, float]): Additional degrees
+            of freedom for the rotational thermostat (:math:`\xi_\mathrm{rot}`,
+            :math:`\eta_\mathrm{rot}`)
```
vyasr, line 90: We may want to include a sentence in the docstring explaining that these variables control the internal state of the thermostat. Not sure how much detail is appropriate, but at least mentioning that they're explained in detail in the paper.

vyasr: Aha these are well documented in the thermalize documentation. So maybe just add a note 
```suggestion
            :math:`\eta_\mathrm{rot}`) (see `~.thermalize_extra_dof` for details).
```
Also add to the translational DOF.

hoomd/md/methods.py
```diff
@@ -272,6 +320,18 @@ class NPT(_Method):
...
+
+        barostat_dof (tuple[float, float, float, float, float, float]):
+            Additional degrees of freedom for the barostat (:math:`\nu_{xx}`,
+            :math:`\nu_{xy}`, :math:`\nu_{xz}`, :math:`\nu_{yy}`,
+            :math:`\nu_{yz}`, :math:`\nu_{zz}`)
```
vyasr, line 334: Same documentation suggestion from the NVT class, add links to the `thermalize*` method that describes these variables.

hoomd/md/methods.py
```diff
@@ -339,6 +409,37 @@ def __preprocess_stress(self,value):
         else:
             return (value,value,value,0,0,0)
 
+    def thermalize_extra_dof(self, seed):
```
vyasr, line 412: The meaning of `extra` is more obvious here than in the C++ because the docstring is right there; maybe it's OK as is, I'd still prefer a more meaningful name but I wouldn't waste too long on it if you don't like my suggestion and can't think of a better one.

hoomd/md/pytest/test_methods.py
```diff
@@ -273,6 +285,92 @@ def test_npt_attributes_attached_3d(simulation_factory,
...
+                               kT=constant_t,
+                               tau=2.0,
+                               S=constant_s,
+                               tauS=2.0,
+                               box_dof=[True, True, True, True, True, True],
+                               couple='xyz')
+
+    sim = simulation_factory(two_particle_snapshot_factory())
+    sim.operations.integrator = hoomd.md.Integrator(0.005, methods=[npt])
+    sim.operations.schedule()
```
vyasr, line 317: This isn't necessary anymore, right? Or has the relevant PR not been merged yet? In which case maybe just need to keep this in mind when merging things together.

hoomd/md/pytest/test_methods.py
```diff
@@ -273,6 +285,92 @@ def test_npt_attributes_attached_3d(simulation_factory,
...
+
+    npt.thermalize_extra_dof(100)
+    xi, eta = npt.translational_thermostat_dof
+    assert xi != 0.0
+    assert eta == 0.0
+
+    xi_rot, eta_rot = npt.rotational_thermostat_dof
+    assert xi_rot == 0.0
+    assert eta_rot == 0.0
```
vyasr, line 326: You can always validate the tuple directly (since IIRC it is a tuple and not a list), but no need to change it now, more for future reference.

hoomd/md/pytest/test_methods.py
```diff
@@ -273,6 +285,92 @@ def test_npt_attributes_attached_3d(simulation_factory,
...
+    npt.thermalize_extra_dof(100)
+    xi, eta = npt.translational_thermostat_dof
+    assert xi != 0.0
+    assert eta == 0.0
+
+    xi_rot, eta_rot = npt.rotational_thermostat_dof
+    assert xi_rot != 0.0
+    assert eta_rot == 0.0
+
+    print(npt.barostat_dof)
```
vyasr, line 370:
```suggestion

```

hoomd/md/pytest/test_methods.py
vyasr: Same applies to other calls to `schedule` below.

vyasr (changes requested):
This looks quite good! The user is now responsible for making sure that they use the same temperature for the thermostat and the randomization, correct? I think that's in line with what we originally discussed, just verifying.

hoomd/ParticleGroup.cc
joaander: Done.

hoomd/ParticleGroup.cc
joaander: The fast and slow sqrt's have more performance implications on the GPU - they should be identical on the CPU. Still, this code may be copied to the GPU some day and a few precise sqrts per particle will not slow performance significantly in this rarely called method. I replaced them with slow::.

hoomd/ParticleGroup.cc
joaander: This method acts locally on each rank. To prevent the same random numbers from being used we either need to use tags or include the MPI rank as a seed. Using the tags is consistent with how BD and DPD seed their RNGs, so that was propagated here. I don't think one additional memory lookup is going to hurt performance much in a rarely called method. Plus, as you say it does provide some level of consistency across serial and MPI runs.

hoomd/ParticleGroup.cc
joaander: I'm not a fan of the ternary operator. I'll run clang-format on this function and see what the style rules we have set do with these blocks.

hoomd/md/methods.py
joaander: I'm open to suggestions for better names. "Extra" is a catch-all that applies to both the thermostat and barostat. I didn't put too much thought into it. Reservoir works for a thermostat but I'm not sure that is the most natural name for the barostat. Maybe "additional"? Would be best if it isn't so generic. Will think on it.

hoomd/md/pytest/test_methods.py
joaander: Right, I made it explicit when I was being careful to ensure that the two variables appeared in the correct order in the tuple. They have different layouts in C++ between NVT and NPT.

hoomd/ParticleGroup.cc
vyasr: Agreed, no need to optimize here really. The consistency is good.

hoomd/md/methods.py
vyasr: @b-butler any thoughts?

hoomd/ParticleGroup.h
joaander: Yes it is standard, I don’t know why my spell checker didn’t catch it.

hoomd/md/methods.py
joaander: We could be really explicit and avoid fluid representations using `NVT.thermalize_thermostat_dof` and `NPT.thermalize_thermostat_and_barostat_dof`. That matches the names of the parameter attriburtes.

hoomd/md/pytest/test_methods.py
joaander: Yes I should remove this now that we decided run(0) is the preferred solution.

hoomd/ParticleGroup.cc
joaander: clang-format leaves these simple if's as is with the current configuration. We can discuss at a later date what the style settings should be before we run clang-format on the code base.

hoomd/md/TwoStepNPTMTK.cc
joaander: I moved the introduction of the thermostat and barostat degrees of freedom up into the main NVT and NPT documentation, and renamed the methods.

hoomd/md/methods.py
joaander: I moved the longer description up into the main class documentation so it appears first and references the attribute doc.

joaander:
Thanks for the review! These are all great comments and I have addressed them.

hoomd/ParticleGroup.cc
vyasr: OK sounds good.

vyasr (approved):

PR #790, opened 2020-10-01
Add documentation to hoomd.trigger

syjlee:
## Description

There should be documentation for Trigger classes for public facing properties, and methods. 
In addition, class docstrings need to be created

## Motivation and context
Resolves: #718 

## How has this been tested?
Should be checked with locally installed sphinx package, and double check with generated html

## Checklist:


syjlee:
This branch is not ready to be reviews. Will request review once it's ready.

syjlee:
1. I'm getting the error: `WARNING: py:class reference target not found: hoomd.Trigger`
What is the right reference target for a trigger or a list of triggers as arguments?

2. What would be an appropriate example for `hoomd.trigger.Not`?

Also, could you leave comments what to fix, add or remove?
@joaander @b-butler

joaander:
> 1. I'm getting the error: `WARNING: py:class reference target not found: hoomd.Trigger`
>    What is the right reference target for a trigger or a list of triggers as arguments?
`hoomd.trigger.Trigger`

> 2. What would be an appropriate example for `hoomd.trigger.Not`?
Perhaps not(on(0))? I can't think of any common use cases for not. Some complex logic can sometimes best be represented by not(and(...)) or not(or(...)).

hoomd/trigger.py
```diff
@@ -2,15 +2,45 @@
 # This file is part of the HOOMD-blue project, released under the BSD 3-Clause
 # License.
 
+""" :py:class:`hoomd.trigger` enables users to design time points when and while
```
joaander, line 5:
```suggestion
"""Triggers enable users to design time points when and while
```

hoomd/trigger.py
```diff
@@ -2,15 +2,45 @@
...
+    Triger returns `True`.
+
+"""
+
 from hoomd import _hoomd
 from inspect import isclass
 
 
-class Trigger(_hoomd.Trigger):
+class Trigger(_hoomd.Trigger):  # noqa: D101
```
joaander, line 16: We should document this class.
Something like:
```
"""Base class trigger.

Provides methods common to all triggers.

Attention:
    Users should instantiate the subclasses, using `Trigger` directly
    will result in an error.
"""
```

We also need to document the `compute` and `__call__` methods. I need to do some checking and see if I can figure out how to do that with the way we set these classes up in sphinx.

hoomd/trigger.py
```diff
@@ -2,15 +2,45 @@
...
     pass
 
 
 class Periodic(_hoomd.PeriodicTrigger, Trigger):
+    R""" Set periodicity of trigger.
```
joaander, line 21:
```suggestion
    """Trigger periodically.
```

hoomd/trigger.py
```diff
@@ -2,15 +2,45 @@
 # This file is part of the HOOMD-blue project, released under the BSD 3-Clause
 # License.
 
+""" :py:class:`hoomd.trigger` enables users to design time points when and while
+    :py:class:`hoomd.Operations` operates. The composite of triggers takes
```
joaander, line 6:
```suggestion
   `hoomd.Operations` operates. The composite of triggers takes 
```
You don't need to preface links. Sphinx can find the right target without the identifier and this is easier to read.

hoomd/trigger.py
```diff
@@ -2,15 +2,45 @@
...
 
 class Periodic(_hoomd.PeriodicTrigger, Trigger):
+    R""" Set periodicity of trigger.
+
+    Args:
+        period (float): timesteps for periodicity
+        phase (float): timesteps for phase
+    
+    :py:class:`hoomd.Operations` will operate when the Periodic trigger returns
+    `True` every `period` steps ((timestep - phase)/period = 0)
```
joaander, line 28:
```suggestion
    `True` every `period` steps ``((t - phase) % period = 0``)
```

hoomd/trigger.py
```diff
@@ -20,6 +50,29 @@ def __str__(self):
 
 
 class Before(_hoomd.BeforeTrigger, Trigger):
+    R""" Set the timepoint to finish triger.
```
joaander, line 53:
```suggestion
    """Trigger every step before a given step.
```

hoomd/trigger.py
```diff
@@ -2,15 +2,45 @@
 # This file is part of the HOOMD-blue project, released under the BSD 3-Clause
 # License.
 
+""" :py:class:`hoomd.trigger` enables users to design time points when and while
+    :py:class:`hoomd.Operations` operates. The composite of triggers takes 
+    time steps and returns `True` or `False`. The operation will perform when 
+    Triger returns `True`.
+
```
joaander, line 9:
```suggestion
```
You shouldn't add these blank lines at the end of the docstrings.

joaander (changes requested):
Thanks! This is a great start. I left some comments on the first few triggers. You can apply the same pattern to the remaining ones.

I'll check tomorrow on how to document `compute` and `__call__`. I'll also add some documentation on user defined triggers.

- [x] Document `compute` and `__call__`
- [ ] Provide an example user defined trigger.

joaander:
@syjlee I saw you are out today so I revised the docs myself so we can get this finished and included in the first beta release. I took out the GSD, CSV, and other classes from the examples, I think it is cleaner with just the trigger specifications.

joaander (approved):
Thanks!

syjlee:
> @syjlee I saw you are out today so I revised the docs myself so we can get this finished and included in the first beta release. I took out the GSD, CSV, and other classes from the examples, I think it is cleaner with just the trigger specifications.

Thanks Josh!

PR #793, opened 2020-10-02
Document particle filters

### LINKED ISSUE #719: Add documentation to hoomd.filter ParticleFilters
## Purpose

This module needs to have better documentation of its `ParticleFilter` fclasses or public facing properties, and methods. In addition, class docstrings need to be added.

## Tasks

- [ ] Expand docstring for `hoomd.State`.
- [ ] Add docstrings for pubilic (not leading with an underscore) methods and properties of `hoomd.filter` `ParticleFilter` subclasses, or after discussion explicitly mark code with the appropriate error code.
- [ ] Add files to CI pydocstyle checking.
- [ ] Enable inheriting on sphinx doc builds for `ParticleFilter` subclasses. 

## Pertinent Information

* Information about how particle filters works can be found in the new HOOMD-blue tutorial [here](https://github.com/glotzerlab/hoomd-examples/pull/12). This should provide some idea about how the class in question and many of its methods work.
* The main code for the classes can be seen in the C++ files located in the `hoomd/filter` directory.
````python
def foo(bar):  # noqa: D103
    pass

class Baz:  # noqa: D101
    def egg(self):  # noqa: D102
        pass
````

### PR DESCRIPTION
**Pull request by joaander**
## Description

Document particle filters and make the base class `ParticleFilter` (from `_ParticleFilter`)

## Motivation and context

HOOMD-blue functionality should be documented.

Resolves #719

## How has this been tested?

I added filter/*.py to the pydocstyle tests.

The sphinx documentation builds and renders correctly. Note that I explicitly name the base classes because the sphinx option `:show-inerheritance` does not work. 1) It includes the base class with a full name (e.g. `hoomd.filter.filter_.ParticleFilter` which fails to link because this is not where the class appears in the documentation. 2) It lists all base classes, including the multiply inherited C++ base class which the user does not need to be aware of.

## Checklist:

### DIFF
--- a/hoomd/filter/filter_.py
+++ b/hoomd/filter/filter_.py
@@ -1,21 +1,32 @@
-from hoomd._hoomd import ParticleFilter
+"""Define ParticleFilter."""
+
+from hoomd import _hoomd
 
 
-class _ParticleFilter(ParticleFilter):
-    """Base class for all particle filters.
-
-    Should not be instantiated or subclassed by users.
-    """
+class ParticleFilter(_hoomd.ParticleFilter):
+    """Base class for all particle filters."""
     def __hash__(self):
+        """Return a hash of the filter parameters."""
         return NotImplementedError("Must implement hash for ParticleFilters.")
 
     def __eq__(self, other):
+        """Test for equality between two particle filters."""
         raise NotImplementedError(
             "Equality between {} is not defined.".format(self.__class__))
 
     def __str__(self):
+        """Format a human readable string describing the filter."""
         return "ParticleFilter.{}".format(self.__class__.__name__)
 
     def __call__(self, state):
-        """Needs to interact with state to get particles across MPI rank."""
+        """Evaluate the filter.
+
+        Returns:
+            list[int]: The particle tags selected by this filter.
+
+        Note:
+            This method may return tags that are only present on the local MPI
+            rank. The full set of particles selected is the combination of
+            these the lists across ranks with a set union operation.
+        """
         return self._get_selected_tags(state._cpp_sys_def)
--- a/hoomd/filter/filter_.py
+++ b/hoomd/filter/filter_.py
@@ -4,7 +4,14 @@ from hoomd import _hoomd
 
 
 class ParticleFilter(_hoomd.ParticleFilter):
-    """Base class for all particle filters."""
+    """Base class for all particle filters.
+
+    This class provides methods common to all particle filters.
+
+    Attention:
+        Users should instantiate one of the subclasses. Calling `ParticleFilter`
+        directly may result in an error.
+    """
     def __hash__(self):
         """Return a hash of the filter parameters."""
         return NotImplementedError("Must implement hash for ParticleFilters.")
--- a/hoomd/filter/set_.py
+++ b/hoomd/filter/set_.py
@@ -1,11 +1,14 @@
-from hoomd.filter.filter_ import _ParticleFilter
+"""Define particle filter set operations."""
+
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: hoomd/filter/set_.py [hunk 1]
```diff
@@ -49,16 +52,49 @@ def __eq__(self, other):
 
 class SetDifference(_ParticleFilterSetOperations,
                     _hoomd.ParticleFilterSetDifference):
+    r"""Set difference operation.
+
+    Args:
+        f (ParticleFilter): First set in the difference.
+        g (ParticleFilter): Second set in the difference.
+
+    `SetDifference` is a comosite filter. It selects particles in the set
```
b-butler, line 61:
```suggestion
    `SetDifference` is a composite filter. It selects particles in the set
```

hoomd/filter/set_.py
```diff
@@ -49,16 +52,49 @@ def __eq__(self, other):
...
 
 
 class Union(_ParticleFilterSetOperations, _hoomd.ParticleFilterUnion):
+    r"""Set union operation.
+
+    Args:
+        f (ParticleFilter): First set in the union.
+        g (ParticleFilter): Second set in the union.
+
+    `Union` is a comosite filter. It selects particles in the set
```
b-butler, line 77:
```suggestion
    `Union` is a composite filter. It selects particles in the set
```

hoomd/filter/set_.py
```diff
@@ -49,16 +52,49 @@ def __eq__(self, other):
...
 
 class Intersection(_ParticleFilterSetOperations,
                    _hoomd.ParticleFilterIntersection):
+    r"""Set intersection operation.
+
+    Args:
+        f (ParticleFilter): First set in the intersection.
+        g (ParticleFilter): Second set in the intersection.
+
+    `Intersection` is a comosite filter. It selects particles in the set
```
b-butler, line 94:
```suggestion
    `Intersection` is a composite filter. It selects particles in the set
```

hoomd/filter/tags.py
```diff
@@ -1,21 +1,34 @@
...
+"""Define the Tags filter."""
+
+from hoomd.filter.filter_ import ParticleFilter
 from hoomd._hoomd import ParticleFilterTags
 import numpy as np
 
 
-class Tags(_ParticleFilter, ParticleFilterTags):
+class Tags(ParticleFilter, ParticleFilterTags):
+    """Select particles by tag.
```
b-butler, line 9: It may be worth going into what a particle tag is briefly, and how to get it from the state.

hoomd/filter/filter_.py
```diff
@@ -1,21 +1,32 @@
...
 
+from hoomd import _hoomd
 
-class _ParticleFilter(ParticleFilter):
-    """Base class for all particle filters.
 
-    Should not be instantiated or subclassed by users.
-    """
+class ParticleFilter(_hoomd.ParticleFilter):
+    """Base class for all particle filters."""
```
b-butler, line 7: I think we should still document that users should not create instances of this class. If they do and use it as a particle filter, it won't break right away. It will only break when attaching and raise a pybind11 error about the constructor of `_hoomd.ParticleFilter` from what I can glean from the code.

b-butler (changes requested):
I have left a few comments.

hoomd/filter/tags.py
joaander: Done.

hoomd/filter/filter_.py
joaander: Done.

PR #795, opened 2020-10-05
Revise docs for LoadBalancer, Clusters, and HPMC integrators

### LINKED ISSUE #715: Review HPMC integrator documentation
## Purpose

Review HPMC integrator documentation. For example, list and tuple are capitalized in some places in the docs where they should be lowercase.

## Tasks

- [ ] Review docstrings for classes.

## Pertinent Information

* Information about how the class works can be found in the new HOOMD-blue tutorial [here](https://github.com/glotzerlab/hoomd-examples/pull/12). This should provide some idea about how the class in question and many of its methods work.
````python
def foo(bar):  # noqa: D103
    pass

class Baz:  # noqa: D101
    def egg(self):  # noqa: D102
        pass
````

### PR DESCRIPTION
**Pull request by joaander**
## Description

Apply the latest documentation patterns and otherwise revise the documentation for LoadBalancer, Clusters, and HPMC integrators.

## Motivation and context

This makes the HOOMD documentation more consistent.

Resolves #715, #723

## How has this been tested?

The sphinx docs render correctly.

## Checklist:

### CODE REVIEW

File: hoomd/hpmc/integrate.py [hunk 1]
```diff
@@ -206,12 +206,14 @@ def map_energies(self):
...
 
+        Attention:
+            `map_energies` is not yet implemented in HOOMD v3.x.
+
         Example:
             mc = hpmc.integrate.shape(...)
             mc.shape_param.set(...)
             energy_map = np.asarray(mc.map_energies())
         """
-
```
b-butler, line 214: I think we need a pass here?

hoomd/hpmc/integrate.py
```diff
@@ -254,6 +255,9 @@ def test_overlap(self,
 
         For two-dimensional shapes, pass the third dimension of **rij** as zero.
 
+        Attention:
+            `test_overlap` is not yet implemented in HOOMD v3.x.
```
b-butler, line 259: We should probably comment the code out or just use `return` to prevent the code from running if a user uses the method.

hoomd/hpmc/integrate.py
```diff
@@ -748,14 +747,20 @@ class Polyhedron(_HPMCIntegrator):
...
         nselect (int): Number of trial moves to perform per particle per
             timestep.
 
-    Note:
+    Perform hard particle Monte Carlo of general polyhedra defined by their
+    vertices and faces (see `shape`). `Polyhedron` supports triangle meshes and
+    spheres only. The mesh must be free of self-intersections.
+
+    See Also:
+        Use `ConvexPolyhedron` for faster performance with convex polyhedra.
```
b-butler, line 761: Did we ever decide if we should encourage users to just use the `ConvexSpheropolyhedron` class? If I remember correctly it is the same under the hood when the sweep radius = 0.

hoomd/hpmc/update.py
```diff
@@ -744,7 +744,18 @@ def __init__(self, mc, external_lattice, period=1):
 
 
 class Clusters(_Updater):
-    R""" Equilibrate the system according to the geometric cluster algorithm (GCA).
+    """Apply geometric cluster algorithm (GCA) moves.
+
+    Args:
+        seed (int): Random number seed.
+        swap_types(list): A pair of two types whose identities may be swapped.
```
b-butler, line 751:
```suggestion
        swap_types (list[tuple[str, str]]): A pair of two types whose identities may be swapped.
```

hoomd/hpmc/update.py
```diff
@@ -744,7 +744,18 @@ def __init__(self, mc, external_lattice, period=1):
...
+        move_ratio(float): Set the ratio between pivot and reflection moves.
```
b-butler, line 752:
```suggestion
        move_ratio (float): Set the ratio between pivot and reflection moves.
```

hoomd/hpmc/update.py
```diff
@@ -744,7 +744,18 @@ def __init__(self, mc, external_lattice, period=1):
...
+        flip_probability(float): Set the probability for transforming an
```
b-butler, line 753:
```suggestion
        flip_probability (float): Set the probability for transforming an
```

hoomd/hpmc/update.py
```diff
@@ -744,7 +744,18 @@ def __init__(self, mc, external_lattice, period=1):
...
+                                 individual cluster.
+        swap_move_ratio(float): Set the ratio between type swap moves and
```
b-butler, line 755:
```suggestion
        swap_move_ratio (float): Set the ratio between type swap moves and
```

hoomd/hpmc/update.py
```diff
@@ -760,31 +771,25 @@ class Clusters(_Updater):
...
+    The `Clusters` updater support threaded execution on multiple CPU cores.
+
+    Attributes:
         seed (int): Random number seed.
         swap_types(list): A pair of two types whose identities may be swapped.
         move_ratio(float): Set the ratio between pivot and reflection moves.
         flip_probability(float): Set the probability for transforming an
                                  individual cluster.
         swap_move_ratio(float): Set the ratio between type swap moves and
                                 geometric moves.
```
b-butler, line 790: Spaces between name and type info.

hoomd/hpmc/update.py
```diff
@@ -854,15 +859,15 @@ def _attach(self):
...
 
         Returns:
             A counter object with pivot, reflection, and swap properties. Each
             property is a list of accepted moves and rejected moves since the
             last run.
 
-        Note::
-            if the updater is not attached None will be returned.
+        Note:
+            `None` when the simulation run has not stated.
```
b-butler, line 870:
```suggestion
            `None` when the simulation run has not started.
```

hoomd/hpmc/update.py
```diff
@@ -949,9 +951,9 @@ class QuickCompress(_Updater):
         L_current``
 
     Tip:
-        Use the MoveSizeTuner (TODO: make reference) in conjunction with
-        `QuickCompress` to adjust the move sizes to maintain a constant
-        acceptance ratio as the density of the system increases.
+        Use the `MoveSizeTuner` in conjunction with `QuickCompress` to adjust
```
b-butler, line 954: Have we decided to use shorter links like `MoveSizeTuner` or longer ones like `hoomd.hpmc.tune.MoveSizeTuner`. I am fine either way, but I have been using the longer format.

hoomd/tune/balance.py
```diff
@@ -1,48 +1,41 @@
...
 from hoomd.parameterdicts import ParameterDict
-from hoomd.typeconverter import OnlyType
 from hoomd.trigger import Trigger
 from hoomd import _hoomd
+import hoomd
 
 
 class LoadBalancer(_Tuner):
-    R""" Adjusts the boundaries of a domain decomposition on a regular 3D grid.
+    R""" Adjusts the boundaries of the domain decomposition.
```
b-butler, line 11:
```suggestion
    r""" Adjusts the boundaries of the domain decomposition.
```

b-butler (changes requested):
The documentation is definitely improved. I have left some comments to address.

hoomd/hpmc/integrate.py
joaander: I added a `NotImplementedError` exception.

hoomd/hpmc/integrate.py
joaander: I added a `NotImplementedError` exception.

hoomd/hpmc/integrate.py
joaander: Yes, the overlap check is identical. ConvexSpheropolyhedron implements a few methods differently, but I think these are strict superset of ConvexPolyhedron when `N_verts > 2` and `sweep_radius=0`.  We can deal with this removal in a later PR: #798

hoomd/hpmc/update.py
joaander: Explicit is better than implicit and the full namespace is copy and pastable into user scripts. We should use the full namespace.

hoomd/hpmc/update.py
joaander: Done.

hoomd/hpmc/update.py
joaander: I do use the class name only when referring to the class currently being documented. I think that is fine given the local context.

PR #805, opened 2020-10-13
Remove underscore from public classes

b-butler:
## Description

Many classes implemented for v3 are base classes that should not be instantiated, but since we expose many of these classes to avoid duplication of documentation and accurately document the type expected by certain attributes and methods, we have decided to make them fully public facing with the caveat they should not be instantiated except through their subclasses.

This is not as big a concern as it may seem in exposing internals as the real meat of the internal object structure is mediated by `hoomd.operation._BaseHOOMDObject` which is still completely private.

## Motivation and context

Makes classes that are user facing follow Python class name standards.

## How has this been tested?

Since this is just renaming, existing tests pass.

## Change log

```

```

## Checklist:


joaander (approved):
Thanks!

PR #808, opened 2020-10-16
Switch CMake to use C++14 standard. Resolves #784.

### LINKED ISSUE #784: Build with C++14 for Thrust
## Description

I am using a new version of Thrust (`THRUST_VERSION 100909`) that requires C++14. Compiler support is available in gcc 5, clang 3.4, nvcc 9, xlc 13.1.2. https://en.cppreference.com/w/cpp/compiler_support

## Proposed solution

Upgrade CMake files to use `--std=c++14` instead of `c++11`.

## Additional context

@joaander mentioned the potential drawback of no `manylinux` support with C++14. If that is relevant, there are newer images `manylinux2014` and `manylinux2020` that can be used with full C++14 support. https://cibuildwheel.readthedocs.io/en/stable/cpp_standards/#manylinux1-and-c14

## Developer

@joaander recommends investigating this after releasing 3.0 beta 1.

bdice:
## Description

This resolves #784 by using the C++14 standard.

## Motivation and context

Newer version of Thrust and TBB require C++14. Compiler support is available in gcc 5, clang 3.4, nvcc 9, xlc 13.1.2. https://en.cppreference.com/w/cpp/compiler_support
Resolves: #784

## How has this been tested?

I'm letting CI try this with GCC 4.8. It may support enough C++14 features for HOOMD to build correctly, but I will remove GCC 4.8 from CI in this PR if it fails.

## Change log

```
Changed:
- Building from source requires a C++14 compatible compiler.
```

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,7 +31,7 @@ functionality.
 
 *Changed*
 
-- The `run` method has minimal overhead
+- The ``run`` method has minimal overhead
 - All loggable quantities are directly accessible as object properties.
 - Operation parameters are always synchronized.
 - Operations can be instantiated without a device or MPI communicator.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,10 +56,10 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     "MinSizeRel" "RelWithDebInfo")
 endif()
 
-# enable c++11
-set(CMAKE_CXX_STANDARD 11)
+# enable c++14
+set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CUDA_STANDARD 11)
+set(CMAKE_CUDA_STANDARD 14)
 
 # Enable compiler warnings on gcc and clang (common compilers used by developers)
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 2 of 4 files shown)


joaander (commented):
Thanks. There is a `if (CMAKE_CXX_STANDARD LESS 14)` block in `jit/CMakeLists.txt` that is no longer needed.

Also, we should verify that the `set(CMAKE_CXX_STANDARD 11)` lines in `hipper` and `neighbor` don't reset this. These are both header-file only sumobuldes but on quick glance, it looks like we just set the include directories and don't run their embedded cmake code. Still, it is worth grepping for `--std` in the `Makefile` or `rules.ninja` to make sure it is set properly for both CC and NVCC compiled files.

mphoward:
> Also, we should verify that the `set(CMAKE_CXX_STANDARD 11)` lines in `hipper` and `neighbor` don't reset this.

I believe I did not include either of these as submodules because it got very complicated to control installation. I don't mind bumping those submodules to C++14 standard as well if you'd like to open an issue or PR.

bdice:
@joaander Should this be updated to `c++14` too? I am not as familiar with this part of the code.
https://github.com/glotzerlab/hoomd-blue/blob/f51fa112cc0d0caba9fc06dfd502a9477bc5b579/hoomd/jit/patch.py#L125
https://github.com/glotzerlab/hoomd-blue/blob/f51fa112cc0d0caba9fc06dfd502a9477bc5b579/hoomd/jit/patch.py#L229-L232

joaander:
> @joaander Should this be updated to `c++14` too? I am not as familiar with this part of the code.
> https://github.com/glotzerlab/hoomd-blue/blob/f51fa112cc0d0caba9fc06dfd502a9477bc5b579/hoomd/jit/patch.py#L125
> 
> 
> https://github.com/glotzerlab/hoomd-blue/blob/f51fa112cc0d0caba9fc06dfd502a9477bc5b579/hoomd/jit/patch.py#L229-L232

Yes, these should be updated. They JIT code needs to be ABI compatible with the host code.

bdice:
@mphoward I opened preliminary PRs to `neighbor` and `hipper` (there many be other places in those code bases that need to be updated, I only checked CMakeLists.txt). I don't think it's an issue for HOOMD to require C++14 and external dependencies to require C++11, but it's probably helpful to move those libraries up for future-proofing and enabling new features (e.g. if either library uses Thrust or other libraries that require C++14).
https://github.com/bdice/hipper/pull/1
https://github.com/bdice/neighbor/pull/1

joaander (approved):
Thanks!

PR #820, opened 2020-10-20
Handle kernel interrupt signals in Jupyter

### LINKED ISSUE #813: Run won't stop when interrupting Jupyter kernel
## Description

I ran the HPMC tutorial. During the call to `sim.run(2e6)`, I tried to stop the Jupyter kernel, but the SIGINT doesn't seem to pass through to HOOMD. I used the "stop kernel" button which is supposed to trigger a `KeyboardInterrupt`. I tested this with Jupyter Notebook and Jupyter Lab and saw the problem in both. The SIGINT / `KeyboardInterrupt` stops the run as expected in a Python shell and in an IPython shell. It's entirely possible this is a bug in Jupyter, I don't know.

## Script

See here: https://github.com/glotzerlab/hoomd-examples/blob/v3-api/00-Introducing-HOOMD-blue/06-Equilibrating-the-System.ipynb

## Output

Expected the run to stop. It did not.

```
[I 14:50:41.831 LabApp] Kernel interrupted: e051d5c2-ab07-4f30-bd18-26476a5d27d2
```

## Additional context

This snippet might help debug things. Be aware that it catches all signals. If you run `time.sleep(10)`, it captures and prints the signals as they arrive during the sleep. However, I didn't see it printing the captured signals during `sim.run`. Could be related?

```python
import signal
import time

def sighandler(signal, frame):
    print("signal", signal, frame)
    return

catchable_sigs = set(signal.Signals)
for sig in catchable_sigs:
    try:
        signal.signal(sig, sighandler)
        print(f"Setting {repr(sig)}, {sig}")
    except (ValueError, OSError, RuntimeError) as m:
        print(f"Skipping {repr(sig)}, {sig}")
```

## Configuration

Platform:
- WSL2 (Ubuntu), CPU

Installation method:
- Compiled from source

Jupyter / related package versions:
```
ipython                   7.18.1           py38h1cdfbd6_0    conda-forge
jupyter-server-proxy      1.5.0                      py_0    conda-forge
jupyter_client            6.1.7                      py_0    conda-forge
jupyter_core              4.6.3            py38h32f6830_2    conda-forge
jupyter_sphinx            0.3.1            py38h32f6830_0    conda-forge
jupyterlab                2.2.8                      py_0    conda-forge
jupyterlab_code_formatter 1.3.6                      py_0    conda-forge
jupyterlab_pygments       0.1.2              pyh9f0ad1d_0    conda-forge
jupyterlab_server         1.2.0                      py_0    conda-forge
```

Versions:
- Python version: 3.8.3
- **HOOMD-blue** version: 3.0.0-beta1

joaander:
## Description

Install the `SIGINT` signal handler at the start of every call to `run`.

## Motivation and context

See the discussion in #813 

Resolves #813

## How has this been tested?

I tested this manually with a jupyter notebook.

## Change log

```
Fixed:
- ``Simulation.run`` now ends with a ``KeyboardInterrupt`` exception when Jupyter interrupts the kernel.
```

## Checklist:

### DIFF
--- a/hoomd/SignalHandler.cc
+++ b/hoomd/SignalHandler.cc
@@ -7,6 +7,7 @@
 #include <signal.h>
 #include "SignalHandler.h"
 #include <iostream>
+#include "string.h"
 
 using namespace std;
 
@@ -14,8 +15,8 @@ using namespace std;
     \brief Defines variables and functions related to handling signals
 */
 
-//! Tracks the previous signal handler that was set to make a chain
-void (*prev_sigint_handler)(int) = NULL;
+//! Tracks the previous signal handler
+static struct sigaction g_oldact;
 
 volatile sig_atomic_t g_sigint_recvd = 0;
 
@@ -26,33 +27,38 @@ extern "C" void sigint_handler(int sig)
     if (sig != SIGINT)
         return;
 
-    // call the previous signal handler, but only if it is well defined
-    if (prev_sigint_handler && prev_sigint_handler != SIG_ERR && prev_sigint_handler != SIG_DFL && prev_sigint_handler != SIG_IGN)
-        prev_sigint_handler(sig);
+    std::cout << "HOOMD caught SIGINT" << std::endl;
 
     // set the global
     g_sigint_recvd = 1;
     }
 
-/*! Call only once at the start of program execution. This method
-    installs a signal handler for SIGINT that will set \c g_sigint_recvd
-    to 1. It will also call the previously set signal handler.
+/*! This method installs a signal handler for SIGINT that will set \c g_sigint_recvd to 1.
 */
 void InstallSIGINTHandler()
     {
-    void (*retval)(int) = NULL;
-    retval = signal(SIGINT, sigint_handler);
+    struct sigaction newact;
+    newact.sa_handler = sigint_handler;
+    sigemptyset(&newact.sa_mask);
+    newact.sa_flags = 0;
 
-    if (retval == SIG_ERR)
+    int retval = sigaction(SIGINT, &newact, &g_oldact);
+
+    if (retval != 0)
         {
-        cerr << "Error setting signal handler" << endl;
+        cerr << "Error setting signal handler: " << strerror(errno) << endl;
         return;
         }
+    }
 
-    // set the previous signal handler, but only if it is not the same as the
-    // one we just set. That would make for a fun infinite loop!
-    if (retval != sigint_handler)
-        prev_sigint_handler = retval;
-    else
-        prev_sigint_handler = NULL;
+void RemoveSIGINTHandler()
+    {
+    struct sigaction dummy_action;
+    int retval = sigaction(SIGINT, &g_oldact, &dummy_action);
+
+    if (retval != 0)
+        {
+        cerr << "Error setting signal handler: " << strerror(errno) << endl;
+        return;
+        }
     }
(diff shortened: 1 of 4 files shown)


hoomd/SignalHandler.cc
```diff
@@ -7,16 +7,14 @@
 #include <signal.h>
 #include "SignalHandler.h"
 #include <iostream>
+#include "string.h"
```
bdice, line 10: According to [this post](https://stackoverflow.com/questions/8380805/difference-between-string-h-and-cstring), it's probably better to use `<cstring>` in C++. `<cstring>` _must_ place the functions in the standard namespace and _may_ place the functions in the global namespace, but that's not going to cause any differences here since you have `using namespace std;` below this.
```suggestion
#include <cstring>
```

hoomd/SignalHandler.cc
```diff
@@ -7,16 +7,14 @@
 #include <signal.h>
```
bdice, line 7: This relies on the POSIX standard for `<signal.h>` to have `sigaction`, so unlike in my comment below, I don't think this can be replaced by `<csignal>`.

hoomd/SignalHandler.cc
```diff
@@ -26,33 +24,34 @@ extern "C" void sigint_handler(int sig)
...
-        prev_sigint_handler = NULL;
+ScopedSignalHandler::~ScopedSignalHandler()
+    {
+    struct sigaction dummy_action;
+    int retval = sigaction(SIGINT, &m_old_action, &dummy_action);
+
+    if (retval != 0)
+        {
+        cerr << "Error setting signal handler: " << strerror(errno) << endl;
+        return;
```
bdice, line 55: Is this line necessary?
```suggestion
```

hoomd/SignalHandler.cc
```diff
@@ -26,33 +24,34 @@ extern "C" void sigint_handler(int sig)
...
+    newact.sa_flags = 0;
 
-    if (retval == SIG_ERR)
+    int retval = sigaction(SIGINT, &newact, &m_old_action);
+
+    if (retval != 0)
         {
-        cerr << "Error setting signal handler" << endl;
+        cerr << "Error setting signal handler: " << strerror(errno) << endl;
         return;
```
bdice, line 43: Unnecessary.
```suggestion
```

hoomd/SignalHandler.h
```diff
@@ -22,7 +22,22 @@
 */
 extern volatile sig_atomic_t g_sigint_recvd;
 
-//! Installs the signal handler
-void InstallSIGINTHandler();
+/** Manage the signal handler within a scope
+
+    This allows System::run to install the signal handler and have in removed when it returns
+    or an excpetion is thrown.
```
bdice, line 28:
```suggestion
    This allows System::run to install the signal handler and have it removed when it returns
    or an exception is thrown.
```

bdice (approved):
This worked great for me! Really clean fix, the RAII style is nice. I have a few minor comments that can be addressed how you see fit, otherwise I approve this.

hoomd/SignalHandler.cc
joaander: When working on this, I noticed that the man page for `signal` suggests using `sigaction` instead. I also found a reference that `signal` and `sigaction` might not interoperate which I thought might be causing the issues with Jupyter.

Issue #732, opened 2020-08-21
Unit test dump.gsd


Charlottez112:
Unit test the class dump.gsd

Issue #811, opened 2020-10-19
Use specific exception classes where appropriate


bdice:
## Description

I tried to read a file that doesn't exist with `sim.create_state_from_gsd`. Should GSD file reading errors be `OSError` and/or `FileNotFoundError` instead of `RuntimeError`?

```pytb
**ERROR**: dump.gsd: Bad file descriptor - lattice.gsd

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-3-3623b04ad812> in <module>
     11 
     12 sim.operations.integrator = mc
---> 13 sim.create_state_from_gsd(filename='lattice.gsd')

~/miniconda3/envs/dice/lib/python3.8/site-packages/hoomd/simulation.py in create_state_from_gsd(self, filename, frame)
    107                                         self.device._cpp_exec_conf)
    108         # Grab snapshot and timestep
--> 109         reader = _hoomd.GSDReader(self.device._cpp_exec_conf,
    110                                   filename, abs(frame), frame < 0)
    111         snapshot = Snapshot._from_cpp_snapshot(reader.getSnapshot(),

RuntimeError: Error reading GSD file
```

## Proposed solution

Use another exception instead of `RuntimeError`, perhaps `FileNotFoundError` or `OSError`.

## Additional context

I think this may apply to other exceptions as well - we use `RuntimeError` pretty liberally in our code bases and it seems like it's a catch-all for things that have more specific causes. The [docstring](https://docs.python.org/3/library/exceptions.html#RuntimeError) says
> Raised when an error is detected that doesn’t fall in any of the other categories. The associated value is a string indicating what precisely went wrong.

I frequently refer to this list of all built-in exceptions when choosing what to raise. It may be useful to other developers. https://docs.python.org/3/library/exceptions.html#exception-hierarchy

vyasr:
[...]

PR #822, opened 2020-10-21
Improve GSD API and documentation

### LINKED ISSUE #674: Mode API for GSD
## Description

The GSD file writer should take python style mode strings.

## Proposed solution

Replace the `overwrite` arg to GSD with a mode string that accepts `wb`, `xb`, and `ab` modes.

## Additional context

It is more Pythonic.

### LINKED ISSUE #814: Add example for logging type_shapes
## Description

I would like a code example for logging `mc.type_shapes` data in the documentation. This may also be addressed in the logging tutorial that @joaander is planning. If possible, I would suggest that we include a snippet in the docs as well as the tutorial.

It would also be nice to show how to load a saved HPMC integrator state with `from_state`.

### LINKED ISSUE #815: Inconsistency with filter and filter_
## Description

It seems like there are some issues with Sphinx rendering `filter` and `filter_` correctly in the docstrings.

Example: https://hoomd-blue.readthedocs.io/en/v3.0.0-beta.1/module-hoomd-write.html#hoomd.write.GSD

It looks like this is written correctly in the source, but Sphinx is parsing it differently than anticipated.
https://github.com/glotzerlab/hoomd-blue/blob/376eab8c92b4b4c62b21abb68222efe06a6b81b9/hoomd/write/gsd.py#L23

There are also cases where the argument name is `filter` and might need to be changed to `filter_` for consistency. Some instances grepped with `"def.*\(.*filter[^_]"` (:warning: this regex won't catch instances where `filter` is on a different line than `def`):

https://github.com/glotzerlab/hoomd-blue/blob/376eab8c92b4b4c62b21abb68222efe06a6b81b9/hoomd/md/methods.py#L100
https://github.com/glotzerlab/hoomd-blue/blob/376eab8c92b4b4c62b21abb68222efe06a6b81b9/hoomd/md/force.py#L296
https://github.com/glotzerlab/hoomd-blue/blob/376eab8c92b4b4c62b21abb68222efe06a6b81b9/hoomd/state.py#L375

joaander:
## Description

* Use a string file `mode` instead of `overwrite` (breaking change).
* Clean up and make the documentation consistent with the rest of HOOMD.
* Include the exception reason in the exception and don't write to the messenger error() stream.

## Motivation and context

String file modes are more Pythonic and enable a new mode: exclusive create. The documentation had errors and was not consistent with the rest of HOOMD. Users expect to find the exception reason in the exception output that Python writes.

Resolves #674, resolves #814 and resolves #815.

## How has this been tested?

Unfortunately, there are no unit tests for GSD yet (#732). I tested this with the tutorial notebooks.

## Change log

```
Added:

- Support pybind11 2.6.0
- Exclusive creation file mode for ``write.GSD``.

Changed:

- [breaking] Replace ``write.GSD`` argument ``overwrite`` with ``mode``.
```
(doc improvement change is already in the log).

## Checklist:

### DIFF
--- a/hoomd/GSD.h
+++ b/hoomd/GSD.h
@@ -77,7 +77,7 @@ class GSDUtils
         else if (retval == GSD_ERROR_FILE_MUST_BE_WRITABLE)
             {
             std::ostringstream s;
-            s << "GSD: File must be writeable"
+            s << "GSD: File must be writable"
                  " - "
               << fname;
             throw std::runtime_error(s.str());
--- a/hoomd/GSD.h
+++ b/hoomd/GSD.h
@@ -1,0 +1,102 @@
+// Copyright (c) 2009-2019 The Regents of the University of Michigan
+// This file is part of the HOOMD-blue project, released under the BSD 3-Clause License.
+
+#pragma once
+
+#include "hoomd/extern/gsd.h"
+#include <string>
+#include <sstream>
+
+namespace hoomd
+    {
+namespace detail
+    {
+/// Utility class to collect common GSD file operations.
+class GSDUtils
+    {
+    public:
+    /// Check and raise an exception if an error occurs
+    static void checkError(int retval, const std::string& fname)
+        {
+        // checkError prints errors and then throws exceptions for common gsd error codes
+        if (retval == GSD_ERROR_IO)
+            {
+            std::ostringstream s;
+            s << "GSD: " << strerror(errno) << " - " << fname;
+            throw std::runtime_error(s.str());
+            }
+        else if (retval == GSD_ERROR_INVALID_ARGUMENT)
+            {
+            std::ostringstream s;
+            s << "GSD: Invalid argument"
+                 " - "
+              << fname;
+            throw std::invalid_argument(s.str());
+            }
+        else if (retval == GSD_ERROR_NOT_A_GSD_FILE)
+            {
+            std::ostringstream s;
+            s << "GSD: Not a GSD file"
+                 " - "
+              << fname;
+            throw std::runtime_error(s.str());
+            }
+        else if (retval == GSD_ERROR_INVALID_GSD_FILE_VERSION)
+            {
+            std::ostringstream s;
+            s << "GSD: Invalid GSD file version"
+                 " - "
+              << fname;
+            throw std::runtime_error(s.str());
+            }
+        else if (retval == GSD_ERROR_FILE_CORRUPT)
+            {
+            std::ostringstream s;
+            s << "GSD: File corrupt"
+                 " - "
+              << fname;
+            throw std::runtime_error(s.str());
+            }
+        else if (retval == GSD_ERROR_MEMORY_ALLOCATION_FAILED)
+            {
+            std::ostringstream s;
+            s << "GSD: Memory allocation failed"
+                 " - "
+              << fname;
+            throw std::runtime_error(s.str());
+            }
+        else if (retval == GSD_ERROR_NAMELIST_FULL)
+            {
+            std::ostringstream s;
+            s << "GSD: Namelist full"
+                 " - "
+              << fname;
+            throw std::runtime_error(s.str());
+            }
+        else if (retval == GSD_ERROR_FILE_MUST_BE_WRITABLE)
+            {
+            std::ostringstream s;
+            s << "GSD: File must be writeable"
+                 " - "
+              << fname;
+            throw std::runtime_error(s.str());
+            }
+        else if (retval == GSD_ERROR_FILE_MUST_BE_READABLE)
+            {
+            std::ostringstream s;
+            s << "GSD: File must be readable"
+                 " - "
+              << fname;
+            throw std::runtime_error(s.str());
+            }
+        else if (retval != GSD_SUCCESS)
+            {
+            std::ostringstream s;
+            s << "GSD: "
+              << "Unknown error " << retval << " writing: " << fname;
+            throw std::runtime_error(s.str());
+            }
+        }
+    };
+    } // namespace detail
+    } // namespace hoomd
--- a/hoomd/GSD.h
+++ b/hoomd/GSD.h
@@ -6,6 +6,7 @@
 #include "hoomd/extern/gsd.h"
 #include <string>
 #include <sstream>
+#include <stdexcept>
 
 namespace hoomd
     {
--- a/hoomd/GSDDumpWriter.h
+++ b/hoomd/GSDDumpWriter.h
@@ -26,8 +26,7 @@
     every time analyze() is called. When a group is specified, only write out the
     particles in the group.
 
-    On the first call to analyze() \a fname is created with a dcd header. If it already
-    exists, append to the file (unless the user specifies overwrite=True).
+    The file is not openend until the first call to
 
     \ingroup analyzers
 */
@@ -38,7 +37,7 @@ class PYBIND11_EXPORT GSDDumpWriter : public Analyzer
         GSDDumpWriter(std::shared_ptr<SystemDefinition> sysdef,
                       const std::string &fname,
                       std::shared_ptr<ParticleGroup> group,
-                      bool overwrite=false,
+                      std::string mode="ab",
                       bool truncate=false);
 
         //! Control attribute writes
@@ -70,9 +69,9 @@ class PYBIND11_EXPORT GSDDumpWriter : public Analyzer
             return m_fname;
             }
 
-        bool getOverwrite()
+        std::string getMode()
             {
-            return m_overwrite;
+            return m_mode;
             }
 
         bool getTruncate()
@@ -139,7 +138,7 @@ class PYBIND11_EXPORT GSDDumpWriter : public Analyzer
 
     private:
         std::string m_fname;                //!< The file name we are writing to
-        bool m_overwrite;                   //!< True if file should be overwritten
+        std::string m_mode;                 //!< The file open mode
         bool m_truncate;                    //!< True if we should truncate the file on every analyze()
         bool m_is_initialized;              //!< True if the file is open
         bool m_write_attribute;             //!< True if attributes should be written
@@ -147,6 +146,8 @@ class PYBIND11_EXPORT GSDDumpWriter : public Analyzer
         bool m_write_momentum;              //!< True if momenta should be written
         bool m_write_topology;              //!< True if topology should be written
         gsd_handle m_handle;                //!< Handle to the file
+
+        static std::list<std::string> particle_chunks;
 
         /// Callback to write log quantities to file
         pybind11::object m_log_writer;
--- a/hoomd/GSDDumpWriter.h
+++ b/hoomd/GSDDumpWriter.h
@@ -26,7 +26,7 @@
     every time analyze() is called. When a group is specified, only write out the
     particles in the group.
 
-    The file is not openend until the first call to
+    The file is not opened until the first call to analyze().
 
     \ingroup analyzers
 */
--- a/hoomd/hpmc/integrate.py
+++ b/hoomd/hpmc/integrate.py
@@ -66,6 +66,16 @@ class HPMCIntegrator(BaseIntegrator):
     TODO: Describe implicit depletants algorithm. No need to write this now,
     as Jens is rewriting the implementation.
 
+    .. rubric:: Writing type_shapes to GSD files.
+
+    Use a Logger in combination with a HPMC integrator and a GSD writer to write
+    ``type_shapes`` to the GSD file for use with OVITO::
+
+        mc = hoomd.hpmc.integrate.<Shape>(...)
+        log = hoomd.logging.Logger()
+        log.add(mc, quantities=['type_shapes'])
+        gsd = hoomd.write.GSD(..., log=log)
+
     .. rubric:: Parameters
 
     Attributes:
--- a/hoomd/hpmc/integrate.py
+++ b/hoomd/hpmc/integrate.py
@@ -69,12 +69,13 @@ class HPMCIntegrator(BaseIntegrator):
     .. rubric:: Writing type_shapes to GSD files.
 
     Use a Logger in combination with a HPMC integrator and a GSD writer to write
-    ``type_shapes`` to the GSD file for use with OVITO::
+    ``type_shapes`` to the GSD file for use with OVITO. For example::
 
-        mc = hoomd.hpmc.integrate.<Shape>(...)
+        mc = hoomd.hpmc.integrate.Sphere(seed=123)
         log = hoomd.logging.Logger()
         log.add(mc, quantities=['type_shapes'])
-        gsd = hoomd.write.GSD(..., log=log)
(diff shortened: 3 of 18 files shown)


hoomd/write/gsd.py
```diff
@@ -177,30 +180,37 @@ def _attach(self):
...
 
     @property
     def log(self):
+        """hoomd.logging.Logger: Provide log quantities to write.
+
+        May be `None`.
+        """
         return self._log
 
     @log.setter
```
b-butler, line 210: We should probably fix this to be settable to `None`.

hoomd/write/gsd.py
```diff
@@ -177,30 +180,37 @@ def _attach(self):
...
         writer = _hoomd.GSDDumpWriter(state._cpp_sys_def,
                                       filename,
                                       state._get_group(filter),
-                                      True,
+                                      mode,
                                       False)
 
         if log is not None:
-            writer.log_writer = GSDLogWriter(log)
+            writer.log_writer = _GSDLogWriter(log)
```
b-butler, line 199: This should be underscored since a user shouldn't access it.

b-butler (changes requested):
I like the mode argument and the `type_shapes` documentation. The use of other C++ exceptions to raise different Python exceptions is nice too. I do have a question about why we are using `dump.gsd` in our error messages since the Python module hierarchy is `hoomd.write.GSD`. Also, I am not sure if this is meant to be addressed in this PR or if we will address it, but we still raise `RuntimeError`s where a `IOError` would be more appropriate. I know pybind11 does not have any bindings for an IOError. It looks like we could define our own exception that pybind11 would raise and could catch that in `hoomd.write.GSD` and then raise the appropriate `IOError`. This doesn't have to be done in this PR though, just a thought about the feedback given by @bdice.

joaander:
> Also, I am not sure if this is meant to be addressed in this PR or if we will address it, but we still raise `RuntimeError`s where a `IOError` would be more appropriate. 

I started down this path thinking it would be a quick fix. It is not. We should solve #811 and the broader issue of using more specific exceptions in another PR. There aren't enough built in C++ execeptions and they don't all map naturally to Python, so if we want Pythonic exceptions we will need to create them all as custom exceptions and use them there.

hoomd/hpmc/integrate.py
```diff
@@ -66,6 +66,16 @@ class HPMCIntegrator(BaseIntegrator):
...
 
+    .. rubric:: Writing type_shapes to GSD files.
+
+    Use a Logger in combination with a HPMC integrator and a GSD writer to write
+    ``type_shapes`` to the GSD file for use with OVITO::
+
+        mc = hoomd.hpmc.integrate.<Shape>(...)
+        log = hoomd.logging.Logger()
+        log.add(mc, quantities=['type_shapes'])
+        gsd = hoomd.write.GSD(..., log=log)
```
bdice, line 77: Sometimes I think it's easier to understand "runnable" code and guess the general case from a specific example.
```suggestion
    ``type_shapes`` to the GSD file for use with OVITO. For example::

        mc = hoomd.hpmc.integrate.Sphere(seed=123)
        log = hoomd.logging.Logger()
        log.add(mc, quantities=['type_shapes'])
        gsd = hoomd.write.GSD(
            'trajectory.gsd', hoomd.trigger.Periodic(1000), log=log)
```

hoomd/write/gsd.py
```diff
@@ -15,28 +15,20 @@
...
-        log (hoomd.logging.Logger): A ``Logger`` object for GSD
-            logging, defaults to ``None``.
-
-    .. note::
-
-        All parameters are also available as instance attributes. Only
-        *trigger* and *log* may be modified after construction.
+        filter (hoomd.filter.ParticleFilter): Select the particles to write.
+            Defaults to `hoomd.filter.All`.
+        mode (str): The file open mode. Defaults to `ab`.
```
bdice, line 25: Probably needs to be quoted, since it's a string. Not sure if double-backticks are needed here, but I _think_ so?
```suggestion
        mode (str): The file open mode. Defaults to ``'ab'``.
```

hoomd/write/gsd.py
```diff
@@ -46,15 +38,33 @@ class GSD(Writer):
...
+    +==================+=============================================+
+    | ``'wb'``         | Open the file for writing. Creates the file |
+    |                  | if needed, or overwrites an existing file.  |
+    +------------------+---------------------------------------------+
+    | ``'xb'``         | Create a GSD file exclusively and opens it  |
+    |                  | for writing.                                |
+    |                  | Raise an exception when it already exists.  |
+    +------------------+---------------------------------------------+
+    | ``'ab'``         | Appends to the file when it already exists. |
+    |                  | Creates the file when it does not exist.    |
```
bdice, line 54: Tense is inconsistent in this table: create/creates, raise/raises.

hoomd/write/gsd.py
```diff
@@ -90,43 +100,36 @@ class GSD(Writer):
         * constraints/
         * pairs/
 
-
-    .. seealso::
-
+    See Also:
         See the `GSD documentation <http://gsd.readthedocs.io/>`_ and `GitHub
         project <https://github.com/glotzerlab/gsd>`_ for more information on
```
bdice, line 105: I recommend making these anonymous links with two underscores. Otherwise if you have another link somewhere in HOOMD with the text "GSD documentation" or "GitHub project," it will throw a warning.
```suggestion
        See the `GSD documentation <http://gsd.readthedocs.io/>`__ and `GitHub
        project <https://github.com/glotzerlab/gsd>`__ for more information on
```

hoomd/write/gsd.py
```diff
@@ -90,43 +100,36 @@ class GSD(Writer):
...
-    .. note::
+    Note:
+        When you use ``filter`` to select a subset of the whole system, `GSD`
+        will write out all of the selected particles in ascending tag order and
+        will **not** write out **topology**.
 
+    Tip:
         All logged data chunks must be present in the first frame in the gsd
         file to provide the default value. Some (or all) chunks may be omitted
         on later frames.
```
bdice, line 116: I had a question about this -- how do you write a quantity to only frame 0? Like `type_shapes` for a system with fixed shape parameters.

hoomd/write/gsd.py
```diff
@@ -177,30 +177,34 @@ def _attach(self):
...
+    def write(state, filename, filter=All(), mode='wb', log=None):
         """Write the given simulation state out to a GSD file.
 
         Args:
             state (State): Simulation state.
             filename (str): File name to write.
-            filter_ (``hoomd.ParticleFilter``): Select the particles to write.
-            log (``hoomd.logger.Logger``): A ``Logger`` object for GSD logging.
+            filter (`hoomd.ParticleFilter`): Select the particles to write.
+            mode (str): The file open mode. Defaults to `wb`.
```
bdice, line 187:
```suggestion
            mode (str): The file open mode. Defaults to ``'wb'``.
```

hoomd/write/gsd.py
```diff
@@ -177,30 +177,34 @@ def _attach(self):
...
+            log (`hoomd.logger.Logger`): Provide log quantities to write.
 
-        Note:
-            The file is always overwritten.
+        The valid file modes for `write` are ``'wb'`` and ``'xb'``.
```
bdice, line 190: Why is append not allowed?

bdice (changes requested):
Suggestions in comments.

hoomd/write/gsd.py
joaander: Good point. You ensure that the logger is present on the first frame written. Namely, don't create the writer, write a few frames, and then set the log attribute.

hoomd/write/gsd.py
joaander: This function is intended for use when writing single frames (i.e. a restart file at the end of a script). It would be inefficient and cause significant metadata I/O if users made repeated calls to `write` appending to the same file in place of using one writer operation that gets triggered (it keeps the file open).

hoomd/write/gsd.py
joaander: Yes, the double back ticks are needed.

hoomd/write/gsd.py
joaander: Done.

hoomd/write/gsd.py
joaander: `writer` is the C++ object and is only used as a local variable within this static method. The user should not have access to it.

hoomd/GSDDumpWriter.h
```diff
@@ -26,8 +26,7 @@
     every time analyze() is called. When a group is specified, only write out the
     particles in the group.
 
-    On the first call to analyze() \a fname is created with a dcd header. If it already
-    exists, append to the file (unless the user specifies overwrite=True).
+    The file is not openend until the first call to
```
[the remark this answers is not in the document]
joaander: fixed.

hoomd/write/gsd.py
joaander: fixed

hoomd/write/gsd.py
bdice: @joaander I am having trouble understanding what you mean in the second half of your comment. Do you just mean that it will automatically skip writing `type_shapes` on later frames if the data is identical to that of frame 0? If that's not what you mean, can you provide an example?

hoomd/write/gsd.py
bdice: That's reasonable but not obvious to me. I am thinking about a use case that @ramanishsingh asked me about yesterday. He had a GSD file and needed to convert it to DCD for compatibility with some other tool. I wasn't aware of a good method to do that (garnett doesn't support writing DCD), so I helped him create a HOOMD v2 script that loaded the GSD system state frame-by-frame and wrote the DCD file that way. I had to do the same thing for a simulation a while ago where I needed the output of the HPMC SDF to compute a pressure, so I loaded the whole trajectory frame by frame and dumped the SDF. I could imagine wanting to do post-processing like that where HOOMD could write additional log quantities to the GSD file (adding a logger that wasn't originally present in the simulation run), because I do that explicitly for my current pathways analysis code, adding new metadata fields that cannot be computed on the fly (only after knowing the whole trajectory). I am not sure if the single-frame write would be appropriate for that, or if the standard writer approach allows for that.

hoomd/write/gsd.py
joaander: The writer will not automatically skip writing log entries. They have no defaults. What I am trying to say is do this:
```
gsd = hoomd.write.GSD(...)
gsd.log = log
sim.run(1e6)
```

Not this:
```
gsd = hoomd.write.GSD(...)
sim.run(1e6)
gsd.log = log
sim.run(1e6)
```

And not this:
```
gsd = hoomd.write.GSD(...)
gsd.log = log
sim.run(1e6)
log.add(something)
sim.run(1e6)
```

In the second and third cases the writer will write log entries to later frames in the file that are not present in the earlier frames. Logged entries have no defaults, so readers cannot guess at what they might be in earlier frames.

hoomd/write/gsd.py
joaander: Conversion to a DCD file would be best achieved with a file format tool that can write DCD files. However, scanning through a GSD file's frames and recomputing HOOMD compute properties is something best handled by HOOMD itself. This came up in a recent mailing list discussion as well.

The single frame writer is not appropriate to write out multi-frame GSD files because of the overhead of opening and closing the file each time. Files like this should be written in append mode.

I propose a `GSDTrajectoryReplay` updater that reads frames from a GSD file in order and replaces the system state every time it is triggered. The canonical usage would run without an integrator and use this updater with a period of 1 so that each timestep corresponds to 1 frame in the GSD file. What do you think?

bdice: The `GSDTrajectoryReplay` sounds like a great idea. I agree with all the comments about the other user needs that I mentioned. I felt there was something "related" about them which was kind of hard to put my finger on.

joaander: I opened #823 to track this feature request.

hoomd/write/gsd.py
```diff
@@ -187,19 +188,16 @@ def write(state, filename, filter=All(), mode='wb', log=None):
             state (State): Simulation state.
             filename (str): File name to write.
             filter (`hoomd.ParticleFilter`): Select the particles to write.
-            mode (str): The file open mode. Defaults to `wb`.
+            mode (str): The file open mode. Defaults to ``'wb'``.
             log (`hoomd.logger.Logger`): Provide log quantities to write.
```
b-butler, line 192:
```suggestion
            log (`hoomd.logging.Logger`): Provide log quantities to write.
```

hoomd/write/gsd.py
```diff
@@ -187,19 +188,16 @@ def write(state, filename, filter=All(), mode='wb', log=None):
             state (State): Simulation state.
             filename (str): File name to write.
             filter (`hoomd.ParticleFilter`): Select the particles to write.
```
b-butler, line 190:
```suggestion
            filter (`hoomd.filter.ParticleFilter`): Select the particles to write.
```

hoomd/write/gsd.py
```diff
@@ -215,7 +213,7 @@ def log(self):
 
     @log.setter
     def log(self, log):
-        if isinstance(log, Logger):
+        if (log is not None) and isinstance(log, Logger):
```
b-butler, line 216:
```suggestion
        if log is not None and isinstance(log, Logger):
```

b-butler (changes requested):
Overall looks good. I have a few small documentation corrections and a stylistic preference.

b-butler (approved):
LGTM

hoomd/GSD.h
```diff
@@ -0,0 +1,103 @@
...
+            std::ostringstream s;
+            s << "GSD: Namelist full"
+                 " - "
+              << fname;
+            throw std::runtime_error(s.str());
+            }
+        else if (retval == GSD_ERROR_FILE_MUST_BE_WRITABLE)
+            {
+            std::ostringstream s;
+            s << "GSD: File must be writeable"
```
bdice, line 80: The spelling of writable/writeable doesn't match the GSD error name. Both spellings are common. I get a red underline in my browser for writeable and Google suggests writable, but NumPy arrays use "WRITEABLE." 🤷‍♂️ 
```suggestion
            s << "GSD: File must be writable"
```

hoomd/write/gsd.py
```diff
@@ -157,16 +163,14 @@ def _attach(self):
         if self.dynamic is not None:
             for v in self.dynamic:
                 if v not in categories:
-                    raise RuntimeError("GSD: dynamic quantity " + v +
-                                       " is not valid")
+                    raise RuntimeError("GSD: dynamic quantity " + v
+                                       + " is not valid")
```
bdice, line 167:
```suggestion
                    raise RuntimeError(
                        f"GSD: dynamic quantity {v} is not valid")
```

bdice (approved):
LGTM, with two minor suggestions.

PR #829, opened 2020-10-29
Document Loggable properties and methods

### LINKED ISSUE #771: Document loggable properties
## Description

Every loggable property in HOOMD should state that it is loggable in the documentation. They should possibly also state the flags, data type, and default status so that users are aware of how to add them to a `Logger.

## Motivation and context

Users need to know what properties will be logged by a `Logger`.

b-butler:
## Description

Adds a method to `hoomd.logging.Loggable` to add documentation to property and method docstrings that displays that a property is loggable along with its logging type and default status.

## Motivation and context

Resolves #771

## How has this been tested?

Documentation was built locally and reviewed.

## Change log

```
- document loggable quantities in docstring
```

## Checklist:

### DIFF
--- a/hoomd/logging.py
+++ b/hoomd/logging.py
@@ -247,8 +247,30 @@ class Loggable(type):
     @classmethod
     def _get_current_cls_loggables(cls, new_cls):
         """Gets the current class's new loggables (not inherited)."""
-        return {name: _LoggerQuantity(name, new_cls, entry.flag, entry.default)
-                for name, entry in cls._meta_export_dict.items()}
+        current_loggables = {}
+        for name, entry in cls._meta_export_dict.items():
+            current_loggables[name] = _LoggerQuantity(
+                name, new_cls, entry.flag, entry.default)
+            cls._add_loggable_docstring_info(
+                new_cls, name, entry.flag, entry.default)
+        return current_loggables
+
+    @classmethod
+    def _add_loggable_docstring_info(cls, new_cls, attr, flag, default):
+        doc = getattr(new_cls, attr).__doc__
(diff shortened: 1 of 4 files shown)


joaander (commented):
Nice! It's great that there is an automated way to do this.

It formats like this: (**Loggable**: type “sequence”, default True)

Can we link Loggable to `hoomd.logging.Logger`?
Also, we call "type" "flag" in `Logger`, we should be consistent here.
Should we always show default=X? Maybe we should not list default when True and only add "not logged by default" when not default.

@bdice do you have any good suggestions for formatting and language to use?

bdice:
> Should we always show default=X? Maybe we should not list default when True and only add "not logged by default" when not default.

I think "default True" is a little confusing. This suggestion might be better, or just specify behavior for both cases: "[not] logged by default."

> Can we link Loggable to `hoomd.logging.Logger`?

This would be nice. However the text is currently bold, and the solutions for "bold links" aren't great. You might have to decide between "link" and "bold." https://docutils.sourceforge.io/FAQ.html#is-nested-inline-markup-possible

b-butler:
@joaander, I have made the changes discussed to the documentation formatting.

joaander (changes requested):
Great! I like the new format.

We just need one minor fix. The loggable output for `state` does not seem to render properly: https://hoomd-blue--829.org.readthedocs.build/en/829/module-hpmc-integrate.html#hoomd.hpmc.integrate.HPMCIntegrator.state

I guess that this is because the `state` docstring is empty so that (Loggable: type="state") gets interpreted with `(Loggable` as the type and `type="state"` as the description.

joaander (approved):
Thanks!

PR #830, opened 2020-10-30
Pass the variable directly to the if so that its value is checked.

vyasr:
## Description

Change the macro for finding packages to correctly evaluate the not found variable.

## Motivation and context

CMake's if statement seems more robust to empty variables than to the evaluated form of the variable. I suspect that in this case it is because it doesn't like negating a constant (NOT ...). 

## How has this been tested?

This was tested when porting this solution to freud.

## Change log

N/A

## Checklist:


joaander:
@vyasr This looks fine. Did you mean to leave it as a draft?

vyasr:
Yes, I left it as a draft because I made this while getting the freud CMake setup working and figured I would port back any and all related changes at once. I'll convert this to a normal PR as soon as I get a chance to review and see if there are any other changes (or if this one actually ended up being necessary). IIRC the reason I had to make this change was because a variable being set to the string NOTFOUND and a variable not being defined at all behave differently depending on the syntax change I made here, and the latter case stopped being an issue for me once I improved TBB finding.

joaander:
OK, let me know when this is ready to merge.

joaander:
Should we merge this?

vyasr:
@joaander yes I think this can be merged. Nothing else needs to be done here.

joaander (approved):
Thanks!

PR #834, opened 2020-11-06
Snapshot dimension behavior change

b-butler:
`dimensions` can no longer be set, and is dependent on box.Lz. If `box.Lz
== 0` then `dimensions == 2` otherwise it equals 3.

## Description

Changes `ConfigurationData.dimensions` to be unsettable and depend on the snapshot box. If `box.Lz == 0` then `dimensions == 2` otherwise 3.

## Motivation and context

This helps keep behavior consistent with `hoomd.Box` and `hoomd.State.box`.

## How has this been tested?

Has not been tested directly.

## Change log

```
hoomd.ConfigurationData.dimensions cannot be set and depends on box's z dimension.
```

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,7 +15,7 @@ v3.0.0-beta.2 (not yet released)
 *Changed*
 
 - ``hoomd.snapshot.ConfigurationData.dimensions`` is unsettable
-  and depends on the snapshot box. If :math:`box.Lz = 0` then
+  and depends on the snapshot box. If ``box.Lz = 0`` then
   the dimensions are 2 otherwise 3.
 - Building from source requires a C++14 compatible compiler.
 - Improved documentation.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,9 @@ v3.0.0-beta.2 (not yet released)
 
 *Changed*
 
+- ``hoomd.snapshot.ConfigurationData.dimensions`` is unsettable
+  and depends on the snapshot box. If :math:`box.Lz = 0` then
+  the dimensions are 2 otherwise 3.
 - Building from source requires a C++14 compatible compiler.
 - Improved documentation.
 - [breaking] Replace ``write.GSD`` argument ``overwrite`` with ``mode``.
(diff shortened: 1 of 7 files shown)


CHANGELOG.rst
```diff
@@ -14,6 +14,9 @@ v3.0.0-beta.2 (not yet released)
 
 *Changed*
 
+- ``hoomd.snapshot.ConfigurationData.dimensions`` is unsettable
+  and depends on the snapshot box. If :math:`box.Lz = 0` then
```
bdice, line 18: This is more like code than math.
```suggestion
  and depends on the snapshot box. If ``box.Lz == 0`` then
```

hoomd/snapshot.py
```diff
@@ -31,6 +27,7 @@ def box(self, box):
             raise ValueError(
```
bdice, line 27: On the line above this, I would recommend catching a limited set of exceptions that we know can be thrown by `from_box`.

hoomd/snapshot.py
```diff
@@ -31,6 +27,7 @@ def box(self, box):
...
                 "{} is not convertible to a hoomd.Box object. "
                 "using hoomd.Box.from_box".format(box))
```
bdice, line 29:
```suggestion
                f"{box} is not convertible to a hoomd.Box object "
                f"using hoomd.Box.from_box()")
```

bdice (changes requested):

joaander (commented):
Thanks! This looks good to me. Address @bdice's comments and we can merge.

hoomd/snapshot.py
b-butler: There is a
```
except:
     raise
````
in the method like in `freud` so technically we don't know what errors will be raised necessarily. I generally agree with the limited catching of exceptions, but without converting the error in `from_box` to a `ValueError` I don't think we can do this completely safely.

bdice: Ok, sounds fine.

bdice (approved):
I pushed a couple small changes to edit the changelog / error message. This is ready to merge.

PR #841, opened 2020-11-13
Escape $ORIGIN

joaander:
## Description

Escape the $ in $ORIGIN in CMakeLists.txt.

## Motivation and context

I couldn't replicate it on my system, but the freud developers found issues (https://github.com/glotzerlab/freud/pull/695)

## How has this been tested?

I confirmed the rpath setting with `readelf`
```
readelf -d _mpcd.cpython-38-x86_64-linux-gnu.so | grep PATH
 0x000000000000000f (RPATH)              Library rpath: [$ORIGIN/..:$ORIGIN/../md:$ORIGIN]
```

## Change log

Not needed.

## Checklist:

### DIFF
--- a/hoomd/hpmc/CMakeLists.txt
+++ b/hoomd/hpmc/CMakeLists.txt
@@ -99,7 +99,7 @@ add_library(HOOMD::_hpmc ALIAS _hpmc)
 if (APPLE)
 set_target_properties(_hpmc PROPERTIES INSTALL_RPATH "@loader_path/..;@loader_path")
 else()
-set_target_properties(_hpmc PROPERTIES INSTALL_RPATH "$ORIGIN/..;$ORIGIN")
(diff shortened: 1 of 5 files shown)


bdice:
@joaander @vyasr I found an explicit recommendation for this backslash, which increased my confidence in this as the correct thing to do.

> In `CMAKE_INSTALL_RPATH` lines, [`$ORIGIN`] should have its dollar sign escaped with a backslash to have it end up with proper syntax in the final executable. See also the [CMake and $ORIGIN discussion](http://www.cmake.org/pipermail/cmake/2008-January/019290.html).

https://gitlab.kitware.com/cmake/community/-/wikis/doc/cmake/RPATH-handling#recommendations

vyasr:
Yes that's where I found that as well.

joaander:
Yes, it should. Years ago, CMake would always interpret `$ORIGIN` incorrectly. Modern CMake versions seem to handle it without the `\`, but not always.

PR #846, opened 2020-11-23
Fix error in removing operations

b-butler:
## Description

Fixes an error that would occur if trying to remove a triggered operation after attaching. This will likely be fixed when `Updater` and `Writer` are made to own their own triggers.

## Motivation and context

Fix behavior of hoomd.

Resolves #???

## How has this been tested?

Tested locally to see if it fixes the bug I had.

## Change log

```
Fix bug that caused an error when removing specific operations from a simulation's operations attribute.
```

## Checklist:

### DIFF
--- a/hoomd/operation.py
+++ b/hoomd/operation.py
@@ -534,6 +534,13 @@ class _TriggeredOperation(Operation):
     def _attach(self):
         super()._attach()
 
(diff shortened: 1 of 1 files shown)


joaander (approved):
Thanks!

PR #849, opened 2020-11-24
Loggable Simulation

### LINKED ISSUE #797: Make Simulation loggable
## Description

`Simulation` should be loggable so that information about the simulation run can be logged.

## Proposed solution

Quantities to consider including:
* tps
* elapsed time (of the current run)
* timestep
* end timestep (of the current run)

## Additional context

If we choose to fully replicate the HOOMD 2.x run() output, we will also need an ETA estimate and a way to pretty print time in seconds as HH:MM:SS in `Table`.

## Developer

Seeking volunteers!

joaander:
## Description

Add useful loggable quantities to `Simulation`.

This PR will also add a tutorial for the logging subsystem and fixes a number of logging issues discovered when writing the tutorial.

## Motivation and context

Users will find a variety of loggable quantities, such as timestep, tps, walltime, etc.. useful.

Resolves #797

## How has this been tested?

Tested in the tutorial jupyter notebook.

## Change log

```
Fixed:

* ``Simulation.timestep`` reports the correct value when analyzers are called.
* ``Logger`` names quantities with the documented namespace name.

Added:

* ``walltime`` and ``final_timestep`` loggable properties in ``Simulation``.
* ``Null`` particle filter.
* Logging tutorial.

Changed:

* [breaking] Rename ``flags`` to ``categories`` in ``Logger``
```

## Checklist:


b-butler:
I don't think the force loggables should be non-default. Shouldn't the way to specify not to log those be to restrict the logger types? For instance `TypeFlags.All - TypeFlags.particle` would do this.

joaander:
> I don't think the force loggables should be non-default. Shouldn't the way to specify not to log those be to restrict the logger types? For instance `TypeFlags.All - TypeFlags.particle` would do this.

I agree. At which point I wonder what are we going to use non-default for?

b-butler:
I am not sure, we talked about this for performance loggables like `mps`, `tps`, `walltime`, and quantities like that since they aren't useful in analytical calculations.

b-butler (approved):
I think this is good, we can have a future discussion about what should be made non-default with respect to logging another time.

PR #855, opened 2020-12-02
Make ParticleFilter objects picklable.

b-butler:
## Description

Makes `ParticleFilter` objects picklable using Python using __reduce__ to reconstruct the object. This prevents problems in the set filters and prevents needing to construct getter and setters in C++ to be able to pickle through pybind11's facilities.

## Motivation and context

Moving towards most HOOMD-blue object being picklable for restartability.

## How has this been tested?

Has been tested locally. Will be tested through unit tests before merging.

## Change log

```
Added: ParticleFilters are picklable.
```

## Checklist:

### DIFF
--- a/hoomd/filter/all_.py
+++ b/hoomd/filter/all_.py
@@ -21,3 +21,6 @@ class All(ParticleFilter, ParticleFilterAll):
     def __eq__(self, other):
         """Test for equality between two particle filters."""
         return type(self) == type(other)
+
+    def __reduce__(self):
+        return (type(self), tuple())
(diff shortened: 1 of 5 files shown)


joaander (approved):
Thanks! Looks good to me. 
@bdice or @vyasr is this a good use of the pickling APIs?

vyasr:
This looks fine to me. @b-butler I assume the reason you went for using `__reduce__` rather than the higher level `__getstate__` and `__setstate__` APIs is that you need to actually call the constructor in order to call all parent constructors (since these use multiple inheritance) and propagate the data down to the C++ mirror classes?

vyasr:
Posting here for future reference: @b-butler and I have talked and I'm on board with the changes here as well as in #885 and #886.

PR #886, opened 2020-12-07
Make Variants Picklable

b-butler:
## Description

Makes `hoomd.variant.Variant` objects picklable. This is to facilitate the eventual support for pickling `hoomd.Operations` objects. The variant tests were also reworked to use pytest's facilities for parametrization.

## Motivation and context

This is to facilitate the eventual support for pickling `hoomd.Operations` objects.

## How has this been tested?

A test has been added to test for correct pickling.

## Change log

```
Added:
    - ``hoomd.variant.Variant`` subclasses are picklable.

Fixed:
    - ``hoomd.variant.Power`` objects now have a ``t_ramp`` attribute as documented.
```

## Checklist:


hoomd/variant.py
```diff
@@ -47,6 +47,13 @@ def max(self):
         """The maximum value of this variant."""
         return self._max()
 
+    def __getstate__(self):
```
joaander, line 50: pydocstyle is erroring because `__geststate__` and `__setstate__` don't have documentation.

joaander (changes requested):
Thanks! This looks fine to me. We just need to fix the pydocstyle errors.

joaander (approved):
Thanks!

Issue #869, opened 2020-12-04
Remove oneD, sphere, and ellipsoid constraints


joaander:
## Description

Remove constrain.oneD, constrain.sphere, and the ellipsoid active matter constraint.

## Motivation and context

These are all being replaced with integrators that support constraints on generic manifolds.

Issue #210, opened 2016-12-06
Additional validation tests


joaander:
**[Original report](https://bitbucket.org/glotzer/hoomd-blue/issue/210) by me.**

----------------------------------------

Pull request #278 along with a Jenkins reconfiguration paves the way for long running validation tests as part of hoomd's unit test suite. We need to expand the validation tests with additional tests for MD.

As part of this effort, I would like to simplify the cmake files. Currently we have a LOT of copied and pasted add_test code around. We could simplify that down to a simple macro which all the various cmake scripts call when they want to add a test.

The doumentation needs to be updated to describe the validation testing procedures, including a list of compile options and what hardware / compiler versions we officially validate.

joaander:
Once #888 and #996 are merged, we should port over the rigid body validation tests from hoomd v2.

joaander:
While we can use the validation framework for some simulations, others require run times so long that we cannot run them in the CI framework  - e.g. the hard disk equation of state in #985. We need to consider options for organizing these tests to run manually on HPC resources. One option would be a separate repository and use signac-flow to manage the jobs.

joaander:
As the plan is now to write a separate framework, this is not critical to complete before the 3.0.0 release.

joaander:
Some of the tests to include in validation:
* Hard disk equation of state
* Lennard-Jones equation of state (including via off-center rigid spheres to test rigid body pressure)
* Spherical depletants in HPMC.
* Energy conservation in MD.
* Energy drift in MD.

PR #888, opened 2020-12-09
Update constrain.Rigid to Hoomd 3.0

### LINKED ISSUE #871: Port constrain.rigid to the v3 API
## Description

Port constrain.rigid to the v3 API and add tests.

jennyfothergill:
## Description

The goal of this PR is to update the constrain.Rigid to the v3 api.

## Motivation and context

Resolves #871 

## Change log

```

```

## Checklist:


joaander:
You don't need to update sphere and oned constraint forces. We are going to remove those and replace them with integrators that support manifold constraints: #869. You also don't need to update the `distance` constraint here - that can be handled in another PR.

jennyfothergill:
I have a question about how best to resolve a cyclic dependency: If `ConstraintForce` inherits from `Force`, then `Force` needs to by imported in constrain.py, but `ConstraintForce` is also imported in force.py. Should the import for `ConstraintForce` be put inside the `Active` class? Or should I do something totally different?

joaander:
You could move `ConstraintForce` into `force.py`.

joaander:
Sorry, I meant just the `ConstraintForce` class, not `distance`, `rigid` and the others. They should remain in `constrain` to identify them as constraint forces.

jennyfothergill:
OK, so building the docs locally, it seems like `md.constrain` still has the documentation for `ConstraintForce` but not `distance` or `rigid`. The sphinx docs builds from the docstrings, right? How do I fix this?
I'm also working on updating the rigid rods tutorial to the new api as a way to check that rigid is working.

jennyfothergill:
@b-butler So I have gotten the workflow shown in the [rigid rods tutorial](https://github.com/jennyfothergill/hoomd-examples/blob/feature/new-rigid/Tutorial%20-%20MD%20-%20Rigid%20rods.ipynb) to run without error messages, but it is clearly not working correctly: when the simulation runs, the body centers move without the constituents being tied to them.

I'm including a snippet from a conversation with Josh to ask your preference for how we choose to solve this problem:
>OK, it looks like I was wrong about the composite force just needing to be added to the constraint forces: https://github.com/glotzerlab/hoomd-blue/blob/d21e9c781c9e948d2bcb4fa68fbe41cfae60b15d/hoomd/md/IntegratorTwoStep.cc#L479-L484
>There is no need for a list of composite forces, just one. But it does need to handle being attached/detached. I would replace `m_composite_forces` with `std::shared_ptr<ForceComposite> m_composite_force`. If you need help handling with attach/detach, Brandon would be the best person to ask.
>Then, instead of `integrator.constraints.append(rigid)` it would be `integrator.composite = rigid`.
>Alternately, we could use type checking and instead loop through `m_constraint_forces` in `updateRigidBodies` and `updateCompositeParticles` if the constraint force is a `ForceComposite`. Brandon might have thoughts on what solution he would prefer from an API standpoint.

joaander:
@b-butler What do you think? Should the composite force be special and have its own property? Or should it be an item in the constraint forces list and get discovered by RTTI?

Note that it never makes sense to have more than one composite force. It can make sense to have multiple constraint forces (i.e. one composite force for rigid bodies and one distance constraint force for fixed bonds). **However**, the constraints do not work together so this only works if the constraints operate on disjoint sets of particles. In this example, polymers created with distance constraints could not contain any particles that are parts of rigid bodies.

Given that we are working on the RATTLE integrators, rigid bodies and distance constraints will soon be the only two types of constraint forces.

b-butler:
@joaander is there potential for other constraints in the future? My instincts say that we add the composite force to the `constraints` list.

hoomd/md/force.py
```diff
@@ -360,25 +452,50 @@ def __init__(self, field_x,field_y,field_z,p):
...
+            cpp_cls = getattr(_md, self._cpp_class_name + "GPU")
+
+        self._cpp_obj = cpp_cls(self._simulation.state._cpp_sys_def)
+
+        super()._attach()
+
+    def update_coeffs(self):
+        pass
+        # does nothing: this is for derived classes to implement
+
```
b-butler, line 501: This is unnecessary.

b-butler (changes requested):
@jennyfothergill The finished `Rigid` class should not have a `set_params` function, but it should use `_param_dict` and `_type_params` to set object attributes. It is difficult to debug the class given all the formatting changes. It is hard to tell what has actually changed. I am also not too familiar with the old method of setting parameters especially for rigid bodies.

I also cannot debug the failure without seeing the script that fails.

jennyfothergill:
> @jennyfothergill The finished `Rigid` class should not have a `set_params` function, but it should use `_param_dict` and `_type_params` to set object attributes. 

OK, is the `Active` force a good example of how to use `_param_dict`? 

>It is difficult to debug the class given all the formatting changes. It is hard to tell what has actually changed. I am also not too familiar with the old method of setting parameters especially for rigid bodies.

Yeah... I'm sorry about that. I used black but should've done the reformatting in a different PR. It does make it harder to see the changes. I can undo it without too much difficulty if you like.

> I also cannot debug the failure without seeing the script that fails.

I am working on the rigid rods example with the updated code. [This notebook](https://github.com/jennyfothergill/hoomd-examples/blob/feature/new-rigid/Tutorial%20-%20MD%20-%20Rigid%20rods.ipynb) run with the [feat rigid branch of hoomd](https://github.com/jennyfothergill/hoomd-blue/tree/feature/new-rigid) shows the problem in the last cell.

b-butler:
@jennyfothergill I would look at `hoomd/write/gsd.py`'s `GSD` class for an example of `_param_dict`.

jennyfothergill:
In the `hoomd/write/gsd.py` example, the `_param_dict` is set in the `__init__` method, but I think this might not work for rigid bodies. From what I remember Josh saying `Rigid` uses the `_simulation` attribute which isn't set until after attaching? (I might be getting this wrong-- @joaander please correct me if so.)
Anyway trying to set the parameters in the init method results in 
```AttributeError: Object <class 'hoomd.md.constrain.Rigid'> has no attribute _simulation```

**Comment by b-butler** [comment]
The parameters should be stored in pure Python until attaching. If you set `pybind11` properties, the setting and getting from C++ will be handled automatically from there.

**Comment by joaander** [comment]
> The parameters should be stored in pure Python until attaching. If you set `pybind11` properties, the setting and getting from C++ will be handled automatically from there.

I suggested that Jenny get the code working first, then we could refactor the parameter setting in a later step. @b-butler what API do you suggest for setting complex parameter dictionaries for some types but not others?

**Comment by b-butler** [comment]
@joaander @jennyfothergill I would suggest using a type parameter with a default acceptable value of `None`.

**Comment by jennyfothergill** [comment]
OK, I know I'm not doing this right, but I know I'm supposed to do something where I have some typeparam (or typeparamdict or paramdict?) that defaults to None and it's not supposed to be a list (like forces or methods). I think I am supposed to add a rigid property to the DynamicIntegrator. Do I add the param_dict to the Rigid class?

Sorry...

**Comment by b-butler** [comment]
@jennyfothergill don't worry about having to potentially ask something again. So here are some first steps.

1. Create an entry in `hoomd.md.integrate.DynamicIntegrator` instances `_param_dict` attribute called `rigid`. This will house the rigid body the specification should be something like `{'rigid': hoomd.data.typeconverter.OnlyType(hoomd.md.constrain.Rigid, allow_none=True)}`.
2. Create a `_setattr_param` method in `DynamicIntegrator`. This should check if the key is `rigid`, and if so it should run `self._cpp_obj.rigid = self.rigid._cpp_obj` otherwise call the parent's method with super.
3. Make sure that the `Rigid` can properly attach and send information to C++. Here it may be easier to start by letting `_attach` not set any properties and setting them after a `sim.run(0)` call. Otherwise, you would need to fill out the `_param_dict` and `_type_paramdict` attributes.
4. Fill out the `_param_dict` and `_typeparam_dict` attributes for `Rigid`. Here looking at the HPMC integrator should help. You want the entire entry for a type to be settable to `None` so you should look at, `hoomd.typeconverter.OnlyIf`specifically the `allow_none` argument and `hoomd.typeconverter.to_type_converter` for setting up the values of the `body` type parameter correctly.

Feel free to ask more questions. A tip on learning to read code better, though. I have found that grepping a code base for functions or classes I don't understand can help a lot since often time you can see them used in the ways they were intended in other parts of the code base.

**Comment by jennyfothergill** [comment]
Thank you so much for taking the time to write out detailed steps! I think I've done the first two steps.
How would I go about attaching the rigid constraint at this point? I've tried
```
integrator._param_dict["rigid"] = rigid
```
and
```
integrator._rigid = rigid
```
where integrator and rigid are
```
rigid = hoomd.md.constrain.Rigid('R', 
                                 types=['A']*8,
                                 positions=[(-4,0,0),(-3,0,0),(-2,0,0),(-1,0,0),
                                            (1,0,0),(2,0,0),(3,0,0),(4,0,0)])
```

**Comment by joaander** [comment]
`integrator.rigid = rigid`

**Comment by jennyfothergill** [comment]
Hmm... I'm still getting an error when trying to initialize the rigid class. Full traceback:
```
AttributeError                            Traceback (most recent call last)
<ipython-input-8-855bc3afc6fb> in <module>
      2                                  types=['A']*8,
      3                                  positions=[(-4,0,0),(-3,0,0),(-2,0,0),(-1,0,0),
----> 4                                             (1,0,0),(2,0,0),(3,0,0),(4,0,0)])
      5 integrator.rigid = rigid

~/Projects/hoomd-blue/build/hoomd/md/constrain.py in __init__(self, type_name, types, positions, orientations, charges, diameters)
    237         # get a list of types from the particle data
    238         ntypes = (
--> 239             self._simulation.state._cpp_sys_def.getParticleData().getNTypes()
    240         )
    241         type_list = []

~/Projects/hoomd-blue/build/hoomd/operation.py in __getattr__(self, attr)
     99         else:
    100             raise AttributeError("Object {} has no attribute {}".format(
--> 101                 type(self), attr))
    102 
    103     def _getattr_param(self, attr):

AttributeError: Object <class 'hoomd.md.constrain.Rigid'> has no attribute _simulation
```

**Comment by b-butler** [comment]
The `Rigid` object doesn't have a `_simulation` attribute by default.

File: hoomd/md/constrain.py [hunk 2]
```diff
@@ -233,138 +98,180 @@ def set_params(self,rel_tol=None):
...
-            rigid = constrain.rigid()
-            rigid.set_param('A', types = ['A_const', 'A_const'], positions = [(0,0,1),(0,0,-1)])
-            rigid.set_param('B', types = ['B_const', 'B_const'], positions = [(0,0,.5),(0,0,-.5)])
+    """
 
-        """
+    _cpp_class_name = "ForceComposite"
+    def __init__(
+        self,
+        type_name,
```
joaander, line 230: `__init__` should not take in `type_name`, `types`, etc... arguments. These parameters will be set later. The only thing that `__init__` should do now is update the `param_dict` and later will define the type parameter body.

jennyfothergill:
OK, I've moved those parameter setting steps to a different function called `set_params` for now. 

In the parameter dict, right now I just have the types as `list`--they should be more specific and orientation, charges, and diameter should be allowed to be none--I think this can be done similarly to how it is done in hpmc.integrate.SphereUnion: 
```
from hoomd.data.typeconverter import OnlyIf, to_type_converter 
...
typeparam_shape = TypeParameter(
...
orientations=OnlyIf(to_type_converter([(float, float, float, float)]), allow_none=True), 
```
but I'm not sure how to specify the length of the lists or if it needs to be specified.

Also I get an error when running 
```
rigid = hoomd.md.constrain.Rigid()
integrator.rigid = rigid
``` 
where the rigid's `_cpp_obj` is not set. Is it set by the attach method in ConstraintForce? Is the attach method not called until `sim.run(0)` is called? Traceback:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-10-1ddef83666b2> in <module>
      1 rigid = hoomd.md.constrain.Rigid()
----> 2 integrator.rigid = rigid

~/Projects/hoomd-blue/build/hoomd/operation.py in __setattr__(self, attr, value)
    113             super().__setattr__(attr, value)
    114         elif attr in self._param_dict.keys():
--> 115             self._setattr_param(attr, value)
    116         elif attr in self._typeparam_dict.keys():
    117             self._setattr_typeparam(attr, value)

~/Projects/hoomd-blue/build/hoomd/md/integrate.py in _setattr_param(self, attr, value)
     99     def _setattr_param(self, attr, value):
    100         if "rigid" in self._param_dict:
--> 101             self._cpp_obj.rigid = self.rigid._cpp_obj
    102         else:
    103             super()._setattr_param("rigid", None)

AttributeError: 'NoneType' object has no attribute '_cpp_obj'
```

b-butler:
@jennyfothergill The last error is my fault. You need to check if the `integrator` is attached before trying to set the `self._cpp_obj.rigid`. You only need to set that if you are already attached.

Currently we do not support setting the size of the list.

jennyfothergill:
>@jennyfothergill The last error is my fault. You need to check if the integrator is attached before trying to set the self._cpp_obj.rigid. You only need to set that if you are already attached.

No worries. How can I make sure that after the integrator is attached that rigid cpp_obj is set?

b-butler:
By making the `_setattr_param` hook check for being attached or not. In addition, the C++ rigid body class may need to store a reference to the Python rigid object to return when queried by a user (e.g. `rigid = integrator.rigid`) after attaching. We could also provide a `_from_cpp` classmethod to create `Rigid` objects from the C++ object, and overwrite `_getattr_param` to convert to the Python type before returning.

jennyfothergill:
OK I think I've done this in ca4ba92:
> By making the _setattr_param hook check for being attached or not. 

But I'm not sure what you mean here:
>In addition, the C++ rigid body class may need to store a reference to the Python rigid object to return when queried by a user (e.g. rigid = integrator.rigid) after attaching. We could also provide a _from_cpp classmethod to create Rigid objects from the C++ object, and overwrite _getattr_param to convert to the Python type before returning.

hoomd/md/integrate.py
```diff
@@ -93,6 +100,13 @@ def _children(self):
 
         return children
 
+    def _setattr_param(self, attr, value):
+        if attr == "rigid":
```
joaander, line 104:
```suggestion
        if attr == "rigid":
            self._rigid = value
```

hoomd/md/integrate.py
```diff
@@ -93,6 +100,13 @@ def _children(self):
...
+            if self._attached:
+                self._rigid._attach()
+                self._cpp_obj.rigid = self._rigid._cpp_obj
+        else:
+            super()._setattr_param("rigid", None)
```
joaander, line 109:
```suggestion
            super()._setattr_param(attr, value)
```
Maybe? @b-butler what should this be?

hoomd/md/integrate.py
```diff
@@ -51,10 +52,16 @@ def __init__(self, forces, constraints, methods):
...
                                    iterable=methods)
 
+        self._rigid = None
+        self._param_dict = ParameterDict(rigid=OnlyType(Rigid, allow_none=True))
+
     def _attach(self):
         self.forces._sync(self._simulation, self._cpp_obj.forces)
         self.constraints._sync(self._simulation, self._cpp_obj.constraints)
         self.methods._sync(self._simulation, self._cpp_obj.methods)
+        if self._rigid != None:
```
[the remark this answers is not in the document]
joaander: @b-butler When/how will `self._rigid` get the value of `self._simulation`?

jennyfothergill:
OK, the Rigid constraint is currently working! (Example notebook [here](https://github.com/jennyfothergill/hoomd-examples/blob/feature/new-rigid/Tutorial%20-%20MD%20-%20Rigid%20rods.ipynb).) Thank you for your help today. When you have time, could you help me plan the next steps? @joaander @b-butler

joaander:
@jennyfothergill the next steps would be to implement the `body` `TypeParameter`. This would be a `dict` or `None`. You will need to implement in C++ the methods `ForceComposite::setBody(std::string typ, pybind::object v)` (which replaces `setParam`) and `ForceComposite::getBody(pybind::object v)` (new method). Start with `setBody` and you can add `getBody` later after you get set working.

`setBody` will unpack the dictionary provided like `set_param` does now, except in C++. For examples, see `IntegratorHPMCMono::setShape` which defers the dict unpacking to another classes constructor (you can do the unpacking directly in `setBody`). For a good example, see the constructor `PolyhedronVertices(pybind11::dict v, ...` in `ShapeConvexPolyhedron.h`.

While `setParam` takes a number of arguments with C++ types, `setBody` will take one Python object as an argument. The various parameters `type`, `position`, `orientation`, `charge`, and `diameter` become keys in the dictionary. When working the new API equivalent of (from your example):
```
rigid.set_params('R', 
                types=['A']*8,
                positions=[(-4,0,0),(-3,0,0),(-2,0,0),(-1,0,0),
                           (1,0,0),(2,0,0),(3,0,0),(4,0,0)])
```
will be:
```
rigid.body['R'] = dict(type=['A']*8, positions=[(-4,0,0),(-3,0,0),(-2,0,0),(-1,0,0), (1,0,0),(2,0,0),(3,0,0),(4,0,0)]))
```

b-butler:
@jennyfothergill could you create another PR that just formats the files that you formatted in this one? We can merge that and then the diffs will be much more helpful.

jennyfothergill:
OK, I have a rough draft of what I think you intended, @joaander. I followed the example in `PolyhedronVertices(pybind11::dict v, ... in ShapeConvexPolyhedron.h`. I get an error during build `error: use of undeclared identifier 'OverlapReal'` that I'm not sure how to fix--I think it's one of the include files, but I couldn't tell which one.

hoomd/md/ForceComposite.h
```diff
@@ -73,7 +73,50 @@ class PYBIND11_EXPORT ForceComposite : public MolecularForceCompute
...
+        
+        /// Construct from a Python dictionary
+        void setBody(std::string typ, pybind11::object v)
+            {   
+            pybind11::list types = v["types"];
+            pybind11::list positions = v["positions"];
+            int N = (unsigned int)len(positions);
+
+            // extract the positions from the python list
+            std::vector<vec3<OverlapReal>> pos_vector;
```
joaander, line 85:
```suggestion
            std::vector<vec3<Scalar>> pos_vector;
```
`OverlapReal` is an HPMC specific data type. You want `Scalar` here.

hoomd/md/ForceComposite.h
```diff
@@ -73,7 +73,50 @@ class PYBIND11_EXPORT ForceComposite : public MolecularForceCompute
...
+            for (unsigned int i = 0; i < N; i++)
+                {   
+                pybind11::list pos_i = positions[i];
+                if (len(pos_i) != 3)
+                    throw std::runtime_error("Each position must have 3 coordinates");
+                vec3<OverlapReal> pos = vec3<OverlapReal>(pybind11::cast<OverlapReal>(pos_i[0]),
```
joaander, line 91:
```suggestion
                vec3<OverlapReal> pos = vec3<Scalar>(pybind11::cast<Scalar>(pos_i[0]),
```

hoomd/md/ForceComposite.h
```diff
@@ -73,7 +73,50 @@ class PYBIND11_EXPORT ForceComposite : public MolecularForceCompute
...
+                                                          pybind11::cast<OverlapReal>(pos_i[1]),
+                                                          pybind11::cast<OverlapReal>(pos_i[2]));
+                pos_vector.push_back(pos);
+                }   
+
+            }   
+
+        /// Convert parameters to a python dictionary
+        pybind11::dict asDict()
```
joaander, line 100:
```suggestion
        pybind11::dict getBody(std::string typ)
```

joaander (commented):
Here are a few suggestions to get you past the current compile errors.

hoomd/md/constrain.py
```diff
@@ -279,18 +113,18 @@ class rigid(ConstraintForce):
     orientation quaternion defines the rotation from the body space into the
     simulation box. In body space, the center of mass of the body is at 0,0,0
     and the moment of inertia is diagonal. You specify the constituent particles
-    to :py:class:`rigid` for each type of body in body coordinates. Then,
-    :py:class:`rigid` takes control of those particles, and sets their position
+    to :py:class:`Rigid` for each type of body in body coordinates. Then,
```
tommy-waltmann, line 116: I believe we are changing how these are done. This occurs in a lot of places throughout this file.
```suggestion
    to `Rigid` for each type of body in body coordinates. Then,
```

hoomd/md/constrain.py
```diff
@@ -527,71 +345,17 @@ def create_bodies(self, create=True):
...
         simulation run.
         """
-        self.cpp_force.validateRigidBodies(False)
+        self._cpp_obj.validateRigidBodies(False)
 
     ## \internal
     # \brief updates force coefficients
     def update_coeffs(self):
         # validate copies of rigid bodies
         self.create_bodies(False)
```
tommy-waltmann, line 360: I believe this can be removed.

hoomd/md/constrain.py
```diff
@@ -342,85 +176,69 @@ class rigid(ConstraintForce):
...
-                'B',
-                types = ['B_const', 'B_const'],
-                positions = [(0,0,.5),(0,0,-.5)]
-                )
+    _cpp_class_name = "ForceComposite"
+    def __init__(self):
+        pass
 
-        """
+    def set_params(
```
tommy-waltmann, line 230: Let's add documentation for this method.

hoomd/md/ForceComposite.h
```diff
@@ -73,7 +73,50 @@ class PYBIND11_EXPORT ForceComposite : public MolecularForceCompute
...
+                vec3<OverlapReal> pos = vec3<Scalar>(pybind11::cast<Scalar>(pos_i[0]),
+                                                          pybind11::cast<OverlapReal>(pos_i[1]),
+                                                          pybind11::cast<OverlapReal>(pos_i[2]));
+                pos_vector.push_back(pos);
+                }   
+
+            }   
+
+        /// Convert parameters to a python dictionary
+        pybind11::dict getBody(std::string typ)
```
tommy-waltmann, line 100: Why is there a `getBody` if there is no way to query this info in python?

hoomd/md/constrain.py
```diff
@@ -342,85 +176,69 @@ class rigid(ConstraintForce):
...
+        ``system.particles.types.add('A_const')``.
 
-        self.composite = True
+    Example::
 
-        # create the c++ mirror class
-        if not hoomd.context.current.device.cpp_exec_conf.isCUDAEnabled():
-            self.cpp_force = _md.ForceComposite(
-                hoomd.context.current.system_definition
+        rigid = constrain.rigid()
```
tommy-waltmann, line 212:
```suggestion
        rigid = constrain.Rigid()
```

tommy-waltmann (changes requested):
I want to comment also on a few things I saw in the example notebook:

1) I'd like to see the `aniso` property of the integrator be renamed to something more explicit, perhaps `anisotropic_mode`.

2) There is a call in the script to a `create_bodies` method which I did not see in the Rigid class. I'd prefer to make scripts run without the user explicitly calling `create_bodies` (I think its a bit clunky), perhaps this could be done implicitly once we attach the operations to a simulation?

joaander:
There are conflicts with `master`. Let me know if you have any questions about merging `master` up into your branch.

hoomd/md/constrain.py
jennyfothergill: I think this method is going to be replaced.

hoomd/md/ForceComposite.h
jennyfothergill: Check Josh's comment from Jan 26. I'm not familiar with C++ so my progress on this is very slow. I think the plan is to add methods in python that query them.

hoomd/md/ForceComposite.h
```diff
@@ -73,6 +73,49 @@ class PYBIND11_EXPORT ForceComposite : public MolecularForceCompute
...
+            {
+            pybind11::dict v;
+            pybind11::list types;
+            pybind11::list positions;
+            for(unsigned int i = 0; i < N; i++)
```
jennyfothergill, line 105: OK, I know I am doing something wrong, but I don't understand how variables are moving between python and C. I am trying to get the properties of (some C object I think?) set by `setBody` N should be the length of the positions and x, y, and z should be the coordinates. How to I access those correctly?

hoomd/md/ForceComposite.h
```diff
@@ -73,6 +73,49 @@ class PYBIND11_EXPORT ForceComposite : public MolecularForceCompute
...
+                {   
+                pybind11::list pos_i = positions[i];
+                if (len(pos_i) != 3)
+                    throw std::runtime_error("Each position must have 3 coordinates");
+                vec3<Scalar> pos = vec3<Scalar>(pybind11::cast<Scalar>(pos_i[0]),
+                                                pybind11::cast<Scalar>(pos_i[1]),
+                                                pybind11::cast<Scalar>(pos_i[2]));
+                pos_vector.push_back(pos);
+                }   
+
```
joaander, line 96: You need a call to `setParam` here with the type id, unpacked positions, orientations, etc...

hoomd/md/ForceComposite.h
joaander: `setBody` moves values from the Python dictionary the user provides and stores them in the internal C++ data structures that `ForceComposite` manges.

`getBody` pulls them out of the internal C++ data structures and packs them in a dictionary.

The data structures are dense arrays that store the per-type-per-constituent-particle data. Get the number of particles in a given body with:
```
ArrayHandle<unsigned int> h_body_len(m_body_len, access_location::host, access_mode::readwrite);
unsigned int N h_body_len.data[body_type_id];
```

To get the position of particle `i` in this body and pack it in a list, use:
```
ArrayHandle<Scalar3> h_body_pos(m_body_pos, access_location::host, access_mode::read);
pybind11::list pos;
pos.append(h_body_pos.data[m_body_idx(body_typeid,i)].x);
pos.append(h_body_pos.data[m_body_idx(body_typeid,i)].y);
pos.append(h_body_pos.data[m_body_idx(body_typeid,i)].z);
```

See the contents of `setParam` for details.

b-butler (approved):
I have to approve since I previously requested changes. We should wait until @joaander and @tommy-waltmann review though.

hoomd/md/pytest/test_rigid.py
```diff
@@ -0,0 +1,242 @@
...
+    rigid = md.constrain.Rigid()
+    rigid.body["A"] = valid_body_definition
+
+    initial_snapshot = two_particle_snapshot_factory()
+    initial_snapshot.particles.types = ["A", "B"]
+    sim = simulation_factory(initial_snapshot)
+
+    rigid.create_bodies(sim.state)
+    snapshot = sim.state.snapshot
+    if sim.device.communicator.rank == 0:
```
tommy-waltmann, line 124: Did we get rid of the `snaphost.exists` property? I believe this line is the case where we would normally use that.

hoomd/Integrator.h
```diff
@@ -130,7 +130,7 @@ class PYBIND11_EXPORT Integrator : public Updater
             }
 
         /// Count the total number of degrees of freedom removed by all constraint forces
-        Scalar getNDOFRemoved(std::shared_ptr<ParticleGroup> query);
+        virtual Scalar getNDOFRemoved(std::shared_ptr<ParticleGroup> query);
```
tommy-waltmann, line 133: I may have missed something, but I don't think this method needed to be declared virtual for this PR.

tommy-waltmann (changes requested):
I left some comments, most are just little things that I think could be be changed for consistency, cleaner code, etc.

Also, what is the status on a rigid bodies tutorial?

hoomd/filter/ParticleFilterRigid.h
```diff
@@ -0,0 +1,121 @@
...
+constexpr enum RigidBodySelection operator&(
+    const enum RigidBodySelection flag1, const enum RigidBodySelection flag2)
+    {
+        return static_cast<enum RigidBodySelection>(
+            static_cast<unsigned int>(flag1) & static_cast<unsigned int>(flag2));
+    }
+
+constexpr bool toBool (const enum RigidBodySelection& flag)
+    {
+        return static_cast<unsigned short>(flag) > 0;
```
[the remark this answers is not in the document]
b-butler: That is an artifact of a previous commit that was never updated.

hoomd/filter/ParticleFilterRigid.h
```diff
@@ -0,0 +1,121 @@
...
+/** \b Overview
+    Filters particles based on them being rigid body centers or not.
+*/
+
+enum class RigidBodySelection
+    {
+    NONE = 0,
+    CENTERS = 1,
+    CONSTITUENT = 2,
+    FREE = 4
```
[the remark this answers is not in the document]
b-butler: It is a bit flag effectively which means we need to use powers of two or the | and & operators will not work properly.

hoomd/md/pytest/test_rigid.py
```diff
@@ -0,0 +1,242 @@
...
+            assert np.allclose(value, current_body_definition[key])
+        else:
+            assert value == current_body_definition[key]
+
+    for key, values in invalid_body_definitions.items():
+        for invalid_value in values:
+            current_body_definition[key] = invalid_value
+            with pytest.raises(hoomd.data.typeconverter.TypeConversionError):
+                rigid.body["A"] = current_body_definition
+        current_body_definition[key] = valid_body_definition[key]
```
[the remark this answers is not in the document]
b-butler: Yes, I am writing a comment as to why.

tommy-waltmann (approved):

hoomd/md/IntegratorTwoStep.cc
```diff
@@ -478,6 +473,62 @@ void IntegratorTwoStep::setAutotunerParams(bool enable, unsigned int period)
         method->setAutotunerParams(enable, period);
     }
 
+/// helper function to compute net force/virial
+void IntegratorTwoStep::computeNetForce(uint64_t timestep)
+    {
+    if (m_rigid_bodies)
+        {
+        m_rigid_bodies->validateRigidBodies();
```
joaander, line 481: I see that `validateRigidBodies` has an early exit condition. What conditions invalidate that? It is an expensive operation, so we wouldn't want to run it too often.

hoomd/md/constrain.py
```diff
@@ -4,192 +4,49 @@
 # Maintainer: joaander / All Developers are free to add commands for new
 # features
 
-R""" Constraints.
+r""" Constraints.
```
joaander, line 7:
```suggestion
r"""Constraints.
```
Normally flake8-docstrings is picky about this.

hoomd/md/constrain.py
```diff
@@ -4,192 +4,49 @@
...
 
-Constraint forces constrain a given set of particle to a given surface, to have
-some relative orientation, or impose some other type of constraint.
+Constraint forces can constrain particles to be a set distance from each other,
+to have some relative orientation, or impose other types of constraint.
 
 As with other force commands in hoomd, multiple constrain commands can be issued
 to specify multiple constraints, which are additively applied.
```
joaander, line 13:
```suggestion
```
Force objects are not "commands" any longer. In any case, this is explained in the `md.Integrator` docs and doesn't need to be mentioned here.

hoomd/md/constrain.py
```diff
@@ -4,192 +4,49 @@
...
 
-Warning: Constraints will be invalidated if two separate constraint commands
-apply to the same particle.
+The `Rigid` class is special in that only one is allowed in a system and is set
+to an `hoomd.md.Integator` object separately in the `rigid` attribute.
```
joaander, line 16:
```suggestion
to an `hoomd.md.Integator` object separately in the `rigid <hoomd.md.Integrator.rigid>` attribute.
```
I think this will fix it to link correctly.

hoomd/md/constrain.py
```diff
@@ -4,192 +4,49 @@
...
+
+Warning:
+    Constraints will be invalidated if two separate constraint commands
```
joaander, line 19:
```suggestion
    Constraints will be invalid if two separate constraint forces
```
We don't check for this, so nothing is invalidated - but neither of the constraints will be applied correctly. Also, it's not a command.

hoomd/md/constrain.py
```diff
@@ -4,192 +4,49 @@
...
+    apply to the same particle.
 
 The degrees of freedom removed from the system by constraints are correctly
 taken into account when computing the temperature.
```
joaander, line 23:
```suggestion
The degrees of freedom removed from the system by constraints are
accounted for in `hoomd.md.ThermodynamicQuantities`.
```

hoomd/md/constrain.py
```diff
@@ -273,49 +130,43 @@ class rigid(ConstraintForce):
...
     In this way, each particle of type *R* in the system configuration defines
     a body of type *R*.
 
     As a convenience, you do not need to create placeholder entries for all of
     the constituent particles in your initial configuration. You only need to
     specify the positions and orientations of all the central particles. When
-    you call :py:meth:`create_bodies()`, it will create all constituent
-    particles that do not exist. (those that already exist e.g. in a restart
-    file are left unchanged).
+    you call `~.create_bodies`, it will create all constituent particles.
```
joaander, line 152:
```suggestion
    you call `create_bodies`, it will create all constituent particles.
```
I think this will link correctly now.

hoomd/md/constrain.py
```diff
@@ -273,49 +130,43 @@ class rigid(ConstraintForce):
...
 
-    .. danger:: Automatic creation of constituent particles can change particle
-    tags. If bonds have been defined between particles in the initial
-    configuration, or bonds connect to constituent particles, rigid bodies
-    should be created manually.
+    Warning:
+        Automatic creation of constituent particles can change particle tags. If
+        bonds have been defined between particles in the initial configuration,
+        or bonds connect to constituent particles, rigid bodies should be
+        created manually.
```
joaander, line 158:
```suggestion
        Automatic creation of constituent particles changes particle tags. When
        there are bonds between particles in the initial configuration, or bonds 
        connect to constituent particles, include the constituent particles in the
        initial configuration manually.
```
Active voice.

hoomd/md/constrain.py
```diff
@@ -336,237 +189,103 @@ class rigid(ConstraintForce):
...
-    def update_coeffs(self):
-        # validate copies of rigid bodies
-        self.create_bodies(False)
-
-
-class oneD(ConstraintForce):
-    R"""Constrain particles to move along a specific direction only
+    Set constituent particle types and coordinates for a rigid body.
 
     Args:
```
joaander, line 205: We need to document `body` as an `Attribute:` like we do `shape` in HPMC.

hoomd/md/constrain.py
```diff
@@ -336,237 +189,103 @@ class rigid(ConstraintForce):
...
+        type_name (str): The type of the central particle
+        types (list): List of types of constituent particles
+        positions (list): List of relative positions of constituent
+            particles
+        orientations (list): List of orientations of constituent particles
+            (**optional**)
+        charge (list): List of charges of constituent particles
+            (**optional**)
+        diameters (list): List of diameters of constituent particles
+            (**optional**)
```
joaander, line 215:
```suggestion
        types (list[str]): List of types of constituent particles
        positions (list[ tuple[float, float, float] ]): List of relative positions of constituent
            particles
        orientations (list[ tuple[float, float, float, float]]): List of orientations of constituent particles
            (**optional**)
        charge (list[float]): List of charges of constituent particles
            (**optional**)
        diameters (list[float]): List of diameters of constituent particles
            (**optional**)
```
Fix the typing syntax. My suggestion might require some tweaks to render properly in sphinx and may not be wrapped properly.

joaander (changes requested):
Thanks to Jenny for getting this started and Brandon for finishing it off! This was a big effort.

I have a few change requests related to style and documentation, along with one question about performance. Once those are resolved, this PR is good to merge.

Follow up work for future PRs:
1. Add validation tests for rigid bodies (#210)
2. Add tutorial to hoomd-examples (https://github.com/glotzerlab/hoomd-examples/issues/29).
3. Revise the documentation for clarity and consistency with the tutorial once it is written.

hoomd/filter/export_filters.h
```diff
@@ -1,6 +1,17 @@
...
+#include "ParticleFilter.h"
+#include "ParticleFilterAll.h"
+#include "ParticleFilterCustom.h"
+#include "ParticleFilterIntersection.h"
+#include "ParticleFilterNull.h"
+#include "ParticleFilterRigid.h"
+#include "ParticleFilterSetDifference.h"
+#include "ParticleFilterTags.h"
+#include "ParticleFilterType.h"
+#include "ParticleFilterUnion.h"
```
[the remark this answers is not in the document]
b-butler: I can't remember why I did it. I think it had something to do with consistency, but I was not thinking of the recompilation, but I think `export_filters.h` is only included in `module.cc`.

hoomd/md/ForceComposite.cc
```diff
@@ -398,406 +372,305 @@ Scalar ForceComposite::requestExtraGhostLayerWidth(unsigned int type)
...
-                        else
-                            {
-                            // wipe out body flag
-                            assert(snap.body[i] == i);
-                            snap.body[i] = NO_BODY;
-                            }
-                        }
+            // Particle is of a type with a non-zero body definition. This should only happen if it
+            // is a central particle.
+            bool is_central_particle = h_body_len.data[snap.type[i]] != 0;
```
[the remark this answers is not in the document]
b-butler: Good catch.

hoomd/md/IntegratorTwoStep.cc
b-butler: A change in rigid body definitions or the addition or removal of particles so infrequent events in most simulations. Without this though, either of these could leave the simulation in an invalid state without an error potentially, The addition or removal of particles could lead to error or no error or potentially segfaults.

joaander: OK. So this is only a performance penalty if the user dynamically changes their rigid bodies or if they dynamically add particles to the system. For this PR, let's document that these cases will degrade performance. Later, if any user has a strong need for such simulations we can relax it.  For example, only changing the rigid body positions doesn't require a full revalidation - only changing the numbers. New free particles added to the system doesn't need to trigger revalidation, but new rigid bodies does.

hoomd/md/IntegratorTwoStep.cc
```diff
@@ -142,7 +142,19 @@ void IntegratorTwoStep::setDeltaT(Scalar deltaT)
...
+        {
         method->setDeltaT(deltaT);
+        }
+    if (m_rigid_bodies)
+        {
+        m_rigid_bodies->setDeltaT(deltaT);
+        }
+    }
+
+void IntegratorTwoStep::removeForceComputes()
```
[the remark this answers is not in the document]
b-butler: Probably not I will do a grep and remove if it isn't since it isn't a big change.

joaander (changes requested):
Thanks, I think all my previous comments have been addressed.

We do need to fix the pre-commit and unit test failures before we can merge:
pre-commit:
```
hoomd/data/parameterdicts.py:61:49: F821 undefined name 'TypeConverterMapping'
hoomd/data/parameterdicts.py:63:61: F821 undefined name 'OnlyIf'
hoomd/data/parameterdicts.py:65:61: F821 undefined name 'Either'
hoomd/data/parameterdicts.py:67:52: F821 undefined name 'TypeConverterMapping'
```
unit test:
```
NameError: name 'TypeConverterMapping' is not defined
```

joaander (approved):
Thanks!

PR #889, opened 2020-12-14
doc/moved environment variable warning before cmake

cbkerr:
## Description
Following the build directions rendered on github, I skipped setting the `CMAKE_PREFIX_PATH` because the warning appeared after the step at which it is needed. (I thought it only applied to the `make` step.) Move warning one paragraph earlier.

## Checklist:


INSTALLING.rst
```diff
@@ -180,19 +180,19 @@ Run ``python3 install-prereq-headers.py -h`` to see a list of the command line o
 Compile HOOMD-blue
 ------------------
 
-Configure::
-
-    $ cd /path/to/hoomd-blue
-    $ cmake -B build
-    $ cd build
-
 .. warning::
```
joaander, line 183: The section *Configure a virtual environment* includes instructions to set `CMAKE_PREFIX_PATH`. If that is too far from the `cmake` command that requires, I think we should move that here and delete this warning block.

joaander (changes requested):
Thanks. I have a small suggestion to make things more clear.

INSTALLING.rst
cbkerr: Thanks for pointing that out!

cbkerr:
I moved the section about virtual environments after the section on install prerequisites, allowing us to delete the warning. Instead, I added a normal text reminder about the environment variable.

While reading for flow, I moved another bit of text above the corresponding command.

I also built it locally to make sure it builds.

joaander:
Do you find it a problem that now *Install prerequisites* mentions installing into a virtual environment that doesn't exist yet?

cbkerr:
> Do you find it a problem that now _Install prerequisites_ mentions installing into a virtual environment that doesn't exist yet?

Good point. I was treating "install prerequisites" as "the install prerequisites" and it felt better to have the "prerequisites" first. A virtual environment is kind of a prerequisite, but it is different from a system package manager or module system. I think it makes sense to discuss the general case before talking about virtual environments.

I've tried restructuring some of the end of the "dependencies" section and merged it with the virtual environment section. It helped to split discussion of the module system from the discussion of virtual environments.

joaander (approved):
Thanks!

PR #893, opened 2020-12-15
Create `hoomd.Snapshot` from `gsd.hoomd.snapshot` and allow `Simulation.create_state_from_snapshot` to accept gsd snapshotss

### LINKED ISSUE #891: Allow snapshots to be duck-typed in state creation
## Description

Some users have run into errors by creating a `gsd.hoomd.Snapshot` object, and attempting to use it for `hoomd.Simulation.create_state_from_snapshot`. This leads to an error since we currently only allow for a `hoomd.Snapshot` object. I would propose that we duck-type this function to allow for `gsd.hoomd.Snapshot` objects. While this is less performant that properly using a `hoomd.Snapshot` object, given that a user would have already created the `gsd.hoomd.Snapshot` object, this is better than an error. We could raise a `UserWarning` to let the user know that they may have made a sub-optimal decision in doing this (e.g. check if the `num_ranks > 1` for wasted memory usage).

## Proposed solution

I propose adding a `hoomd.Snapshot._from_gsd` class method for `hoomd.Snapshot` to create a `hoomd.Snapshot` from a `gsd.hoomd.Snapshot` object. This approach would also prepare a path in the future for other factory methods to create a `hoomd.Snapshot` object from other similar objects. For instance, given a box array and particle positions a simple snapshot could be created. If we decide to support this in the future, we could utilize the ability of `hoomd.Snapshot` to be initialized from a variety of objects to allow for state creation that way.

To also give useful performance warnings in MPI configurations, we should likely check after grabbing the communicator, if using multiple ranks that the passed snapshot like object is `None` on all but the root rank otherwise throw a warning (this does not need to be the case for `hoomd.Snapshot` objects as they only store data on the root rank). I believe this should be safe with the exception hook HOOMD-blue uses by default.

jennyfothergill:
## Description

The goal of this PR is to add duck-typing to `Simulation.create_state_from_snapshot` to allow for the snapshot to be a `gsd.hoomd.snapshot` or a `hoomd.snapshot`.  A function to create `hoomd.Snapshots` from gsd snapshots will also be made: `Snapshot._from_gsd_snapshot`.

## Motivation and context

Resolves #891 

## How has this been tested?

no unit test yet, but I've checked that the attribute of a snapshot created with `hoomd.Snapshot._from_gsd_snapshot()` are the same as in the gsd snapshot.

sphinx docs builds correctly -- there's not much change because `_from_gsd_snapshot` is not public, but `create_state_from_snapshot` docstring is updated.

## Change log

```
allow `hoomd.Snapshot` to be created from `gsd.hoomd.Snapshot` and `Simulation.create_state_from_snapshot` to automatically convert `gsd.hoomd.Snapshots`
```

## Checklist:


hoomd/snapshot.py
```diff
@@ -109,3 +109,9 @@ def replicate(self, nx, ny, nz):
 
     def _broadcast_box(self):
         self._cpp_obj._broadcast_box(self._comm.cpp_mpi_conf)
+
+    @classmethod
+    def _from_gsd_snapshot(cls, gsd_snapshot, communicator):
+        sp = cls()
+        snapshot = sp._from_cpp_snapshot(gsd_snapshot, communicator)
```
b-butler, line 116: While this works, I feel this may be a bit to fragile. Doesn't really create a standard snapshot, but merely wraps the gsd snapshot with a hoomd snapshot. We want to transfer the data to a C++ snapshot too. Something like
```python
_from_gsd_snapshot(cls, gsd_snapshot, communicator):
    snap = cls()
    snap.particles.N = gsd_snapshot.particles.N
    snap.particles.position[:] = snap.particles.position[:]
```
is more what I was thinking. It is more verbose and work, but I think this will be less likely to break expectations. For instance, I think this would fail if passed in to `create_state_from_snapshot` since the `self._cpp_obj` is not a `_hoomd.Snapshot` object.

hoomd/simulation.py
```diff
@@ -137,7 +137,11 @@ def create_state_from_snapshot(self, snapshot):
         if self.state is not None:
             raise RuntimeError("Cannot initialize more than once\n")
 
-        self._state = State(self, snapshot)
+        try:
```
b-butler, line 140: Some developer comments that state we only support `hoomd.Snapshot` and `gsd.hoomd.Snapshot` would be helpful here.

hoomd/snapshot.py
jennyfothergill: Yeah, your idea is much better and more understandable! I'm thinking all particle, bond, and box attributes should be set. I'm not sure what to do with the communicator.

hoomd/simulation.py
jennyfothergill: Good point. I'll also modify the docstring.

hoomd/simulation.py
```diff
@@ -129,15 +129,21 @@ def create_state_from_snapshot(self, snapshot):
         When `timestep` is `None` before calling, `create_state_from_snapshot`
         sets `timestep` to 0.
 
-        Warning:
-            *snapshot* must be a `hoomd.Snapshot`. Use `create_state_from_gsd`
-            to read GSD files. `create_state_from_snapshot` does not support
-            ``gsd.hoomd.Snapshot`` objects from the ``gsd`` Python package.
+        Note:
+            *snapshot* must be either a `hoomd.Snapshot` or a
+            ``gsd.hoomd.Snapshot``. A ``gsd.hoomd.Snapshot`` will first be
```
joaander, line 134: Explicitly document the types accepted in the  `Args:` section - they are more important than a `Note:` block signifies.

hoomd/simulation.py
```diff
@@ -129,15 +129,21 @@ def create_state_from_snapshot(self, snapshot):
...
+            converted to a `hoomd.Snapshot`.
         """
         if self.state is not None:
             raise RuntimeError("Cannot initialize more than once\n")
 
-        self._state = State(self, snapshot)
+        try:
+            # snapshot is hoomd.Snapshot
+            self._state = State(self, snapshot)
+        except AttributeError:
```
joaander, line 143: There a lot of ways for `AttributeError`s to be raised that do not imply the type is a `gsd.hoomd.Snapshot`. I'd prefer to see more explicit type checks and to raise a `TypeError` when the caller provides an invalid type.

hoomd/snapshot.py
```diff
@@ -109,3 +109,38 @@ def replicate(self, nx, ny, nz):
...
+            if val is not None:
+                try:
+                    setattr(snap.particles, key, val)
+                except AttributeError:
+                    try:
+                        # Some attributes exist but aren't setable without [:]
+                        x = getattr(snap.particles,key)
+                        x[:] = val
+                    except AttributeError:
+                        # The attribute doesn't exist in hoomd.Snapshot
```
joaander, line 128: Why is this needed? What attributes are in `gsd.hoomd.Snapshot` that are not in `hoomd.Snapshot`?

hoomd/snapshot.py
```diff
@@ -109,3 +109,38 @@ def replicate(self, nx, ny, nz):
 
     def _broadcast_box(self):
         self._cpp_obj._broadcast_box(self._comm.cpp_mpi_conf)
+
+    @classmethod
+    def _from_gsd_snapshot(cls, gsd_snapshot):
+        snap = cls()
```
joaander, line 115:
```suggestion
        snap = cls(communicator=self._device.communicator)
```

hoomd/snapshot.py
```diff
@@ -109,3 +109,38 @@ def replicate(self, nx, ny, nz):
...
+        # Set all particle attributes in snap from gsd_snapshot
```
joaander, line 116: The whole set routine needs to be wrapped in an: `if self._device.communicator.rank == 0:` block.

hoomd/snapshot.py
```diff
@@ -109,3 +109,38 @@ def replicate(self, nx, ny, nz):
...
+            if val is not None:
+                try:
+                    setattr(snap.bonds, key, val)
+                except AttributeError:
+                    try:
+                        x = getattr(snap.bonds,key)
+                        x[:] = val
+                    except AttributeError:
+                        pass
+
```
joaander, line 143: You need to also copy over all angles, dihedrals, impropers, special pairs, etc...

hoomd/snapshot.py
```diff
@@ -109,3 +109,38 @@ def replicate(self, nx, ny, nz):
...
+    @classmethod
+    def _from_gsd_snapshot(cls, gsd_snapshot):
+        snap = cls()
+        # Set all particle attributes in snap from gsd_snapshot
+        for key in vars(gsd_snapshot.particles):
+            val = vars(gsd_snapshot.particles)[key]
+            if val is not None:
+                try:
+                    setattr(snap.particles, key, val)
+                except AttributeError:
```
joaander, line 122: I'd prefer to see more explicit code here rather than exception tests which could be fragile. For example, set `type` and `N` explicitly and then loop over the known numpy arrays that need to be copied with `[:]`.

hoomd/snapshot.py
joaander, line 115: Call `gsd_snapshot.validate()` to ensure the snapshot is correct before trying the conversion.

hoomd/snapshot.py
```diff
@@ -109,3 +109,38 @@ def replicate(self, nx, ny, nz):
...
+                    setattr(snap.bonds, key, val)
+                except AttributeError:
+                    try:
+                        x = getattr(snap.bonds,key)
+                        x[:] = val
+                    except AttributeError:
+                        pass
+
+        # Set box attribute
+        snap.configuration.box = gsd_snapshot.configuration.box
```
joaander, line 145: Boxes work slightly differently in gsd and hoomd. Specifically, you need to check if `gsd_snapshot.configuration.dimensions == 2` and if it is force `box[3]=0`.

joaander (changes requested):
This is moving forward. In addition to the inline change requests, you also need to add unit tests to ensure that this implementation works.

hoomd/simulation.py
jennyfothergill: The way I'm thinking to do type checking is like this:
```
import gsd.hoomd
isinstance(snapshot, gsd.hoomd.Snapshot)
```
Is there a better way to do this that won't add gsd as a dependency?

hoomd/snapshot.py
jennyfothergill: I think the only only is `type_shapes`

hoomd/snapshot.py
jennyfothergill: yeah, I thought I was being clever, but you're right that it's less error-prone (and easier to follow) if I just set each attribute manually.

hoomd/simulation.py
b-butler: Better is subjective here. What we can do is look at the name of the object something like

```python
def _match_class_path(obj, *matches):
     return any(cls.__module__ + '.' + cls.__name__ in matches
                       for cls in inspect.getmro(type(obj)))
# Example usage
if _match_class_path(system, 'gsd.hoomd.Snapshot'):
    pass        
```
would work without introducing even soft dependencies (this is what freud does to support so many ways to define a system.

hoomd/snapshot.py
```diff
@@ -109,3 +109,65 @@ def replicate(self, nx, ny, nz):
 
     def _broadcast_box(self):
         self._cpp_obj._broadcast_box(self._comm.cpp_mpi_conf)
+
+    @classmethod
+    def _from_gsd_snapshot(cls, gsd_snap):
+        gsd_snap.validate()
+        if self._device.communicator.rank == 0:
```
b-butler, line 116: There is no self for a classmethod

b-butler (changes requested):
I left a couple of comments to address for improving the PR,

In addition, we might want to check that we only look at the data on the first rank. `Snapshot._from_gsd_snaphshot` should then take an optional device and use that device to only set arrays (or use the passed snapshot at all) on that rank. That way in all other ranks it can be `None` that was passed.

I would like to use the freud model here as I think it is better than using a try except block or adding `gsd` as a dependency.

hoomd/snapshot.py
jennyfothergill: OK, I must've misunderstood @joaander's suggestion:
>The whole set routine needs to be wrapped in an: `if self._device.communicator.rank == 0:` block.

I'm not sure what was the correct way to add this.

hoomd/simulation.py
jennyfothergill: Oh that's very cool! Thank you!

hoomd/snapshot.py
```diff
@@ -109,3 +110,104 @@ def replicate(self, nx, ny, nz):
 
     def _broadcast_box(self):
         self._cpp_obj._broadcast_box(self._comm.cpp_mpi_conf)
+
+    @classmethod
+    def _from_gsd_snapshot(cls, gsd_snap, communicator):
+        gsd_snap.validate()
+        if communicator.rank == 0:
+            snap = cls(communicator=communicator)
```
b-butler, line 118: This needs to be constructed outside the conditional.

hoomd/snapshot.py
```diff
@@ -109,3 +110,104 @@ def replicate(self, nx, ny, nz):
...
+            if gsd_snap.particles.moment_inertia is not None:
+                snap.particles.moment_inertia[:] = gsd_snap.particles.moment_inertia
+            if gsd_snap.particles.orientation is not None:
+                snap.particles.orientation[:] = gsd_snap.particles.orientation
+            if gsd_snap.particles.position is not None:
+                snap.particles.position[:] = gsd_snap.particles.position
+            if gsd_snap.particles.typeid is not None:
+                snap.particles.typeid[:] = gsd_snap.particles.typeid
+            if gsd_snap.particles.velocity is not None:
+                snap.particles.velocity[:] = gsd_snap.particles.velocity
```
b-butler, line 146: For these it would be more concise to do something like

```python

for property_ in ('positions', 'angmom', 'body', 'charge', 'diameter', ...):
    gsd_prop = getattr(gsd_snap.particles, property_)
    if gsd_prop is not None:
        setattr(snap.particles, gsd_prop)
```

I don't care strongly either way but I do prefer the more concise style. If you decide to implement this, something similar can be done for bonds, angles, etc.

hoomd/pytest/test_snapshot.py
```diff
@@ -209,6 +262,92 @@ def test_constraints(s):
...
         assert s.constraints.group.shape == (3, 2)
-        assert s.constraints.group.dtype == numpy.uint32
+        assert s.constraints.group.dtype == np.uint32
+
+
+@skip_gsd
+def test_from_gsd_snapshot_empty(s, device):
+    if s.exists:
+        gsd_snap = make_gsd_snapshot(s)
+        hoomd_snap = Snapshot._from_gsd_snapshot(gsd_snap, device.communicator)
```
b-butler, line 274: These should be outside the conditional.

hoomd/pytest/test_snapshot.py
```diff
@@ -209,6 +262,92 @@ def test_constraints(s):
...
+        assert_equivalent_snapshots(gsd_snap, hoomd_snap)
+
+
+@skip_gsd
+def test_from_gsd_snapshot_configuration(s, device):
+    if s.exists:
+        s.configuration.box = [10, 12, 7, 0.1, 0.4, 0.2]
+        gsd_snap = make_gsd_snapshot(s)
+        hoomd_snap = Snapshot._from_gsd_snapshot(gsd_snap, device.communicator)
```
b-butler, line 283: These should be outside the conditional.

hoomd/pytest/test_snapshot.py
```diff
@@ -209,6 +262,92 @@ def test_constraints(s):
...
+        assert_equivalent_snapshots(gsd_snap, hoomd_snap)
+
+
+@skip_gsd
+def test_from_gsd_snapshot_particles(s, device):
+    if s.exists:
+        s.particles.N = 5
+        s.particles.types = ['A', 'B']
+        gsd_snap = make_gsd_snapshot(s)
+        hoomd_snap = Snapshot._from_gsd_snapshot(gsd_snap, device.communicator)
```
b-butler, line 293: Same as above.

hoomd/pytest/test_snapshot.py
```diff
@@ -209,6 +262,92 @@ def test_constraints(s):
...
+        assert_equivalent_snapshots(gsd_snap, hoomd_snap)
+
+
+@skip_gsd
+def test_from_gsd_snapshot_bonds(s, device):
+    if s.exists:
+        s.bonds.N = 5
+        s.bonds.types = ['A', 'B']
+        gsd_snap = make_gsd_snapshot(s)
+        hoomd_snap = Snapshot._from_gsd_snapshot(gsd_snap, device.communicator)
```
b-butler, line 303: Same as above.

hoomd/pytest/test_snapshot.py
```diff
@@ -209,6 +262,92 @@ def test_constraints(s):
...
+        assert_equivalent_snapshots(gsd_snap, hoomd_snap)
+
+
+@skip_gsd
+def test_from_gsd_snapshot_angles(s, device):
+    if s.exists:
+        s.angles.N = 5
+        s.angles.types = ['A', 'B']
+        gsd_snap = make_gsd_snapshot(s)
+        hoomd_snap = Snapshot._from_gsd_snapshot(gsd_snap, device.communicator)
```
b-butler, line 313: Same as above.

hoomd/pytest/test_snapshot.py
```diff
@@ -209,6 +262,92 @@ def test_constraints(s):
...
+        assert_equivalent_snapshots(gsd_snap, hoomd_snap)
+
+
+@skip_gsd
+def test_from_gsd_snapshot_dihedrals(s, device):
+    if s.exists:
+        s.dihedrals.N = 5
+        s.dihedrals.types = ['A', 'B']
+        gsd_snap = make_gsd_snapshot(s)
+        hoomd_snap = Snapshot._from_gsd_snapshot(gsd_snap, device.communicator)
```
b-butler, line 323: Same as above.

hoomd/pytest/test_snapshot.py
```diff
@@ -209,6 +262,92 @@ def test_constraints(s):
...
+        assert_equivalent_snapshots(gsd_snap, hoomd_snap)
+
+
+@skip_gsd
+def test_from_gsd_snapshot_impropers(s, device):
+    if s.exists:
+        s.impropers.N = 5
+        s.impropers.types = ['A', 'B']
+        gsd_snap = make_gsd_snapshot(s)
+        hoomd_snap = Snapshot._from_gsd_snapshot(gsd_snap, device.communicator)
```
b-butler, line 333: Same as above.

hoomd/pytest/test_snapshot.py
```diff
@@ -209,6 +262,92 @@ def test_constraints(s):
...
+        assert_equivalent_snapshots(gsd_snap, hoomd_snap)
+
+
+@skip_gsd
+def test_from_gsd_snapshot_pairs(s, device):
+    if s.exists:
+        s.pairs.N = 5
+        s.pairs.types = ['A', 'B']
+        gsd_snap = make_gsd_snapshot(s)
+        hoomd_snap = Snapshot._from_gsd_snapshot(gsd_snap, device.communicator)
```
b-butler, line 343: Same as above.

hoomd/pytest/test_snapshot.py
```diff
@@ -209,6 +262,92 @@ def test_constraints(s):
...
+        assert_equivalent_snapshots(gsd_snap, hoomd_snap)
+
+
+@skip_gsd
+def test_from_gsd_snapshot_constraints(s, device):
+    if s.exists:
+        s.constraints.N = 3
+        gsd_snap = make_gsd_snapshot(s)
+        hoomd_snap = Snapshot._from_gsd_snapshot(gsd_snap, device.communicator)
```
b-butler, line 352: Same as above.

b-butler (changes requested):
I left a couple of comments. For testing, I think populating an entire snapshot with values on all the different properties and then performing a single test on all the properties equivalence would be sufficient. If you want to test what happens when some properties are none that could be handled in a second test.

Overall though this is looking good. We also have already gotten a request to make `_from_gsd_snapshot` public, so you can remove the underscore from this method.

hoomd/pytest/test_snapshot.py
```diff
@@ -1,6 +1,59 @@
 from hoomd.snapshot import Snapshot
-import numpy
+import numpy as np
```
joaander, line 2: This change introduces many unnecessary changes to the test. Please revert.

joaander (changes requested):
Thanks!

This is looking pretty good now. We just need to address @b-butler's comments and this will be good to merge.

jennyfothergill:
OK, I think I've implemented all suggestions. I'm a little confused about the `if communicator.rank == 0:` wrapper. Am I using it correctly and what is it for? Now that this is a public method, I should add a note about it in the docstring.

hoomd/pytest/test_snapshot.py
```diff
@@ -212,3 +265,31 @@ def test_constraints(s):
...
+    s.bonds.types = ['A', 'B']
+    s.angles.N = 5
+    s.angles.types = ['A', 'B']
+    s.dihedrals.N = 5
+    s.dihedrals.types = ['A', 'B']
+    s.impropers.N = 5
+    s.impropers.types = ['A', 'B']
+    s.pairs.N = 5
+    s.pairs.types = ['A', 'B']
+    s.constraints.N = 3
```
b-butler, line 292: Can we test that some of the array quantities get set correctly too?

b-butler (changes requested):
@jennyfothergill I made a couple of small changes if that is okay. I also left a comment regarding the test. I think the code should work, but I would like to test the array quantities as well.

hoomd/pytest/test_snapshot.py
jennyfothergill: I like the changes you made :) I updated that test to follow a similar format and set all array quantities as well. Should I be more careful to keep the datatype the same or does it not matter for this test? (e.g. `s.particles.body` gets set to a random float which doesn't make sense...)

jennyfothergill: mmm... I just went ahead and fixed it just in case

b-butler:
@jennyfothergill there are some MPI errors causing the tests to fail. Placing these sections in a `if snapshot.exists:` block should fix the errors.

joaander (approved):
Thanks! This looks good. 

`master` has a fix for the failing readthedocs builds.

b-butler (approved):
LGTM

b-butler:
\azp run validate

PR #894, opened 2020-12-16
Use 64-bit counter for time steps

### LINKED ISSUE #229: Use 64-bit time step counters

joaander:
## Description

* Switch from 32-bit to 64-bit counter for time steps.
* Refactor RNG seeding to include:
  *  8-bit class id
  * 16-bit user seed
  * The lower 40-bits of the time step (~ 1 trillion steps)
* Move user seed to the `Simulation` class

## Motivation and context

* Allow users to run simulations with more than 2^32 steps.
* Simplify seed setting so that users only need to set the seed in one place.

Resolves #229

## How has this been tested?

* Added unit tests to ensure correct seeding
* #809 enabled compiler warnings to flag when developers pass incorrect data types to the seed.

## Change log

```
Added:

* `Simulation.seed` - one place to set random number seeds for all operations.

Changed:

* Support timestep values in the range [0,2**64-1].
* [breaking] Removed *seed* argument from ``State.thermalize_particle_momenta``
* [breaking] Removed *seed* argument from ``md.methods.NVT.thermalize_thermostat_dof``
* [breaking] Removed *seed* argument from ``md.methods.NPT.thermalize_thermostat_and_barostat_dof``
* [breaking] Removed *seed* argument from ``md.methods.NPH.thermalize_barostat_dof``
* [breaking] Removed *seed* argument from ``md.methods.Langevin``
* [breaking] Removed *seed* argument from ``md.methods.Brownian``
* [breaking] Removed *seed* argument from ``md.force.Active``
* [breaking] Removed *seed* argument from ``md.pair.DPD``
* [breaking] Removed *seed* argument from ``md.pair.DPDLJ``
* [breaking] Removed *seed* argument from all HPMC integrators.
* [breaking] Removed *seed* argument from ``hpmc.update.Clusters``
* [breaking] Removed *seed* argument from ``hpmc.update.BoxMC``
* [breaking] Removed *seed* argument from ``hpmc.update.QuickCompress``
```

## Checklist:


hoomd/hpmc/IntegratorHPMCMonoGPU.cuh
```diff
@@ -978,7 +986,8 @@ __global__ void hpmc_insert_depletants(const Scalar4 *d_trial_postype,
         while (s_depletant_queue_size < max_depletant_queue_size && i_dep < n_depletants)
             {
             // one RNG per depletant
-            hoomd::RandomGenerator rng(hoomd::RNGIdentifier::HPMCDepletants, seed+i, i_dep, select*num_types + depletant_type, timestep);
+            hoomd::RandomGenerator rng(hoomd::Seed(hoomd::RNGIdentifier::HPMCDepletants, timestep, seed + (uint16_t)i_dep),
```
joaander, line 989: @jglaser do we expect `n_depletants`  to be larger than 65535 at rational simulation parameters?

jglaser: that would be quite a slow simulation, but if there is a large number in the process, it would be this one. Can shuffle order of arguments once #894 gets merged

jglaser: I would prefer to have that RNG argument to have a least 32 bits

joaander: Thanks for looking at this. It looks like these RNG are only used to generate a handful of numbers in the stream so we can safely use 16 more bits in the counter leaving only 16 bits for the stream.
How about I put the full 32-bit i_dep in the counter and then cast `select*num_types + depletant_type` down to 16 bits? Someone would either need a massive n_select or run 1 type per particle to overflow that one.

Or I could keep mixing 16-bits with the user seed but I'm not sure I like that now that I think about it. There would be some correlation between simulations with consecutive seeds. I suppose we could mix with the timestep instead - but that would lead to repeated depletion insertions after (2**24) ~16 million steps. 

`rank` also doesn't get that high in practice. How about I make `select` cast down to 16 bits and then mix the type with `rank`.

joaander:
Reviewers,

This PR is large in the number of files modified, but the vast majority of the changes come from a source code wide replacement of `unsigned int` with `uint64_t` for all `timestep` arguments and related values.

Things to focus on in a review:
* The new documentation for `Simulation.seed`
* Changes to the `RandomGenerator` seeding - ensure that the same seeds are kept before/after even if the order changed. In some cases (HPMC and depletants), I improved the seeding by making inputs that were mixed with `+` independent.

@mphoward You'll want to double check the seeding changes in mpcd.

hoomd/simulation.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, device):
         self._operations = Operations()
         self._operations._simulation = self
         self._timestep = None
+        self._seed = 0
```
mphoward, line 35: Does it make sense to default the seed to zero or to a random seed chosen from the system entropy? The former is more explicit for the user, but the latter is more like what is done by numpy. (i.e., don't set the seed unless you care about reproducing the simulation.)

mphoward (changes requested):
This is excellent Josh! I think it will make maintaining the RNG streams much easier. I have a few comments and questions, noted below.

hoomd/DCDDumpWriter.cc
```diff
@@ -212,7 +212,7 @@ void DCDDumpWriter::write_file_header(std::fstream &file)
     char cord_data[] = "CORD";
     file.write(cord_data, 4);
     write_int(file, 0);      // Number of frames in file, none written yet
-    write_int(file, m_start_timestep); // Starting timestep
+    write_int(file, (uint32_t)m_start_timestep); // Starting timestep
```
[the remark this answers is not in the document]
joaander: I think most tools will still read the DCD file, but the reported time step will be incorrect. A warning will help users understand why this is the case. I added one.

hoomd/DCDDumpWriter.cc
```diff
@@ -398,13 +398,13 @@ void DCDDumpWriter::write_frame_data(std::fstream &file, const SnapshotParticleD
...
 */
-void DCDDumpWriter::write_updated_header(std::fstream &file, unsigned int timestep)
+void DCDDumpWriter::write_updated_header(std::fstream &file, uint64_t timestep)
     {
     file.seekp(NFILE_POS);
     write_int(file, m_num_frames_written);
 
     file.seekp(NSTEP_POS);
-    write_int(file, timestep);
+    write_int(file, (unsigned int)timestep);
```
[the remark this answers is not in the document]
joaander: done.

hoomd/IMDInterface.cc
```diff
@@ -442,7 +442,7 @@ void IMDInterface::sendCoords(unsigned int timestep)
 
     // setup and send the energies structure
     IMDEnergies energies;
-    energies.tstep = timestep;
+    energies.tstep = (int32_t)timestep;
```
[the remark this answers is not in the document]
joaander: Yes, the IMD protocol as defined in `imd.h` (taken from VMD source code) has this as an `int32` which it defines as:
```#if ( INT_MAX == 2147483647 )
typedef int     int32;
#else
typedef short   int32;
#endif```

Realistically, I don't think anyone has used IMD in years and I may remove it before the v3 final release but for now I need to get it to compile with `-Wconversion`.

File: hoomd/RNGIdentifiers.h [hunk 6]
```diff
@@ -2,63 +2,63 @@
...
-    static const uint32_t TwoStepNPTMTK = 0x9db2f0ab;
-    static const uint32_t TwoStepNVTMTK = 0x451234b9;
-    static const uint32_t ATCollisionMethod = 0xf4009e6a;
-    static const uint32_t CollisionMethod = 0x5af53be6;
-    static const uint32_t SRDCollisionMethod = 0x7b61fda0;
-    static const uint32_t SlitGeometryFiller = 0xdb68c12c;
-    static const uint32_t SlitPoreGeometryFiller = 0xc7af9094;
-    static const uint32_t UpdaterQuickCompress = 0x00981234;
-    static const uint32_t ParticleGroupThermalize = 1;
+    ComputeFreeVolume,
```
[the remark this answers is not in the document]
joaander: I didn't consider the limitations of external plugins and this system. I will change it back to manually defined values. So much for letting the compiler do some work for us.

hoomd/md/TwoStepBD.cc
```diff
@@ -84,7 +84,8 @@ void TwoStepBD::integrateStepOne(unsigned int timestep)
         unsigned int ptag = h_tag.data[j];
 
         // Initialize the RNG
-        RandomGenerator rng(RNGIdentifier::TwoStepBD, m_seed, ptag, timestep);
+        RandomGenerator rng(hoomd::Seed(RNGIdentifier::TwoStepBD, timestep, seed),
```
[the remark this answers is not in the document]
joaander: I had not thought about that side effect of centralizing the seed. I will add some documentation to the `Seed` class regarding this.

hoomd/md/TwoStepNVTMTK.cc
```diff
@@ -451,7 +451,10 @@ void TwoStepNVTMTK::thermalizeThermostatDOF(unsigned int seed, unsigned int time
     Scalar sigmasq_t = Scalar(1.0)/((Scalar) g*m_tau*m_tau);
 
     bool master = m_exec_conf->getRank() == 0;
-    hoomd::RandomGenerator rng(hoomd::RNGIdentifier::TwoStepNVTMTK, seed, timestep);
+    hoomd::RandomGenerator rng(hoomd::Seed(hoomd::RNGIdentifier::TwoStepNVTMTK,
+                                           timestep,
+                                           m_sysdef->getSeed()),
+                                      hoomd::Counter());
```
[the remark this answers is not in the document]
joaander: In this case, I think I will solve it by including one of the particle tags from the group since those are disjoint. I suppose a static counter could be used in other cases if there were no other option. I can't think of any cleaner general solution.

hoomd/md/pytest/test_thermo.py
```diff
@@ -50,12 +50,12 @@ def _assert_thermo_properties(thermo, npart, rdof, tdof, pe, rke, tke, ke, p, pt
                                         thermo.rotational_degrees_of_freedom)
 
     np.testing.assert_allclose(thermo.potential_energy, pe)
-    np.testing.assert_allclose(thermo.rotational_kinetic_energy, rke, rtol=1e-5)
-    np.testing.assert_allclose(thermo.translational_kinetic_energy, tke, rtol=1e-5)
-    np.testing.assert_allclose(thermo.kinetic_energy, ke, rtol=1e-5)
-    np.testing.assert_allclose(thermo.kinetic_temperature, 2*thermo.kinetic_energy/thermo.degrees_of_freedom, rtol=1e-5)
-    np.testing.assert_allclose(thermo.pressure, p, rtol=1e-5)
-    np.testing.assert_allclose(thermo.pressure_tensor, pt, rtol=1e-5, atol=5e-5)
+    np.testing.assert_allclose(thermo.rotational_kinetic_energy, rke, rtol=1e-4)
```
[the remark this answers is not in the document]
joaander: The runs with different seeds produced values outside the initial tight tolerances.
These quantities fluctuate with a std dev of 1/sqrt(N) so I guess I could have increased N to compensate.

hoomd/mpcd/ATCollisionMethod.cc
```diff
@@ -13,14 +13,13 @@
 #include "hoomd/RNGIdentifiers.h"
 
 mpcd::ATCollisionMethod::ATCollisionMethod(std::shared_ptr<mpcd::SystemData> sysdata,
-                                           unsigned int cur_timestep,
-                                           unsigned int period,
+                                           uint64_t cur_timestep,
+                                           uint64_t period,
```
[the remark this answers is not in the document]
joaander: We don't have to convert periods to 64-bit values. But periods used in math operations with the timestep, so to get compile without warnings with `-Wconversion` we either need to allow 64-bit periods or add manual casting to any timestep/period math. I took what I thought was the shorter route. I doubt anyone will every attempt to set that large of a period.

hoomd/mpcd/SlitGeometryFiller.h
```diff
@@ -29,13 +29,12 @@ namespace mpcd
  * subject to the grid shift.
  */
 class PYBIND11_EXPORT SlitGeometryFiller : public mpcd::VirtualParticleFiller
-    {
+{
```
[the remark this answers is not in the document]
joaander: fixed.

hoomd/simulation.py
joaander: My preference is that "explicit is better than implicit" and script runs should be as reproducible as possible.  In hoomd v2, I required the seed argument to every constructor. I don't want to do that now because some users may run simulations that don't use random numbers and being forced to provide a seed would seem pointless to them. 

However, if others in the dev community prefer system entropy selected seeds I can be outvoted. What does everyone think? @glotzerlab/hoomd-api-reviewers

I should at least document the default seed behavior.

mphoward: I am OK either way. I can imagine the post, "Help! My Brownian dynamics simulation is always giving the same answer!" popping up quite a bit :-), but I myself always use a fixed seed when I am debugging and writing my production scripts. For playing around, though, it is nice to have numpy generate a new random stream automatically.

On the whole, I think I would vote to randomize the seed and let the user override this if they want to set it for reproducibility, but I'm OK with whatever gets decided.

bdice: > I think I would vote to randomize the seed and let the user override this if they want to set it for reproducibility

Agreed. This is what I would expect by default.

hoomd/md/TwoStepBD.cc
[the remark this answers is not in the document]
joaander: Classes where this could apply:
* HPMC integrators - These are effectively singletons as there is only one integrator per Simulation. I don't think this is an issue.
* HPMC BoxMC - This should be used as a singleton, but in reality users *could* instantiate multiple BoxMC updaters.
* Same for HPMC Cluster, MuVT, and QuickCompress updaters.
* mpcd ColissionMethod gridshift

@glotzerlab/hoomd-api-reviewers what do you think on the HPMC singleton updaters. Should we force them to be singletons in the Python API? Or should we allow multiple and communicate their position in the Updaters list to use as an instance id? Do you have another suggestion?

The thing I don't like about using a static variable solution is that then the seed is dependent on the number of classes and the order in which they are initialized. It is also global and not tied to the specific Simulation. If we want to go that route, I might just use the `this` pointer as an instance id - however that will not be reproducible.

hoomd/md/TwoStepNVTMTK.cc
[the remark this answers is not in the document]
joaander: done for NVT and NPT

hoomd/md/EvaluatorPairDPDLJThermo.h
```diff
@@ -242,7 +242,8 @@ class EvaluatorPairDPDLJThermo
                    m_oj = m_j;
                    }
 
-                hoomd::RandomGenerator rng(hoomd::RNGIdentifier::EvaluatorPairDPDThermo, m_seed, m_oi, m_oj, m_timestep);
+                hoomd::RandomGenerator rng(hoomd::Seed(hoomd::RNGIdentifier::EvaluatorPairDPDThermo, m_timestep, m_seed),
```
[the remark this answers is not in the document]
joaander: Yes, this is a good idea going forward. I'm also not sure it is worth the effort given the multiple layers in between the DPD evaluator and the caller.

hoomd/GetarInitializer.cc
```diff
@@ -521,7 +521,7 @@ namespace getardump{
                 }
             }
 
-        m_timestep = max(m_timestep, (unsigned int) atoi(selectedFrame.c_str()));
+        m_timestep = max(m_timestep, (uint64_t)atoi(selectedFrame.c_str()));
```
bdice, line 524:
```suggestion
        m_timestep = max(m_timestep, static_cast<uint64_t>(atoi(selectedFrame.c_str())));
```

hoomd/IMDInterface.cc
bdice, line 445:
```suggestion
    energies.tstep = static_cast<int32_t>(timestep);
```

hoomd/RandomNumbers.h
```diff
@@ -120,18 +120,87 @@ R123_CUDA_DEVICE R123_STATIC_INLINE Ftype uneg11(Itype in)
...
+        Code inside HOOMD should name the class ID value in RNGIdentifiers.h and ensure that each
+        class has a unique id. External plugins should use values 200 or larger.
+
+        id seed1 seed0 timestep4 | timestep3 timestep2 timestep1 timestep0
+    */
+    DEVICE Seed(uint8_t id, uint64_t timestep, uint16_t seed)
+        {
+        m_key = {{uint32_t(id) << 24 | uint32_t(seed) << 8
+                      | uint32_t((timestep & 0x000000ff00000000) >> 32),
+                  uint32_t(timestep & 0x00000000ffffffff)}};
```
bdice, line 153:
```suggestion
        m_key = {{static_cast<uint32_t>(id) << 24 | static_cast<uint32_t>(seed) << 8
                      | static_cast<uint32_t>((timestep & 0x000000ff00000000) >> 32),
                  static_cast<uint32_t>(timestep & 0x00000000ffffffff)}};
```

hoomd/RandomNumbers.h
```diff
@@ -120,18 +120,87 @@ R123_CUDA_DEVICE R123_STATIC_INLINE Ftype uneg11(Itype in)
...
+    {
+    public:
+    /** Default constructor.
+
+    Constructs a 0 valued counter.
+
+    Note: Only use the 4th argument when absolutely necessary and when you know that the resulting
+    RNG stream will not need to sample more than 65536 values.
+    */
+    DEVICE Counter(uint32_t a=0, uint32_t b=0, uint32_t c=0, uint16_t d=0) : m_ctr({{(uint32_t)d << 16, c, b, a}})
```
bdice, line 181:
```suggestion
    DEVICE Counter(uint32_t a=0, uint32_t b=0, uint32_t c=0, uint16_t d=0) : m_ctr({{static_cast<uint32_t>(d) << 16, c, b, a}})
```

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -795,21 +793,20 @@ void IntegratorHPMCMono<Shape>::update(unsigned int timestep)
...
-        return hoomd::RandomGenerator(this->m_seed,
-            timestep,
-            this->m_exec_conf->getRank(),
-            (uint32_t)(hash(std::this_thread::get_id())),
-            hoomd::RNGIdentifier::HPMCDepletants);
+        return hoomd::RandomGenerator(hoomd::Seed(hoomd::RNGIdentifier::HPMCDepletants,
+                                                  timestep,
+                                                  this->m_sysdef->getSeed()),
+                                      hoomd::Counter(this->m_exec_conf->getRank(),
+                                                     (uint32_t)hash(std::this_thread::get_id())));
```
bdice, line 809:
```suggestion
                                                     static_cast<uint32_t>(hash(std::this_thread::get_id()))));
```

hoomd/hpmc/IntegratorHPMCMonoGPU.cuh
```diff
@@ -978,7 +986,8 @@ __global__ void hpmc_insert_depletants(const Scalar4 *d_trial_postype,
         while (s_depletant_queue_size < max_depletant_queue_size && i_dep < n_depletants)
             {
             // one RNG per depletant
-            hoomd::RandomGenerator rng(hoomd::RNGIdentifier::HPMCDepletants, seed+i, i_dep, select*num_types + depletant_type, timestep);
+            hoomd::RandomGenerator rng(hoomd::Seed(hoomd::RNGIdentifier::HPMCDepletants, timestep, seed),
+                                       hoomd::Counter(i, rank*num_types + depletant_type, i_dep, (uint16_t)select));
```
bdice, line 990:
```suggestion
                                       hoomd::Counter(i, rank*num_types + depletant_type, i_dep, static_cast<uint16_t>(select)));
```

hoomd/hpmc/IntegratorHPMCMonoGPU.cuh
```diff
@@ -1051,7 +1060,8 @@ __global__ void hpmc_insert_depletants(const Scalar4 *d_trial_postype,
             // regenerate depletant using seed from queue, this costs a few flops but is probably
             // better than storing one Scalar4 and a Scalar3 per thread in shared mem
             unsigned int i_dep_queue = s_queue_didx[group];
-            hoomd::RandomGenerator rng(hoomd::RNGIdentifier::HPMCDepletants, seed+i, i_dep_queue, select*num_types + depletant_type, timestep);
+            hoomd::RandomGenerator rng(hoomd::Seed(hoomd::RNGIdentifier::HPMCDepletants, timestep, seed),
+                                       hoomd::Counter(i, rank*num_types + depletant_type, i_dep_queue, (uint16_t)select));
```
bdice, line 1064:
```suggestion
                                       hoomd::Counter(i, rank*num_types + depletant_type, i_dep_queue, static_cast<uint16_t>(select)));
```

hoomd/simulation.py
```diff
@@ -63,13 +64,43 @@ def timestep(self):
 
     @timestep.setter
     def timestep(self, step):
-        if step < 0:
-            raise ValueError("Timestep must be positive.")
+        if int(step) < 0 or int(step) > 2**64 - 1:
```
bdice, line 67: Should we define a module-level constant `TIMESTEP_MAX = 2**64 - 1` and `SEED_MAX = 2**16 - 1` and re-use those values in this file?

bdice (commented):
I reviewed about 2/3 of the code (need to move on to other tasks for now). I would recommend checking for (and removing) C-style casts. GCC can do this with `-Wold-style-cast`.

joaander:
> I reviewed about 2/3 of the code (need to move on to other tasks for now). I would recommend checking for (and removing) C-style casts. GCC can do this with `-Wold-style-cast`.

Certainly, we could use C++ casts. I'm not going to enable the warning on this PR as there are many more unrelated C-style casts throughout the codebase.

hoomd/simulation.py
joaander: done

hoomd/simulation.py
joaander: I talked with a few other developers offline. Many of them like the additional reproducibility that an explicit seed provides. Users who wish to use system entropy can opt in with `seed=secrets.randbits(16)`.

bdice: Is a simulation reproducible with the same seed and different devices? (My understanding is that the answer is “no.”) I have concerns that the implications of perfect-reproducibility-by-default-seed may be too strong.

mphoward: If this is the goal, it seems like it would be better to force the user to set the seed (reproducible and explicit) rather than take some default value (reproducible but implicit). I agree that it's confusing to require a seed in simulations that don't consume random numbers, though, so this would require registering classes that use random numbers and checking that the seed has been properly set.

joaander: @bdice These are always reproducible across versions and across compilers and devices:
* Randomized velocities
* Randomized thermostats
* Random forces and trial moves in the first time step

Due to their chaotic behavior, simulation trajectories will diverge due to floating point errors after a few hundred or thousand steps (HPMC trajectories will be stable for longer than MD). Different compilers and/or different devices will lead to different orders for floating point operations that will cause trajectory divergence. However, for multiple runs with the same build of HOOMD (e.g. container) on the same device (CPU or GPU), the following will be binary reproducible:

* HPMC trajectories on the CPU and GPU
* MD trajectories on the CPU
* MD trajectories on the GPU (but only when the user opts into a slower deterministic mode).

@mphoward That is an interesting idea and may be doable. I wonder how users with working scripts will react when they add a new operation or method that requires a seed and then get an error. Is that better or worse than requiring users to set a seed even when they know it won't be used? 

We could make it a warning instead of an error. Then the simulation would run with a default seed and notify the user that this is the case.

bdice: Great info, @joaander, that helps. I really like your idea of warning the user if the seed is left at the default value. That seems like the best tradeoff of user responsibility and user expectations.

hoomd/md/TwoStepBD.cc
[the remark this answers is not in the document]
joaander: @b-butler and I discussed this offline. We will add back in a settable instance id for these classes so that users can set unique values in the rare case that they are required.

joaander:
Reviewers, this is complete and ready for final review and approval.

I've added the warning we discussed when seed is unset as well as RNG instance id values for BoxMC, Clusters, and QuickCompress. I'll add it for MuVT after merging with the depletants branch.

hoomd/mpcd/CollisionMethod.cc
```diff
@@ -148,7 +141,8 @@ void mpcd::CollisionMethod::drawGridShift(unsigned int timestep)
     else
         {
         // PRNG using seed and timestep as seeds
-        hoomd::RandomGenerator rng(hoomd::RNGIdentifier::CollisionMethod, m_seed, timestep / m_period);
+        hoomd::RandomGenerator rng(hoomd::Seed(hoomd::RNGIdentifier::CollisionMethod, timestep, seed),
+                                   hoomd::Counter());
```
mphoward, line 145: With the current implementation of MPCD, there is only allowed to be one collision method in the system, so this does not technically need an instance ID. However, I am worried that I will forget about this discussion if I ever change that :-), so could you please add a setable instance for this class too?

mphoward (approved):
Thanks for addressing my comments! I left one more comment on the MPCD part (I reviewed hoomd, hoomd/md, and hoomd/mpcd), but I think this is otherwise ready to go from my side so I am approving.

I did grep for `hoomd::Counter()` and a couple more instances turned up in hpmc (IntegratorHPMCMono, IntegratorHPMCMonoGPU, UpdaterExternalFieldWall). Are those classes OK without an instance?

joaander:
Thanks for finding the corner cases I didn't consider!

There is only ever one HPMC integrator so `Counter()` is fine there. `UpdaterExternalFieldWall` will need a major update, but I might as well add the instance variable now so we don't forget.

hoomd/pytest/test_simulation.py
```diff
@@ -258,3 +242,64 @@ def act(self, timestep):
...
+
+def test_seed(simulation_factory, lattice_snapshot_factory):
+    sim = simulation_factory()
+    assert sim.seed is None
+
+    sim.seed = 42
+    assert sim.seed == 42
+
+    sim.seed = 0x123456789abcdef
+    assert sim.seed == 0xcdef
```
b-butler, line 299: We should also test the correct warning is generated.

b-butler (changes requested):
I looked at all the Python code and some of the C++ code. I only left one comment, in general this looks good to me.

hoomd/pytest/test_simulation.py
joaander: I tested this manually. If you know a way to write pytest scripts that check for particular output on stderr, feel free to add it.

joaander: Done.

joaander: Apparently, the `capsys` fixture in pytest causes `python` to segfault in the CI environment. https://dev.azure.com/glotzerlab/hoomd-blue/_build/results?buildId=9549&view=logs&j=fdb24454-8954-52a4-1b2d-7de6a8c2d450&t=50c16307-6c3f-5086-6e7e-00e0a5068b16&l=2291

@b-butler Is it critical that we unit test for these warnings that we need to resolve the testing issue, or should I just revert the commit.

bdice: You might be able to try `contextlib.redirect_stderr` and redirect to an internal StringIO object. https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stderr

We use that in signac/signac-flow with pytest.

b-butler: I wouldn't say it is critical, I just thought it would be a nicety. I forgot this was generated in C++ and not a Python warning.

b-butler: Although it is strange that it causes a segfault.

hoomd/hpmc/UpdaterMuVT.h
```diff
@@ -1979,11 +1971,8 @@ bool UpdaterMuVT<Shape>::moveDepletantsIntoNewPosition(unsigned int timestep, un
     #endif
 
     // initialize another rng
-    #ifdef ENABLE_MPI
-    hoomd::RandomGenerator rng(hoomd::RNGIdentifier::UpdaterMuVTDepletants2, timestep, this->m_seed, this->m_exec_conf->getPartition());
-    #else
-    hoomd::RandomGenerator rng(hoomd::RNGIdentifier::UpdaterMuVTDepletants2, timestep, this->m_seed);
-    #endif
+    hoomd::RandomGenerator rng(hoomd::Seed(hoomd::RNGIdentifier::UpdaterMuVTDepletants2, timestep, this->m_sysdef->getSeed()),
+                               hoomd::Counter(this->m_exec_conf->getPartition()));
```
b-butler, line 1975: I noticed you used a ternary operator with `getPartition` some where else for non-MPI runs, but here you just do the call.

b-butler (commented):
I left a few comments, but nothing major. The biggest thing would be that I think it may be better in some places where a random number is only needed for the duration of a function call to all a user to pass in a seed still such as in `NPT` or `NVT` thermo/baro-stat randomization. I know I made a comment about the instance member, but I realize these would not have that. I prefer this explicit disambiguation route rather than looking at particle tag implicitly in C++.

hoomd/md/TwoStepNPTMTK.cc
```diff
@@ -981,14 +981,21 @@ TwoStepNPTMTK::couplingMode TwoStepNPTMTK::getRelevantCouplings()
...
     {
     m_exec_conf->msg->notice(6) << "TwoStepNPTMTK randomizing thermostat and barostat DOF"
         << std::endl;
 
     IntegratorVariables v = getIntegratorVariables();
 
-    hoomd::RandomGenerator rng(hoomd::RNGIdentifier::TwoStepNPTMTK, seed, timestep);
+    unsigned int instance_id = 0;
+    if (m_group->getNumMembersGlobal() > 0)
+        instance_id = m_group->getMemberTag(0);
```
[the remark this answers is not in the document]
joaander: Integration methods need to select different particle groups, so the first tag in each group will be a unique value that will remain the same if the user reruns and keeps the same groups, even if objects have been instantiated in a different order (i.e. running cells multiple times in a single jupyter notebook session).

hoomd/hpmc/UpdaterMuVT.h
joaander: Jens wrote the ternary operators. I tried to convert as much of the seeding code as-is without rewriting to avoid errors.

joaander:
> I left a few comments, but nothing major. The biggest thing would be that I think it may be better in some places where a random number is only needed for the duration of a function call to all a user to pass in a seed still such as in `NPT` or `NVT` thermo/baro-stat randomization. I know I made a comment about the instance member, but I realize these would not have that. I prefer this explicit disambiguation route rather than looking at particle tag implicitly in C++.

It seems weird to me to ask for user seeds in some places but not others. Using multiple integration methods is common, so I think that HOOMD should assign independent random velocities to the different particle groups by default. Users shouldn't have to know the internal implementation details and know that they need to choose different seed values each time they call a function.

joaander:
Let's get this merged.

jglaser:
> Let's get this merged.

Yup, I will review. The seeding in the depletants branch is important to get right, have to check.

hoomd/md/methods.py
```diff
@@ -619,17 +604,13 @@ def thermalize_barostat_dof(self, seed):
...
-
-        Note:
-            The seed for the pseudorandom number stream includes the
-            simulation timestep and the provided *seed*.
         """
         if not self._attached:
             raise RuntimeError("Call Simulation.run(0) before"
                                "thermalize_thermostat_and_barostat_dof")
 
         self._cpp_obj.thermalizeThermostatAndBarostatDOF(
```
b-butler, line 612: We should raise a warning here if the seed is unset.

b-butler (commented):
I found one place where a warning needs to be added, but otherwise I think it is ready to merge.

joaander:
The macOS 10.15 builds are failing with an access error: https://dev.azure.com/glotzerlab/hoomd-blue/_build/results?buildId=10152&view=logs&j=7300be9e-b968-5142-d980-c4de163f8aef&t=da7f8c71-b9af-555f-68f2-7ddca1dafd10&l=14 . This is happening on all branches currently.

PR #899, opened 2020-12-18
Add oscillating pair potential

b-butler:
## Description

Adds the oscillating pair potential to HOOMD-blue.

## Motivation and context

I potentially plan on running simulations with this potential in the future,
so I went ahead and added it.

## How has this been tested?

Tested with the current pair potential testing infrastructure.

## Change log

```
Added: Oscillating pair potential to MD.
```

## Checklist:


CHANGELOG.rst
```diff
@@ -9,6 +9,7 @@ v3.0.0-beta.3 (not yet released)
 
 *Added*
 
+- Oscillating pair potential for MD in ``hoomd.md.pair.OPP``.
```
joaander, line 12: Please don't modify the change log in PRs. It almost guarantees conflicts which then requires that PRs modifying the change log need to be merged one at a time resolving conflicts in between. I update the change log with the message from the PR description directly on the master branch.

hoomd/md/pair.py
```diff
@@ -2377,3 +2377,75 @@ def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
...
+    .. math::
+        :nowrap:
+
+        \\begin{equation*}
+        V_{\\mathrm{OPP}}(r) = C_1 r^{-\\eta_1}
+            + C_2 r^{-\\eta_2} \\cos{\\left(k r - \\phi\\right)}
+        \\end{equation*}
+
+    See :py:class:`Pair` for details on how forces are calculated and the
+    available energy shifting and smoothing modes.  Use ~`params` dictionary
```
joaander, line 2404: ~`params` does not render correctly.

hoomd/md/pair.py
```diff
@@ -2377,3 +2377,75 @@ def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
...
+
+    Args:
+        nlist (:py:mod:`hoomd.md.nlist.NList`): Neighbor list
+        r_cut (float): Default cutoff radius (in distance units).
+        r_on (float): Default turn-on radius (in distance units).
+        mode (str): energy shifting/smoothing mode
+
+    `OPP` specifies that an oscillating pair potential should be applied between
+    every non-excluded particle pair in the simulation. The OPP potential can
+    be used to model metallic interactions.
```
joaander, line 2393: Add a citation to the original paper that introduced this potential.

joaander (changes requested):
Thanks! This has been on my list to implement for a while now.

Glotzer group work has previously used fixed values for eta1 and eta2. I see that you generalized the potential to take arbitrary values. Is there an interest in other values or in making this more general? `pow` is more expensive than a few multiplications, but that probably won't make much of a noticeable difference on the GPU where performance is limited by memory bandwidth.

CHANGELOG.rst
b-butler: My bad. I forget what packages to do this for.

hoomd/md/pair.py
b-butler: Will do.

b-butler:
@joaander, I took the paper on the isocohedral quasicrystal and used the general formulation from the supplementary information. I figured that while it would be slower (due to the pow function calls), the generality was wise to be more future proof. Plus the OPP, I think more technically means this more general form.

joaander (approved):
Thanks!

PR #903, opened 2021-01-04
Make Local Snapshots more Safe

b-butler:
## Description

Prevent `hoomd.Simulation.run` from being called in a local snapshot context manager. This prevents the likely to be most common error in using local snapshots as the device that the data is on may switch multiple times during a run. This does not prevent operations that can operate on a simulation outside of a run from invalidating buffers (for instance if ``BoxResize`` could run on the GPU and a user was accessing the data on a CPU this could overwrite any changes they made). More crucial is this prevents a user from accessing data that has been freed as the result of setting the simulation's state to another snapshot, changing particle number, etc.

## Motivation and context

Helps improve safety when using local snapshots.

## How has this been tested?

A test was added to ensure the proper errors were raised when attempting to run in a local snapshot context manager. A similar test for setting snapshots will soon be added.

## Change log

```
Changed: attempting to run in a local snapshot context manager will now raise a RuntimeError.
Changed: attempting to set the state to a new snapshot in a local snapshot context manager will now raise a RuntimeError.
```

## Checklist:

### DIFF
--- a/hoomd/pytest/test_local_snapshot.py
+++ b/hoomd/pytest/test_local_snapshot.py
@@ -466,12 +466,19 @@ class _TestLocalSnapshots:
                 tags = getattr(snapshot_section, tag_name)
                 property_check(hoomd_buffer, property_dict, tags)
 
-    def test_run_failure(self, simulation_factory, base_snapshot):
+    def test_run_failure(self, simulation_factory):
         sim = simulation_factory()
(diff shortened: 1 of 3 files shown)


b-butler:
@joaander The docs build locally for me, so I am not sure what is happening here.

b-butler:
Doc builds are fixed on #905.

joaander (approved):
Thanks!

PR #909, opened 2021-01-17
Convert dump.DCD to v3 API

### LINKED ISSUE #881: Port dump.DCD to v3 API
## Description

Port dump.DCD to the v3 API.

danevans09:
## Description

I created a _dcd.py_ folder in the _hoomd/write_ directory that contains the converted `DCD` class.

## Motivation and context

This pull request converts dump.DCD into the v3 API.

Resolves #881
## How has this been tested?

I tested that the `hoomd.write.DCD` object can be attached to a simulation, and if garnett is installed, I tested that the output in the DCD file is correct.

## Change log

```

```

## Checklist:


hoomd/pytest/test_dcd.py
```diff
@@ -0,0 +1,66 @@
+import hoomd
+import pytest
+import numpy as np
+
+
+def test_attach(simulation_factory, two_particle_snapshot_factory, tmp_path):
+    d = tmp_path / "sub"
+    d.mkdir()
```
joaander, line 8:
```suggestion
    d = tmp_path
```
`tmp_path` should already be a clean subdirectory - no need to create another subdirectory inside it.

Same for the other tests that follow pattern.

hoomd/pytest/test_dcd.py
```diff
@@ -0,0 +1,66 @@
...
+    filename = d / "temporary_test_file.dcd"
+    sim = simulation_factory(two_particle_snapshot_factory())
+    dcd_dump = hoomd.write.DCD(filename, hoomd.trigger.Periodic(1))
+    sim.operations.add(dcd_dump)
+    sim.operations._schedule()
+    for i in range(10):
+        sim.run(1)
```
joaander, line 15:
```suggestion
    sim.run(10)
```
Scheduling is automatic at the first call to `run`.

hoomd/write/dcd.py
```diff
@@ -0,0 +1,104 @@
+import hoomd
+from hoomd import _hoomd
+from hoomd.filter import ParticleFilter, All
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.operation import Writer
+
+
+class DCD(Writer):
+    R""" Writes simulation snapshots in the DCD format
```
joaander, line 9:
```suggestion
    """Writes simulation snapshots in the DCD format.
```
PEP8 docstring formatting conventions.

hoomd/write/dcd.py
```diff
@@ -0,0 +1,104 @@
...
+                          file *filename* will be overwritten.
+        unwrap_full (bool): When False, (the default) particle coordinates are always written inside the simulation box.
+                            When True, particles will be unwrapped into their current box image before writing to the dcd file.
+        unwrap_rigid (bool): When False, (the default) individual particles are written inside the simulation box which
+               breaks up rigid bodies near box boundaries. When True, particles belonging to the same rigid body will be
+               unwrapped so that the body is continuous. The center of mass of the body remains in the simulation box, but
+               some particles may be written just outside it. *unwrap_rigid* is ignored when *unwrap_full* is True.
+        angle_z (bool): When True, the particle orientation angle is written to the z component (only useful for 2D simulations)
+
+    Every *period* time steps a new simulation snapshot is written to the
```
joaander, line 26:
```suggestion
    On each timestep where `DCD` triggers, it writes the simulation snapshot the
```

hoomd/write/dcd.py
```diff
@@ -0,0 +1,104 @@
...
+    specified file in the DCD file format. DCD only stores particle positions,
+    in distance units.
+
+    Due to constraints of the DCD file format, once you stop writing to
+    a file via ``disable``, you cannot continue writing to the same file,
+    nor can you change the period of the dump at any time. Either of these tasks
+    can be performed by creating a new dump file with the needed settings.
+
```
joaander, line 34:
```suggestion
```
HOOMD v3 does not have `enable` / `disable`.

hoomd/write/dcd.py
```diff
@@ -0,0 +1,104 @@
...
+                 unwrap_rigid=False,
+                 angle_z=False):
+
+        # initialize base class
+        super().__init__(trigger)
+        self._param_dict.update(
+            ParameterDict(filename=str(filename),
+                          filter=ParticleFilter,
+                          overwrite=bool(overwrite),
+                          write_settings=[bool],
```
joaander, line 76: The way this is implemented, `unwrap_full`, `unwrap_rigid`, and `angle_z` are not accessible as separate parameters. You shouldn't need `write_settings`. Add the appropriate properties to the pybind11 export code for DCD to access these individually.

You'll also need a property for `overwrite` but make it a read only property.

hoomd/write/dcd.py
```diff
@@ -0,0 +1,104 @@
...
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.operation import Writer
+
+
+class DCD(Writer):
+    R""" Writes simulation snapshots in the DCD format
+
+    Args:
+        filename (str): File name to write.
+        trigger (hoomd.trigger.Trigger): Select the timesteps to write.
```
joaander, line 13:
```suggestion
        trigger (hoomd.trigger.Periodic): Select the timesteps to write.
```
Document explicitly that only periodic triggers are allowed.

hoomd/write/dcd.py
```diff
@@ -0,0 +1,104 @@
...
+        self._cpp_obj.setAngleZ(write_settings[2])
+        super()._attach()
+
+    def enable(self):
+        """ The DCD dump writer cannot be re-enabled """
+        if self.enabled is False:
+            raise RuntimeError('DCD output cannot be re-enabled after it has been disabled');
+
+    def set_period(self, period):
+        raise RuntimeError('The period of a DCD writer cannot be changed')
```
joaander, line 104:
```suggestion
```
Not needed in v3.

hoomd/write/dcd.py
```diff
@@ -0,0 +1,104 @@
...
+        group = self._simulation.state._get_group(self.filter)
+
+        self._cpp_obj = _hoomd.DCDDumpWriter(self._simulation.state._cpp_sys_def,
+                                             self.filename,
+                                             int(self.trigger.period),
+                                             group,
+                                             self.overwrite)
+        self._cpp_obj.setUnwrapFull(write_settings[0])
+        self._cpp_obj.setUnwrapRigid(write_settings[1])
+        self._cpp_obj.setAngleZ(write_settings[2])
```
joaander, line 95:
```suggestion
```
Once these are available as properties in the pybind11 export, `super.attach()` will set them.

hoomd/pytest/test_dcd.py
```diff
@@ -0,0 +1,66 @@
...
+def test_enabled(tmp_path):
+    d = tmp_path / "sub"
+    d.mkdir()
+    filename = d / "temporary_test_file.dcd"
+    dcd_dump = hoomd.write.DCD(filename, hoomd.trigger.Periodic(1))
+    dcd_dump.enabled = False
+    with pytest.raises(RuntimeError):
+        dcd_dump.enable()
+
+
```
joaander, line 36:
```suggestion
```
Not needed in v3.

hoomd/pytest/test_dcd.py
```diff
@@ -0,0 +1,66 @@
...
+def test_write(simulation_factory, two_particle_snapshot_factory, tmp_path):
+    garnett = pytest.importorskip("garnett")
```
joaander, line 38: Nice! I didn't know this API existed. This is very clean.

I will rebuild the CI images with garnett so this test can run.

joaander (changes requested):
Thanks! I have a few changes to request. Most of them are suggestions that you can commit directly in the GitHub PR. One request requires adding some code to `DCDDumpWriter.cc`

joaander (approved):
Thanks! This looks great.

I made one small change to add DCD to the documentation. @b-butler do you want to review before merging?

hoomd/pytest/test_dcd.py
```diff
@@ -0,0 +1,40 @@
...
+    sim.operations.add(dcd_dump)
+    snap = sim.state.snapshot
+    positions = []
+    for i in range(10):
+        snap = sim.state.snapshot
+        position1 = np.asarray(snap.particles.position[0])
+        position2 = np.asarray(snap.particles.position[1])
+        position1 += 0.1 * i * (-1)**i
+        position2 += 0.1 * (i + 1) * (-1)**(i - 1)
+        if snap.exists:
```
joaander, line 29: Move this `if snap.exists` up to include all code that accesses snap.particles.position.

hoomd/pytest/test_dcd.py
```diff
@@ -0,0 +1,40 @@
...
+            snap.particles.position[0] = position1
+            snap.particles.position[1] = position2
+        sim.state.snapshot = snap
+        sim.run(1)
+        positions.append([list(position1), list(position2)])
+    with open(filename, 'rb') as dcdfile:
```
joaander, line 35: Wrap the whole open and assert in a `if sim.device.communicator.rank == 0:` check. With the change above, `positions` will only exist on rank 0.

joaander (changes requested):
I spoke too soon. MPI tests are failing. Here is how to fix them:

hoomd/write/dcd.py
```diff
@@ -0,0 +1,83 @@
...
+               some particles may be written just outside it. *unwrap_rigid* is ignored when *unwrap_full* is True.
+        angle_z (bool): When True, the particle orientation angle is written to the z component (only useful for 2D simulations)
+
+    On each timestep where `DCD` triggers, it writes the simulation snapshot to the
+    specified file in the DCD file format. DCD only stores particle positions,
+    in distance units.
+
+    Examples::
+
+        hoomd.write.DCD("trajectory.dcd", hoomd.trigger.Periodic(1000))
```
b-butler, line 32: This needs to be set to a variable.

hoomd/write/dcd.py
```diff
@@ -0,0 +1,83 @@
...
+        dcd = hoomd.write.DCD(filename="data/dump.dcd", trigger=hoomd.trigger.Periodic(100, 10))
+
+    Warning:
+        When you use dump.dcd to append to an existing dcd file:
```
b-butler, line 36:
```suggestion
        When you use `hoomd.writer.DCD` to append to an existing dcd file:
```
This needs to be changed in a few places.

hoomd/write/dcd.py
```diff
@@ -0,0 +1,83 @@
...
+        # initialize base class
+        super().__init__(trigger)
+        self._param_dict.update(
+            ParameterDict(filename=str(filename),
+                          filter=ParticleFilter,
+                          overwrite=bool(overwrite),
+                          unwrap_full=bool(unwrap_full),
+                          unwrap_rigid=bool(unwrap_rigid),
+                          angle_z=bool(angle_z),
+                          _defaults=dict(filter=filter)))
```
b-butler, line 74: Don't use `_defaults` here as this will not check for errors, set the filter after the fact.

hoomd/pytest/test_dcd.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_write(simulation_factory, two_particle_snapshot_factory, tmp_path):
+    garnett = pytest.importorskip("garnett")
+    dcd_reader = garnett.reader.DCDFileReader()
+    filename = tmp_path / "temporary_test_file.dcd"
+    sim = simulation_factory(two_particle_snapshot_factory())
+    dcd_dump = hoomd.write.DCD(filename, hoomd.trigger.Periodic(1))
+    sim.operations.add(dcd_dump)
+    snap = sim.state.snapshot
```
b-butler, line 21: Why do we have this line?

hoomd/pytest/test_dcd.py
```diff
@@ -0,0 +1,41 @@
...
+        if snap.exists:
+            position1 = np.asarray(snap.particles.position[0])
+            position2 = np.asarray(snap.particles.position[1])
+            position1 += 0.1 * i * (-1)**i
+            position2 += 0.1 * (i + 1) * (-1)**(i - 1)
+            snap.particles.position[0] = position1
+            snap.particles.position[1] = position2
+        sim.state.snapshot = snap
+        sim.run(1)
+        positions.append([list(position1), list(position2)])
```
b-butler, line 34: `position1` and `position2` won't exist on all ranks here.

b-butler: This needs to be in the conditional.

b-butler (changes requested):
I have a few comments that need to be addressed otherwise, the PR looks good.

joaander (approved):
I couldn't figure out why the MPI tests were failing. Something about the particle positions set in the snapshot caused the simulation run step to deadlock. This isn't an issue that we need to solve right now. I removed the multiple time steps from the test so it should pass now.

Thanks for working on this! @b-butler were all your requests addressed?

b-butler (approved):
LGTM, I went ahead an applied yapf to the new files and made the docstring match our 80 character line standard.

@joaander should we make an issue about the deadlock? I don't know if we can expect this to happen elsewhere.

joaander:
@b-butler You can file an issue if you want. I suspect it has something to do with running a simulation without an integrator because all of our other MPI tests pass without issues.

PR #917, opened 2021-01-29
Convert Berendsen to v3 API

### LINKED ISSUE #864: Port Berendsen to the v3 API
## Description

Port the Berendsen integrator to the v3 API and add tests.

danevans09:
## Description

This converts the `Berendsen` method to the v3 API. I added an `_attach `method and added the parameter dict properties to the pybind export of the C++ class. I removed the `randomize_velocities` method, as there is no function related to randomizing velocities in _hoomd/md/TwoStepBerendsen.cc_.
## Motivation and context

The Berendsen method has not yet been converted to the v3 API

Resolves #864 

## How has this been tested?

I added the same tests as the other method classes for testing the parameters before and after attaching.

## Change log

```

```

## Checklist:

### DIFF
--- a/hoomd/md/methods.py
+++ b/hoomd/md/methods.py
@@ -955,7 +955,7 @@ class Berendsen(_Method):
 
         tau (float): Time constant of thermostat. (in time units)
     """
-    def __init__(self, group, kT, tau):
+    def __init__(self, filter, kT, tau):
         # store metadata
         param_dict = ParameterDict(
             filter=ParticleFilter,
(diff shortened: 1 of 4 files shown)


hoomd/md/methods.py
```diff
@@ -913,15 +913,19 @@ def _attach(self):
         super()._attach()
 
 
-class berendsen(_Method):
+class Berendsen(_Method):
     R""" Applies the Berendsen thermostat.
```
b-butler, line 918: Should we warn that this does not sample the canonical distribution properly, @joaander?

vyasr: I think HOOMD generally assumes that users choosing a thermostat know what they're doing wrt to the physics at this level, right? Or are we trying to provide guidance on what thermostat/barostat/etc to use v3? That seems out of scope for library documentation, but I might have missed that decision.

joaander: Agreed. The library documentation should state what it is and what the parameters are.

hoomd/md/pytest/test_methods.py
```diff
@@ -68,6 +68,57 @@ def test_brownian_attributes_attached(simulation_factory,
     assert brownian.alpha == 0.125
 
 
+def test_berendsen_attributes():
```
joaander, line 71:
```suggestion
@pytest.mark.serial
def test_berendsen_attributes():
```
Berendsen fails in MPI.

hoomd/md/pytest/test_methods.py
```diff
@@ -68,6 +68,57 @@ def test_brownian_attributes_attached(simulation_factory,
...
+
+    ramp = hoomd.variant.Ramp(1, 2, 1000000, 2000000)
+    berendsen.kT = ramp
+    assert berendsen.kT is ramp
+
+    berendsen.tau = 1.2
+    assert berendsen.tau == 1.2
+
+
+def test_berendsen_attributes_attached(simulation_factory,
```
joaander, line 93:
```suggestion
@pytest.mark.serial
def test_berendsen_attributes_attached(simulation_factory,
```

joaander (changes requested):
Thanks!
I've made some change requests below.

joaander (approved):
Thanks!

b-butler (approved):
LTGM

PR #923, opened 2021-02-09
Convert compute.free_volume to v3 API

### LINKED ISSUE #879: Port compute.free_volume to the v3 API
## Description

Port compute.free_volume to the v3 API and add tests.

## Motivation and context

The compute should provide its results as loggable quantities.

danevans09:
## Description

This converts the free volume compute to the v3 API. I added a parameter dict to the `__init__` method of the python class, and changed `num_samples` and `test_particle_type` to properties of the pybind class. I also added a `log` method that calls the C++ class to return the free volume. 
## Motivation and context
The free volume compute has not yet been converted to the v3 API.

Resolves #879 

## How has this been tested?

I tested that the properties of the python class are set properly before attaching. I then tested that the properties of the class are set properly after attaching and tested that the compute could run on a simulation.

## Change log

```

```

## Checklist:


hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,134 @@
...
-        # initialize base class
-        _compute.__init__(self);
-
+    def __init__(self, mc, seed, test_type=None, nsample=None):
+        # store metadata
+        param_dict = ParameterDict(
+            mc=integrate.HPMCIntegrator,
+            seed=int,
+            test_particle_type=Either([to_type_converter(str),
+                                      to_type_converter(int)]),
```
b-butler, line 50:
```suggestion
            test_particle_type=OnlyTypes((str, int)),
```

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,134 @@
...
+            num_samples=int
+        )
+        param_dict.update(
+            dict(mc=mc,
+                 seed=int(seed),
+                 test_particle_type=str(test_type),
+                 num_samples=int(nsample)))
```
b-butler, line 57: No need for explicit conversions here. That will be handled by the param_dict.

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,134 @@
...
-            elif isinstance(mc, integrate.faceted_ellipsoid):
-                cls =_hpmc.ComputeFreeVolumeFacetedEllipsoid;
-            elif isinstance(mc, integrate.polyhedron):
-                cls =_hpmc.ComputeFreeVolumePolyhedron;
-            elif isinstance(mc, integrate.sphinx):
-                cls =_hpmc.ComputeFreeVolumeSphinx;
-            elif isinstance(mc, integrate.convex_spheropolyhedron_union):
+        cls = None
+        if isinstance(self._simulation.device, hoomd.device.CPU):
+            if isinstance(self.mc, integrate.Sphere):
```
b-butler, line 66: I would prefer using a construct like that in `hoomd/hpmc/update.py` for the `Cluster` classes `attach` method.

hoomd/hpmc/ComputeFreeVolume.h
```diff
@@ -303,9 +320,33 @@ Scalar ComputeFreeVolume<Shape>::getLogValue(const std::string& quantity, unsign
     throw std::runtime_error("Undefined log quantity");
     }
 
+// \return the free volume.
+template<class Shape>
+Scalar ComputeFreeVolume<Shape>::getFreeVolume()
```
b-butler, line 325: The way this is written and implemented, it seems that we will never from Python run `computeFreeEnergy`. Is there a reason we have separated the computation from the accessing of the computed quantity?

b-butler (changes requested):
@joaander, admittedly I have not looked at the old code too closely, but would it be possible to support a mode using a diameter rather than an existent type to test for the free volume? I could see users wanting that feature. If not we should add documentation that users can add non-existent (in the simulation) types to improve free volume calculations.

@danevans09 we should have some validation checks that the free volume calculation for simple systems is accurate to a certain degree.

joaander:
> @joaander, admittedly I have not looked at the old code too closely, but would it be possible to support a mode using a diameter rather than an existent type to test for the free volume? I could see users wanting that feature. If not we should add documentation that users can add non-existent (in the simulation) types to improve free volume calculations.

That would require writing new sphere/Shape overlap checks for every shape class in HPMC. This is doable, but would be a very large project in the implementation and testing and non-trivial for a number of shape classes (i.e. simple polygons, general polyhedra, all of the union shapes).

I haven't looked at the details of this code, but it should be possible to reengineer it to accept its own shape definition - however this will mean duplicating all of the TypeParameter shape logic from the HPMC integrators in ComputeFreeVolume. Though a bit kludgey, it seems cleaner and a lot less work to store the shape parameters for the query shape in the HPMC integrator in one of the (possibly unused) particle types. 

I believe the main use-case for ComputeFreeVolume is to in systems with implicit depletants where one of the particle types defines the shape of the depletant particles.

hoomd/hpmc/ComputeFreeVolume.h
danevans09: I copied the code from the `getLogValue` method into the `getFreeVolume` method to have the Python log `free_volume` method return a `free_volume` pybind property, because it looks like that was what was done in `md/ComputeThermo.h` and `compute.ThermodynamicQuantities`, I can change it to directly call `computeFreeVolume` from Python though

b-butler:
>Though a bit kludgey, it seems cleaner and a lot less work to store the shape parameters for the query shape in the HPMC integrator in one of the (possibly unused) particle types.
@joaander given the work necessary to add that feature, I agree. I think just documenting the practice of adding a non-existent type to the state/integrator for querying free volume would be enough.

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,111 @@
...
-            elif isinstance(mc, integrate.sphere_union):
-                cls = _hpmc.ComputeFreeVolumeSphereUnion;
-            else:
-                hoomd.context.current.device.cpp_msg.error("compute.free_volume: Unsupported integrator.\n");
-                raise RuntimeError("Error initializing compute.free_volume");
+    def __init__(self, mc, test_type=None, nsample=None):
+        # store metadata
+        param_dict = ParameterDict(
+            mc=integrate.HPMCIntegrator,
+            test_particle_type=OnlyTypes((str, int)),
```
joaander, line 47:
```suggestion
            test_particle_type=OnlyTypes(str),
```
Particle types should be strings in user-facing code.

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,111 @@
...
-            elif isinstance(mc, integrate.convex_spheropolyhedron_union):
-                cls = _hpmc.ComputeFreeVolumeConvexPolyhedronUnion
-            elif isinstance(mc, integrate.faceted_ellipsoid_union):
-                cls = _hpmc.ComputeFreeVolumeFacetedEllipsoidUnion
-            elif isinstance(mc, integrate.sphere_union):
-                cls = _hpmc.ComputeFreeVolumeSphereUnion;
-            else:
-                hoomd.context.current.device.cpp_msg.error("compute.free_volume: Unsupported integrator.\n");
-                raise RuntimeError("Error initializing compute.free_volume");
+    def __init__(self, mc, test_type=None, nsample=None):
```
joaander, line 43:
```suggestion
    def __init__(self, test_particle_type, num_samples):
```
`None` is not a valid setting for `test_type` or `nsample`. Also, the names of the arguments should match those of the attributes.

In hoomd v3, we don't require users to pass in `mc` objects in the constructor - the code finds the system's integrator on attaching. See `update.Clusters` for an example.

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,111 @@
...
-from hoomd.operation import Compute
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.data.typeconverter import OnlyTypes
+from hoomd.logging import log
 import hoomd
 
-class free_volume(Compute):
-    R""" Compute the free volume available to a test particle by stochastic integration.
+class FreeVolume(Compute):
+    """Compute the free volume available to a test particle by stochastic integration.
```
joaander, line 19: Update the `Args:` list to match the new constructor arguments (see below). Also, add an `Attributes:` section that lists the param_dict entries.

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,111 @@
...
     It works together with an HPMC integrator, which defines the particle types used in the simulation.
     As parameters it requires the number of MC integration samples (*nsample*), and the type of particle (*test_type*)
     to use for the integration.
 
-    Once initialized, the compute provides a log quantity
-    called **hpmc_free_volume**, that can be logged via ``hoomd.analyze.log``.
-    If a suffix is specified, the log quantities name will be
-    **hpmc_free_volume_suffix**.
+    A :py:class`FreeVolume` object can be added to a logger for logging during a simulation,
+    see :py:class:`hoomd.logging.Logger` for more details.
```
joaander, line 32:
```suggestion
```
All HOOMD operations can be added to loggers. We don't need to mention that explicitly here.

hoomd/hpmc/ComputeFreeVolume.h
```diff
@@ -45,13 +45,27 @@ class ComputeFreeVolume : public Compute
...
+
         //! Set the number of MC samples to perform
+        //! \param n_sample number of MC steps to sample
         void setNumSamples(unsigned int n_sample)
             {
             m_n_sample = n_sample;
             }
 
+        //! Get the type of depletant particle
+        unsigned int getTestParticleType()
```
joaander, line 62: For this to work with the associated `param_dict` entry `getTestParticleType` needs to return the string name of the particle type and `setParticleType` needs to take in a string argument. Perform the needed string<->int mappings in C++.

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,111 @@
...
+                                (integrate.SimplePolygon, _hpmc.ComputeFreeVolumeSimplePolygonGPU),
+                                (integrate.ConvexPolyhedron, _hpmc.ComputeFreeVolumeConvexPolyhedronGPU),
+                                (integrate.ConvexSpheropolyhedron, _hpmc.ComputeFreeVolumeSpheropolyhedronGPU),
+                                (integrate.Polyhedron, _hpmc.ComputeFreeVolumePolyhedronGPU),
+                                (integrate.Ellipsoid, _hpmc.ComputeFreeVolumeEllipsoidGPU),
+                                (integrate.ConvexSpheropolygon, _hpmc.ComputeFreeVolumeSpheropolygonGPU),
+                                (integrate.FacetedEllipsoid, _hpmc.ComputeFreeVolumeFacetedEllipsoidGPU),
+                                (integrate.SphereUnion, _hpmc.ComputeFreeVolumeSphereUnionGPU),
+                                (integrate.ConvexSpheropolyhedronUnion, _hpmc.ComputeFreeVolumeConvexPolyhedronUnionGPU),
+                                (integrate.FacetedEllipsoidUnion, _hpmc.ComputeFreeVolumeFacetedEllipsoidUnionGPU)]
```
joaander, line 91: Use getattr and string concatenation to simplify this code. This is what another PR uses in a similar way:
```
                my_class = getattr(_md, 'TwoStepRATTLENVE' + self.manifold_constraint.name)
            else:
                my_class = getattr(_md, 'TwoStepRATTLENVE' + self.manifold_constraint.name + 'GPU')
```

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,111 @@
...
-
-        if test_type is not None:
-            itype = hoomd.context.current.system_definition.getParticleData().getTypeByName(test_type)
-            self.cpp_compute.setTestParticleType(itype)
-        if nsample is not None:
-            self.cpp_compute.setNumSamples(int(nsample))
-
-        hoomd.context.current.system.addCompute(self.cpp_compute, self.compute_name)
-        self.enabled = True
+            return None
```
joaander, line 117: @bdice I recall you some alternate to returning `None` in these cases. What was that suggestion?

hoomd/hpmc/ComputeFreeVolume.h
joaander: `computeFreeVolume` is the internal implementation of the method called at the appropriate time in `compute`. 

`computeFreeVolume` should be a `protected` member. The method that Python needs to call is `compute`, which @danevans09 calls in the `free_volume` property.

hoomd/hpmc/pytest/test_compute_free_volume.py
```diff
@@ -0,0 +1,64 @@
...
+    free_volume = (7**3) * (1 - (4 / 3) * np.pi * (radius1 + radius2)**3)
+    free_volume = max([0.0, free_volume])
+    sim = simulation_factory(lattice_snapshot_factory(particle_types=['A', 'B']))
+    mc = hoomd.hpmc.integrate.Sphere()
+    mc.shape["A"] = {'diameter': radius1 * 2}
+    mc.shape["B"] = {'diameter': radius2 * 2}
+    mc.depletant_fugacity["B"] = 1.5
+    sim.operations.add(mc)
+
+    sim.operations._schedule()
```
joaander, line 60:
```suggestion
```
Is this necessary? There is a `sim.run(0)` below.

hoomd/hpmc/pytest/test_compute_free_volume.py
```diff
@@ -0,0 +1,64 @@
...
+          (0.4, 0.05),
+          (0.7, 0.17)]
+
+@pytest.mark.parametrize("radius1, radius2", _radii)
+def test_validation_systems(simulation_factory,
+                            two_particle_snapshot_factory,
+                            lattice_snapshot_factory,
+                            radius1,
+                            radius2):
+    free_volume = (7**3) * (1 - (4 / 3) * np.pi * (radius1 + radius2)**3)
```
joaander, line 51: Nice test!

I'm not sure why it is failing in some configurations. I am going to investigate.

joaander (changes requested):
Thanks! I have some change requests to improve this PR.

I am going to investigate why the tests fail in some configurations.

joaander:
Bug solved, there was a missing `setCommunicator`. I also fixed a few other performance related issues in the GPU free volume code I found along the way. The tests on this branch should pass now. @danevans09 once you make the API related change requests, we can merge this in.

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,85 @@
...
-            elif isinstance(mc, integrate.sphere_union):
-                cls = _hpmc.ComputeFreeVolumeGPUSphereUnion;
-            elif isinstance(mc, integrate.faceted_ellipsoid_union):
-                cls = _hpmc.ComputeFreeVolumeGPUFacetedEllipsoidUnion;
-            elif isinstance(mc, integrate.convex_spheropolyhedron_union):
-                cls = _hpmc.ComputeFreeVolumeGPUConvexPolyhedronUnion;
-            else:
-                hoomd.context.current.device.cpp_msg.error("compute.free_volume: Unsupported integrator.\n");
-                raise RuntimeError("Error initializing compute.free_volume");
+                cpp_cls = getattr(_hpmc, 'ComputeFreeVolume' + integrator_name + 'GPU')
```
b-butler, line 71: It seems the appropriate class name would be `'ComputeFreeVolumeGPU' + integrator_name`.

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,85 @@
...
+        # set defaults
+        self._param_dict.update(param_dict)
+
+    def _attach(self):
+        integrator = self._simulation.operations.integrator
+        if not isinstance(integrator, integrate.HPMCIntegrator):
+            raise RuntimeError("The integrator must be an HPMC integrator.")
+
+        # Extract 'Shape' from '<hoomd.hpmc.integrate.Shape object>'
+        integrator_name = str(integrator).split()[0].split('.')[-1]
```
b-butler, line 66: You could just use this for the same effect.
```suggestion
        integrator_name = integrator.__class__.__name__
```

hoomd/hpmc/compute.py
```diff
@@ -7,123 +7,85 @@
...
-            elif isinstance(mc, integrate.faceted_ellipsoid_union):
-                cls = _hpmc.ComputeFreeVolumeGPUFacetedEllipsoidUnion;
-            elif isinstance(mc, integrate.convex_spheropolyhedron_union):
-                cls = _hpmc.ComputeFreeVolumeGPUConvexPolyhedronUnion;
-            else:
-                hoomd.context.current.device.cpp_msg.error("compute.free_volume: Unsupported integrator.\n");
-                raise RuntimeError("Error initializing compute.free_volume");
+                cpp_cls = getattr(_hpmc, 'ComputeFreeVolume' + integrator_name + 'GPU')
+        except AttributeError:
+            raise RuntimeError("Unsupported integrator.\n")
```
b-butler, line 73: No need to end with a new line here.

hoomd/hpmc/pytest/test_compute_free_volume.py
```diff
@@ -0,0 +1,74 @@
+import hoomd
+import pytest
+import numpy as np
+
+
+def test_before_attaching():
+    mc = hoomd.hpmc.integrate.Sphere()
+    mc.shape["A"] = {'diameter': 1.0}
+    mc.shape["B"] = {'diameter': 0.2}
+    mc.depletant_fugacity["B"] = 1.5
```
b-butler, line 10: These lines are unnecessary.

hoomd/hpmc/pytest/test_compute_free_volume.py
```diff
@@ -0,0 +1,74 @@
...
+    sim.operations.add(mc)
+
+    free_volume_compute = hoomd.hpmc.compute.FreeVolume(test_particle_type='B',
+                                                        num_samples=10000)
+    sim.operations.add(free_volume_compute)
+    sim.run(0)
+    np.testing.assert_allclose(free_volume,
+                               free_volume_compute.free_volume,
+                               rtol=2e-2)
```
b-butler, line 74: Add a comment explaining that the `rtol` is fairly high since the free volume for sized particles won't be all the available free volume.

b-butler (changes requested):
Overall it is looking good. I have left some comments to address.

hoomd/hpmc/compute.py
danevans09: Should I change this in the pybind export in all of the `module.cc` files? Because for example `module_convex_polyhedron.cc` has this:

`export_ComputeFreeVolumeGPU< ShapeConvexPolyhedron >(m, "ComputeFreeVolumeConvexPolyhedronGPU");`

joaander: Let's make an issue to clean this up later to avoid conflicts with #659 which rewrites how all of these module files.

joaander (approved):
Thanks, this is good to merge!

I made made some final tweaks to the documentation. @Tobias-Dwyer you recently asked me about free volume computations. Could you review the new docs https://hoomd-blue--923.org.readthedocs.build/en/923/module-hpmc-compute.html and let me know if there is anything missing?

hoomd/hpmc/compute.py
```diff
@@ -1,129 +1,110 @@
...
+    where :math:`V_\mathrm{free}` is the estimated free volume `free_volume`,
+    :math:`n_\mathrm{samples}` is the number of samples `num_samples`,
+    :math:`n_\mathrm{overlaps}` is the number of overlapping test placements,
+    and :math:`V_\mathrm{box}` is the volume of the simulation box.
+
+    Note:
+
+        The test particle type must exist in the simulation state and its shape
+        parameters must be set in the simulation's HPMC integrator. Particles
+        with this type may be present in the simulation state.
```
b-butler, line 48:
```suggestion
        with this type may or may not be present in the simulation state.
```

b-butler (commented):
@danevans09 I have one change to documentation that I would like and then I will approve to merge.

hoomd/hpmc/compute.py
```diff
@@ -1,129 +1,110 @@
...
+    def __init__(self, test_particle_type, num_samples):
+        # store metadata
+        param_dict = ParameterDict(
+            test_particle_type=str,
+            num_samples=int
+        )
+        param_dict.update(
+            dict(test_particle_type=test_particle_type,
+                 num_samples=num_samples))
+        # set defaults
```
b-butler, line 76: This isn't really setting defaults, but the user provided values.
```suggestion
```

b-butler (changes requested):
LGTM, I made one suggestion regarding an inline comment, that I will push and then approve.

b-butler (approved):

Issue #922, opened 2021-02-08
Test for logging names


joaander:
## Description

Each operation should have tests that 
1. Add it to a `Logger`
2. Check that the namespace is generated properly

## Motivation and context

On https://github.com/glotzerlab/hoomd-examples/pull/19 I found that the LJ pair energy quantity changed names. Unit tests in hoomd would detect regressions such as these.

PR #924, opened 2021-02-09
Fix Logging Namespace

b-butler:
## Description

Refactor the logging namespace generation as well as introduce a new mechanism for detecting consecutive duplicates in the namespace such as `('md', 'pair', 'pair', 'LJ')` and remove them. This also adds a check in `hoomd/conftest.py` that can be used to check that the logging for a particular class is set up correctly. This check is then used in `hoomd/md/pytest/test_potential.py` to show how to use it with `pytest`.

## Motivation and context

Fixes a regression in the logging namespace and partially addresses #922 by providing an infrastructure for writing quick tests and showing an example.

## How has this been tested?

The refactor of namespace generation does not fail any extant tests, and a new test was added to specifically test the namespace and loggables of two body and three body potentials.

## Change log

Internal changes only.

```

```

## Checklist:

### DIFF
--- a/hoomd/conftest.py
+++ b/hoomd/conftest.py
@@ -244,3 +244,35 @@ def pytest_sessionfinish(session, exitstatus):
 
     if exitstatus != 0 and hoomd.version.mpi_enabled:
         atexit.register(abort, exitstatus)
+
+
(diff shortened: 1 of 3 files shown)


joaander (approved):
Thanks!

PR #929, opened 2021-02-15
Feature/nlist v3

### LINKED ISSUE #861: Port tree and stencil neighbor lists to v3.0 API
## Description

Port the tree and stencil neighbor lists to v3 and include unit tests.

tommy-waltmann:
## Description

This PR covers converting all the neighbor lists from v2 to v3.

## Motivation and context

Resolves #861 

## How has this been tested?

Tests for the neighbor lists are in `md/pytest/test_nlist.py`. They are very simplistic, they mostly just involve getting and setting parameters. More in-depth tests of the neighbor finding are done in c++. We may also want to add tests ensuring neighbor lists and pair potentials are communicating the rcut matrix correctly.

## Change log

```
update Tree and Stencil neighbor lists to v3 API
```

## Checklist:

- [ x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.


hoomd/md/nlist.py
```diff
@@ -190,235 +176,138 @@ def _detach(self):
...
-            self.cpp_nlist = _md.NeighborListStencil(hoomd.context.current.system_definition, 0.0, r_buff, self.cpp_cl, cls)
-        else:
-            self.cpp_cl  = _hoomd.CellListGPU(hoomd.context.current.system_definition)
-            hoomd.context.current.system.addCompute(self.cpp_cl , self.name + "_cl")
-            cls = _hoomd.CellListStencil(hoomd.context.current.system_definition, self.cpp_cl)
-            hoomd.context.current.system.addCompute(cls, self.name + "_cls")
-            self.cpp_nlist = _md.NeighborListGPUStencil(hoomd.context.current.system_definition, 0.0, r_buff, self.cpp_cl, cls)
-
-        self.cpp_nlist.setEvery(check_period, dist_check)
+    def __init__(self, cell_width, buffer=0.4, check_dist=True, deterministic=False,
```
joaander, line 231: We should use the same order of arguments for the common parameters between the different neighbor list classes. Put the parameters that are different at the end of the list. The documentation should be updated to match the order in the signature.

joaander (changes requested):
Thanks! This is a great start.

It looks like we need to fix the documentation build failures and some failing unit tests. @b-butler, would you review this as well?

b-butler (commented):
The code looks good, besides the failing tests, although I am not sure what is causing them.

hoomd/md/nlist.py
```diff
@@ -155,7 +140,7 @@ class Cell(NList):
     Examples::
 
         cell = nlist.Cell()
-        lj = md.pair.LJ(nlist=cell)
+        lj = md.pair.LJ(nlist=cell, r_cut=1.1)
```
joaander, line 143:
```suggestion
```
Examples for pair potentials are in their corresponding docs.

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
+        buffer (float): Buffer width.
+        exclusions (tuple[str]): Excludes pairs from the neighbor list, which
+            excludes them from the pair potential calculation.
+        rebuild_check_delay (int): How often to attempt to rebuild the neighbor
+            list.
+        diameter_shift (bool): Flag to enable / disable diameter shifting.
+        check_dist (bool): Flag to enable / disable distance checking.
+        max_diameter (float): The maximum diameter a particle will achieve.
+        deterministic (bool): When `True`, sort neighbors to help provide
+            deterministic simulation runs.
```
joaander, line 192:
```suggestion
```
I don't see `deterministic` in the C++ code for the Stencil neighbor list. I don't think it supports this option.

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
+
+    `Stencil` creates a cell list based neighbor list object to which pair
+    potentials can be attached for computing non-bonded pairwise interactions.
+    Cell listing allows for O(N) construction of the neighbor list. Particles
+    are first spatially sorted into cells based on the largest pairwise cutoff
```
joaander, line 197:
```suggestion
    are first spatially sorted into cells with the given width `cell_width`.
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
+    radius attached to this instance of the neighbor list.
```
joaander, line 198:
```suggestion
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
         super()._detach()
 
 
-class stencil(nlist):
-    R""" Cell list based neighbor list using stencils
+class Stencil(NList):
+    R""" Cell list based neighbor list using stencils.
```
joaander, line 179:
```suggestion
    """Cell list based neighbor list using stencils.
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
     Examples::
 
-        nl_s = nlist.stencil(check_period = 1)
-        nl_s.set_params(r_buff=0.5)
-        nl_s.reset_exclusions([]);
-        nl_s.tune()
-        nl_s.tune_cell_width(min_width=1.5, max_width=3.0)
+        nl_s = nlist.Stencil(cell_width=1.5)
+        nl_s.cell_width = 0.75
+        pair = hoomd.md.pair.LJ(nl_s)
```
joaander, line 223:
```suggestion
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
 
-    Note:
-        *d_max* should only be set when slj diameter shifting is required by a pair potential. Currently, slj
-        is the only pair potential requiring this shifting, and setting *d_max* for other potentials may lead to
-        significantly degraded performance or incorrect results.
+    Attributes:
+        cell_width (float): The underlying stencil bin width for the cell list.
+        deterministic (bool): When `True`, sort neighbors to help provide
+            deterministic simulation runs.
```
joaander, line 228:
```suggestion
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
-            self.cpp_cl  = _hoomd.CellListGPU(hoomd.context.current.system_definition)
-            hoomd.context.current.system.addCompute(self.cpp_cl , self.name + "_cl")
-            cls = _hoomd.CellListStencil(hoomd.context.current.system_definition, self.cpp_cl)
-            hoomd.context.current.system.addCompute(cls, self.name + "_cls")
-            self.cpp_nlist = _md.NeighborListGPUStencil(hoomd.context.current.system_definition, 0.0, r_buff, self.cpp_cl, cls)
-
-        self.cpp_nlist.setEvery(check_period, dist_check)
+    def __init__(self, cell_width, buffer=0.4, exclusions=('bond',),
+                 rebuild_check_delay=1, diameter_shift=False, check_dist=True,
+                 max_diameter=1.0, deterministic=False):
```
joaander, line 232:
```suggestion
                 max_diameter=1.0):
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
-        """
-        if cell_width is not None:
-            self.cpp_nlist.setCellWidth(float(cell_width))
+        super().__init__(buffer, exclusions, rebuild_check_delay,
+                         diameter_shift, check_dist, max_diameter)
 
-    def tune_cell_width(self, warmup=200000, min_width=None, max_width=None, jumps=20, steps=5000):
-        R""" Make a series of short runs to determine the fastest performing bin width.
+        params = ParameterDict(deterministic=bool(deterministic),
+                               cell_width=float(cell_width))
```
joaander, line 238:
```suggestion
        params = ParameterDict(cell_width=float(cell_width))
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
-            steps (int): Number of time steps to run() at each point
+        self._param_dict.update(params)
 
-        :py:class:`tune_cell_width()` executes *warmup* time steps. Then it sets the nlist *cell_width* value to *min_width* and
-        runs for *steps* time steps. The TPS value is recorded, and the benchmark moves on to the next *cell_width*
-        value completing at *max_width* in *jumps* jumps. Status information is printed out to the screen, and the
-        optimal *cell_width* value is left set for further runs() to continue at optimal settings.
+    def _attach(self):
+        if isinstance(self._simulation.device, hoomd.device.CPU):
+            cl_cls = _hoomd.CellList
```
joaander, line 244:
```suggestion
```
Not needed. The C++ code creates this internally.

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
+            nlist_cls = _md.NeighborListStencil
+        else:
+            cl_cls = _hoomd.CellListGPU
```
joaander, line 247:
```suggestion
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
+            nlist_cls = _md.NeighborListGPUStencil
+        self._cpp_cell = cl_cls(self._simulation.state._cpp_sys_def)
+        self._cpp_stencil = _hoomd.CellListStencil(self._simulation.state._cpp_sys_def,
+                                                   self._cpp_cell)
```
joaander, line 251:
```suggestion
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
+        # TODO remove 0.0 r_cut from constructor
+        self._cpp_obj = nlist_cls(self._simulation.state._cpp_sys_def, 0.0,
+                                  self.buffer, self._cpp_cell, self._cpp_stencil)
```
joaander, line 254:
```suggestion
                                  self.buffer)
```

hoomd/md/NeighborListStencil.cc
```diff
@@ -325,5 +325,6 @@ void export_NeighborListStencil(py::module& m)
     {
     py::class_<NeighborListStencil, NeighborList, std::shared_ptr<NeighborListStencil> >(m, "NeighborListStencil")
         .def(py::init< std::shared_ptr<SystemDefinition>, Scalar, Scalar, std::shared_ptr<CellList>, std::shared_ptr<CellListStencil> >())
```
joaander, line 327:
```suggestion
        .def(py::init< std::shared_ptr<SystemDefinition>, Scalar)
```
No need to pass these in from Python. The C++ code manages them internally.

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
+            list.
+        diameter_shift (bool): Flag to enable / disable diameter shifting.
+        check_dist (bool): Flag to enable / disable distance checking.
+        max_diameter (float): The maximum diameter a particle will achieve.
 
-        nl_t = nlist.tree(check_period = 1)
-        nl_t.set_params(r_buff=0.5)
-        nl_t.reset_exclusions([]);
-        nl_t.tune()
+    `Tree` creates a neighbor list using bounding volume hierarchy (BVH) tree
```
joaander, line 276:
```suggestion
    `Tree` creates a neighbor list using a bounding volume hierarchy (BVH) tree
```

hoomd/md/nlist.py
```diff
@@ -190,235 +175,138 @@ def _detach(self):
...
+    if you utilize this neighbor list style in your work.
 
-    Note:
-        *d_max* should only be set when slj diameter shifting is required by a pair potential. Currently, slj
-        is the only pair potential requiring this shifting, and setting *d_max* for other potentials may lead to
-        significantly degraded performance or incorrect results.
+    Examples::
 
+        nl_t = nlist.Tree(check_dist=False)
+        pair = hoomd.md.pair.LJ(nl_t)
```
joaander, line 296:
```suggestion
```

joaander (changes requested):
Thanks! I have some suggested improvements to the documentation and some to simplify the C++/Python interface.

tommy-waltmann:
The pattern of "Neighborlist class that creates its own CellList if nothing is passed through to the constructor" is repeated in the Cell neighborlist as well. I imagine we would want to make the same changes as the ones you are recommending to make the in Stencil nieghborlist in the Cell neighborlist.

joaander:
> The pattern of "Neighborlist class that creates its own CellList if nothing is passed through to the constructor" is repeated in the Cell neighborlist as well. I imagine we would want to make the same changes as the ones you are recommending to make the in Stencil nieghborlist in the Cell neighborlist.

Yes.

This pattern is an artifact of how Computes were handled in v2. There, all Computes were instantiated in Python and handed to the System object through Python. System handled the setCommunicator call which is why we now keep finding cases where it is not being called on these Computes that are internal to other classes.

hoomd/md/nlist.py
tommy-waltmann: After taking a closer look, I think this is something I forgot to write get/setDeterministic methods for and export in the NeighborListStencil.

tommy-waltmann:
> 
> 
> After taking a closer look, I think this is something I forgot to write get/setDeterministic methods for and export in the NeighborListStencil.

Also, is the fact that this doesn't create errors in the unit tests a consequence of how pybind handles attempting to access exported properties that don't exist, or a consequence of how the synced data structures are written?

Ideally, accessing a property that isn't exported in the cpp class should create an error to alert people to this sort of thing.

joaander:
> > After taking a closer look, I think this is something I forgot to write get/setDeterministic methods for and export in the NeighborListStencil.
> 
> Also, is the fact that this doesn't create errors in the unit tests a consequence of how pybind handles attempting to access exported properties that don't exist, or a consequence of how the synced data structures are written?
> 
> Ideally, accessing a property that isn't exported in the cpp class should create an error to alert people to this sort of thing.

Ideally, it should. @b-butler can comment on why the param_dict infrastructure works the way it does. It is convenient, but does make catching and debugging errors like this hard.

hoomd/md/nlist.py
mphoward: > After taking a closer look, I think this is something I forgot to write get/setDeterministic methods for and export in the NeighborListStencil.

I may not be remembering right, but I feel like I did not implement this option for the stencil neighbor list. It would be easy to add it if you wanted to copy the getter / setter from `NeighborListGPUBinned`. I think I just didn't see much of a potential use case at the time. (`deterministic=True` triggers an extra sort of the cell list... I found `NeighborListGPUTree` worked better for most cases where you would use `NeighborListGPUStencil`, and the tree is inherently deterministic too.)

tommy-waltmann: On the v2.9 branch, you can look and see that there is a deterministic argument passed to the stencil constructor which is passed through the setSortCellList method of the cell list class.

mphoward: Ah, I see, I must have done this at a purely python level. I only skimmed the C++ code, which does not handle this option. If you want to keep it but remove the external cell list, you would need to add the C++ methods to forward that flag to the cell list.

vyasr:
I'm removing myself from the reviewer list here since there are multiple other active reviewers. Let me know if my review will be particularly useful here.

hoomd/md/nlist.py
```diff
@@ -174,251 +158,139 @@ def __init__(self, buffer=0.4, exclusions=('bond',), rebuild_check_delay=1,
...
+            nlist_cls = _md.NeighborListGPUStencil
+        # TODO remove 0.0 r_cut from constructor
+        self._cpp_obj = nlist_cls(self._simulation.state._cpp_sys_def, 0.0,
+                                  self.buffer)
+        super()._attach()
 
-        Each benchmark is repeated 3 times and the median value chosen. In total, ``(warmup + 3*jump*steps)`` time steps
-        are run.
+    def _detach(self):
+        super()._detach()
```
tommy-waltmann, line 245: remove this

hoomd/md/nlist.py
```diff
@@ -174,251 +158,139 @@ def __init__(self, buffer=0.4, exclusions=('bond',), rebuild_check_delay=1,
...
-        value completing at *max_width* in *jumps* jumps. Status information is printed out to the screen, and the
-        optimal *cell_width* value is left set for further runs() to continue at optimal settings.
+    def _attach(self):
+        if isinstance(self._simulation.device, hoomd.device.CPU):
+            nlist_cls = _md.NeighborListStencil
+        else:
+            nlist_cls = _md.NeighborListGPUStencil
+        # TODO remove 0.0 r_cut from constructor
+        self._cpp_obj = nlist_cls(self._simulation.state._cpp_sys_def, 0.0,
+                                  self.buffer)
```
tommy-waltmann, line 241: I can add and export a cpp constructor which would only take 2 arguments so we don't have to always pass this 0 through here. Does this sound like a good idea to others?

hoomd/md/nlist.py
```diff
@@ -174,251 +158,139 @@ def __init__(self, buffer=0.4, exclusions=('bond',), rebuild_check_delay=1,
...
-        self._cpp_cell = cell_cls(self._simulation.state._cpp_sys_def)
         # TODO remove 0.0 (r_cut) from constructor
         self._cpp_obj = nlist_cls(self._simulation.state._cpp_sys_def, 0.0,
-                                  self.buffer, self._cpp_cell)
+                                  self.buffer)
         super()._attach()
 
     def _detach(self):
-        del self._cpp_cell
         super()._detach()
```
tommy-waltmann, line 170: remove this

hoomd/md/nlist.py
mphoward: This would be a bit of a weird pattern because you would be eliminating the middle variable from this full constructor. I prefer explicitly setting r_cut to zero.

tommy-waltmann: In v3 API the middle variable (`r_cut`) is more or less a dummy because of how the neighborlist consumer communicates its rcut matrix with the neighborlist. Because of this, there aren't many cases (only cases I'm aware of are cpp unit tests) where this argument is useful on construction of the neighborlists. Because of the above, I don't think this is a weird pattern, its just more in line with how neighborlists are going to be used going forward.

mphoward: I'm pretty sure that variable has been used as a dummy for a while, it is vestigial from when HOOMD did not support per-type cutoffs in the neighbor list. If you don't want to type the zero explicitly, I would propose removing it (and having the class default initialize to 0.0 internally) rather than having two constructors:
```c++
NeighborList(system, r_buff)
```
I find the pattern strange compared to python and most C++ classes where you would want to overload a method with incrementally longer call signatures, but you wouldn't usually want to reorder the variables. This is particularly important here since `r_cut` and `r_buff` have the same type. For example, I would expect:
```c++
NeighborList(system, r_buff);
NeighborList(system, r_buff, r_cut);
```
or even better
```c++
NeighborList(system, r_buff, r_cut=0.0);
```
if you wanted to support both. As you say, there isn't much of a use case for setting `r_cut` though, so the simplest solution would be to remove it.

tommy-waltmann: I see what you meant by "weird pattern" originally now. I agree with your last suggestion to reorder the arguments and make the `r_cut` argument default to 0. Removing it would require a lot of work to update the cpp tests, as there are many of them that make use of the single `r_cut` and don't have consumers that communicate matrices with the neighborlists.

joaander: But, if you reorder the arguments, won't you have to update the unit tests?

I think it would be find to remove rcut entirely. You can rewrite the unit tests with a global regex search and replace.

hoomd/md/NeighborListGPUBinned.cc
```diff
@@ -18,15 +18,9 @@ namespace py = pybind11;
 #endif
 
 NeighborListGPUBinned::NeighborListGPUBinned(std::shared_ptr<SystemDefinition> sysdef,
-                                             Scalar r_cut,
-                                             Scalar r_buff,
-                                             std::shared_ptr<CellList> cl)
-    : NeighborListGPU(sysdef, r_cut, r_buff), m_cl(cl), m_param(0)
+                                             Scalar r_buff)
+    : NeighborListGPU(sysdef, r_buff), m_cl(std::make_shared<CellList>(sysdef)), m_param(0)
```
mphoward, line 22: This should be on the GPU, yes?
```suggestion
    : NeighborListGPU(sysdef, r_buff), m_cl(std::make_shared<CellListGPU>(sysdef)), m_param(0)
```

hoomd/md/NeighborListGPUStencil.cc
```diff
@@ -27,23 +27,14 @@ namespace py = pybind11;
...
 NeighborListGPUStencil::NeighborListGPUStencil(std::shared_ptr<SystemDefinition> sysdef,
-                                               Scalar r_cut,
-                                               Scalar r_buff,
-                                               std::shared_ptr<CellList> cl,
-                                               std::shared_ptr<CellListStencil> cls)
-    : NeighborListGPU(sysdef, r_cut, r_buff), m_cl(cl), m_cls(cls), m_override_cell_width(false),
-      m_needs_restencil(true), m_needs_resort(true)
+                                               Scalar r_buff)
+    : NeighborListGPU(sysdef, r_buff),
+    m_cl(std::make_shared<CellList>(sysdef)),
```
mphoward, line 32:
```suggestion
    m_cl(std::make_shared<CellListGPU>(sysdef)),
```

hoomd/md/NeighborListStencil.cc
```diff
@@ -26,31 +26,19 @@ namespace py = pybind11;
...
-                                         Scalar r_buff,
-                                         std::shared_ptr<CellList> cl,
-                                         std::shared_ptr<CellListStencil> cls)
-    : NeighborList(sysdef, r_cut, r_buff), m_cl(cl), m_cls(cls), m_override_cell_width(false),
-      m_needs_restencil(true)
+                                         Scalar r_buff)
+    : NeighborList(sysdef, r_buff),
+    m_cl(std::make_shared<CellList>(sysdef)),
+    m_cls(std::make_shared<CellListStencil>(sysdef, m_cl)),
+    m_override_cell_width(false)
```
mphoward, line 33: This default is set in the header already, I think
```suggestion
```

hoomd/md/nlist.py
```diff
@@ -174,251 +158,127 @@ def __init__(self, buffer=0.4, exclusions=('bond',), rebuild_check_delay=1,
...
-    disparity in the cutoff radius. The memory demands of :py:class:`stencil` can also be lower than :py:class:`Cell` if your
-    system is large and has many small cells in it; however, :py:class:`tree` is usually a better choice for these systems.
-
-    The performance of the stencil depends strongly on the choice of *cell_width*. The best performance is obtained
-    when the cutoff radii are multiples of the *cell_width*, and when the *cell_width* covers the simulation box with
-    a roughly integer number of cells. The *cell_width* can be set manually, or be automatically scanning through a range
-    of possible bin widths using :py:meth:`tune_cell_width()`.
+        cell_width (float): The underlying stencil bin width for the cell list.
+        buffer (float): Buffer width.
+        exclusions (tuple[str]): Excludes pairs from the neighbor list, which
```
mphoward, line 174: I think this is copied from above, but if `exclusions` has the same meaning as it used to, I find "excludes pairs" to be confusing. I think you mean what styles of exclusions to use, or something like that.

mphoward (changes requested):
Looks overall good to me! Here are a few things.

joaander (approved):
Thanks!

mphoward (approved):

PR #935, opened 2021-02-19
Fixed create_state_from_snapshot and added a test

Charlottez112:
## Description

I wasn't able to create a state using ```create_state_from_snapshot``` with a ```gsd.hoomd.Snapshot()```, and found that there's a bug in the ```simulation.py```

## How has this been tested?

Added a test in ```test_simulation.py```

## Checklist:

### DIFF
--- a/hoomd/pytest/test_simulation.py
+++ b/hoomd/pytest/test_simulation.py
@@ -198,32 +198,25 @@ def test_state_from_gsd(simulation_factory, get_snapshot,
         assert_equivalent_snapshots(snap, sim.state.snapshot)
 
 @skip_gsd
-def test_state_from_snapshot(simulation_factory, get_snapshot,
-                        device, state_args, tmp_path):
+def test_state_from_gsd_snapshot(simulation_factory, get_snapshot,
+                                 device, state_args, tmp_path):
     snap_params, nsteps = state_args
 
-    d = tmp_path / "sub"
-    d.mkdir()
-    filename = d / "temporary_test_file.gsd"
-    with gsd.hoomd.open(name=filename, mode='wb+') as file:
-        sim = simulation_factory(get_snapshot(n=snap_params[0],
-                                              particle_types=snap_params[1]))
-        snap = sim.state.snapshot
+    sim = simulation_factory(get_snapshot(n=snap_params[0],
+                                          particle_types=snap_params[1]))
+    snap = sim.state.snapshot
+    snap = make_gsd_snapshot(snap)
+    gsd_snapshot_list = [snap]
+    box = sim.state.box
+    for _ in range(1, nsteps):
+        particle_type = np.random.choice(snap_params[1])
+        snap = update_positions(sim.state.snapshot)
+        set_types(snap, random_inds(snap_params[0]),
+                  snap_params[1], particle_type)
         snap = make_gsd_snapshot(snap)
-        snapshot_dict = {}
-        snapshot_dict[0] = snap
-        file.append(snap)
-        box = sim.state.box
-        for step in range(1, nsteps):
-            particle_type = np.random.choice(snap_params[1])
-            snap = update_positions(sim.state.snapshot)
-            set_types(snap, random_inds(snap_params[0]),
-                      snap_params[1], particle_type)
-            snap = make_gsd_snapshot(snap)
-            file.append(snap)
-            snapshot_dict[step] = snap
+        gsd_snapshot_list.append(snap)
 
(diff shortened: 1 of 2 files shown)


bdice (dismissed):
Good fix, @Charlottez112. I think the test looks correct but might be shorter. I'll let other reviewers weigh in if it needs shortened.

hoomd/pytest/test_simulation.py
```diff
@@ -197,6 +197,38 @@ def test_state_from_gsd(simulation_factory, get_snapshot,
...
 
+@skip_gsd
+def test_state_from_snapshot(simulation_factory, get_snapshot,
+                        device, state_args, tmp_path):
+    snap_params, nsteps = state_args
+
+    d = tmp_path / "sub"
+    d.mkdir()
+    filename = d / "temporary_test_file.gsd"
+    with gsd.hoomd.open(name=filename, mode='wb+') as file:
```
b-butler, line 208: In your tests there is no reason to write to a file since you only use the stored `gsd.hoomd.Snapshot` objects.

hoomd/pytest/test_simulation.py
```diff
@@ -197,6 +197,38 @@ def test_state_from_gsd(simulation_factory, get_snapshot,
         assert box == sim.state.box
         assert_equivalent_snapshots(snap, sim.state.snapshot)
 
+@skip_gsd
+def test_state_from_snapshot(simulation_factory, get_snapshot,
```
b-butler, line 201: This should be named so that it knows it only tests the use of gsd snapshots.

b-butler, line 201: It would also be nice, though not a must if this also tested that it works when the snapshot data is only passed on the root rank for MPI simulations.

hoomd/pytest/test_simulation.py
```diff
@@ -197,6 +197,38 @@ def test_state_from_gsd(simulation_factory, get_snapshot,
...
+
+    d = tmp_path / "sub"
+    d.mkdir()
+    filename = d / "temporary_test_file.gsd"
+    with gsd.hoomd.open(name=filename, mode='wb+') as file:
+        sim = simulation_factory(get_snapshot(n=snap_params[0],
+                                              particle_types=snap_params[1]))
+        snap = sim.state.snapshot
+        snap = make_gsd_snapshot(snap)
+        snapshot_dict = {}
```
b-butler, line 213: You don't use the step for any purpose other than a `dict` key, we could just store this as a list.

b-butler (changes requested):
Thanks for the PR, and bug fix. I have some comments on how to improve the test you added, then we can get this merged.

hoomd/pytest/test_simulation.py
Charlottez112: Based on our discussion today, I tried to add a test to check that only the root rank gets the data from the snapshot, but I wasn't sure how to do this exactly. I'm thinking something like using an mpi4py communicator but I don't know at which step exactly only the root rank gets the data and all the other ranks don't, so that we can do the check at this step. And I'm also having some trouble rebuilding with MPI enabled. I think it would be better if we merge this PR first. I can add the test for MPI simulations after I fix my hoomd build

joaander: @b-butler hasn't this already done? `make_gsd_snapshot` copies the HOOMD snapshot object which does not have data on ranks >= 1. The GSD snapshots on those ranks will be invalid with `snap.particles.position` and other attributes set to `None`.

hoomd/pytest/test_simulation.py
```diff
@@ -197,6 +197,31 @@ def test_state_from_gsd(simulation_factory, get_snapshot,
...
+    snap = make_gsd_snapshot(snap)
+    gsd_snapshot_list = [snap]
+    box = sim.state.box
+    for _ in range(1, nsteps):
+        particle_type = np.random.choice(snap_params[1])
+        snap = update_positions(sim.state.snapshot)
+        set_types(snap, random_inds(snap_params[0]),
+                  snap_params[1], particle_type)
+        snap = make_gsd_snapshot(snap)
+        gsd_snapshot_list.append(snap)
```
joaander, line 217: I agree that this test could be shorter. However, it is written in the same style as the other tests in this file so I think it is OK for now. Any future clean up in this file can clean up all the tests in a similar way.

joaander (approved):
Thanks!

b-butler (approved):
This is fine for now, better to get the bug fix out there, but some future refactoring for some increased brevity might be good.

PR #937, opened 2021-02-23
Port the ji.patch module to v3 API

lyrivera:
## Description
This PR converts the `jit.patch` submodule to the new v3.0 API. It took more work than I had anticipated, so I will defer the conversion of the  `jit.external` submodule to another PR.

## Motivation and context

Resolves ##875 (partially)

## How has this been tested?

Added unit tests.

## Checklist:


lyrivera:
I realized there's some work left to do, so I converted back to draft.

CHANGELOG.rst
```diff
@@ -4,6 +4,14 @@ Change Log
 v3.x
 ----
 
+v3.0.0-beta.5 (Not released yet)
```
joaander, line 7: Can you revert the changes to the change log? I add these changes only on the `master` branch to avoid merge conflicts. You can draft the change log entry in a PR comment.

hoomd/hpmc/IntegratorHPMC.h
```diff
@@ -205,6 +205,52 @@ class PatchEnergy
...
+        //! Return a flag to trigger the OBB update
+        virtual bool getBuildOBB()
+            {
+            return m_build_obb;
+            }
+
+    protected:
+        bool m_attached = true;   //! Track whether we have attached to the Simulation object
+        bool m_log_only = false;  //! If true, only use patch energy for logging
+        bool m_build_obb = false; //! Flag to update the OBB tree for union of particles
```
joaander, line 253: I don't understand why a HPMC integrator needs these flags? What do they control?

`m_attached` does not seem to be used.

The logging system in v3 has been completely rewritten. Objects shouldn't need a notion of a "logging only" state. You don't have to solve in this PR as we haven't solved it elsewhere, but it will eventually be possible to attach force/energy objects to a Simulation in a way that they are not attached to the integrator and thus don't get evaluated when evolving the system forward in time.

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -333,17 +333,24 @@ class IntegratorHPMCMono : public IntegratorHPMC
...
-                        m_hasOrientation = true;
+                    m_hasOrientation = true;
+                    break;
                     }
                 }
+
+            //! Update the OBB tree if necessary
+            if (m_patch && m_patch->getBuildOBB())
+                {
+                m_patch->buildOBBTree();
```
joaander, line 351: I'm not familiar enough with the patch code to know what this does, but it seems odd to me that a user parameter can influence whether or not the tree is built. I would think that the tree build always needs to be called and the callee can decide whether the call does anything or not.

joaander (commented):
I left a few initial comments. I haven't had chance to read through the whole PR yet.

CHANGELOG.rst
lyrivera: Done.

hoomd/hpmc/IntegratorHPMCMono.h
lyrivera: I believe that is how it was done in the original implementation where the tree was built inside the `setParams` method, so every `patch.set_params(...)` call by the user would trigger the tree build. I can certainly move the build flag check into the `buildOBBTree`method and call it every time within the integrator as is done with the AABB tree.

hoomd/hpmc/IntegratorHPMC.h
lyrivera: Indeed, I tried to implement some hacky attachment protocol but I quickly rescinded. These were leftovers from that.

tcmoore3:
This branch currently (as of commit `b5b0999`) errors out if you try to `import hoomd.hpmc` but built with `-DBUILD_JIT=off`, since `hpmc.pair` only gets installed if you build jit.

joaander (changes requested):
Thanks!

Tim and I have been working on this branch. We've got it running on the GPU and on Summit. I also merged master up into the branch (including the pre-commit styling).

In making my review, I cleaned up a number of issues directly and am leaving comments with tasks for @lyrivera and @b-butler. I'm still working on a few more changes to the CMake configuration and then I'm going to set up GitHub Actions to build and test the JIT builds.

joaander:
FYI, I also removed the source directory as an include directory. This means you need to run the install step in your build tool to get a working module, as v3 no longer copies the headers into the build directory. Including the source directory in the search path is not something we can do safely - if you install hoomd, then switch branches in the source, you could end up with an inconsistent environment.

If there is a strong desire for it, we could consider adding it back in - but conditionally to be either the install or source directory. I'm not aware of a good way to detect that condition right now.

joaander:
This is now building with GPU JIT enabled. Once we fix the pair import issues with ENABLE_LLVM=off and the unit tests, CI will test these configurations to ensure they work.

After we get the unit tests working, we should add a test modeled on the MD LJ validation test to check for thermodynamic correctness.

hoomd/hpmc/external/user.py
```diff
@@ -0,0 +1,142 @@
...
+        integrate.ConvexSpheropolyhedron: _jit.ExternalFieldJITSpheropolyhedron,
+        integrate.Ellipsoid: _jit.ExternalFieldJITEllipsoid,
+        integrate.ConvexSpheropolygon: _jit.ExternalFieldJITSpheropolygon,
+        integrate.FacetedEllipsoid: _jit.ExternalFieldJITFacetedEllipsoid,
+        integrate.Polyhedron: _jit.ExternalFieldJITPolyhedron,
+        integrate.Sphinx: _jit.ExternalFieldJITSphinx
+    }
+
+    def __init__(self, code, clang_exec='clang'):
+        self._llvm_ir = _compile.to_llvm_ir(self._wrap_cpu_(code))
```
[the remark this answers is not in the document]
joaander: Actually, the situation here is worse than I thought. In MPI simulations, the `fork()` to run `clang` causes problems with MPI and/or CUDA. I haven't narrowed down exactly what conditions are needed for a minimal reproducer but the behavior we get is this with 2-rank GPU simulations:
- On `brett`: forking, even to just run `ls` corrupts the particle data memory leading to random errors in `MPI_ISend`, errors that particles are missing when gathering in `getSnapshot`, or `cudaErrorIllegalAddress` errors.
- On `summit`: the fork hangs on both ranks, even when running just `ls`.

We are currently working around this by pre-compiling and specifying the LLVM IR when running. I'm going to push a commit that enables this for now so we have something that works.

Longer term, we should implement a better solution. Options include:
1) Identify the root cause of the failures and work around them somehow to enable forking.
2) Use clang as a library to avoid forking altogether.
3) Accept the llvm IR specification as a solution going forward.
4) Any other ideas?

tcmoore3:
TODO: 
- [ ] choose which header path to include for JIT compilation of user potentials.

joaander:
This PR is no longer a draft.

joaander:
Labelling release to start continue testing older llvm and cuda versions as we finish this PR.

joaander (changes requested):
Great work all, this is coming together very nicely!

I reviewed the C++ code and left a number of comments and suggested changes. I will review the Python code and documentation in another round after my C++ comments have been addressed.

Please add unit tests for the union potential.

hoomd/hpmc/IntegratorHPMC.h
```diff
@@ -120,23 +122,38 @@ struct hpmc_patch_args_t
     \ingroup hpmc_integrators
```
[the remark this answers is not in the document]
tcmoore3: I'll need some help writing this block at the appropriate level of abstraction, but will give it a shot and see how it goes.

hoomd/hpmc/external/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+if (BUILD_JIT)
```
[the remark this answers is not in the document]
tcmoore3: Can we resolve this now?

hoomd/hpmc/external/user.py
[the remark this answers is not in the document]
tcmoore3: Resolving now that Josh figured out how to use clang as a library.

hoomd/hpmc/IntegratorHPMC.h
```diff
@@ -187,6 +204,12 @@ class PatchEnergy
         throw std::runtime_error("PatchEnergy (base class) does not support launchKernel");
         }
 #endif
+
+    //! Update the OBB tree  for union of particles
+    virtual void buildOBBTree() { }
```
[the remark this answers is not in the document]
tcmoore3: Removed and now building the OBBTree everywhere where `m_build_obb_tree` was getting set to `true`.

tcmoore3:
@joaander this is ready for another review

tcmoore3:
Here is a comparisons of a system of 4,096 hard-core + LJ attraction particles run on the gpu vs. on 8 cpus with mpi. These are run at a temperature of kT/epsilon = 0.5. Both systems form solid–vapor coexistence. The RDFs are nearly identical, and it appears the potential energies are approaching the same values.

image
The system from the CPU:
image
The system from the GPU:
image

tcmoore3:
Here snapshots of simulations of tetrahedral clusters of LJ particles run at kT/epsilon = 0.5. These simulations are clearly not equilibrated yet, so I am running them longer to make sure we get the same behavior on the cpu and gpu.

The energies and RDFs:
image

A snapshot of the system run on the CPU after 5e4 steps:
image

A snapshot of the system run on the GPU after 5e4 steps:
image

joaander:
@b-butler This is ready for another review.

tcmoore3:
I was a little concerned at how the potential energies in the union test above look so different, so I ran another example at a higher temperature where kinetic arrest does not occur. The results from the GPU and CPU are more similar in this case.
image

System snapshots for reference (colored by the union identity):
GPU:
image
CPU:
image

tcmoore3:
> One more TODO to remove.
> 
> Also, remove the `hpmc/jit` directory exclusions from `setup.cfg` and `.pre-commit-config.yaml`

Done.

lyrivera:
> My thoughts are:
> 
> > 1. In C++, the total patch energy calculation is being done by the integrator via the `computePatchEnergy` method, not by the PatchEnergy* classes. So, should we move the `energy` attribute from the `CPPPotential` objects and add it to the integrator instead? (`mc.energy` , `mc.potential_energy`, `mc.patch_energy`?)
> 
> No, do not move the energy property to the integrator. It logically makes more sense as a property of the patch energy class.
> 
> > 1. The opposite is true for the field energy, which is computed by the field class class itself since it doesn't need to query for pair interactions. I think that if the patch energy attribute is moved to the integrator so should be the field one (`mc.field_energy` instead of `field.energy`). This can be easily achieved by adding a `computeFieldEnergy` to the integrator that simply returns a call to `m_field->computeEnergy`.
> 
> The energy property makes logical sense as a property of the field class.

Sounds good to me. Thanks

Issue #936, opened 2021-02-22
Active force doesn't work when attached during simulation


syjlee:
## Description

The active force works just fine when a simulation starts with active force in the first place.
But, when the active force is attached after the first run, the next simulation is running without active force.
In this case, it seems like Dr(rotational diffusion) doesn't update the orientation quaternion of particles.

## Script

```
cpu = hoomd.device.CPU()
sim = hoomd.Simulation(device=cpu)
sim.create_state_from_gsd(gsd)
cell = hoomd.md.nlist.Cell()
cell.max_diameter = 0.5*2**(1/6)+1
slj = hoomd.md.pair.SLJ(r_cut=3*2**(1/6), nlist=cell)
slj.params[('A', 'A')] = dict(epsilon=1.0, sigma=0.5) 
npt = hoomd.md.methods.NPT(filter=hoomd.filter.All(), kT=T, tau=0.1, tauS=10, S=[0,0,0,0,0,0], couple='none')
integrator = hoomd.md.Integrator(dt=0.005, methods=[npt], forces=[slj])
sim.operations += integrator 
sim.run(10)

active = hoomd.md.force.Active(filter=hoomd.filter.All(), seed=57, rotation_diff=1.0)
active.active_force['A','A'] = (1.0,0,0)
integrator.forces.append(active)
sim.run(10)
```

## Output

Active force doesn't exert any effect on the integrator.

## Expected output

Active force should be properly attached and particle orientation should reflect the active force vector.

Platform:
- Linux
- CPU

Installation method:
- Conda package

Versions:
- Python version: 3.8.5
- **HOOMD-blue** version: 3.0.0-beta.4

b-butler:
I don't know if this is a bug, my guess is what is happening, is that `aniso` for the integrator is being set to `False` since you don't have any anisotropic interactions when first attaching. After adding the active force after the first run call, you still update the translational degrees of freedom, but not rotational.

syjlee:
@b-butler How can I set `aniso` `True` in the first run when there's no active force?

b-butler:
`integrator = md.Integrator(aniso=True)`. However, I do think it is a bug not to mention this in the documentation. We should have a warning about using the `Active` force without `aniso=True`.

syjlee:
[...]

PR #938, opened 2021-02-23
fixes on active force of less than one

syjlee:
## Description
It changes the numerical type of active force from int to float for active force less than one 
and added consideration of zero active force.

## Motivation and context
Active force smaller than 1 is treated 0 due to variable type int.

## How has this been tested?
The change is tested with a simple running script.

## Change log
1. Active force indicator's numerical type is changed from int to float in md/force.py
2. if statement is added to treat the case with zero active force magnitude in md/ActiveForceCompute.cc

## Checklist:

- [o] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [o] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).

### DIFF
--- a/hoomd/md/ActiveForceCompute.cc
+++ b/hoomd/md/ActiveForceCompute.cc
@@ -109,10 +109,20 @@ void ActiveForceCompute::setActiveForce(const std::string& type_name, pybind11::
 
     Scalar f_activeMag = slow::sqrt(f_activeVec.x*f_activeVec.x+f_activeVec.y*f_activeVec.y+f_activeVec.z*f_activeVec.z);
 
-    f_activeVec.x /= f_activeMag;
-    f_activeVec.y /= f_activeMag;
-    f_activeVec.z /= f_activeMag;
-    f_activeVec.w = f_activeMag;
+    if(f_activeMag >0)
+        {
+        f_activeVec.x /= f_activeMag;
+        f_activeVec.y /= f_activeMag;
+        f_activeVec.z /= f_activeMag;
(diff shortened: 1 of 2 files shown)


joaander:
I don't understand. Where is the `int` cast that causes the problem described? Is this related to #936?

syjlee:
Actually, it's separate issue. 
To avoid the issue I had with #936, I attached active force object from the start but set the force magnitude zero. 
activeforcecomputing.cc is correcting/adding the zero active force case.
And the change in force.py `param_dict=TypeParameterDict((1., 0., 0.), len_keys=1)` is preventing the active force less than 1 from being converted to zero.

joaander:
> And the change in force.py `param_dict=TypeParameterDict((1., 0., 0.), len_keys=1)` is preventing the active force less than 1 from being converted to zero.

Ahh, I see. The `TypeParameter` machinery is making the cast. @b-butler is there a clearer way to mark these as tuples of floats and set the defaults?

b-butler:
The dot in the numbers would work. Although that can be easy to miss, I would go for the explicit `(1.0, 0.0, 0.0)`. We aren't consistent on this, but I think this is to be preferred.

syjlee:
Just made the change and committed.

joaander (approved):
Thanks!

syjlee:
@SchoeniPhlippsn Requesting your review on this PR.

PR #940, opened 2021-02-25
A few missing lines of get methods of particle data are added

syjlee:
## Description
Few missing lines in export_LocalParticleData in ParticleData.h caused attribute error when getting those particle data.
Three lines are added for net_force, net_virial, net_torque, each.

## Motivation and context
Logging particle data: net_force and other more through local snapshot didn't work due to few missing lines.

## How has this been tested?
With simple running script:
```
def log_net_force():
    with sim.state.cpu_local_snapshot as snap:
        indices = np.argsort(snap.particles.tag)
        # indexing by array automatically copies data
        return snap.particles.net_force[indices]
```

- [o] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [o] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).


syjlee:
@b-butler I wonder why getNetForce is Scalar4, not Scalar3. I think net_force should generate three-element vector.

joaander:
HOOMD stores the net per-particle energy in the 4th component of net_force. The local data snapshot should return a view of just the first 3 elements. 

I do not know why net_torque is a scalar4. The 4th component is unused.

joaander:
@syjlee You can leave the code-review auto-assigments in place. They help distribute code reviews among a group of developers. If you need to ping a specific person, feel free to add them.

syjlee:
1. If I want to generate only the first three components of net_force in ParticleLocalAccess, is export_LocalParticleData in ParticleData.h only place I need to tell to export only the first three in the array?
2. net_torque is Scalar4 across other files in hoomd. And with shallow screening, all four components seem to be used in integration calculation.

joaander:
> 1. If I want to generate only the first three components of net_force in ParticleLocalAccess, is export_LocalParticleData in ParticleData.h only place I need to tell to export only the first three in the array?

I don't know, @b-butler should be able to answer this.

> 2. net_torque is Scalar4 across other files in hoomd. And with shallow screening, all four components seem to be used in integration calculation.

The integrator ignores the .w component of the `net_torque` vector: https://github.com/glotzerlab/hoomd-blue/blob/2f0966faa8b6ccd260efbd761c67761511b63f7f/hoomd/md/TwoStepNVE.cc#L178

Integrator does add up all the .w components: https://github.com/glotzerlab/hoomd-blue/blob/2f0966faa8b6ccd260efbd761c67761511b63f7f/hoomd/Integrator.cc#L372-L375
But if you look at the forces, they always set the 4th component to 0.

b-butler (approved):
Thanks @syjlee!

b-butler (changes requested):
Actually we need to wait, there are some residual things that need to be cleaned up in the C++ code.

b-butler:
@syjlee you can make the torques only use the first three entries, just set the second dimension size to 3 in `getNetTorque`.

joaander:
This is failing tests on the GPU: https://dev.azure.com/glotzerlab/hoomd-blue/_build/results?buildId=10526&view=logs&j=fdb24454-8954-52a4-1b2d-7de6a8c2d450&t=1ee31097-8742-507c-39b9-1f83218a9f6b&l=1594

syjlee:
From the error details, does the error have something to do with an incompatibility in the number of particles in test_local_snapshot.py?

joaander:
> From the error details, does the error have something to do with an incompatibility in the number of particles in test_local_snapshot.py?

The test also fails on the CPU. It only fails when the test is run with MPI enabled. It appears that this test has 5 total particles - 4 on rank 0 and 1 on rank 1. You can see this in the output of pytest (rank0):
```
E       ValueError: could not broadcast input array from shape (5,3) into shape (4,3)
```
and the rank1 output in pytest.out.1
```
ValueError: could not broadcast input array from shape (5,3) into shape (1,3)
```

@b-butler It isn't clear to me how the updated test is supposed to work. It appears to be initializing from a global snapshot, and then immediately assigning values to local snapshots.
```
            sim.create_state_from_snapshot(base_snapshot)
            with sim.state.cpu_local_snapshot as snap:
                particle_data = getattr(snap, 'particles')
                for attr, inner_dict in _particle_local_data.items():
                    getattr(particle_data, attr)[:] = inner_dict['value']
```
Should these tests be marked for serial only execution? What is the assignment attempting to achieve?

Also, the test seems to replicate code already in conftest.simulation_factory. We could use that fixture here.

Lastly, we could avoid so many skipped tests (namely, the `TestLocalSnapshotGPUDevice` run on the CPU device) if we moved the implementation of`conftest.simulation_factory.make_simulation` into a non-fixture callable function `make_simulation_impl(device_)` which you could explicitly call in `TestLocalSnapshotGPUDevice` with a GPU device. Now that I explain it, it wouldn't avoid any skipped tests on the CPU only builds (we would have to skip these explicit GPU tests) - but it would avoid the skipped tests in GPU builds.

b-butler:
I messed up on the implementation their for MPI. That isn't a test though, but just an initialization of the simulation for all other tests for the local snapshot. I am attempting to use the `simulation_factory` fixture now, which should shorten the test code a decent bit.

I am not sure I understand your comment about not skipping tests, I could make the code such that it checked the device and ran the appropriate tests which should result in lower numbers of skipped tests, but that would require removing away a fixture.

joaander:
The skipped tests is a minor issue. I don't know why, but it bugs me when I see the test output:
```
hoomd/pytest/test_local_snapshot.py .ssssssssssssss........................................................................... [ 11%]
.............................................................................................................................. [ 27%]
.............................................................................................................................. [ 43%]
...............................................sssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 60%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 76%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 92%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssss  
```
or (on the GPU):
```
hoomd/pytest/test_local_snapshot.py .......................................................................................... [  6%]
.............................................................................................................................. [ 15%]
.............................................................................................................................s [ 25%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 34%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 43%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 52%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 62%]
ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss [ 71%]
sssssssssssssssssssssssssssssssssssssssssssssssssss.................ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss [ 80%]
..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss.. [ 89%]
ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss..ss [ 98%]
..ss..ss..ss..  
```
It just seems that there should be a better way to do this. It comes from the dependency of the `device` fixture which is parameterized on GPU and CPU while you have tests that are specific to one device type or the other. Ideally we wouldn't even generate test cases for invalid configurations - which is why I suggested refactoring the `make_simulation` to make a version callable that doesn't depend on the parameterized fixture.

b-butler:
The latest push should fix invalid configurations and perform the same tests.

b-butler (approved):
LGTM, I also implemented most of the changes to the tests, and some to the implementation, so it would be good to get another set of eyes besides mine on this.

PR #943, opened 2021-02-25
Fix Custom Action Docstring

b-butler:
## Description

Fixes two errors in the custom action docstring.

## Motivation and context

Accurate documentation.

## Checklist:

### DIFF
--- a/hoomd/custom/custom_action.py
+++ b/hoomd/custom/custom_action.py
@@ -68,7 +68,7 @@ class Action(metaclass=_AbstractLoggable):
 
         class ExampleAction(Action):
             def act(self, timestep):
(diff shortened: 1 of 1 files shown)


vyasr:
@b-butler is the state only intended to be accessed by subclasses of Action, not by users writing simulations? I'm just wondering because the `_state` attribute is documented in such a way as to make it seem like public API, but it's prefixed with an underscore.

b-butler:
It is an interesting middle ground, it is public in the sense that users creating subclasses need to know about it. It isn't meant to be used by users using the action though.

vyasr:
OK that's fine then I think that the way it's documented seems appropriate. I followed similar logic throughout the synced collections work in signac.

Issue #736, opened 2020-08-24
Update restartable-docs for HOOMD v3


b-butler:
## Description

`sphinx-docs/restartable-jobs.rst` has not been updated to the new API.

joaander:
We should replace this with a Jupyter notebook tutorial lesson in `hoomd-examples`.

PR #944, opened 2021-02-25
Make Operations Picklable

b-butler:
## Description

This PR implements the necessary changes to make `hoomd.Operations` and `hoomd._HOOMDBaseObject` objects picklable. Overall there are many individual changes that needed to be made, but I tried to keep the commits modular, and if necessary I can move this into multiple PRs.

Note due to the limitations of pickling, if a user uses lambdas or certain file type objects in their operations, they will not be picklable. 

## Motivation and context

We want to make operations picklable for restartability and enabling simulation metadata storing.

Resolves #???

## How has this been tested?

Tests for all existing tested operations have been added.

## Change log

```
* Added: `hoomd.md.Force`, `hoomd.md.Operation`, and `hoomd.md.Operations` objects are now picklable.
```

## Checklist:


vyasr:
A feature to consider for way down the road is allowing users to provide custom pickle functions (e.g. using `dill` or `cloudpickle`) if they want to pickle something that the default `pickle` library doesn't support.

joaander:
Some of the tests need to be updated to remove `seed`: https://dev.azure.com/glotzerlab/hoomd-blue/_build/results?buildId=10531&view=logs&j=c21dfe3d-6dd2-5969-8cf5-3e196ca72a91&t=ee49db16-ba03-55b0-4481-6e0b297f3809&l=24

I would recommend merging master up into this branch.

hoomd/data/typeparam.py
```diff
@@ -1,14 +1,17 @@
 from hoomd.data.parameterdicts import AttachedTypeParameterDict
-from copy import deepcopy
 
 
 class TypeParameter:
+    __slots__ = ('name', 'type_kind', 'param_dict')
```
bdice, line 5: I can't recall the details, but there are a lot of potential problems with `__slots__` if class inheritance is involved. I believe that using `__slots__` for filtering class attributes in `__getattr__` wasn't an option in signac's synced collections for that reason. There are some notes here: https://docs.python.org/3/reference/datamodel.html#notes-on-using-slots

hoomd/data/typeparam.py
```diff
@@ -52,14 +55,15 @@ def to_dict(self):
...
-        if self.param_dict._len_keys > 1:
-            state = {str(key): value for key, value in state.items()}
-        state['__default__'] = self.default
+    def __getstate__(self):
+        state = {'name': self.name,
+                 'type_kind': self.type_kind,
+                 'param_dict': self.param_dict
+                 }
+        if isinstance(self.param_dict, AttachedTypeParameterDict):
+            state['param_dict'] = self.param_dict.to_dettached()
```
bdice, line 64: Typo? (Missing test?)
```suggestion
            state['param_dict'] = self.param_dict.to_detached()
```

hoomd/hpmc/pytest/test_quick_compress.py
```diff
@@ -170,3 +171,16 @@ def test_disk_compression(phi, simulation_factory, lattice_snapshot_factory):
     assert qc.complete
     assert mc.overlaps == 0
     assert sim.state.box == target_box
+
+
+def test_pickling(simulation_factory, two_particle_snapshot_factory):
+    """Test that QuickCompress can compress (and expand) simulation boxes."""
```
bdice, line 177: Test comment seems inaccurate.

hoomd/hpmc/pytest/test_clusters.py
```diff
@@ -224,3 +225,19 @@ def test_pivot_moves(delta_mu, simulation_factory,
         assert acc > 0
 
     assert cl.reflection_moves[0]/np.sum(cl.reflection_moves) > 0
+
+
+def test_pickling(simulation_factory, two_particle_snapshot_factory):
+    """Test that Clusters always accept pivot moves."""
```
bdice, line 231: Test comment seems inaccurate.

hoomd/hpmc/tune/move_size.py
```diff
@@ -259,6 +238,30 @@ def _update_tunables_attr(self, attr, value):
         for tunable in self._tunables:
             setattr(tunable, attr, value)
 
+    def _target_postprocess(self, target):
+        if not (0 <= target <= 1):
+            return target
```
bdice, line 243: Is this `return` a mistake?
```suggestion
```

hoomd/md/integrate.py
```diff
@@ -7,16 +7,17 @@
 # Maintainer: joaander / All Developers are free to add commands for new
 # features
 
+import itertools
+from functools import partial
```
bdice, line 11: Is this import used? It's newly added but I don't see where it it utilized.

hoomd/operation.py
```diff
@@ -371,11 +241,14 @@ class _HOOMDBaseObject(_StatefulAttrBase, _DependencyRelation):
     detaching is removing an object from its simulation.
     """
     _reserved_default_attrs = {**_HOOMDGetSetAttrBase._reserved_default_attrs,
-                               '_cpp_obj': lambda: None,
-                               '_dependents': lambda: [],
-                               '_dependencies': lambda: []}
+                               '_cpp_obj': _ReturnCopies(None),
+                               '_dependents': _ReturnCopies([]),
```
bdice, line 245: An alternative would be to define something like:

```python
class _Instantiator:
    def __init__(self, instance_class):
        self.instance_class = instance_class
    def __call__(self):
        return self.instance_class()

_dependents = _Instantiator(list)
```

Or even better yet, just provide a class that will create a new instance when called. 😉 
```
_dependents = list
```

Then you could just call the item to get a value if it's callable and return values like `None` directly.

```python
# Replace whatever calls _ReturnCopies with:
return value() if callable(value) else value
```

hoomd/operations.py
```diff
@@ -301,3 +305,10 @@ def computes(self):
         be modified as a standard Python list.
         """
         return self._computes
+
+    def __getstate__(self):
+        # ensure that top level changes to self.__dict__ are not propegated
```
bdice, line 310:
```suggestion
        # ensure that top level changes to self.__dict__ are not propagated
```

hoomd/pytest/test_table.py
```diff
@@ -154,7 +166,13 @@ def test_only_string_and_scalar_quantities(device):
     logger = hoomd.logging.Logger()
     output = StringIO("")
     with pytest.raises(ValueError):
-        _ = hoomd.write.Table(0, logger, output)
+        _ = hoomd.write.Table(1, logger, output)
     logger = hoomd.logging.Logger(categories=['sequence'])
     with pytest.raises(ValueError):
-        _ = hoomd.write.Table(0, logger, output)
+        _ = hoomd.write.Table(1, logger, output)
```
bdice, line 172: Some linters will complain about this syntax and prefer that the output value is never assigned (because it's not used). Assignment to a variable cannot have side-effects as far as I am aware, so the test behavior of raising an error should be equivalent.
```suggestion
        hoomd.write.Table(1, logger, output)
```

hoomd/simulation.py
```diff
@@ -240,8 +240,10 @@ def operations(self, operations):
                 self._operations._unschedule()
                 self._operations._simulation = None
                 operations._simulation = self
-                operations._schedule()
                 self._operations = operations
+                # This must currently be done after setting _operations so that
+                # self.state.update_group_dof runs without erroring
```
bdice, line 245: Suggestion on the comment text. But more importantly, I don't know if this gives a developer enough info to understand why an error would occur. I haven't looked at `update_group_dof` but if it's not obvious, that would be good to describe.
```suggestion
                # The _schedule method must be called after setting _operations so that
                # self.state.update_group_dof runs without errors.
```

hoomd/util.py
```diff
@@ -37,7 +38,9 @@ def is_iterable(obj):
 
 def bad_iterable_type(obj):
     '''Returns True if str or dict.'''
-    return isinstance(obj, str) or isinstance(obj, dict)
+    return isinstance(
+        obj,
+        (str, dict, io.TextIOBase, io.BufferedIOBase, io.RawIOBase, io.IOBase))
```
bdice, line 43: `io.TextIOBase, io.BufferedIOBase, io.RawIOBase` are already subclasses of `io.IOBase`.
```suggestion
        (str, dict, io.IOBase))
```

bdice (changes requested):
Comments below. I probably won't have time to get back for another round of feedback on the entire PR, so feel free to merge this without waiting for a second round of review from me. Of course you can ping me in the comments if you need further guidance on anything specific that I mentioned. 👍

> 
> 
> A feature to consider for way down the road is allowing users to provide custom pickle functions (e.g. using `dill` or `cloudpickle`) if they want to pickle something that the default `pickle` library doesn't support.

I wondered about this as well -- there are a lot of private "partial" classes that had to be introduced for this behavior to work. If you think that's detrimental to readability or maintenance, I think it is fine to require users to install cloudpickle.

hoomd/operation.py
b-butler: Good thought, the last suggestion is way less over engineered.

hoomd/md/integrate.py
b-butler: This is a remnant of an intermediate commit.

hoomd/hpmc/tune/move_size.py
b-butler: Yes, I do not know how that happened, maybe a botched copy and paste. Good catch.

hoomd/hpmc/pytest/test_quick_compress.py
b-butler: Good catch.

hoomd/data/typeparam.py
b-butler: Typo, a consistent one.

hoomd/data/typeparam.py
b-butler: Yes, for classes that will likely be inherited slots are not a good option. This is not the case here though.

hoomd/data/syncedlist.py
```diff
@@ -1,15 +1,41 @@
-class SyncedList:
+from collections.abc import MutableSequence
+import inspect
+from copy import copy
+
+
+class _PartialIsInstance:
```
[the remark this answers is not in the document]
b-butler: They can but I need to use partial for the second positional only argument which I do not think is doable.

hoomd/data/syncedlist.py
```diff
@@ -19,14 +45,17 @@ class SyncedList:
...
+    def __init__(self, validation,
                  to_synced_list=None,
                  iterable=None):
         if to_synced_list is None:
-            def identity(x):
-                return x
             to_synced_list = identity
-        self._validate = validation_func
+
+        if inspect.isclass(validation):
```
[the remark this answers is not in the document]
b-butler: For minimal changes, I will add an argument to the constructor `callable_class` which when set to `True` defaults to `False` will not wrap the class, instead treating it as a callable.

hoomd/simulation.py
```diff
@@ -240,8 +240,16 @@ def operations(self, operations):
...
                 operations._simulation = self
-                operations._schedule()
                 self._operations = operations
+                # This _schedule method must be called  after setting
+                # _operations so that self.state.update_group_dof runs without
+                # errors. This is called by the integrator when it attaches and
+                # requires that the state have access to the current integrator,
+                # so if we do not add the correct operations object to the
+                # simulation first, the new integrator will call
+                # `state.update_group_dof` but the state will still point to the
```
joaander, line 250: Not for this PR, but we still need to implement hooks to call `update_group_dof` as needed to keep things up to date after attaching.

joaander (commented):
This all appears fine to me. I have some high level questions which we could discuss at the next meeting.

One of my questions is about version compatibility. Say we add a new type parameter to an operation in 3.1. For API compatibility, this parameter would have a default value consistent with the behavior in 3.0. Will 3.1 be able to read a pickled dump of this operation written by 3.0?

b-butler:
@joaander Yes and no. We would need to modify the `__setstate__` function, but otherwise adding to the spec would be fine. We just need to check if the dict in the pickle has the new key.

hoomd/hpmc/tune/move_size.py
```diff
@@ -259,6 +238,29 @@ def _update_tunables_attr(self, attr, value):
         for tunable in self._tunables:
             setattr(tunable, attr, value)
 
+    def _target_postprocess(self, target):
+        if not (0 <= target <= 1):
+            raise ValueError(
+                "target {} should be between 0 and 1.".format(target))
```
bdice, line 244:
```suggestion
                f"target {target} should be between 0 and 1.")
```

hoomd/data/syncedlist.py
```diff
@@ -1,32 +1,75 @@
...
+        validation (function or class): A function that takes one argument
             and returns a boolean based on whether the value is appropriate for
             the list. Can raise ValueError for helpful diagnosis of the problem
-            with validation.
+            with validation. Alternatively validation can be a class which
+            indicates the expected type of items of the list.
         to_synced_list (function, optional): A function that takes one
             argument (a valid SyncedList python value) and does necessary
             conversion before adding to the cpp list. Defaults to a function
             that grabs the ``_cpp_obj`` attribute from the value.
```
bdice, line 52: It appears that the default value is an identity function?

hoomd/md/pair/pair.py
```diff
@@ -1659,7 +1653,7 @@ class LJ1208(Pair):
         \\begin{eqnarray*}
         V_{\\mathrm{LJ}}(r)
           = & 4 \\varepsilon \\left[ \\left( \\frac{\\sigma}{r} \\right)^{12} -
-          \\alpha \\left( \\frac{\\sigma}{r} \\right)^{8} \\right]
+          \\left( \\frac{\\sigma}{r} \\right)^{8} \\right]
```
bdice, line 1656: The changes to documentation of potential parameters seem unrelated to the purpose of this PR. This PR is very large and I would advocate for smaller PRs in the future, because I can't review this in one pass.

hoomd/md/pytest/test_potential.py
```diff
@@ -983,3 +983,22 @@ def test_force_energy_accuracy(simulation_factory,
...
+                                      nlist=md.nlist.Cell(),
+                                      r_cut=2.5)
+    for pair in valid_params.pair_potential_params:
+        pot.params[pair] = valid_params.pair_potential_params[pair]
+    pickling_check(pot)
+    integrator = hoomd.md.Integrator(0.05, forces=[pot])
+    sim.operations.integrator = integrator
+    sim.run(0)
+    # if isinstance(pot, hoomd.md.pair.SLJ):
+    #     breakpoint()
```
bdice, line 1003:
```suggestion
```

hoomd/operation.py
```diff
@@ -144,153 +95,15 @@ def _setattr_typeparam(self, attr, value):
...
-            if obj._typeparam_dict[name]._len_keys > 1:
-                tp_dict = str_to_tuple_keys(tp_dict)
-            setattr(obj, name, tp_dict)
-        return obj
+    def __eq__(self, other):
+        if not isinstance(other, _HOOMDGetSetAttrBase):
+            return NotImplemented
+        return (self._param_dict == other._param_dict
+                and all(
+                    self._typeparam_dict.keys() == other._typeparam_dict.keys()
```
bdice, line 103: I'm not sure if this needs to be a `set` -- are dictionary key views treated as ordered or unordered? Also probably depends on the implementation of `keys()` for this object.

hoomd/operation.py
```diff
@@ -394,15 +211,15 @@ def _setattr_param(self, attr, value):
                                      " initialization".format(attr))
 
     def __eq__(self, other):
-        other_keys = set(other.__dict__.keys())
+        other_keys = other.__dict__.keys()
         for key in self.__dict__.keys():
             if key in self._skip_for_equality:
```
bdice, line 216: Wondering if `other._skip_for_equality` matters?

hoomd/operation.py
```diff
@@ -394,15 +211,15 @@ def _setattr_param(self, attr, value):
...
                 continue
             else:
                 if key not in other_keys \
```
bdice, line 219: This can just be a single-line `elif`.

hoomd/simulation.py
```diff
@@ -240,8 +240,16 @@ def operations(self, operations):
                 self._operations._unschedule()
                 self._operations._simulation = None
                 operations._simulation = self
-                operations._schedule()
                 self._operations = operations
+                # This _schedule method must be called  after setting
```
bdice, line 244:
```suggestion
                # This _schedule method must be called after setting
```

hoomd/variant.py
```diff
@@ -73,6 +73,14 @@ def __init__(self, value):
         Variant.__init__(self)
         _hoomd.VariantConstant.__init__(self, value)
 
+    def __eq__(self, other):
+        """Return whether two variants are equivalent."""
```
bdice, line 77: These `__eq__` methods are highly redundant. I wonder if you can do something less redundant, by defining the relevant attributes in each subclass and only implementing this `__eq__` once in the parent class?

hoomd/write/table.py
```diff
@@ -38,6 +39,12 @@ def __subclasshook__(cls, C):
             return NotImplemented
 
 
+def _writable(fh):
```
bdice, line 42: Indicate that this is a "check" function in its name.
```suggestion
def _ensure_writable(fh):
```

hoomd/write/table.py
```diff
@@ -275,6 +277,30 @@ def act(self, timestep=None):
             self._write_row(output_dict)
             self.output.flush()
 
+    def __getstate__(self):
+        state = copy.copy(self.__dict__)
+        state.pop('_comm', None)
+        # This is to handle when the output specified is just stdout. By default
+        # file objects like this are not picklable, so we need to handle it
+        # differently. We let `None` represent stdout in the state dictionary.
```
bdice, line 285: This is confusing. So objects using `stdout` are picklable, because that value is mapped to `None`. Are objects using other outputs (file outputs, stderr, etc.) not picklable?

b-butler:
@bdice sorry I know it is long. :sweat_smile: Most of these changes are directly relevant to picklability, though. I could have done many smaller PRs, but I originally did not split PR's because I thought this would get merged faster than a bunch of smaller PR's given some difficulties we have had in getting a second reviews for some Python based hoomd PR's. Given the lifetime of this one though, I don't think that panned out. I do agree though in general smaller atomic PR's are better, sorry for this behemoth.

hoomd/write/table.py
b-butler: No there are not currently. It may be implementable to some extent, but I didn't attempt to do so here.

hoomd/custom/custom_operation.py
```diff
@@ -54,13 +54,13 @@ def __init__(self, action, trigger=1):
     def __getattr__(self, attr):
         """Allows pass through to grab attributes/methods of the wrapped object.
         """
+        if attr == '_action':
+            raise AttributeError(
+                f"{type(self).__name__} object has no attribute _action")
```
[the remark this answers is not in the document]
b-butler: This is only called if `_action` isn't an attribute by the nature of `__getattr__`. This is to prevent infinite recursion in some edge cases where `_action` is not set.

hoomd/md/pair/pair.py
b-butler: Yes this one is unrelated. I ended up fixing it in this PR because I only noticed the error due to other changes I was making.

hoomd/operation.py
b-butler: A set would be more reliable.

hoomd/operation.py
b-butler: It shouldn't since this is set at the class level, and objects of different classes shouldn't be equal in the first place.

hoomd/data/syncedlist.py
```diff
@@ -1,32 +1,75 @@
...
+def identity(obj):
+    return obj
+
+
+class SyncedList(MutableSequence):
     """Provides syncing and validation for a python and cpp list.
 
     Used to ensure once synced that standard list operations effect both
     python and cpp. Also guarentees that list remain in same order when using
     the public API.
```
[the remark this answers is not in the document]
b-butler: This is from when it was a private class

b-butler:
@bdice, I addressed all your comments. If you don't want to review again given the size of the PR, can you remove yourself from the reviewer pull, so the PR does not say changes requested.

joaander (commented):
Does this PR add any user documentation that indicates what HOOMD objects are pickleable?

I added a note to #979 to add dev docs later, but we do need to document it at the reference level for the user now. 

We can later add a full fledged tutorial with restartable jobs in #736 . I've considered, but haven't decided if it is a good idea to introduce pickling in the early introductory tutorials. It would reduce the code length of the tutorials but may be too abstract for new users. What do you think?

bdice:
> I've considered, but haven't decided if it is a good idea to introduce pickling in the early introductory tutorials. It would reduce the code length of the tutorials but may be too abstract for new users. What do you think?

For me, this decision depends on how widely pickling is supported, because it sets the user’s expectation that pickling is (more or less) universally allowed. I’m not really sure what objects don’t support pickling but if that list is short, I’d say it is fine to use in a tutorial.

joaander:
> > I've considered, but haven't decided if it is a good idea to introduce pickling in the early introductory tutorials. It would reduce the code length of the tutorials but may be too abstract for new users. What do you think?
> 
> For me, this decision depends on how widely pickling is supported, because it sets the user’s expectation that pickling is (more or less) universally allowed. I’m not really sure what objects don’t support pickling but if that list is short, I’d say it is fine to use in a tutorial.

@b-butler correct me if I'm wrong, but this PR essentially makes all (or almost all) `Operations` objects pickleable. We could use it to replace the repeated instantiations of the HPMC integrator in the tutorial, for example: https://hoomd-blue.readthedocs.io/en/latest/tutorial/00-Introducing-HOOMD-blue/06-Equilibrating-the-System.html#Equilibration

b-butler:
@bdice @joaander The PR does make basically everything picklable that can be. The only exception that I am aware of is the `Table` writer when using a filehandle. We could even allow this by changing it to take in a filename and mode instead of a file handle.

b-butler:
@joander I added a basic architecture file regarding pickling in `hoomd/architecture.rst`.

hoomd/architecture.rst
```diff
@@ -0,0 +1,67 @@
+ARCHITECTURE.rst
```
joaander, line 1: Why are we adding this in `ARCHITECTURE.rst`? Is there something wrong with the existing `ARCHITECTURE.md` file? I would prefer not to maintain two separate files with similar goals.

joaander (changes requested):
Please resolve the merge conflicts and provide only one `ARCHITECTURE.*` file. The only reason to use `.rst` over `.md` is if we plan to also include it in the sphinx generated docs.

joaander:
> Edit: I did not add it to the documentation as I had forgotten what we decided on about that and its very incomplete status.

It should be mentioned at least in the base class: `hoomd.operation.Operation`.

hoomd/architecture.rst
b-butler: Honestly I somehow just missed it, I will fix that immediately.

hoomd/operation.py
```diff
@@ -497,6 +306,13 @@ class Operation(_HOOMDBaseObject):
     `Writer`, `Compute`, `Tuner`, and `hoomd.integrate.BaseIntegrator`. All
     HOOMD-blue operations inherit from one of these five base classes. To find
     the purpose of each class see its documentation.
+
+    Note:
```
joaander, line 310: I think you may have misunderstood my comment. I don't think there should be a link to `ARCHITECTURE.md` here, but there should be a brief description that all `Operation` instances are picklable. Without this mention in the documentation, the change log is the only place that users will be able to see pickling mentioned.

b-butler: I did misunderstand... I will fix that.

vyasr:
This question isn't 100% relevant to this PR and I may have missed when this decision was made, but is the plan going forward for all of the meatadata/restartability features of HOOMD to be handled via pickling? I ask because of the many portability concerns associated with pickle that also make it an unsuitable format for long-term archival or sharing of such restartability information. The original JSON-based format would have avoided this problem, but at the cost of having to develop a suitable schema. Was this tradeoff deemed acceptable?

joaander:
> This question isn't 100% relevant to this PR and I may have missed when this decision was made, but is the plan going forward for all of the meatadata/restartability features of HOOMD to be handled via pickling?

Not exactly. 

We decided to drop work on the JSON format because it was not able to handle many of the nuances our object schema needs (tuples vs lists, object references). Also, we could not identify any other suitable format other than pickle.

However, I wouldn't say that the pickle format enabled in this PR is intended to be the solution to long term archival or sharing for the reasons you state. The best format for long term archival and sharing is still the Python script itself, in combination with metadata stored by tools like signac. I see the pickle file as a temporary storage and convenience for users with multiple workflow stages. For example, one workflow stage could use the Python API to create a list of operations and save it to a pickle file. Later stages in the workflow could read that pickle file instead of duplicating code.

Users could also avoid code duplication by putting the operation generating code into a reusable function, but there are cases like the original use-case for the JSON files (mbuild/foyer) where the code that generates the force field may not be easy to install on a cluster. With pickle files, the user could run that workflow step on their workstation, transfer the pickle files to the cluster, and continue the later stages (provided that they write the pickle files in a form compatible with both systems).

joaander:
@b-butler tests are failing after merging from master.

vyasr:
@joaander got it. I think that makes sense and is all that we should try to support, but probably needs to be clarified for the MoSDeF people. I know you're working on the mbuild->HOOMD conversion, is the intention to create any non-temporary intermediate files, or is it just meant for including HOOMD and mbuild in the same script? In the latter case it's completely fine, but in the former we probably need to document the limitations.

joaander (approved):
Thanks!

PR #945, opened 2021-02-25
Improve Mie Pair Potential Performance

### LINKED ISSUE #939: Improve Mie pair potential performance
## Description

Improve the performance of the Mie pair potential.

## Proposed solution

It would involve modifying EvaluatorPairMie.h and HOOMDMath.h - 

* param_type(pybind11::dict v) should precompute mie3=log2(n)/2.0 and mie4=log2(m)/2.0
* asDict needs to recover n and m from mie3 and mie4
* evalForceAndEnergy should replace pow(x, y) with fast::exp2(x * log2f(y)) - note that mie3 and mie4 contain the precomputed log2f factor
* add fast::exp2 to HOOMDMath.h

## Additional context

In the current implementation, the calls to pow() are always double precision. Also, pow is extremely expensive on the GPU.

## Developer

Seeking contributors.

jbdelony:
## Description

Improve the performance of the Mie pair potential relative to the Lennard-Jones potential.

## Motivation and context

The Lennard-Jones potential was observed to perform ~5x better than the equivalent Mie pair potential (m = 6, n = 12) due to  several inefficient pow() operations within the Mie potential.

Resolves #939 

## How has this been tested?

Performance of HOOMD's LJ-liquid benchmark was compared before and after making the proposed changes. Performance improved by a factor of 2 on the GTX 1080Ti.

## Change log

Replace `pow(x, y)` operations with the equivalent `exp2(y * log2(x))` operation in `EvaluatorPairMie.h`.  Added single-precision `exp2` and `log2` compute capabilities to `HOOMDMath.h`. 

## Checklist:


hoomd/md/HOOMDMath.h
```diff
@@ -0,0 +1,830 @@
+// Copyright (c) 2009-2021 The Regents of the University of Michigan
```
joaander, line 1: You should modify the existing `hoomd/HOOMDMath.h`. Don't create a new identical one in the `md` directory.

hoomd/md/EvaluatorPairMie.h
```diff
@@ -58,14 +58,16 @@ class EvaluatorPairMie
             Scalar m2;
             Scalar m3;
             Scalar m4;
+	    Scalar mie_n;
```
joaander, line 61: There should be a way to implement this without storing additional parameters. If you transform m and n and when you store them in `m3` and `m4`, then you can apply the reverse transform to the numbers in `asDict`

joaander (changes requested):
Thanks for doing this! There are a few changes needed before we can merge.

vyasr (commented):
This is a nice refactoring. It appears that the only changes in the new HOOMDMath.h file relative to the original are the addition of the `log2` and `exp2` functions. If this method is only faster when the exponent is large (which I think would be the case?) then I would suggest implementing this optimization as a new function in the `fast` namespace rather than rewriting every line to use the exp/log trick. That would also be self-documenting, since the math behind this choice may not be immediately evident to everyone.

If this approach is always faster and does not have meaningfully worse robustness characteristics, then @joaander would it make sense to redefine `fast::pow` directly? That would be a quick way to apply this optimization across the codebase.

Is the choice of log2 significant? I assume that the main log functions (log2, log10, log) are all GPU intrinsics and equivalent from a performance perspective, is there any reason to prefer one to the others?

joaander:
The CUDA best practices guide mentions that pow() has to handle many special cases and that is whey it is slow and consumes many registers. The exp/log trick will work well for all values I expect users to pass into a mie potential
 
> If this approach is always faster and does not have meaningfully worse robustness characteristics, then @joaander would it make sense to redefine `fast::pow` directly? That would be a quick way to apply this optimization across the codebase.

That would be a more readable modification to the codebase. @jbdelony feel free to make this change instead. The diff will be much smaller.

I originally thought to precompute the log2 factor on the CPU so that the GPU only needed to call exp2 but that may be premature optimizaiton.

> 
> Is the choice of log2 significant? I assume that the main log functions (log2, log10, log) are all GPU intrinsics and equivalent from a performance perspective, is there any reason to prefer one to the others?

This is what CUDA uses for its fast __pow (which is only available for single precision). I don't know, but I'd guess maybe that log2 and exp2 are slightly faster than log and pow.

hoomd/md/EvaluatorPairMie.h
jbdelony: For now I renamed the transformed `n` and `m` to `mie_log2_n` and `mie_log2_m` for clarity.  Since `m3` and `m4` are used to directly compute the prefactor `outFront`, `m1`, and `m2`, I'm not sure if there's a way to store the transformed values without additional parameters.  Perhaps I'm misunderstanding.

jbdelony:
> That would be a more readable modification to the codebase. @jbdelony feel free to make this change instead. The diff will be much smaller.

@vyasr @joaander While the redefinition of `fast::pow` certainly makes the codebase more readable, unfortunately it is not as robust as anticipated as this build no longer passes the pair potential unit tests due to an assertion error.  Assuming this can be attributed to the augmented loss of precision via refactoring, what do you think is the best solution?

vyasr:
@jbdelony @joaander it appears at a glance that the only potential that this causes failures for is the Fourier potential. Unless we for some reason expect nondeterministic failures for other potentials (@joaander will have a better idea if that is the case from previous runs), what this really indicates to me is that the Fourier potential should be using `slow::pow` rather than `fast::pow`. Up until now those two functions have been defined identically (at least on CUDA, if not on HIP), but using `slow::pow` would be an explicit declaration that that potential requires a more precise output. If that change is sufficient to make tests pass I would be happy with it.

joaander:
The issue with the Fourier pair potential is that computes `fast::pow(Scalar(-1),Scalar(i))`. The new fast::pow takes the log of -1 which results in NaN. The Fourier pair potential has no need to call `pow` at all. It can use an even/odd test and set the  factor to -1 or 1 accordingly.

vyasr:
That does raise an interesting point, is it fine that `fast::pow` is no longer defined for negative bases? Are we OK with requiring calling `slow::pow` for negative bases? If so, this should at least be documented.

jbdelony:
> Is the choice of log2 significant? I assume that the main log functions (log2, log10, log) are all GPU intrinsics and equivalent from a performance perspective, is there any reason to prefer one to the others?

For the sake of rigor, I refactored and profiled `fast::pow` as a function of `log` and `log10` as well; results are attached here.  Although all perform within a 10% margin, `log` performs best.  This would also alleviate the need for defining `fast::exp2` and `fast::log2` (same with base 10); however, our issue with negative numbers would still remain.

[profile_Mie_1E6_log10.log](https://github.com/glotzerlab/hoomd-blue/files/6063659/profile_Mie_1E6_log10.log)
[profile_Mie_1E6_loge.log](https://github.com/glotzerlab/hoomd-blue/files/6063660/profile_Mie_1E6_loge.log)
[profile_Mie_1E6_log2.log](https://github.com/glotzerlab/hoomd-blue/files/6063661/profile_Mie_1E6_log2.log)

joaander:
> > Is the choice of log2 significant? I assume that the main log functions (log2, log10, log) are all GPU intrinsics and equivalent from a performance perspective, is there any reason to prefer one to the others?
> 
> For the sake of rigor, I refactored and profiled `fast::pow` as a function of `log` and `log10` as well; results are attached here. Although all perform within a 10% margin, `log` performs best. This would also alleviate the need for defining `fast::exp2` and `fast::log2` (same with base 10); however, our issue with negative bases would still remain.
> 
> [profile_Mie_1E6_log10.log](https://github.com/glotzerlab/hoomd-blue/files/6063659/profile_Mie_1E6_log10.log)
> [profile_Mie_1E6_loge.log](https://github.com/glotzerlab/hoomd-blue/files/6063660/profile_Mie_1E6_loge.log)
> [profile_Mie_1E6_log2.log](https://github.com/glotzerlab/hoomd-blue/files/6063661/profile_Mie_1E6_log2.log)

Certainly, you could remove the log2 code and just use log and exp. That does seem to be faster in your benchmarks. I'll make a few more suggestions in a review.

hoomd/HOOMDMath.h
```diff
@@ -492,20 +492,20 @@ inline HOSTDEVICE void sincospi(double x, double& s, double& c)
     #endif
     }
 
-//! Compute the pow of x,y
+//! Compute the pow of x,y with single precison via exp2(log2) refactoring
```
joaander, line 495: As Vyas said, this should document that this approximation is invalid for negative x. I don't think the loss of generality is a problem. Developers that use the fast math functions have to opt-in to them and should understand the limitations before doing so.

hoomd/md/EvaluatorPairMie.h
```diff
@@ -58,14 +58,16 @@ class EvaluatorPairMie
...
+            Scalar mie_log2_m;
 
             #ifdef ENABLE_HIP
             // set CUDA memory hints
             void set_memory_hint() const {}
             #endif
 
             #ifndef __HIPCC__
-            param_type() : m1(0), m2(0), m3(0), m4(0) {}
+            param_type() : m1(0), m2(0), m3(0), m4(0), mie_log2_n(0), mie_log2_m(0) {}
```
joaander, line 70:
```suggestion
            param_type() : m1(0), m2(0), m3(0), m4(0) {}
```
With the new fast::pow refactoring, `mie_log2_n/m` are no longer used in `evalForceAndEnergy` and are not needed.

hoomd/md/EvaluatorPairMie.h
```diff
@@ -75,9 +77,11 @@ class EvaluatorPairMie
...
                 Scalar sigma = v["sigma"].cast<Scalar>();
 
-                Scalar outFront = (m3/(m3-m4)) * pow(m3/m4, m4/(m3-m4));
-                m1 = outFront * epsilon * pow(sigma, m3);
-                m2 = outFront * epsilon * pow(sigma, m4);
+                Scalar outFront = (m3/(m3-m4)) * fast::pow(m3/m4, m4/(m3-m4));
+                m1 = outFront * epsilon * fast::pow(sigma, m3);
+                m2 = outFront * epsilon * fast::pow(sigma, m4);
+                mie_log2_n = fast::log2(m3) / Scalar(2.0);
+                mie_log2_m = fast::log2(m4) / Scalar(2.0);
```
joaander, line 84:
```suggestion
                mie_log2_n = fast::log2(m3) / Scalar(2.0);
                mie_log2_m = fast::log2(m4) / Scalar(2.0);
```
With the new fast::pow refactoring, `mie_log2_n/m` are no longer used in `evalForceAndEnergy` and are not needed.

hoomd/md/EvaluatorPairMie.h
```diff
@@ -86,12 +90,16 @@ class EvaluatorPairMie
...
                 v["m"] = m4;
 
-                Scalar sigma = pow(m1 / m2, 1 / (m3 - m4));
-                Scalar epsilon = m1 / pow(sigma, m3) * (m3 - m4) / m3 * pow(m3 / m4, m4 / (m4 - m3));
-
+                Scalar sigma = fast::pow(m1 / m2, 1 / (m3 - m4));
+                Scalar epsilon = m1 / fast::pow(sigma, m3) * (m3 - m4) / m3 * fast::pow(m3 / m4, m4 / (m4 - m3));
+		
+                Scalar n = fast::exp2(Scalar(2.0) * mie_log2_n);
+                Scalar m = fast::exp2(Scalar(2.0) * mie_log2_m);
```
joaander, line 97:
```suggestion
                Scalar n = fast::exp2(Scalar(2.0) * mie_log2_n);
                Scalar m = fast::exp2(Scalar(2.0) * mie_log2_m);
```
n and m are stored directly above:
```
v["n"] = m3;
v["m"] = m4;
```

hoomd/md/EvaluatorPairMie.h
```diff
@@ -86,12 +90,16 @@ class EvaluatorPairMie
...
+		
                 v["epsilon"] = epsilon;
                 v["sigma"] = sigma;
-
+                v["n"] = n;
+                v["m"] = m;
```
joaander, line 102:
```suggestion
```
`n` and `m` were already set above.

hoomd/md/EvaluatorPairMie.h
```diff
@@ -141,16 +149,16 @@ class EvaluatorPairMie
...
                 force_divr= r2inv * (mie3 * mie1 * rninv - mie4 * mie2 * rminv);
 
                 pair_eng = mie1 * rninv - mie2 * rminv;
 
                 if (energy_shift)
                     {
-                    Scalar rcutninv = Scalar(1.0)/pow(rcutsq,mie3/Scalar(2.0));
-                    Scalar rcutminv = Scalar(1.0)/pow(rcutsq,mie4/Scalar(2.0));
+                    Scalar rcutninv = Scalar(1.0)/fast::pow(rcutsq,mie3/Scalar(2.0));
+                    Scalar rcutminv = Scalar(1.0)/fast::pow(rcutsq,mie4/Scalar(2.0));
```
joaander, line 161:
```suggestion
                    Scalar rcutninv = fast::pow(rcutsq,-mie3/Scalar(2.0));
                    Scalar rcutminv = fast::pow(rcutsq,-mie4/Scalar(2.0));
```
You can save a few cycles using a negative exponent in place of division.

You might try the same to avoid the 1.0/rsq above and see if it makes any measurable change in performance.

joaander (changes requested):
Thanks! Just a few more changes and we can merge this improvement.

To fix the failing tests, replace this code in `EvaluatorPairFourier.h`:
```
                    a1 = a1 + fast::pow(Scalar(-1),Scalar(i)) * params.a[i-2];
                    b1 = b1 + i * fast::pow(Scalar(-1),Scalar(i)) * params.b[i-2];
```
with:
```
                    Scalar pow_neg1_i = (i & 1) ? -1.0 : 1.0;
                    a1 = a1 + pow_neg1_i * params.a[i-2];
                    b1 = b1 + i * pow_neg1_i * params.b[i-2];
```

hoomd/md/EvaluatorPairMie.h
jbdelony: I tested the performance both `1/rsq` and the negative exponent in the non-energy shifted calculation, and `1/rsq` performs marginally better, presumably because `r2inv` is calculated and stored for the following force calculation anyway.

hoomd/HOOMDMath.h
jbdelony: I also added a note that this redefinition is undefined for negative bases for the sake of clarity.

vyasr (approved):
Thanks for going through all these iterations! This version looks good to me.

hoomd/md/EvaluatorPairFourier.h
```diff
@@ -162,6 +162,9 @@ class EvaluatorPairFourier
                 Scalar b1 = 0;
                 for (int i=2; i<5; i++)
                     {
+                    Scalar pow_neg1_i = (i & 1) ? -1.0 : 1.0;
+                    a1 = a1 + pow_neg1_i * params.a[i-2];
+                    b1 = b1 + i * pow_neg1_i * params.b[i-2];
                     a1 = a1 + fast::pow(Scalar(-1),Scalar(i)) * params.a[i-2];
                     b1 = b1 + i * fast::pow(Scalar(-1),Scalar(i)) * params.b[i-2];
```
joaander, line 169:
```suggestion
```

joaander (changes requested):
One more change needed to the fourier potential to get the tests to pass.

joaander (approved):
Thanks!

PR #946, opened 2021-02-27
Convert Mueller plathe flow to v3 API

### LINKED ISSUE #858: Port mueller_plathe_flow to v3.0 API
## Description

Port the mueller_plathe_flow updater to the v3.0 API, including unit tests.

danevans09:
## Description

I converted the mueller plathe flow updater to the v3 API by by adding a parameter dict to the initialization method and adding an attach method. I also added all of the python properties in the parameter dict to the pybind properties. Additionally, I initialized and added flow_epsilon to the parameter dict and then passed it as an argument to the C++ constructor, so it could be treated like a python property.

## Motivation and context

The Mueller plathe flow updater has not yet been converted to the v3 API

Resolves #858 

## How has this been tested?

I tested that the python class initialized properly before attaching. I then tested that the class attached and set the properties correctly. I also tested setting flow_epsilon from the python property.

## Change log

```

```

## Checklist:


hoomd/md/pytest/test_mueller_plathe_flow.py
```diff
@@ -0,0 +1,63 @@
...
+    assert mpf.slab_direction == slab_direction
+    assert mpf.flow_direction == flow_direction
+    assert mpf.n_slabs == n_slabs
+    assert mpf.max_slab == n_slabs / 2
+    assert mpf.min_slab == 0
+    assert mpf.trigger == hoomd.trigger.Periodic(1)
+    assert isinstance(mpf.has_max_slab, bool)
+    assert isinstance(mpf.has_min_slab, bool)
+    assert mpf.summed_exchanged_momentum == 0
+    sim.run(1)
```
SchoeniPhlippsn, line 57: I would recommend also to test that the other parameters cannot be changed after attachment. See `test-methods.py`

hoomd/md/update.py
```diff
@@ -221,100 +224,144 @@ class mueller_plathe_flow(_updater):
...
-        slab_direction (``X``, ``Y``, or ``Z``): Direction perpendicular to the slabs..
-        flow_direction (``X``, ``Y``, or ``Z``): Direction of the flow..
-        n_slabs (int): Number of slabs. You want as many as possible for small disturbed volume, where the unphysical swapping is done. But each slab has to contain a sufficient number of particle.
-        max_slab (int): Id < n_slabs where the max velocity component is search for. If set < 0 the value is set to its default n_slabs/2.
-        min_slab (int): Id < n_slabs where the min velocity component is search for. If set < 0 the value is set to its default 0.
+        filter (`hoomd.filter.ParticleFilter`): Subset of particles on which to
+            apply this updater.
+
+        flow_target (`hoomd.variant.Variant`): Integrated target flow. The unit
+            is the in the natural units of the simulation: [flow_target] = [timesteps] x :math:`\mathcal{M}` x :math:`\frac{\mathcal{D}}{\tau}`.
```
SchoeniPhlippsn, line 231:
```suggestion
            is in the natural units of the simulation: [flow_target] = [timesteps] x :math:`\mathcal{M}` x :math:`\frac{\mathcal{D}}{\tau}`.
```

hoomd/md/update.py
```diff
@@ -221,100 +224,144 @@ class mueller_plathe_flow(_updater):
...
+            The unit of [timesteps] is your discretization dt x :math:`\mathcal{\tau}`.
+
+        slab_direction (``X``, ``Y``, or ``Z``): Direction perpendicular to the slabs.
+
+        flow_direction (``X``, ``Y``, or ``Z``): Direction of the flow.
+
+        n_slabs (int): Number of slabs. You want as many as possible for small
+            disturbed volume, where the unphysical swapping is done. But each
+            slab has to contain a sufficient number of particle.
```
SchoeniPhlippsn, line 240: What does "sufficient" mean here?

hoomd/md/update.py
```diff
@@ -221,100 +224,144 @@ class mueller_plathe_flow(_updater):
...
+            for. If set < 0 the value is set to its default n_slabs/2.
+
+        min_slab (int): Id < n_slabs where the min velocity component is search
+            for. If set < 0 the value is set to its default 0.
 
     .. attention::
         * This updater has to be always applied every timestep.
         * This updater works currently only with orthorhombic boxes.
 
     .. note:
```
SchoeniPhlippsn, line 252:
```suggestion
    Note:
```

hoomd/md/update.py
```diff
@@ -221,100 +224,144 @@ class mueller_plathe_flow(_updater):
...
+            disturbed volume, where the unphysical swapping is done. But each
+            slab has to contain a sufficient number of particle.
+
+        max_slab (int): Id < n_slabs where the max velocity component is search
+            for. If set < 0 the value is set to its default n_slabs/2.
+
+        min_slab (int): Id < n_slabs where the min velocity component is search
+            for. If set < 0 the value is set to its default 0.
 
     .. attention::
```
SchoeniPhlippsn, line 248:
```suggestion
    Attention:
```

hoomd/md/update.py
```diff
@@ -221,100 +224,144 @@ class mueller_plathe_flow(_updater):
...
         * This updater has to be always applied every timestep.
         * This updater works currently only with orthorhombic boxes.
 
     .. note:
-        If you set this updater with unrealistic values, it algorithm might not terminate,
+        If you set this updater with unrealistic values, the algorithm might not terminate,
         because your desired flow target can not be achieved.
 
     .. versionadded:: v2.1
```
SchoeniPhlippsn, line 256: Probably not needed anymore

hoomd/md/update.py
```diff
@@ -221,100 +224,144 @@ class mueller_plathe_flow(_updater):
...
-        #const integrated flow with 0.1 slope for max 1e8 timesteps
-        const_flow = hoomd.variant.linear_interp( [(0,0),(1e8,0.1*1e8)] )
-        #velocity gradient in z direction and shear flow in x direction.
-        update.mueller_plathe_flow(all,const_flow,md.update.mueller_plathe_flow.Z,md.update.mueller_plathe_flow.X,100)
+        # const integrated flow with 0.1 slope for max 1e8 timesteps
+        ramp = hoomd.variant.Ramp(0.0, 0.1e8, 0, int(1e8))
+        # velocity gradient in z direction and shear flow in x direction.
+        slab_direction = hoomd.md.update.MullerPlatheFlow.Z
+        flow_direction = hoomd.md.update.MullerPlatheFlow.X
+        mpf = hoomd.md.update.MuellerPlatheFlow(hoomd.filter.All(), ramp, slab_direction, flow_direction, 20)
```
SchoeniPhlippsn, line 265: For consistency, please set the parameter explicitly in the examples
```suggestion
        mpf = hoomd.md.update.MuellerPlatheFlow(filter = hoomd.filter.All(), flow_target = ramp, slab_direction  = slab_direction, flow_direction = flow_direction, n_slabs = 20)
```

hoomd/md/update.py
```diff
@@ -221,100 +224,144 @@ class mueller_plathe_flow(_updater):
...
+        params = ParameterDict(
+            filter=hoomd.filter.ParticleFilter,
+            flow_target=hoomd.variant.Variant,
+            slab_direction=_md.MuellerPlatheFlow.Direction,
+            flow_direction=_md.MuellerPlatheFlow.Direction,
+            n_slabs=int(n_slabs),
+            max_slab=int(max_slab),
+            min_slab=int(min_slab),
+            flow_epsilon=float(1e-2))
+        params["filter"] = filter
```
SchoeniPhlippsn, line 308: Here you can use `params.update( dict(filter=filter,...))`

SchoeniPhlippsn (commented):
Thanks for converting it to v3 API. I only have a couple of comments of mostly cosmetic nature.

hoomd/md/MuellerPlatheFlow.cc
```diff
@@ -453,15 +454,18 @@ void export_MuellerPlatheFlow(py::module& m)
...
+        .def_property_readonly("n_slabs",&MuellerPlatheFlow::get_N_slabs)
+        .def_property_readonly("min_slab",&MuellerPlatheFlow::get_min_slab)
+        .def_property_readonly("max_slab",&MuellerPlatheFlow::get_max_slab)
+        .def_property_readonly("has_min_slab",&MuellerPlatheFlow::has_min_slab)
+        .def_property_readonly("has_max_slab",&MuellerPlatheFlow::has_max_slab)
+        .def_property_readonly("flow_target",&MuellerPlatheFlow::get_flow_target)
+        .def_property_readonly("slab_direction",&MuellerPlatheFlow::get_slab_direction)
+        .def_property_readonly("flow_direction",&MuellerPlatheFlow::get_flow_direction)
+        .def_property("flow_epsilon",&MuellerPlatheFlow::get_flow_epsilon, &MuellerPlatheFlow::set_flow_epsilon)
+        // .def("setFlowEpsilon",&MuellerPlatheFlow::set_flow_epsilon)
```
joaander, line 467:
```suggestion
```

hoomd/md/update.py
```diff
@@ -203,7 +205,8 @@ def __init__(self, group, r=None, rx=None, ry=None, rz=None, P=(0,0,0)):
         self.rz = rz
         self.metadata_fields = ['group','P', 'rx', 'ry', 'rz']
 
-class mueller_plathe_flow(_updater):
+
+class MuellerPlatheFlow(Updater):
     R""" Updater class for a shear flow according
```
joaander, line 210:
```suggestion
    """Mueller/Plathe method to establish shear flow.
```

hoomd/md/update.py
```diff
@@ -203,7 +205,8 @@ def __init__(self, group, r=None, rx=None, ry=None, rz=None, P=(0,0,0)):
...
     to an algorithm published by Mueller Plathe.:
```
joaander, line 211:
```suggestion
```

hoomd/md/update.py
joaander: @mphoward Could you recommend some language here?

hoomd/md/update.py
```diff
@@ -221,100 +224,140 @@ class mueller_plathe_flow(_updater):
...
-        assert (max_slab>-1 and max_slab < n_slabs),"Invalid max_slab in [0,"+str(n_slabs)+")."
-        assert (min_slab>-1 and min_slab < n_slabs),"Invalid min_slab in [0,"+str(n_slabs)+")."
-        assert (min_slab != max_slab),"Invalid min/max slabs. Both have the same value."
+        min_slab (int): Id < n_slabs where the min velocity component is search
+            for. If set < 0 the value is set to its default 0.
 
-        # initialize the base class
-        _updater.__init__(self);
+    Attention:
+        * This updater has to be always applied every timestep.
```
joaander, line 249: @b-butler Can we force an Updater to have a certain trigger?

hoomd/md/update.py
```diff
@@ -221,100 +224,140 @@ class mueller_plathe_flow(_updater):
...
+        * This updater works currently only with orthorhombic boxes.
 
-        self._flow_target = hoomd.variant._setup_variant_input(flow_target);
+    .. versionadded:: v2.1
```
joaander, line 252:
```suggestion
```

hoomd/md/update.py
```diff
@@ -221,100 +224,140 @@ class mueller_plathe_flow(_updater):
...
+        if max_slab < 0:
+            max_slab = n_slabs / 2
+        assert (max_slab > -1 and max_slab < n_slabs), "Invalid max_slab in [0,"+str(n_slabs)+")."
+        assert (min_slab > -1 and min_slab < n_slabs), "Invalid min_slab in [0,"+str(n_slabs)+")."
+        assert (min_slab != max_slab), "Invalid min/max slabs. Both have the same value."
+
+        params = ParameterDict(
+            filter=hoomd.filter.ParticleFilter,
+            flow_target=hoomd.variant.Variant,
+            slab_direction=_md.MuellerPlatheFlow.Direction,
```
joaander, line 298: Let's encode the directions as strings at the Python level. Pass them as std::string to C++ methods which can set the underlying enum values appropriately.

joaander (changes requested):
Thanks! I have a few change requests.

@mphoward You have mentioned several times on the mailing list that this implementation doesn't follow the parameters in the paper. It is beyond the scope of this particular PR, but would you suggest we update or replace this implementation?

hoomd/md/update.py
mphoward: Sufficient = "big enough volume that there are enough particles to apply momentum swap to". So, it depends on particle density, volume of the slab, and how much momentum to swap.

Not sure if this belongs here or maybe in the "long description", but some language along these lines should be in the docs somewhere:
```
n_slabs (int): Number of slabs used to divide the simulation box along the shear gradient. Using too few slabs will lead to a larger volume being disturbed by the momentum exchange, while using too many slabs may mean that there are not enough particles to exchange the target momentum.
```

hoomd/md/update.py
mphoward: This is actually one person with a hyphenated name: "Müller-Plathe" or "Mueller-Plathe"

joaander: > This is actually one person with a hyphenated name: "Müller-Plathe" or "Mueller-Plathe"

Thanks, I edited my suggestion accordingly. 

Perhaps we should also include the umlaut in the class name for correctness?

mphoward:
> @mphoward You have mentioned several times on the mailing list that this implementation doesn't follow the parameters in the paper. It is beyond the scope of this particular PR, but would you suggest we update or replace this implementation?

Yes, I would support converting this updater to the "standard" algorithm with an implementation more like this one:

https://github.com/mphowardlab/azplugins/blob/d54fe88808d1d65fdcd4764e84c0310b6b4e8ca7/azplugins/flow.py#L535-L656

Our plugin more closely matches the original implementation of the algorithm and the interface in lammps:

https://lammps.sandia.gov/doc/fix_viscosity.html

I think Ludwig Schneider implemented this one because he was using a new nonstandard method that he proposed:

https://doi.org/10.1016/j.commatsci.2019.109107

He was interested in doing shear in different directions & also having time varying shear rates. To this end, I think this implementation uses some sort of control logic to try to drive the stress toward a Variant value. This can be really confusing to users because it isn't what the original algorithm was meant to do, and also the control can fail pretty spectacularly.

The "standard" Müller-Plathe algorithm produces a steady shear. You prescribe the number of momentum swaps and optionally the target momentum to swap. (By default, the particles with the fastest speeds should be swapped.) Hence there is no need for this time-varying control... the system just comes to a steady state eventually. It also can run like a normal updater, with whatever period the user wants.

Our plugin (written by @astatt) has the guts of this implemented with a fixed coordinate system (you can't select the direction that is the shear gradient / flow direction... I think we chose z/x for those, but it could be rotated as needed). It also does not currently work in MPI. Maybe this and Ludwig's code could be combined though.

Feel free to paste this into a future issue. I'm not sure when I would have time to do this, but it would be a welcome change from my end. (We also have a matching updater for MPCD that we could add.) Maybe others prefer the current interface though.

hoomd/md/update.py
mphoward: I'm not a fan of unicode characters in class names, but maybe I am old fashioned. I think both spellings are acceptable, so I'm OK either way. This method can also be called RNE ("reverse nonequilibrium") or reverse perturbation, so `RNEFlow` and `ReversePerturbationFlow` would also be good names.

joaander: Sounds good. Let's call it `ReversePerturbationFlow`. @danevans09 can you rename the class at the Python level? No need to make C++ modifications at this time, we can defer that until the proposed later refactoring of this class.

hoomd/md/update.py
b-butler: Yes, through proper validation.

joaander: Given Michael Howard's comments below, this may not be necessary. With a different implementation, the period does not need to be 1.

joaander:
@danevans09 I merged master into this branch. Could you fix the pre-commit errors and compiler warnings?

Even if we later rewrite this with @astatt, I'd like to get this cleaned up and merged. We can break APIs during the beta period and this one is not very widely used.

hoomd/md/update.py
```diff
@@ -149,133 +150,140 @@ class mueller_plathe_flow:
...
+                 filter,
                  flow_target,
                  slab_direction,
                  flow_direction,
                  n_slabs,
                  max_slab=-1,
                  min_slab=-1):
-        period = 1  # This updater has to be applied every timestep
+
         assert (n_slabs > 0), "Invalid negative number of slabs."
```
joaander, line 222: Switch these to use `if` checks and raise `ValueError` or other appropriate exceptions. `assert` is meant to be used to check for errors in the implementation, not validate user parameters.

hoomd/md/update.py
```diff
@@ -149,133 +150,140 @@ class mueller_plathe_flow:
...
+        if self._attached:
+            return self._cpp_obj.summed_exchanged_momentum
+        else:
+            return None
 
+    @property
     def has_min_slab(self):
-        """Returns, whether this MPI instance is part of the min slab."""
-        return self.cpp_updater.hasMinSlab()
+        R"""Returns, whether this MPI instance is part of the min slab."""
```
joaander, line 277: Remove `has_min_slab` and `has_max_slab`. I don't see an obvious reason why the user would need to query this. If it gets requested later, this is easy enough to add back in.

hoomd/md/update.py
joaander: Since you changed the code to always set the trigger to a Periodic(1), update the attention block to state that this is the trigger used.

hoomd/md/update.py
```diff
@@ -149,133 +150,140 @@ class mueller_plathe_flow:
...
-            slabs..
-        flow_direction (``X``, ``Y``, or ``Z``): Direction of the flow..
-        n_slabs (int): Number of slabs. You want as many as possible for small
-            disturbed volume, where the unphysical swapping is done. But each
-            slab has to contain a sufficient number of particle.
+        filter (`hoomd.filter.ParticleFilter`): Subset of particles on which to
+            apply this updater.
+
+        flow_target (`hoomd.variant.Variant`): Integrated target flow. The unit
+            is in the natural units of the simulation: [flow_target] = [timesteps] x :math:`\mathcal{M}` x :math:`\frac{\mathcal{D}}{\tau}`.
```
joaander, line 157: Update the units to match the format used in #1019

joaander (changes requested):
Thanks, I have some suggestions for improvement.

hoomd/md/update.py
```diff
@@ -149,133 +150,134 @@ class mueller_plathe_flow:
...
-            slabs..
-        flow_direction (``X``, ``Y``, or ``Z``): Direction of the flow..
-        n_slabs (int): Number of slabs. You want as many as possible for small
-            disturbed volume, where the unphysical swapping is done. But each
-            slab has to contain a sufficient number of particle.
+        filter (`hoomd.filter.ParticleFilter`): Subset of particles on which to
+            apply this updater.
+
+        flow_target (`hoomd.variant.Variant`): Integrated target flow.
+            :math:`[\\delta t \\cdot \\mathrm{mass} \\cdot \\mathrm{length}
```
joaander, line 157: I updated the style here so that it renders correctly. This is a direct translation of the old documentation. It doesn't make sense to me because `\delta t \cdot \mathrm{time}^{-1}` cancels out and there is no time unit here.

hoomd/md/MuellerPlatheFlow.h
```diff
@@ -55,81 +33,134 @@ class PYBIND11_EXPORT MuellerPlatheFlow : public Updater
...
+        return m_flow_target;
+        }
+    std::string getSlabDirection(void) const
+        {
+        return getStringFromDirection(m_slab_direction);
+        }
+    std::string getFlowDirection(void) const
+        {
+        return getStringFromDirection(m_flow_direction);
+        }
```
b-butler, line 77: I would prefer `Python` or `String` suffix these functions to indicate that they are specifically for getting the Python associated string.

hoomd/md/MuellerPlatheFlow.h
```diff
@@ -55,81 +33,134 @@ class PYBIND11_EXPORT MuellerPlatheFlow : public Updater
...
+            return "Y";
+            }
+        else if (direction == flow_enum::Direction::Z)
+            {
+            return "Z";
+            }
+        else
+            {
+            throw std::runtime_error("Direction must be X, Y, or Z");
+            }
```
b-butler, line 96: should be lowercase (change needs to be made everywhere.

hoomd/md/update.py
b-butler: This has not been resolved.

hoomd/md/update.py
```diff
@@ -149,133 +150,134 @@ class mueller_plathe_flow:
...
+
         assert (n_slabs > 0), "Invalid negative number of slabs."
         if min_slab < 0:
             min_slab = 0
         if max_slab < 0:
             max_slab = n_slabs / 2
+        if max_slab <= -1 or max_slab > n_slabs:
+            raise ValueError("Invalid max_slab in [0," + str(n_slabs) + ").")
+        if min_slab <= -1 or min_slab > n_slabs:
+            raise ValueError("Invalid min_slab in [0," + str(n_slabs) + ").")
```
b-butler, line 240: We should include the value of `min_slab` and `max_slab` here, also use f-strings.

hoomd/md/update.py
```diff
@@ -149,133 +150,134 @@ class mueller_plathe_flow:
...
-    """Direction Enum Y for this class"""
+            self._cpp_obj = _md.MuellerPlatheFlowGPU(
+                sys_def, group, self.flow_target, self.slab_direction,
+                self.flow_direction, self.n_slabs, self.min_slab, self.max_slab,
+                self.flow_epsilon)
+        super()._attach()
 
-    Z = _md.MuellerPlatheFlow.Direction.Z
-    """Direction Enum Z for this class"""
+    @property
```
b-butler, line 277: Should this be loggable? If so, we should use `hoomd.logging.log` also, this should use a `DataAccessError` not return `None`. If using `hoomd.logging.log` this should be done by setting the `requires_run` flag to true.

hoomd/md/update.py
```diff
@@ -149,133 +150,134 @@ class mueller_plathe_flow:
...
+        # const integrated flow with 0.1 slope for max 1e8 timesteps
+        ramp = hoomd.variant.Ramp(0.0, 0.1e8, 0, int(1e8))
+        # velocity gradient in z direction and shear flow in x direction.
+        mpf = hoomd.md.update.ReversePerturbationFlow(filter=hoomd.filter.All(),
+                                                      flow_target=ramp,
+                                                      slab_direction="Z",
+                                                      flow_direction="X",
+                                                      n_slabs=20)
+
+        self._flow_target = hoomd.variant._setup_variant_input(flow_target)
```
b-butler, line 200: Users should not access or need to know about underscored attributes.

hoomd/md/update.py
```diff
@@ -130,8 +131,8 @@ def __init__(self, group, r=None, rx=None, ry=None, rz=None, P=(0, 0, 0)):
         self.metadata_fields = ['group', 'P', 'rx', 'ry', 'rz']
 
 
-class mueller_plathe_flow:
-    r"""Updater for a shear flow via an algorithm published by Mueller Plathe.
+class ReversePerturbationFlow(Updater):
+    """Reverse Perturbation (Müller-Plathe) method to establish shear flow.
```
b-butler, line 135: Why aren't we using a raw string here?

setup.cfg
```diff
@@ -19,6 +19,8 @@ ignore =
     RST201,RST203,RST301,
     # Allow :nowrap: in math directives
     RST307
+    # Support LaTex in docstrings
+    D301
```
b-butler, line 23: This does not support LaTeX, it is to enable escaped backslashes in non-raw strings. This could be to enable line wrapping for long type names in docstrings that require escaped backslashes. However, I am not sure if the new Python updater couldn't be a raw string here.

hoomd/md/update.py
```diff
@@ -149,133 +150,134 @@ class mueller_plathe_flow:
...
+            raise ValueError("Invalid max_slab in [0," + str(n_slabs) + ").")
+        if min_slab <= -1 or min_slab > n_slabs:
+            raise ValueError("Invalid min_slab in [0," + str(n_slabs) + ").")
+        if min_slab == max_slab:
+            raise ValueError("Invalid min/max slabs. Both have the same value.")
+
+        params = ParameterDict(filter=hoomd.filter.ParticleFilter,
+                               flow_target=hoomd.variant.Variant,
+                               slab_direction=str,
+                               flow_direction=str,
```
b-butler, line 247: Use `hoomd.data.typeconverter.OnlyTypes(str, strict=True, postprocessing=func)` where `func` returns the string but lowercase (to allow for `'X'` and `'x'`).

hoomd/md/pytest/test_reverse_perturbation_flow.py
```diff
@@ -0,0 +1,77 @@
...
+@pytest.mark.parametrize("slab_direction, flow_direction", _directions)
+def test_before_attaching(slab_direction, flow_direction):
+    filt = hoomd.filter.All()
+    ramp = hoomd.variant.Ramp(0.0, 0.1e8, 0, int(1e8))
+    n_slabs = 20
+    mpf = hoomd.md.update.ReversePerturbationFlow(filt, ramp, slab_direction,
+                                                  flow_direction, n_slabs)
+
+    assert mpf.filter is filt
+    assert mpf.flow_target is ramp
```
b-butler, line 17: filters and variants should both support equality.

hoomd/md/pytest/test_reverse_perturbation_flow.py
```diff
@@ -0,0 +1,77 @@
...
+    assert mpf.slab_direction is slab_direction
+    assert mpf.flow_direction is flow_direction
```
b-butler, line 19: Likewise, this should be an equality check, not an identity check.

hoomd/md/pytest/test_reverse_perturbation_flow.py
```diff
@@ -0,0 +1,77 @@
...
+    assert mpf.n_slabs == n_slabs
+    assert mpf.max_slab == n_slabs / 2
+    assert mpf.min_slab == 0
+    assert mpf.trigger == hoomd.trigger.Periodic(1)
+    assert mpf.summed_exchanged_momentum is None
```
b-butler, line 24: Change to a `pytest.raises` when coverted to `DataAccessError`.

hoomd/md/pytest/test_reverse_perturbation_flow.py
```diff
@@ -0,0 +1,77 @@
...
+    mpf = hoomd.md.update.ReversePerturbationFlow(filt, ramp, slab_direction,
+                                                  flow_direction, n_slabs)
+
+    nve = hoomd.md.methods.NVE(filter=hoomd.filter.All())
+    sim = simulation_factory(two_particle_snapshot_factory())
+    sim.operations.integrator = hoomd.md.Integrator(0.005, methods=[nve])
+    sim.operations.add(mpf)
+    sim.run(0)
+
+    assert mpf.flow_target is ramp
```
b-butler, line 43: Variants should support equality.

b-butler (changes requested):
I left some comments to address before merging. Thanks for all the work.

setup.cfg
joaander: This has come up on many PRs where the long text strings needed to produce the desired formatting conflicts with the Google docstring requirement that certain descriptions fit on one line. To work around this, I suggested that we use `\` as a line continuation character which unfortunately cannot be used in a raw string. Therefore, to enable LaTeX in strings that need this line continuations, I added this ignore. This is discussed in detail in #1001.

If you would like to go and reformat all the existing documentation that needs this and that on all open PRs so we can remove the D301 ignore, feel free to. The alternative presented in #1001 is to apply a noqa to the entire string which is far too heavy handed in my opinion. It would allow any line in the docstring to run over the line length limit which would result in many future style debates in PRs and would make reviewing diffs of the documentation on small screens difficult.

b-butler: Fine, by me then. I would just like the comment to describe more of why this supports LaTeX in conjunction with line continuation.

hoomd/md/update.py
SchoeniPhlippsn: I agree that this should be loggable. Then we don't have to worry about  `self._attached` here and let do `hoomd.logging.log` its job.

hoomd/md/update.py
SchoeniPhlippsn: I don't get this unit either. I am not perfectly familiar with Mueller-Plathe flow but ussually the units of shear flow are `N \cdot \mathrm{length}^{-1} = \mathrm{mass} \cdot \mathrm{time}^{-2}`.

SchoeniPhlippsn (changes requested):
Not a lot more comments here. However, the issues raised by @b-butler have to be addressed.

hoomd/md/update.py
danevans09: The formula in the description of flow_target does not render if a raw string is used

joaander (approved):
The latest changes look good to me. @danevans09 please re-request reviews from the other reviewers when this is ready for another round.

hoomd/md/update.py
danevans09: I think the units should either be `mass * length^(-1)` or `mass * time^(-1) * length^(-1)`.

I believe the `flow_target` argument is defining a target value for the _time-integrated momentum flux_, because in the paper where this method was implemented https://doi.org/10.1016/j.commatsci.2019.109107 they say:

> Therefore we have chosen to control the time-integrated momentum flux

Since momentum flux is in units of `mass * length * time^(-1) * length^(-2)` I think the time-integrated momentum flux should be in units of `delta_t * mass * time^(-1) * length^(-1)` or just `mass * length^(-1)`.

However, it looks like the code is expecting something in units of `mass * length * time^(-1) * length^(-2)` or `mass * time^(-1) * length^(-1)`. This is the line of code that is using flow_target:

`while (fabs((*m_flow_target)(timestep) - this->getSummedExchangedMomentum()/area) > this->getFlowEpsilon())`

where `getSummedExchangedMomentum()` returns `m_exchanged_momentum` which is found in this line:

`m_exchanged_momentum += sign * (m_last_max_vel.x - m_last_min_vel.x);`

where `m_last_max_vel` is actually storing the momentum, and `sign` is just +/- 1 based on how the min and max slabs are defined.

So it looks like the code is summing momenta and dividing the sum by an area, and then comparing this to `flow_target`. I don't think it is multiplying by `dt` during this sum.

SchoeniPhlippsn (approved):
Looks good to me. I don't have more comments and I think this PR is ready to be merged.

joaander:
@b-butler your change request is preventing the merge.

hoomd/md/update.py
```diff
@@ -130,152 +133,160 @@ def __init__(self, group, r=None, rx=None, ry=None, rz=None, P=(0, 0, 0)):
...
             min_slab = 0
         if max_slab < 0:
             max_slab = n_slabs / 2
+        if max_slab <= -1 or max_slab > n_slabs:
+            raise ValueError(f"Invalid max_slab of {max_slab}")
+        if min_slab <= -1 or min_slab > n_slabs:
+            raise ValueError(f"Invalid min_slab of {min_slab}")
+        if min_slab == max_slab:
+            raise ValueError(f"Min and max slab are equal. \
+                              min_slab = max_slab = {min_slab}")
```
b-butler, line 246: This validation would be helpful to have in the `ParameterDict`.

b-butler (commented):
I have one comment about validation in `__init__` rather than `self._param_dict`, but otherwise looks good.

b-butler (approved):

PR #952, opened 2021-03-10
Detect when m_comm is not set.

joaander:
## Description

Check on every call to `compute`, `update` and `analyze` that `m_comm` is set when needed.

## Motivation and context

In the current API for HOOMD, it is possible to instantiate and use many objects in MPI mode without calling `setCommunicator()`. When doing so, this can lead to seg faults (#929) or silent but incorrect results (#923). I don't have a good general solution to this issue. This PR is an effort to detect when `setCommunicator` calls are missing and notify the developer, but it is not foolproof as it relies on calling the base class methods. in `compute`, `update` and `analyze`.

We can discuss later if there is a better way to handle this. This PR is a stopgap fix.

## How has this been tested?

This PR should fail tests on submission as it finds a missing `setCommunicator` call in the neighbor list. It will start passing once #929 is merged.

## Change log

Internal changes only.

## Checklist:

### DIFF
--- a/example_plugin/example_plugin/ExampleUpdater.cc
+++ b/example_plugin/example_plugin/ExampleUpdater.cc
@@ -26,6 +26,7 @@ ExampleUpdater::ExampleUpdater(std::shared_ptr<SystemDefinition> sysdef)
 */
 void ExampleUpdater::update(uint64_t timestep)
     {
+    Updater::update(timestep);
     if (m_prof) m_prof->push("ExampleUpdater");
 
     // access the particle data for writing on the CPU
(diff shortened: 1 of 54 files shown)


joaander:
After thinking on this, I think we need an `isDomainDecomposed` method and use `if (m_sysdef->isDomainDecomposed)` in most places that need to check for domain decomposition. Then we can remove `m_comm` and `setCommunicator` from all the classes that don't need it. Those that do need to actually work with the `Communicator` object can get it from a central data structure. 

Until we implement those changes, this PR is needed to avoid bugs when `setCommunicator` is not called.

PR #954, opened 2021-03-10
Updated the test for write.GSD

Charlottez112:
## Description

Updated the test for ```hoomd.write.GSD```

Some problems I'm seeing:
* For ```hoomd.write.GSD.write```, the ```filename``` parameter only takes strings, but when ```hoomd.write.GSD``` is used as an operation, it takes paths. So we may want to improve this. 
* ```hoomd.write.GSD.write``` does not have the ```dynamic``` keyword, and it seems like the GSD file created does not store ```angles```. I tried adding a ```logger``` that logs the ```angle``` quantity, but it still gives the error ```AttributeError: 'HOOMDTrajectory' object has no attribute 'angles'```.
* Another thing I'm not sure about:   
  Originally I defined two parts for the ```filter``` test (first part tested with ```filter.Null```, the second part with ```filter.Type```), and used ```filename = tmp_path / "temporary_test_file.gsd"``` for the GSD filename for both parts, and set the ```mode``` to ```wb```.  
 But it seems like when the second part of the test was running, it wasn't able to overwrite the file created by the first part of the test, and gave me the error that the filename was invalid. So I made two separate tests for ```filter```. I'm wondering if this is because we cannot use temprary files in this way.

## Checklist:


hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+
+    sim = simulation_factory(hoomd_snapshot)
+    hoomd.write.GSD.write(state=sim.state, mode='wb',
+                          filename=str(filename))
+
+    with gsd.hoomd.open(name=filename, mode='wb') as traj:
+        assert_equivalent_snapshots(traj, hoomd_snapshot)
+
+
+def test_write_gsd_trigger(simulation_factory, device,
```
Charlottez112, line 78: @joaander I think you mentioned that we could check some additional stuff for the trigger test, but I don't remember what exactly it was. Could you let me know if you still remember?

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def test_write(simulation_factory, device,
+               hoomd_snapshot, tmp_path):
+
+    filename = tmp_path / "temporary_test_file.gsd"
+
+    sim = simulation_factory(hoomd_snapshot)
+    hoomd.write.GSD.write(state=sim.state, mode='wb',
+                          filename=str(filename))
```
b-butler, line 72: This won't write out anything besides particle positions and orientations. If you want to write out all the data you need to change the `dynamic` setting... which currently doesn't exist for `write`. If you could add the capability for `hoomd.write.GSD.write` to take in a `dynamic` argument like the constructor does that would be great, and make this test possible.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+                                 trigger=hoomd.trigger.Periodic(1),
+                                 mode='wb', dynamic=['momentum'])
+    sim.operations.writers.append(gsd_writer)
+
+    snapshot_list = []
+    for _ in range(5):
+        sim.run(1)
+        snapshot_list.append(sim.state.snapshot)
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
```
b-butler, line 105: You should add a length check after this rather than assume they are the same length `assert len(traj) = len(snapshot_list)`.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+
+    filename = tmp_path / "temporary_test_file.gsd"
+
+    sim = simulation_factory(hoomd_snapshot)
+
+    integrator = hoomd.md.Integrator(dt=0.005)
+    lj = hoomd.md.pair.LJ(nlist=hoomd.md.nlist.Cell(), r_cut=2.5)
+    lj.params[('p1', 'p1')] = {'sigma': 1, 'epsilon': 5e-200}
+    lj.params[('p1', 'p2')] = {'sigma': 1, 'epsilon': 5e-200}
+    lj.params[('p2', 'p2')] = {'sigma': 1, 'epsilon': 5e-200}
```
b-butler, line 89:
```suggestion
    lj.params.default = {'sigma': 1, 'epsilon': 5e-200}
```

b-butler: Also why is the number so small?..

hoomd/pytest/test_gsd.py
b-butler, line 78: This is really a test on `hoomd.write.GSD` working in running simulations, not just the trigger.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    integrator.forces.append(lj)
+    langevin = hoomd.md.methods.Langevin(hoomd.filter.All(), kT=1, seed=1)
+    integrator.methods.append(langevin)
+    sim.operations.integrator = integrator
```
b-butler, line 125: This code is identical to the code above, could you create a fixture that is the integrator, or one that returns a simulation with the integrator added?

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    sim.operations.writers.append(gsd_writer)
+
+    snapshot_list = []
+    for _ in range(5):
+        sim.run(1)
+        snapshot_list.append(sim.state.snapshot)
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
+        for step in range(5):
+            assert_equivalent_snapshots(traj[step + 2], snapshot_list[step])
```
b-butler, line 159:
```suggestion
        for gsd_snap, hoomd_snap in zip(traj[2:], snapshot_list):
            assert_equivalent_snapshots(gsd_snap, hoomd_snap)
```

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    lj.params[('p2', 'p2')] = {'sigma': 1, 'epsilon': 5e-200}
+    integrator.forces.append(lj)
+    langevin = hoomd.md.methods.Langevin(hoomd.filter.All(), kT=1, seed=1)
+    integrator.methods.append(langevin)
+    sim.operations.integrator = integrator
+
+    gsd_writer = hoomd.write.GSD(filename=filename,
+                                 trigger=hoomd.trigger.Periodic(1),
+                                 mode='ab', dynamic=['momentum'])
+    sim.operations.writers.append(gsd_writer)
```
b-butler, line 150:
```suggestion
    # test mode=ab
    sim.writers.clear()
    gsd_writer = hoomd.write.GSD(filename=filename,
                                 trigger=hoomd.trigger.Periodic(1),
                                 mode='ab', dynamic=['momentum'])
    sim.operations.writers.append(gsd_writer)
```

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    langevin = hoomd.md.methods.Langevin(hoomd.filter.All(), kT=1, seed=1)
+    integrator.methods.append(langevin)
+    sim.operations.integrator = integrator
+
+    gsd_writer = hoomd.write.GSD(filename=filename,
+                                 trigger=hoomd.trigger.Periodic(1),
+                                 mode='wb')
+    sim.operations.writers.append(gsd_writer)
+
+    sim.run(2)
```
b-butler, line 132: Explain why we run these two steps (so we can check 'ab').

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+        snapshot_list.append(sim.state.snapshot)
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
+        for step in range(5):
+            assert_equivalent_snapshots(traj[step + 2], snapshot_list[step])
+
+    # test mode=xb exception
+    sim = simulation_factory(hoomd_snapshot)
+
+    integrator = hoomd.md.Integrator(dt=0.005)
```
b-butler, line 164: See above suggestion.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    snapshot_list = []
+    for _ in range(5):
+        sim.run(1)
+        snapshot_list.append(sim.state.snapshot)
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
+        for step in range(5):
+            assert_equivalent_snapshots(traj[step + 2], snapshot_list[step])
+
+    # test mode=xb exception
```
b-butler, line 161: Explain in more detail about how we are testing for failure when the file exists.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    gsd_writer = hoomd.write.GSD(filename=filename_xb,
+                                 trigger=hoomd.trigger.Periodic(1),
+                                 mode='xb', dynamic=['momentum'])
+    sim.operations.writers.append(gsd_writer)
+
+    sim.run(5)
+
+    with gsd.hoomd.open(name=filename_xb, mode='rb') as traj:
+        for step in range(5):
+            assert_equivalent_snapshots(traj[step], snapshot_list[step])
```
b-butler, line 205:
```suggestion
        assert len(traj) == len(snapshot_list)
        for gsd_snap, hoomd_snap in zip(traj, snapshot_list):
            assert_equivalent_snapshots(gsd_snap, hoomd_snap)
```

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+
+    integrator = hoomd.md.Integrator(dt=0.005)
+    lj = hoomd.md.pair.LJ(nlist=hoomd.md.nlist.Cell(), r_cut=2.5)
+    lj.params[('p1', 'p1')] = {'sigma': 1, 'epsilon': 5e-200}
+    lj.params[('p1', 'p2')] = {'sigma': 1, 'epsilon': 5e-200}
+    lj.params[('p2', 'p2')] = {'sigma': 1, 'epsilon': 5e-200}
+    integrator.forces.append(lj)
+    langevin = hoomd.md.methods.Langevin(hoomd.filter.All(), kT=1, seed=1)
+    integrator.methods.append(langevin)
+    sim.operations.integrator = integrator
```
b-butler, line 224: See above comment.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+                                 trigger=hoomd.trigger.Periodic(1),
+                                 filter=hoomd.filter.Null(),
+                                 mode='wb', dynamic=['momentum'])
+    sim.operations.writers.append(gsd_writer)
+
+    sim.run(3)
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
+        for step in range(3):
+            assert traj[step].particles.N == 0
```
b-butler, line 236:
```suggestion
        for frame in traj:
            assert frame.particles.N == 0
```

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+def test_write_gsd_type_filter(simulation_factory, device,
```
b-butler, line 239: I don't think this test is necessary, we already test filters elsewhere. but if you think we should keep this @joaander we can.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+
+    integrator = hoomd.md.Integrator(dt=0.005)
+    lj = hoomd.md.pair.LJ(nlist=hoomd.md.nlist.Cell(), r_cut=2.5)
+    lj.params[('p1', 'p1')] = {'sigma': 1, 'epsilon': 5e-200}
+    lj.params[('p1', 'p2')] = {'sigma': 1, 'epsilon': 5e-200}
+    lj.params[('p2', 'p2')] = {'sigma': 1, 'epsilon': 5e-200}
+    integrator.forces.append(lj)
+    langevin = hoomd.md.methods.Langevin(hoomd.filter.All(), kT=1, seed=1)
+    integrator.methods.append(langevin)
+    sim.operations.integrator = integrator
```
b-butler, line 287: See above comment.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    sim.operations.writers.append(gsd_writer)
+
+    snapshot_list = []
+    for _ in range(2):
+        sim.run(1)
+        snapshot_list.append(sim.state.snapshot)
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
+        assert_equivalent_snapshots(traj[0], snapshot_list[-1])
+        assert_equivalent_snapshots(traj[-1], snapshot_list[-1])
```
b-butler, line 302:
```suggestion
        for gsd_snap in traj:
            assert_equivalent_snapshots(traj, snapshot)
```

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    gsd_writer = hoomd.write.GSD(filename=filename,
+                                 trigger=hoomd.trigger.Periodic(1),
+                                 truncate=True,
+                                 mode='wb')
+    sim.operations.writers.append(gsd_writer)
+
+    snapshot_list = []
+    for _ in range(2):
+        sim.run(1)
+        snapshot_list.append(sim.state.snapshot)
```
b-butler, line 298:
```suggestion
    sim.run(2)
    snapshot = sim.state.snapshot
```

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    lj.params[('p1', 'p2')] = {'sigma': 1, 'epsilon': 5e-200}
+    lj.params[('p2', 'p2')] = {'sigma': 1, 'epsilon': 5e-200}
+    integrator.forces.append(lj)
+    langevin = hoomd.md.methods.Langevin(hoomd.filter.All(), kT=1, seed=1)
+    integrator.methods.append(langevin)
+    sim.operations.integrator = integrator
+
+    gsd_writer = hoomd.write.GSD(filename=filename,
+                                 trigger=hoomd.trigger.Periodic(1),
+                                 mode='wb', dynamic=['momentum'])
```
b-butler, line 324: We could have a base `GSD` writer fixture since we use this exact code so much.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    for _ in range(2):
+        sim.run(1)
+        snapshot_list.append(sim.state.snapshot)
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
+        assert_equivalent_snapshots(traj[0], snapshot_list[-1])
+        assert_equivalent_snapshots(traj[-1], snapshot_list[-1])
+
+
+def test_write_gsd_dynamic(simulation_factory, device,
```
b-butler, line 305: We should test that all the dynamic options work, and that when they are not chosen that the corresponding values don't get written. Also, this would likely be cleaner using a parametrized fixture or `pytest.mark.parametrize`.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+    sim.operations.writers.append(gsd_writer)
+
+    velocity_list = []
+    for _ in range(5):
+        sim.run(1)
+        snap = sim.state.snapshot
+        velocity_list.append(snap.particles.velocity)
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
+        for step in range(5):
```
b-butler, line 334: Use zip and a length check here.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,378 @@
...
+                                 log=logger)
+    sim.operations.writers.append(gsd_writer)
+
+    kinetic_energy_list = []
+    for _ in range(5):
+        sim.run(1)
+        kinetic_energy_list.append(thermo.kinetic_energy)
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
+        for s in range(5):
```
b-butler, line 376: use zip here.

b-butler (changes requested):
Thanks @Charlottez112 for the PR. I left some comments about the tests and how to improve them and fix the errors.

hoomd/pytest/test_snapshot.py
```diff
@@ -48,6 +48,8 @@ def assert_equivalent_snapshots(gsd_snap, hoomd_snap):
                     assert y == []
                 else:
                     assert x == y
+            elif prop == 'acceleration' or prop == 'is_accel_set':
```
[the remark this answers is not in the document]
Charlottez112: Currently we are not able to log acceleration by using either  ```log``` or ```dynamic```

hoomd/pytest/test_gsd.py
Charlottez112: Yeah this answers my second question about this test. I can try to do that.

hoomd/pytest/test_gsd.py
Charlottez112: I copied these several lines from somewhere and didn't really think too much about it...

hoomd/pytest/test_gsd.py
Charlottez112: Good idea. I realize now I copied too much

hoomd/pytest/test_gsd.py
Charlottez112: Yeah I agree, but I'm not sure I completely understood what you mean. Do you mean that we also need to add a test that uses a different trigger?

hoomd/pytest/test_gsd.py
Charlottez112: Although these several lines are very similar in each test, we change different arguments almost every time we use it. We could define a function instead, but it would be similar to what we have now.

hoomd/pytest/test_snapshot.py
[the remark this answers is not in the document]
joaander: GSD files do not store acceleration. It is a calculated quantity, not part of the system configuration. I honestly have no idea how this `assert_` method worked previously without these checks.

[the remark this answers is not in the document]
b-butler: I just assumed that it went with `'momentum'`, but it seems it does not. Is there a reason we only store the velocity and angular momentum even if acceleration is set? It is stored in `ParticleData`.

hoomd/pytest/test_gsd.py
b-butler: I am not sure this is the case, we do that many times, but this is used fairly frequently. If you don't feel that that would improve clarity, then leave it as is, a function would obfuscate the code.

hoomd/pytest/test_gsd.py
b-butler: That is fine, I was just more curious since this is effectively a 0 potential.

b-butler: If that is what you want, you can remove the pair potential entirely.

Charlottez112: Yeah I realized that too and I changed the value for ```epsilon``` to 1.

hoomd/pytest/test_gsd.py
Charlottez112: The exact piece of code was used twice. At other places, we either used different ```mode``` or different ```filter``` etc. I'm not sure how to make a fixture that we can easily change the values passed to these arguments.

hoomd/pytest/test_gsd.py
Charlottez112: Or more specifically, check that when the dynamic option is not enabled, only dynamic quantities are written, static ones are only written in the first frame? In the test I wrote, would velocity still be stored in every frame even if the ```dynamic =None```?

b-butler: I think testing that only `property` attributes get written by default, and when another `dynamic` option is chosen that it adds the correct fields, and when it is not chosen that those fields are not written.

hoomd/pytest/test_gsd.py
b-butler: Then that is fine. Leave it as is.

hoomd/pytest/test_gsd.py
Charlottez112: I'm a little confused about what the doc says, so if ```dynamic``` is set to default, only ```property``` attributes are written and not other attributes that are also changing in each frame? If another```dynamic``` option is chosen, the corresponding fields will be added, but still if one of the newly added attributes is static, it's written only in the first frame?

hoomd/pytest/test_gsd.py
Charlottez112: In addition to whether to keep this test, I have a question about overwriting the same gsd file written with a different filter. Like I mentioned in the description of this PR, I was trying to test both two filters in one test session, but somehow the second ```gsd_writer``` with the type filter cannot overwrite the gsd file created by the ```gsd_writer``` with the null filter. I'm wondering if this is universally true with filters or it's just because I used the null filter first.

joaander: Are you using the 'xb' mode? That will raise an error if the file exists. Use the 'wb' mode to overwrite files.

Charlottez112: I am using the 'wb' mode, and that's why I'm a little confused. I'm seeing a ```RuntimeError``` which says that the filename I provided is an invalid argument. The filename I provided is exactly the file created by the first ```gsd_writer```. When I changed this filename to a new one, it worked.

hoomd/pytest/test_gsd.py
joaander: For example, say you have a simulation where positions and velocities change every frame, but particle orientations stay [1, 0, 0, 0].
With `dynamic=['property']`: `position` and `velocity` will be written on the first frame. When you read the file, `position` will be read for each frame. When you read the file and access `velocity` at frame i (i != 0), GSD will provide the velocity from the first frame. This option exists to save disk space. Users need to opt-in to writing out additional quantities every frame.

With `dynamic=['property', 'momentum']`: `position` and `velocity` will be written on every frame. When you read the file, `position` and `velocity` will both be read for each frame.

In both cases, `orientation` will not be written at all. GSD understands that the default orientation is [1, 0, 0, 0] and will provide that per-particle value on read even though the data is not in the file. This behavior is to save disk space.

hoomd/pytest/test_gsd.py
joaander: A test that the trigger works properly doesn't need to check that all snapshot frames are equivalent. So, you don't need to spend a lot of testing code on generating different configurations. 

The trigger controls which timesteps are written, so you could have a static simulation with only a few particles, no forces, no integrator. Specify a non-trivial trigger to GSD, say Periodic(period=10, phase=5), then read in the GSD file and validate that each frame's step matches: e.g. 
```
assert [frame.configuration.step for frame in traj] = [5, 15, 25]
```

hoomd/pytest/test_gsd.py
joaander: `dynamic` doesn't apply with `GSD.write` - which only writes out one frame to a file. GSD always writes out all non-default quantities to the first frame.

hoomd/pytest/test_gsd.py
joaander: I think it is fine to remove the type filter test. You have the null filter test and all the other tests implicitly check the all filter. That's enough to demonstrate that the `filter` argument has an effect.

joaander (commented):
This is looking excellent. I merged master up into this branch and fixed the one failing test.

You just need to address all @b-butler's comments and then we can merge this.

hoomd/pytest/test_gsd.py
```diff
@@ -0,0 +1,382 @@
...
+               hoomd_snapshot, tmp_path):
+
+    filename = tmp_path / "temporary_test_file.gsd"
+
+    sim = simulation_factory(hoomd_snapshot)
+    hoomd.write.GSD.write(state=sim.state, mode='wb',
+                          filename=str(filename))
+
+    with gsd.hoomd.open(name=filename, mode='rb') as traj:
+        assert_equivalent_snapshots(traj[0], hoomd_snapshot)
```
joaander, line 79:
```suggestion
    if hoomd_snapshot.exists:
        with gsd.hoomd.open(name=filename, mode='rb') as traj:
            assert len(traj) == 1
            assert_equivalent_snapshots(traj[0], hoomd_snapshot)
```

hoomd/pytest/test_gsd.py
Charlottez112: I still have some questions about testing ```dynamic```, especially about ```dynamic=['attribute']``` and ```dynamic=['topology']```. For these, do we need to make simulations where the corresponding quantities change to test them? And for the cases where the quantities are only stored in the first frame, do we need to test that they're not written in every frame, but rather read from the first frame?

hoomd/pytest/test_snapshot.py
```diff
@@ -48,6 +48,8 @@ def assert_equivalent_snapshots(gsd_snap, hoomd_snap):
...
+                next
```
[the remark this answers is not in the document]
Charlottez112: Could you explain a bit more about this suggestion? I thought we added this line to skip the check for acceleration. I'm not sure I understand how to change this.

hoomd/pytest/test_gsd.py
joaander: Ideally, yes. You would need to change e.g. the bonds in the system or particle types from one frame to the next in order to test that these dynamic flags function. The GSD python library does not indicate whether a particular quantity is given in every frame or read from the fist frame. Instead, you have to inspect the frame and see if it hast the expected contents (e.g. that the typeid still match those set in frame 0 even though you are looking at frame 1).

Instead of using `assert_equivalent_snapshots` for this, you can use targeted assert checks to verify the expected data. e.g.
```
# attributes static
assert traj[0].typeid == [0, 0]
assert traj[1].typeid == [0, 0]

# attributes dynamic
assert traj[0].typeid == [0, 0]
assert traj[1].typeid == [1, 0]
```

hoomd/pytest/test_snapshot.py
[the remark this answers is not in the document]
joaander: Replace `next` with `continue`. `continue` is the proper keyword to continue on to the next iteration in the containing loop.

joaander:
I pushed changes that 
1) fix the segfaults (we were assigning type ids that were invalid)
2) Moves `test_gsd.py` to the `hoomd.md/pytest` directory because it relies on the `md` module being compiled.
3) Makes all the tests run in serial only.

I tried to get the tests to run in MPI but ran into a variety of issues, including file open exceptions thrown only on the root rank and a number of others. I don't think we gain much by ensuring that all of these tests can run in both MPI and serial.

hoomd/md/pytest/test_gsd.py
```diff
@@ -0,0 +1,370 @@
...
+
+    # test mode=ab
+    sim.operations.writers.clear()
+
+    gsd_writer = hoomd.write.GSD(filename=filename,
+                                 trigger=hoomd.trigger.Periodic(1),
+                                 mode='ab', dynamic=['momentum'])
+    sim.operations.writers.append(gsd_writer)
+
+    snap_list = []
```
joaander, line 145: This method has both `snap_list` and `snapshot_list`. It does make the code confusing to read and understand.

joaander (approved):
Thanks! These tests are very thorough. While I think we could improve the code clarity, It is good enough and I don't want to delay this PR any further.

joaander:
Never mind on the serial changes. Once I resolved the other issues, I was able to get the tests to run in MPI.

hoomd/md/pytest/test_gsd.py
Charlottez112: Thank you for the suggestion!

Charlottez112:
@b-butler Hey can you take another look? I can improve the clarity as well if that's needed

hoomd/md/pytest/test_gsd.py
```diff
@@ -0,0 +1,369 @@
...
+    lj.params.default = {'sigma': 1, 'epsilon': 1}
+    integrator.forces.append(lj)
+    langevin = hoomd.md.methods.Langevin(hoomd.filter.All(), kT=1)
+    integrator.methods.append(langevin)
+
+    return integrator
+
+
+@pytest.fixture(scope='function')
+def create_md_sim(simulation_factory, device, hoomd_snapshot):
```
b-butler, line 82: I would just name this `simulation` and not bother with the `sim = create_md_sim` lines in the tests.

b-butler (approved):
There are additional changes that could be made, but the tests do test what they are supposed to, and in the interest of getting this merged, I am going to approve.

Thanks @Charlottez112!

PR #956, opened 2021-03-15
Pass communicator down to embedded cell lists

joaander:
## Description

Pass the `Communicator` from the neighbor lists down to the embedded cell list objects.

## Motivation and context

This fixes a bug where the neighbor list would compute incorrect neighbors or seg fault in MPI simulations.

This bug was found on #929 . I cherry picked the fix here so we can git it released in beta.5.

## How has this been tested?

The fix has been tested on #929.

## Change log

```
*Fixed*

- Incorrect MD neighbor lists in MPI simulations with more than 1 rank.
```

## Checklist:

### DIFF
--- a/hoomd/md/NeighborListGPUBinned.h
+++ b/hoomd/md/NeighborListGPUBinned.h
@@ -76,6 +76,15 @@ class PYBIND11_EXPORT NeighborListGPUBinned : public NeighborListGPU
             return m_cl->getSortCellList();
             }
 
+        virtual void setCommunicator(std::shared_ptr<Communicator> comm)
+            {
+            // call base class method
+            NeighborList::setCommunicator(comm);
+
+            // set the communicator on the internal cell lists
+            m_cl->setCommunicator(comm);
+            }
+
     protected:
         std::shared_ptr<CellList> m_cl;   //!< The cell list
         unsigned int m_block_size;          //!< Block size to execute on the GPU
--- a/hoomd/md/NeighborListGPUBinned.h
+++ b/hoomd/md/NeighborListGPUBinned.h
@@ -76,6 +76,8 @@ class PYBIND11_EXPORT NeighborListGPUBinned : public NeighborListGPU
             return m_cl->getSortCellList();
             }
 
+        #ifdef ENABLE_MPI
+
         virtual void setCommunicator(std::shared_ptr<Communicator> comm)
             {
             // call base class method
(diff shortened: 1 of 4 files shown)


hoomd/md/NeighborListGPUBinned.h
```diff
@@ -76,6 +76,19 @@ class PYBIND11_EXPORT NeighborListGPUBinned : public NeighborListGPU
...
+            {
+            // call base class method
+            NeighborList::setCommunicator(comm);
+
+            // set the communicator on the internal cell lists
+            m_cl->setCommunicator(comm);
+            }
+
+        #endif
+
```
tommy-waltmann, line 91: This seems to be the same implementation as `NeighborListBinned`, so we should remove it.

hoomd/md/NeighborListGPUStencil.h
```diff
@@ -69,6 +69,20 @@ class PYBIND11_EXPORT NeighborListGPUStencil : public NeighborListGPU
...
+            {
+            // call base class method
+            NeighborList::setCommunicator(comm);
+
+            // set the communicator on the internal cell lists
+            m_cl->setCommunicator(comm);
+            m_cls->setCommunicator(comm);
+            }
+
+        #endif
```
tommy-waltmann, line 84: Same as above comment. This implementation is the same as in `NeighborListStencil`

tommy-waltmann (changes requested):

hoomd/md/NeighborListGPUBinned.h
joaander: `NeighborListGPUBinned` is not a subclass of `NeighborListBinned`, so `setCommunicator` is not inherited from it.
https://github.com/glotzerlab/hoomd-blue/blob/e8eeb42ecd11d5f36b8e2221c0c0ee79e34385d8/hoomd/md/NeighborListGPUBinned.h#L31

hoomd/md/NeighborListGPUStencil.h
joaander: `NeighborListGPUStencil` is not a subclass of `NeighborListStencil`, so `setCommunicator` is not inherited from it.
https://github.com/glotzerlab/hoomd-blue/blob/e8eeb42ecd11d5f36b8e2221c0c0ee79e34385d8/hoomd/md/NeighborListGPUStencil.h#L32

tommy-waltmann (approved):

b-butler (approved):
@joaander thanks for the fix. :smiley:

PR #957, opened 2021-03-18
Support TBB 2021

### LINKED ISSUE #640: Use TBB task arenas
## Description

HOOMD currently relies on the deprecated TBB feature `task_scheduler_init`. This needs to be refactored for HOOMD v3 to use task arenas instead of the deprecated feature. This is also necessary to enable multiple CPU device objects with different numbers of threads.

One solution is to set the number of threads in the `ExecutionConfiguration` and use a task arena to selectively limit parallel loops to the given number of cores. The task arena approach has been successfully implemented in the `fresnel` package:
- https://github.com/glotzerlab/fresnel/blob/8d27bf93c192a351e228f6ef8b2a22f4bac99be2/fresnel/cpu/TracerPath.cc#L49
- https://github.com/glotzerlab/fresnel/blob/8d27bf93c192a351e228f6ef8b2a22f4bac99be2/fresnel/cpu/TracerPath.cc#L77

## Proposed solution

- [ ] Manage number of threads in `ExecutionConfiguration`.
- [ ] Wrap internal uses of TBB-parallel algorithms with the task arena executor.
- [ ] Remove references to deprecated `task_scheduler_init`.
- [ ] Update and test Python Device API for TBB threading.

joaander:
## Description

Enable support for TBB 2021. Implement this with an object based solution. `ExecutionConfiguration` holds a `task_arena` which TBB enabled computations use to limit the number of threads they use to execute.

## Motivation and context

TBB 2021 removes deprecated functionality. Also, prior the object based solution, only the most recent call to setNumThreads was effective. The object based solution allows multiple `Simulation` objects to coexist and use a different number of threads.

As long as other libraries that users import also use the object based solution, HOOMD will interoperate with them. Those that use `global_control` may still interfere with the number of threads that HOOMD uses.

Resolves #640 

## How has this been tested?

I build locally against TBB 2020 and TBB 2021. The Linux CI tests still use TBB 2020 as that is the package that ubuntu provides. I updated the Mac tests to use TBB 2021 for coverage.

## Change log

```
* Support TBB 2021.
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/UpdaterClusters.h
+++ b/hoomd/hpmc/UpdaterClusters.h
@@ -24,6 +24,11 @@
 #include <tbb/parallel_for.h>
 #include <tbb/task.h>
 #include <atomic>
+
+#if TBB_VERSION_MAJOR < 2021
+#define ENABLE_TBB_TASK
+#endif
+
 #endif
 
 namespace hpmc
@@ -78,7 +83,9 @@ class my_atomic_flag
 class Graph
     {
     public:
-        Graph() {}      //!< Default constructor
+        Graph(std::shared_ptr<tbb::task_arena> task_arena) : m_task_arena(task_arena)
+            {
+            }
 
         inline Graph(unsigned int V);   // Constructor
 
@@ -104,6 +111,9 @@ class Graph
         #else
         std::vector<my_atomic_flag> visited;
         #endif
+
+        /// The TBB task arena
+        std::shared_ptr<tbb::task_arena> m_task_arena;
 
         #ifndef ENABLE_TBB_TASK
         // A function used by DFS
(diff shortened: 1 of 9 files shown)


joaander:
UpdaterClusters will take more work to port. We can do that in a future PR: #958

joaander:
@bdice - you were interested to see how the object based solution to TBB thread limits looks. You can see that in this PR. It involves wrapping all parallel operations to be limited in a `this->m_exec_conf->getTaskArena()->execute([&]{ }`.

hoomd/hpmc/UpdaterClusters.h
```diff
@@ -24,6 +24,11 @@
 #include <tbb/parallel_for.h>
 #include <tbb/task.h>
 #include <atomic>
+
+#if TBB_VERSION_MAJOR < 2021
+#define ENABLE_TBB_TASK
+#endif
+
```
b-butler, line 31: Why do we check the version here, but not elsewhere?

b-butler (commented):
In general it looks fine to me, but I am not sure about some of the lines in the cluster updater compared to the other files.

hoomd/hpmc/UpdaterClusters.h
joaander: UpdaterClusters uses `tbb::task`, which doesn't have a direct replacement in TBB2021. This code will need a more extensive rewrite to support TBB 2021: #958

b-butler (approved):
LGTM

PR #964, opened 2021-03-25
Newtonian event chain monte carlo

joaander:
## Description

Add newtonian event chain Monte Carlo integrators. This implementation is by @maklem .

## Motivation and context

The NEC algorithm can be more efficient than local trial moves. 

## How has this been tested?

There is a validation test. Additional unit tests may be needed.

## Change log

```
`hoomd.hpmc.integrate_nec.*` - Newtonian Event Chain Monte Carlo for hard spheres and convex polyhedron
```

## Checklist:


hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,283 @@
...
+                                (self._min_chain_time, self.max_chain_time))
+        if move_definition not in tune_definitions:
+            self._tunables.append(move_definition)
+
+    def _update_tunables_attr(self, attr, value):
+        for tunable in self._tunables:
+            setattr(tunable, attr, value)
+
+
+class ChainTime(_InternalCustomTuner):
```
b-butler, line 197: Import this directly into `hpmc.tune`.

b-butler (changes requested):
I gave the code a look, and left comments where I thought necessary. For the Python code please run the new files through the `yapf` formatter using the settings for the package that we have in the base directory. Also, I tried to find commented out code everywhere, but I might have missed a few places, please remove any others you can find.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1379 @@
...
+//#include <cfloat>
+
+/*! \file IntegratorHPMCMonoNEC.h
+    \brief Defines the template class for HPMC with Newtonian event chains
+    \note This header cannot be compiled by nvcc
+*/
+
+#ifdef NVCC
+#error This header cannot be compiled by nvcc
+#endif
```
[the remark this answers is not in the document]
b-butler: @joaander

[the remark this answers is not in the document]
SchoeniPhlippsn: Yes, I think it should be `#ifdef __HIPCC__`

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1368 @@
...
+*/
+template< class Shape >
+class IntegratorHPMCMonoNEC : public IntegratorHPMCMono<Shape>
+    {
+    protected:
+        Scalar m_chain_time;        //!< the length of a chain, given in units of time
+        unsigned int m_chain_probability; //!< how often we do a chain. Replaces translation_move_probability
+        Scalar m_update_fraction;   //!< if we perform chains we update several particles as one
+
+        // GlobalArray< hpmc_counters_t >     m_count_total;      // Inherited from base class
```
SchoeniPhlippsn, line 42: Remove
```suggestion

```

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1368 @@
...
+            return m_chain_time;
+            }
+
+        //! Change move ratio
+        /*! \param m_chain_time set duration of a chain
+        */
+        void setChainProbability(Scalar chain_probability)
+            {
+            this->m_exec_conf->msg->notice(10) << "IntegratorHPMCMonoNEC<Shape>::setChainProbability(" << chain_probability << ")" << std::endl;
+            m_chain_probability = static_cast<unsigned int>(chain_probability * 65536.0);
```
SchoeniPhlippsn, line 102: Is there a specific reasoning behind this specific factor?

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1368 @@
...
+    // compute the width of the active region
+    Scalar3 npd = box.getNearestPlaneDistance();
+    Scalar3 ghost_fraction = this->m_nominal_width / npd;
+
+    vec3<Scalar> lattice_x = vec3<Scalar>(box.getLatticeVector(0));
+    vec3<Scalar> lattice_y = vec3<Scalar>(box.getLatticeVector(1));
+    vec3<Scalar> lattice_z = vec3<Scalar>(box.getLatticeVector(2));
+
+    vec3<Scalar> normal_x = cross(lattice_y,lattice_z);
+    normal_x /= sqrt(dot(normal_x,normal_x));
```
SchoeniPhlippsn, line 372: Please the sqrt functions written for Hoomd in `hoomd/HOOMDMath.h`
 ```suggestion
    normal_x *= fast::rsqrt(dot(normal_x,normal_x));
```

File: hoomd/hpmc/IntegratorHPMCMonoNEC.h [hunk 6]
```diff
@@ -0,0 +1,1368 @@
...
+                        // collide with walls
+                        vec3<Scalar> vel_k = vec3<Scalar>(h_velocities.data[k]);
+
+                        vec3<Scalar> vel_change = collisionPlaneVector * (dot(vel_k,collisionPlaneVector) / dot(collisionPlaneVector,collisionPlaneVector));
+                        vel_k = vel_k - 2.0 * vel_change;
+
+                        h_velocities.data[k]    = make_scalar4( vel_k.x, vel_k.y, vel_k.z, h_velocities.data[k].w);
+                        next = k;
+
+                        velocity = sqrt( dot(vel_k,vel_k));
```
SchoeniPhlippsn, line 695:
```suggestion
                        velocity = fast::sqrt( dot(vel_k,vel_k));
```

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1368 @@
...
+                Scalar4 orientation_i = h_orientation.data[i];
+                vec3<Scalar> pos_i    = vec3<Scalar>(postype_i);
+                Shape shape_old(quat<Scalar>(orientation_i), this->m_params[typ_i]);
+
+
+                //move_rotate(shape_i.orientation, rng_i, h_a.data[typ_i], ndim);
+                if (ndim == 2)
+                    move_rotate<2>(shape_i.orientation, rng_i, h_a.data[typ_i]);
+                else
+                    move_rotate<3>(shape_i.orientation, rng_i, h_a.data[typ_i]);
```
SchoeniPhlippsn, line 787:
```suggestion
                move_rotate<ndim>(shape_i.orientation, rng_i, h_a.data[typ_i]);
```

hoomd/hpmc/XenoSweep3D.h
```diff
@@ -0,0 +1,235 @@
...
+                v3 = v4;                   // Inside v1 & outside v2 ==> eliminate v3
+            }
+        else
+            {
+            if (dot(v3, x) > OverlapReal(0.0))
+                v2 = v4;    // Outside v1 & inside v3 ==> eliminate v2
+            else
+                v1 = v4;                   // Outside v1 & outside v3 ==> eliminate v1
+            }
+
```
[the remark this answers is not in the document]
SchoeniPhlippsn: For readability I would set an exit at `if (count >= XENOCOLLIDE_3D_MAX_ITERATIONS)` and write that return statement outside the while loop. Even though you have to break it in the middle, it would be even nicer to use it as a condition for the while loop instead of the infinite loop over true (both for Phase 1 and 2).

hoomd/hpmc/IntegratorHPMCMonoNEC.h
maklem: The chain-probability is the chain equivalent to translation-move-probability in local update HPMC and defined the same way.

https://github.com/glotzerlab/hoomd-blue/blob/96d089d297770d1fade62d9b6f7a60c7340cbd6e/hoomd/hpmc/IntegratorHPMC.h#L313-L316

hoomd/hpmc/IntegratorHPMCMonoNEC.h
maklem: with that change I get an error:
`error: the value of ‘ndim’ is not usable in a constant expression`

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1379 @@
...
+            {
+            return (1+count_pressurevirial/count_movelength)*this->m_pdata->getN()/this->m_pdata->getBox().getVolume();
+            }
+
+
+
+        //! Get the value of a logged quantity
+        virtual Scalar getLogValue(const std::string& quantity, unsigned int timestep);
+
+        Scalar getTunerParticlesPerChain()
```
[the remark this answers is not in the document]
maklem: I used it in the hoomd v2 tuner. As tuners work differently in v3 I removed that part.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1379 @@
...
+        .def("setChainTime", &IntegratorHPMCMonoNEC<Shape>::setChainTime)
+        .def("getChainTime", &IntegratorHPMCMonoNEC<Shape>::getChainTime)
+        .def_property("chain_time", &IntegratorHPMCMonoNEC<Shape>::getChainTime, &IntegratorHPMCMonoNEC<Shape>::setChainTime)
+        .def("setChainProbability", &IntegratorHPMCMonoNEC<Shape>::setChainProbability)
+        .def("getChainProbability", &IntegratorHPMCMonoNEC<Shape>::getChainProbability)
+        .def_property("chain_probability", &IntegratorHPMCMonoNEC<Shape>::getChainProbability, &IntegratorHPMCMonoNEC<Shape>::setChainProbability)
+        .def("setUpdateFraction", &IntegratorHPMCMonoNEC<Shape>::setUpdateFraction)
+        .def("getUpdateFraction", &IntegratorHPMCMonoNEC<Shape>::getUpdateFraction)
+        .def_property("update_fraction", &IntegratorHPMCMonoNEC<Shape>::getUpdateFraction, &IntegratorHPMCMonoNEC<Shape>::setUpdateFraction)
+        .def("getTunerParticlesPerChain", &IntegratorHPMCMonoNEC<Shape>::getTunerParticlesPerChain)
```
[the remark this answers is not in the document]
maklem: Direct access to get/set functions is not needed, so now only def_property and def_property_readonly remain.

hoomd/hpmc/XenoSweep3D.h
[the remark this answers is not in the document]
maklem: @b-butler If we did not finish yet, we need to update the portal. It is the same as in XenoCollide3D:
https://github.com/glotzerlab/hoomd-blue/blob/a5d516041247cfb81d81316dd0d9fa8c409115e3/hoomd/hpmc/XenoCollide3D.h#L250-L277

In each loop we get v4 as support vector to the normal of a triangle v1, v2, v3. If v4 is in the same plane as the triangle, we are at the surface, and can stop. Otherwise we need to continue. v4 with two of v1,v2,v3 form a triangle, that contains the origin in projection along the direction v0. The nested if statements pick that triangle. We need to iterate that triangle until it is at the surface of the Minkowski sum to calculate the distance.

maklem:
Hi,
I have updated the files. If I followed the comment, I marked that part as resolved; otherwise I added replies.

Currently I wonder why the test fails with MPI. I do not have a local installation with MPI though. (MPI either does not work in WSL, or I did not try long enough)

Some details are still missing
- paper identifier (not yet available)
- formatting hoomd/hpmc/ShapeConvexPolyhedron.h

I will add them soon. As both are spaces or comments. They do not need to be tested and validated individually, so I will wait until I got both, or there are further changes.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
b-butler: Yeah, the template has to be evaluated at compile time, so something like you had @maklem is necessary.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def __hash__(self):
+        return hash(( "chain_time", "", self._target, self._domain))
+
+    def __eq__(self, other):
+        return (self._target == other._target and
+                self._domain == other._domain)
+
+
+class _InternalChainTime(_InternalAction):
```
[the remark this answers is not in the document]
b-butler: Or a tuple or something similar, but this is not an expensive operation on the scale of the computation. Plus tuning will not happen every time step or at least should not across an entire simulation, so the cost should be minimal. It may even be faster since there isn't the redirection and iteration within the list.

hoomd/hpmc/XenoSweep3D.h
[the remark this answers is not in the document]
b-butler: I missed the loop for some reason. I thought this was the end of the method.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,276 @@
...
+        def flag_chain_time_update(value):
+            self._update_chain_time = True
+            return value
+
+        self._chain_time_def = _ChainTimeTuneDefinition(
+            target, (self._min_chain_time, max_chain_time))
+
+        # This is a bit complicated because we are having to ensure that we keep
+        # the list of tunables and the solver updated with the changes to
+        # attributes. However, these are simply forwarding a change along.
```
b-butler, line 127: This is no longer true.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,276 @@
...
+        param_dict = ParameterDict(target=OnlyTypes(
+            float, postprocess=target_postprocess),
+                                   solver=SolverStep,
+                                   max_chain_time=OnlyTypes(float,
+                                                            allow_none=True),
+                                   min_chain_time=OnlyTypes(float))
```
b-butler, line 133:
```suggestion
                                   min_chain_time=float)
```

hoomd/hpmc/pytest/test_sphere_eos.py
```diff
@@ -0,0 +1,80 @@
...
+    ci = 1.96
+
+    assert math.isclose(betap, mean, abs_tol=ci * (error + rel_err_cs * betap))
+    assert error < 1
+
+    mcCounts = mc.counters
+    necCounts = mc.nec_counters
+
+    assert mcCounts.overlap_errors == 0
+    assert necCounts.overlap_errors == 0
```
b-butler, line 80: Add pickling test see the end of  `hoomd/hpmc/test_shape.py`

hoomd/hpmc/pytest/test_sphere_eos.py
```diff
@@ -0,0 +1,80 @@
+# Copyright (c) 2009-2021 The Regents of the University of Michigan
```
b-butler, line 1: Rename file to indicate testing `NEC`.

b-butler (changes requested):
We need tests for the new NEC chain tuners. For the new C++  files (that haven't been already) please run through clang-format with the format file in the hoomd base directory. If you have trouble or need help let me know, and I can do that for you. Same with yapf for new Python files.

I have left some more comments to be addressed. In addition, some previous comments were not addressed or resolved, so those remain to be looked at.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1275 @@
...
+                                                    ShapeConvexPolyhedron& shape_i,
+                                                    ArrayHandle<unsigned int>& h_overlaps,
+                                                    ArrayHandle<Scalar4>& h_postype,
+                                                    ArrayHandle<Scalar4>& h_orientation,
+                                                    hpmc_nec_counters_t& nec_counters,
+                                                    vec3<Scalar>& collisionPlaneVector
+                                                   )
+    {
+    double sweepableDistance = maxSweep;
+
```
SchoeniPhlippsn, line 1116: Why is the normalisation of the `direction` missing here compared to the `sweepDistance` function using spheres?

SchoeniPhlippsn (changes requested):
First of all sorry for the very late response.
Next to Brandon's comments I have a few suggestions (see below). However, there is one bigger issue:
The `sweepDistance` functions for spheres and convex pollyhedra is largly equivalent. I have the impression that you can combine both of them by adding a `sweep_distance` function in `ShapeSphere.h`. This would not only avoid duplicated code, it also facilitates to make nec-mc applicable to other shapes later (if necessary).  
Lastly, you should add everything to the sphinx-docs.

joaander:
@maklem let us know if you have any questions about resolving conflicts on merging master into this branch or about the change requests.

maklem:
I'm sorry that I take so long to respond. I am currently busy with other things, but I hope to get back to this soon. The requested changes look fine. My main worry is the test, that works fine in serial computation, but not with MPI; and I have no system to test it with MPI at hand.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1275 @@
...
+            {
+            this->m_exec_conf->msg->notice(10) << "IntegratorHPMCMonoNEC<Shape>::setChainProbability(" << chain_probability << ")" << std::endl;
+            m_chain_probability = static_cast<unsigned int>(chain_probability * 65536.0);
+            }
+
+        //! Get move ratio
+        //! \returns ratio of translation versus rotation move attempts
+        inline Scalar getChainProbability()
+            {
+            return m_chain_probability / 65536.0;
```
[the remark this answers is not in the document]
maklem: Dividing integers will return zero.
What is the preferred way to convert the calculation to floating point?

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1275 @@
...
+                            { // if colliding with an inactive particle.
+                                vec3<Scalar> vel_change = collisionPlaneVector * (dot(vel_k,collisionPlaneVector) / dot(collisionPlaneVector,collisionPlaneVector));
+                                vel_k = vel_k - 2.0 * vel_change;
+
+                                h_velocities.data[k]    = make_scalar4( vel_k.x, vel_k.y, vel_k.z, h_velocities.data[k].w);
+                                next = k;
+
+                                velocity = fast::sqrt( dot(vel_k,vel_k));
+
+                                direction = vel_k / velocity;
```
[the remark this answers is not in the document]
maklem: How does `rsqrt` handle an input of zero?
If the velocities are not initialized propperly, the velocity is zero, and `sqrt` allows testing that.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
maklem: Simplifying the code with a `sweep_distance` for spheres, that difference does not exist anymore.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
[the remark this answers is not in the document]
joaander: @b-butler This code is following the pattern in IntegratorHPMC. We should probably revisit it there as well. The main issue this is solving is that it gives very clear boundary conditions for probabilities of 0 and 1 when drawing a random number to choose the type of move. In earlier versions of HOOMD (before the rewritten RandomNumbers.h file), I had some issues where setting a translation move probability to 1 would still result in some rotation moves.

We could define a class level static constant in integratorHPMC that sets the maximum integer for these probabilities converted to integers or we could test and see if the more stringent bounding conditions in the current RandomNumbers.h works if we leave these as floating point values.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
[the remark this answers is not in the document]
joaander: `rsqrt` only leads to performance boosts on the GPU. I wouldn't be concerned about using it here on the CPU where it is emulated with `1/sqrt`.

joaander:
> I'm sorry that I take so long to respond. I am currently busy with other things, but I hope to get back to this soon. The requested changes look fine. My main worry is the test, that works fine in serial computation, but not with MPI; and I have no system to test it with MPI at hand.

Thanks! We appreciate your efforts. How does the test fail in MPI? We can try and look at it here, but it would be best if you could test locally. I think you mentioned you were running in WSL. We've had some success with that, you need to either use a modern distro like Ubuntu 20.04. If you use an older ubuntu LTS release, the openmpi package doesn't work with HOOMD. You can also work around that by building openmpi from source.

hoomd/hpmc/integrate_nec.py
```diff
@@ -0,0 +1,300 @@
...
+        Note:
+            The counts are reset to 0 at the start of each
+            `hoomd.Simulation.run`.
+        """
+        if self._attached:
+            return self._cpp_obj.getNECCounters(1)
+        else:
+            return None
+
+    @log
```
b-butler, line 94: See the new `requires_run` optional argument to `hoomd.logging.log`. Change the other logged properties that need it too.

hoomd/hpmc/integrate_nec.py
```diff
@@ -0,0 +1,300 @@
...
+from hoomd.hpmc.integrate import HPMCIntegrator
+from hoomd.data.parameterdicts import TypeParameterDict, ParameterDict
+from hoomd.data.typeparam import TypeParameter
+from hoomd.data.typeconverter import OnlyTypes
+
+from hoomd.logging import log
+
+
+class HPMCNECIntegrator(HPMCIntegrator):
+    """HPMC Chain Integrator Meta Class.
```
b-butler, line 16: This is not a metaclass per Python definitions, so I would use parent class.

hoomd/hpmc/integrate_nec.py
```diff
@@ -0,0 +1,300 @@
...
+            return None
+
+
+class Sphere(HPMCNECIntegrator):
+    """HPMC chain integration for spheres (2D/3D).
+
+    Args:
+        d (float): Maximum move displacement, Scalar to set for all types,
+            or a dict containing {type:size} to set by type.
+        chain_time (float): length of a chain in units of time.
```
b-butler, line 143: See new unit documentation formatting (examples can be found in `hoomd/hpmc/integrate.py`.

hoomd/hpmc/integrate_nec.py
```diff
@@ -0,0 +1,300 @@
...
+        tune_nec_a = hoomd.hpmc.tune.MoveSize.scale_solver(triggerTune,
+                        moves=['a'], target=0.30)
+        sim.operations.tuners.append(tune_nec_a)
+
+        import hoomd.hpmc.tune.nec_chain_time
+        tune_nec_ct = hoomd.hpmc.tune.ChainTime.scale_solver(
+                        triggerTune, target=20, tol=1, gamma=20 )
+        sim.operations.tuners.append(tune_nec_ct)
+
+        sim.run(1000)
```
b-butler, line 260: We should probably use a minimal example here rather than a complete working simulation (assuming start.gsd exists). @joaander what do you think of long examples like this?

This applies to the class above too.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,286 @@
...
+        self.max_chain_time = max_chain_time
+        self.min_chain_time = self._min_chain_time
+
+    def _process_chain_time_range(self, target):
+        # check range
+        if 0 <= target:
+            pass
+        else:
+            raise ValueError(
+                "Chain time {} should be between 0 and 1000.".format(target))
```
b-butler, line 135: This is just checking that it is non-negative.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,286 @@
...
+        self.solver = solver
+        self.max_chain_time = max_chain_time
+        self.min_chain_time = self._min_chain_time
+
+    def _process_chain_time_range(self, target):
+        # check range
+        if 0 <= target:
+            pass
+        else:
+            raise ValueError(
```
b-butler, line 134: Instead of using a `pass` why not use the opposite condition and just error in that case without an else clause.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,286 @@
...
+        self._tuned = 0
+        return target
+
+    def _process_chain_time_target(self, target):
+        # check range
+        if 0 <= target <= 1000:
+            pass
+        else:
+            raise ValueError(
+                "Value {} should be between 0 and 1000.".format(target))
```
b-butler, line 145: No need for a full if else, just an if with an error.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,286 @@
...
+        # tune here as we have no other indication the system has changed.
+        elif ((self.previous_start > chain_start)
+              or (self.previous_hit > chain_hit)):
+            particles_per_chain = chain_hit / chain_start
+        else:
+            # flake8 does not allow a single-line calculation
+            # W504 line break after binary operator
+            delta_prev = chain_hit - self.previous_hit
+            delta_start = chain_start - self.previous_start
+            particles_per_chain = delta_prev / delta_start
```
b-butler, line 67: `delta_prev` doesn't make much sense `delta_hit` would be better. Also, you can use parenthesis to enable multi-line mathematical expressions.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,286 @@
...
+        return particles_per_chain
+
+    def _get_x(self):
+        return self.integrator.chain_time
+
+    def _set_x(self, value):
+        self.integrator.chain_time = value
+
+    def __hash__(self):
+        return hash(("chain_time", "", self._target, self._domain))
```
b-butler, line 84:
```suggestion
        return hash(("chain_time", self._target, self._domain))
```
empty string is unnecessary.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,286 @@
...
+    def act(self, timestep=None):
+        """Tune chain time.
+
+        Args:
+            timestep (`int`, optional): Current simulation timestep. Is
+                currently ignored.
+        """
+        if self._is_attached:
+            # update maximum chain time
+            if self._update_chain_time:
```
b-butler, line 185: This is always `False`. I think some logic is missing that sets the `maximum_chain_size` which could go in the `_process_chain_time` method.

hoomd/hpmc/tune/nec_chain_time.py
```diff
@@ -0,0 +1,286 @@
...
+            tuned = self.solver.solve([self._chain_time_def])
+            self._tuned = self._tuned + 1 if tuned else 0
+
+
+class ChainTime(_InternalCustomTuner):
+    """Tunes HPMCNECIntegrator chain time to targeted mean particles per chain.
+
+    For most common creation of a `ChainTime` tuner see
+    `ChainTime.secant_solver` and `ChainTime.scale_solver` respectively.
+
```
b-butler, line 198: I would also provide a recommended `target` value in the documentation.

hoomd/hpmc/pytest/test_nec_sphere_eos.py
```diff
@@ -0,0 +1,104 @@
...
+            # This will fill in default values for the inner shape objects
+            inner_mc.shape["A"] = args["shapes"][i]
+            args["shapes"][i] = inner_mc.shape["A"]
+    mc = integrator()
+    mc.shape["A"] = args
+    # L needs to be ridiculously large as to not be too small for the domain
+    # decomposition of some of the shapes definitions in valid_args which have
+    # shapes with large extent in at least one dimension.
+    sim = simulation_factory(two_particle_snapshot_factory(L=1000))
+    operation_pickling_check(mc, sim)
```
b-butler, line 104: We should either rename the file to `test_nec` or move this test to another file named that.

hoomd/hpmc/pytest/test_nec_sphere_eos.py
```diff
@@ -0,0 +1,104 @@
...
+# from test_sphere_eos.py for hoomd-2.9
+# see for example Guang-Wen Wu and Richard J. Sadus, doi:10.1002.aic10233
+_phi_p_ref = [(0.29054, 0.1), (0.91912, 0.2), (2.2768, 0.3), (5.29102, 0.4),
+              (8.06553, 0.45), (9.98979, 0.475)]
+rel_err_cs = 0.0015
+
+
+@pytest.mark.parametrize("betap,phi", _phi_p_ref)
+@pytest.mark.validate
+def test_sphere_eos_nec(betap, phi, simulation_factory,
```
b-butler, line 24: The validation test looks good, and is nice to have. We should also test the proper validation of getting and setting parameters and type parameters.  We also need to test the new NEC tuner class.

b-butler (changes requested):
I gave the code another quick run through and left my comments. Also, while not necessary, `const` could be used for in more places than it is currently, and would be a nice addition. Thanks for the contributor again!

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1162 @@
...
+*/
+template<class Shape> class IntegratorHPMCMonoNEC : public IntegratorHPMCMono<Shape>
+    {
+    protected:
+    Scalar m_chain_time; //!< the length of a chain, given in units of time
+    unsigned int
+        m_chain_probability;  //!< how often we do a chain. Replaces translation_move_probability
+    Scalar m_update_fraction; //!< if we perform chains we update several particles as one
+
+    GlobalArray<hpmc_nec_counters_t> m_nec_count_total; //!< counters for chain statistics
```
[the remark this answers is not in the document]
joaander: The managed unified memory buffers managed by GlobalArray are the most efficient way to share move counts between the CPU and GPU. `IntegratorHPMC` and `IntegratorHPMCMono` use this pattern.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1162 @@
...
+            Scalar4 postype_i = h_postype.data[i];
+            Scalar4 orientation_i = h_orientation.data[i];
+            Scalar4 velocity_i = h_velocities.data[i];
+            int typ_i = __scalar_as_int(postype_i.w);
+            Shape shape_i(quat<Scalar>(orientation_i), this->m_params[typ_i]);
+
+#ifdef ENABLE_MPI
+            if (this->m_comm)
+                {
+                // only move particle if active
```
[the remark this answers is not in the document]
joaander: See https://doi.org/10.1016/j.cpc.2016.02.024

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1162 @@
...
+            // this->m_update_order.shuffle(...) wants to update the particles in forward or reverse
+            // order. For chains this is an invalid behavior. Instead we have to pick a starting
+            // particle randomly.
+            unsigned int i = hoomd::UniformIntDistribution(this->m_pdata->getN() - 1)(rng_chain_i);
+
+            Scalar4 postype_i = h_postype.data[i];
+            Scalar4 orientation_i = h_orientation.data[i];
+            Scalar4 velocity_i = h_velocities.data[i];
+            int typ_i = __scalar_as_int(postype_i.w);
+            Shape shape_i(quat<Scalar>(orientation_i), this->m_params[typ_i]);
```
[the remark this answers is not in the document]
joaander: This is following the pattern already set in `IntegratorHPMCMono`.

hoomd/hpmc/integrate_nec.py
joaander: I think even minimal examples in the API docs are prone to syntax errors and will likely be taken out of context.

joaander:
@maklem let us know when you expect to address the review comments.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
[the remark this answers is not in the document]
b-butler: Got it thanks!

joaander:
@maklem We cannot merge this in until the review comments have been addressed.

joaander:
@maklem We cannot merge this in until the review comments have been addressed.

hoomd/hpmc/pytest/test_sphere_eos.py
```diff
@@ -0,0 +1,72 @@
...
+    sim.operations.tuners.append(tune_nec_ct)
+
+    #equilibrate
+    sim.run(1000)
+    sim.operations.tuners.clear()
+
+    pressures = []
+    mc.nselect = 50
+    for i in range(100):
+        sim.run(1)
```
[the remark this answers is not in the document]
joaander: I have no idea, but we can revisit and improve when reimplementing this in `hoomd-validation-tests`.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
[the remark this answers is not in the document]
joaander: I have returned this to follow the same style as `IntegratorHPMC.h`. Interested developers can later clean up the whole code base to fix instances like this.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1162 @@
...
+
+    \ingroup hpmc_integrators
+*/
+template<class Shape> class IntegratorHPMCMonoNEC : public IntegratorHPMCMono<Shape>
+    {
+    protected:
+    Scalar m_chain_time; //!< the length of a chain, given in units of time
+    unsigned int
+        m_chain_probability;  //!< how often we do a chain. Replaces translation_move_probability
+    Scalar m_update_fraction; //!< if we perform chains we update several particles as one
```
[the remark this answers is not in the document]
joaander: Yes, one chain move update the position of many particles.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1162 @@
...
+
+    //! Get move ratio
+    //! \returns ratio of translation versus rotation move attempts
+    inline Scalar getChainProbability()
+        {
+        // For use with the random number generator we convert the
+        // translation move probability into a 16bit unsigned integer.
+        static auto INT_PROBABILITY_CEILING = 65536;
+        return Scalar(m_chain_probability) / INT_PROBABILITY_CEILING;
+        }
```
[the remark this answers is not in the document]
joaander: I switched this back to the same style as in `IntegratorHPMC.h` and other files. Interested developers can make a later PR to clean up the code base as a whole.

hoomd/hpmc/tune/nec_chain_time.py
joaander: If you know how to fix, please do so.

hoomd/hpmc/tune/nec_chain_time.py
joaander: Good suggestion. This method isn't widely used yet so there may be no good value to suggest. Users will need to refer to the paper to gain any insight in how to use the method effectively.

hoomd/hpmc/pytest/test_nec_sphere_eos.py
joaander: I fixed the pickling test to apply tests to the NEC integrators (they were testing the normal HPMC integrators). This indirectly tests parameter get and set. Do you have time to implement a test of the tuner? I don't.

joaander:
Test script:

```
import hoomd
import numpy
import itertools

cpu = hoomd.device.CPU()
sim = hoomd.Simulation(device=cpu)
sim.seed = 10

snapshot = hoomd.Snapshot()
K = 10
L = K * 2.4
snapshot.particles.N = K**3
x = numpy.linspace(-L / 2, L / 2, 10, endpoint=False)
snapshot.particles.position[:] = list(itertools.product(x, repeat=3))
snapshot.configuration.box = [L, L, L, 0, 0, 0]

sim.state.thermalize_particle_momenta(hoomd.filter.All(), kT=1)

mc = hoomd.hpmc.nec.integrate.ConvexPolyhedron(default_d=0.05,
                                               default_a=0.05,
                                               chain_probability=0.1,
                                               update_fraction=0.05,
                                               chain_time=0.05,
                                               nselect=10)
mc.shape['A'] = dict(
    vertices=[[1, 1, 1], [1, 1, -1], [1, -1, 1], [1, -1, -1], [-1, 1, 1],
              [-1, 1, -1], [-1, -1, 1], [-1, -1, -1]])
mc.chain_time = 0.05
sim.operations.integrator = mc

tuner_trigger = hoomd.trigger.And(
    [hoomd.trigger.Periodic(50),
     hoomd.trigger.Before(sim.timestep + 1000)])
tune_nec_d = hoomd.hpmc.tune.MoveSize.scale_solver(tuner_trigger,
                                                   moves=['d'],
                                                   target=0.10,
                                                   max_translation_move=0.15)
sim.operations.tuners.append(tune_nec_d)
tune_nec_a = hoomd.hpmc.tune.MoveSize.scale_solver(tuner_trigger,
                                                   moves=['a'],
                                                   target=0.30)
sim.operations.tuners.append(tune_nec_a)

tune_nec_ct = hoomd.hpmc.nec.tune.ChainTime.scale_solver(tuner_trigger,
                                                         target=20,
                                                         tol=1,
                                                         gamma=20)
sim.operations.tuners.append(tune_nec_ct)

logger.add(mc, quantities=['type_shapes'])
gsd_writer = hoomd.write.GSD(filename='trajectory.gsd',
                             trigger=hoomd.trigger.Periodic(100),
                             mode='wb',
                             log=logger)
sim.operations.writers.append(gsd_writer)

sim.run(1000)

sim.run(1000)
rotate_moves = mc.rotate_moves
translate_moves = mc.translate_moves
print('Rotation acceptance:', rotate_moves[0] / sum(rotate_moves), ' a=',
      mc.a['A'])
print('Translation collision hits:', translate_moves[0] / sum(translate_moves),
      ' d=', mc.d['A'])
print('Particles per chain:', mc.particles_per_chain, ' chain_time=',
      mc.chain_time)
```

Test output:
```
Rotation acceptance: 0.301266767842184  a= 0.0766609721785175
Translation collision hits: 0.28297650392109464  d= 0.15
Particles per chain: 19.77170366515203  chain_time= 1.5251671451920268
```
snapshot

With this test complete, I have completed this branch as much as I intend to. @b-butler @SchoeniPhlippsn if you have any important change requests feel free to make them yourself then approve.

b-butler (changes requested):
Review to give me action items. I will address and approve.

hoomd/hpmc/IntegratorHPMCMonoNEC.h
```diff
@@ -0,0 +1,1164 @@
...
+*/
+template<class Shape> class IntegratorHPMCMonoNEC : public IntegratorHPMCMono<Shape>
+    {
+    protected:
+    Scalar m_chain_time; //!< the length of a chain, given in units of time
+    unsigned int
+        m_chain_probability;  //!< how often we do a chain. Replaces translation_move_probability
+    Scalar m_update_fraction; //!< if we perform chains we update several particles as one
+
+    GlobalArray<hpmc_nec_counters_t> m_nec_count_total; //!< counters for chain statistics
```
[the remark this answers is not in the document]
joaander: Yes, good idea.

b-butler (approved):

PR #457, opened 2019-09-11
Refactor/aniso pair structs

vyasr:
## Description

This PR updates all Evaluators used with AnisoPairPotential to use a struct for managing their parameters. All such structs are equipped with the ability to load their data into shared memory. This change is enabled by updates to GlobalArray that allow client code to force it to use shared memory.

## Motivation and Context

These changes allow much more efficient access to pair potential parameters if they are more complex than just a few constants. For example, if any such potential involved a table lookup, the table could now be loaded into shared memory.

## How Has This Been Tested?

These changes are tested on a separate development branch implementing a new anisotropic potential Evaluator.

## Change log

```
Allow client code to force GlobalArrays to use shared memory and update all anisotropic potential Evaluators to use structs that can leverage this feature if needed.
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/hoomd-blue/blob/master/sphinx-doc/credits.rst).
- [x] (Glotzer group members): I am a member of the [hoomd-contributors team](https://github.com/orgs/glotzerlab/teams/hoomd-contributors/members).

PR #408, opened 2019-05-15
Optimize HPMC GPU integrator for strong scaling

jglaser:
## Description

Introduces a new optimized GPU integrator for HPMC.

## Motivation and Context

Improves strong scaling on multiple GPUs, and performance of some systems on a single GPU. I will add more information here soon.

Merge after #405 

## How Has This Been Tested?

Standard unit tests

## Change log

```
- Optimizations for hpmc.integrate.* for improved strong scaling of small systems and on multiple GPUs with MPI
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x] My name is on the [list of contributors](https://github.com/glotzerlab/hoomd-blue/blob/master/sphinx-doc/credits.rst).

PR #965, opened 2021-03-25
Convert Table Potential to v3 API

### LINKED ISSUE #360: Port Table potentials to v3 API
HOOMD's tabulated potential implementation and API are being very widely used. Their implementation, and Python API in particular, could use a thorough review and be improved if possible.

danevans09:
## Description

This converts the Table potential to the v3 API. I created an `EvaluatorPairTable.h` file to follow the format of other pair potentials and changed it so the force and energy tables are directly set as numpy arrays for each pair of particle types.
## Motivation and context

The Table potential has not yet been converted to the v3 API.

Works toward #360 

## How has this been tested?

I added valid and invalid Table potential arguments to `test_potential.py`.

## Change log

```

```

## Checklist:


hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+    \ingroup computes
+*/
+class EvaluatorPairTable
+    {
+    public:
+        //! Define the parameter type used by this pair potential evaluator
+        struct param_type
+            {
+            unsigned int width;
+            std::vector<Scalar> V_table;
```
danevans09, line 76: The old code used a `GlobalArray` to store the force and energy values but I wasn't sure how to access `exec_conf` to initialize a `GlobalArray`

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+
+            \return True if they are evaluated or false if they are not because
+            we are beyond the cutoff
+        */
+        DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& pair_eng, bool energy_shift)
+            {
+            // compute the force divided by r in force_divr
+            if (rsq < rcutsq)
+                {
+                Scalar delta_r = std::sqrt(rcutsq) / width;
```
danevans09, line 183: This should be: 

`(std::sqrt(rcutsq) - std::sqrt(ronsq)) / width` 

but `ronsq` is not in the constructor arguments. Should I add a `needsRon` method to every pair potential or change the constructor for `EvaluatorPairTable`?

joaander: The minimum r for the table potential and r_on used in the `PotentialPair` framework are two different things. You should add `r_min` to the `param_type` struct for the table potential and use that here.

joaander (commented):
Dan, this is a great start! I was thinking we would just write a new interface to the old code - but your solution to re-implement as a `PotentialPair` evaluator is much better. It also solves some of the user requests we had (different widths for different interactions).

It is possible to use `ManagedArray` to manage the table values in this framework, but it will take some careful modification of the GPU kernels in `PotentialPairGPU` to get this working on the GPU. The CPU implementation should work without needing modification.

For now, you can update the code to use `ManagedArray` instead of `std::vector` and it should work on the CPU. `ManagedArray` doesn't get an `ExecutionConfiguration`, instead, it gets a boolean if it should use the GPU or not. You can just leave this as a default `false` for now. I need to look more closely at what changes are needed to `PotentialPair` to get the proper value passed down. I remember that took a while to get right in `IntegratorHPMCMono`.

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,249 @@
...
+                    });
+                pybind11::capsule free_F(F, [](void *f)
+                    {
+                    Scalar *F = reinterpret_cast<Scalar *>(f);
+                    delete[] F;
+                    });
+
+                pybind11::dict v;
+                v["V"] = pybind11::array_t<Scalar>(pybind11::array::ShapeContainer({width,}),
+                                                   pybind11::array::StridesContainer({1*sizeof(Scalar),}),
```
b-butler, line 133: Why the `1 *`?

hoomd/md/module-md.cc
```diff
@@ -225,7 +225,7 @@ PYBIND11_MODULE(_md, m)
     export_OPLSDihedralForceCompute(m);
     export_TableDihedralForceCompute(m);
     export_HarmonicImproperForceCompute(m);
-    export_TablePotential(m);
+    // export_TablePotential(m);
```
b-butler, line 228:
```suggestion
```

hoomd/md/pair/pair.py
```diff
@@ -563,226 +562,50 @@ class table(force._force):
...
-    - :math:`_{\mathrm{max}}` - *rmax* (in distance units)
-
-    .. rubric:: Set table from a given function
-
-    When you have a functional form for V and F, you can enter that
-    directly into python. :py:class:`table` will evaluate the given function over *width* points between
-    *rmin* and *rmax* and use the resulting values in the table::
+    See `Pair` for the available energy shifting and smoothing modes. Use
+    `params` dictionary to set potential coefficients. The coefficients must
+    be set per unique pair of particle types.
```
b-butler, line 567: Table doesn't support any of these though?

hoomd/md/pytest/test_potential.py
```diff
@@ -147,10 +147,10 @@ def _make_invalid_param_dict(valid_dict):
         # Set one invalid argument per dictionary
         # Set two invalid arguments per key
         valid_value = invalid_dicts[count][key]
-        if not isinstance(valid_value, list):
+        if not (isinstance(valid_value, list) or isinstance(valid_value, np.ndarray)):
```
b-butler, line 150:
```suggestion
        if not (isinstance(valid_value, (list, np.ndarray)):
```

hoomd/md/pytest/test_potential.py
```diff
@@ -147,10 +147,10 @@ def _make_invalid_param_dict(valid_dict):
...
             invalid_dicts[count][key] = [1, 2]
             invalid_count += 1
-        if not isinstance(valid_value, str):
+        if not (isinstance(valid_value, str) or isinstance(valid_value, np.ndarray)):
```
b-butler, line 153:
```suggestion
        if not (isinstance(valid_value, (str, np.ndarray)):
```

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,249 @@
...
+    \b Table memory layout
+
+    V(r) and F(r) are specified for each unique particle type pair. They will be indexed so that values of increasing r
+    go along the rows in memory for cache efficiency when reading values. The row index to put the potential at can be
+    determined using an Index2DUpperTriangular (typei, typej), as it will uniquely index each unique pair.
+
+    To improve cache coherency even further, values for V and F will be interleaved like so: V1 F1 V2 F2 V3 F3 ... To
+    accomplish this, tables are stored with a value type of Scalar2, elem.x will be V and elem.y will be F. Since Fn,
+    Vn+1 and Fn+1 are read right after Vn, these are likely to be cache hits. Furthermore, on the GPU a single Scalar2
+    texture read can be used to access Vn and Fn.
```
b-butler, line 51: This is not the case currently.

b-butler (commented):
Overall, the PR looks good. I have a few comments on documentation, and on design potentially, but I want to see what @joaander  thinks of the PR before merging.

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
+// Copyright (c) 2009-2021 The Regents of the University of Michigan
+// This file is part of the HOOMD-blue project, released under the BSD 3-Clause License.
+
+
+// Maintainer: joaander
```
joaander, line 5:
```suggestion
```
We no longer assign maintainers per file.

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+#include "hoomd/ForceCompute.h"
+#include "NeighborList.h"
+#include "hoomd/Index1D.h"
+#include "hoomd/GlobalArray.h"
+#include "hoomd/ManagedArray.h"
+#include <memory>
+
+/*! \file TablePotential.h
+    \brief Declares the TablePotential class
+*/
```
joaander, line 16:
```suggestion
```
We no longer support doxygen documentation builds.

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+    \a rmax. Evaluations are performed by simple linear interpolation, thus why F(r) must be explicitly specified to
+    avoid large errors resulting from the numerical derivative. Note that F(r) should store - dV/dr.
+
+    \b Table memory layout
+
+    V(r) and F(r) are specified for each unique particle type pair. Three parameters need to be stored for each 
+    potential: rmin, rmax, and dr, the minimum r, maximum r, and spacing between r values in the table respectively. 
+    V(0) is the value of V at r=rmin. V(i) is the value of V at r=rmin + dr * i where i is chosen such that r >= rmin
+    and r <= rmax. V(r) for r < rmin and > rmax is 0. The same goes for F. Thus V and F are defined between the region
+    [rmin,rmax], inclusive.
```
joaander, line 48: I don't think this is accurate. If the number of bins is 10 and rmax is 1, then the potential is defined at the points 0.0, 0.1, ... 0.9. V and F are zero for r >= rmax.

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+/*! \b Overview
+    Pair potentials and forces are evaluated for all particle pairs in the system within the given cutoff distances.
+    Both the potentials and forces** are provided the tables V(r) and F(r) at discreet \a r values between \a rmin and
+    \a rmax. Evaluations are performed by simple linear interpolation, thus why F(r) must be explicitly specified to
+    avoid large errors resulting from the numerical derivative. Note that F(r) should store - dV/dr.
+
+    \b Table memory layout
+
+    V(r) and F(r) are specified for each unique particle type pair. Three parameters need to be stored for each 
+    potential: rmin, rmax, and dr, the minimum r, maximum r, and spacing between r values in the table respectively.
```
joaander, line 45: I think rmax is misleading in these comments. There is no rmax variable stored or computed in the new code. I would update the documentation to refer to r_cut instead of rmax.

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+            #endif
+
+            #ifndef __HIPCC__
+            param_type() : width(0), rmin(0.0), V_table({}), F_table({}) {}
+
+            param_type(pybind11::dict v)
+                {
+                width = v["width"].cast<unsigned int>();
+                rmin = v["r_min"].cast<Scalar>();
+                unsigned int align_size = 8; //for AVX
```
joaander, line 82: We aren't using AVX in the table potentials. You can allocate the arrays with size `width` and remove the `, 32` and all the alignment size computations.

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+                unsigned int N_align =((width + align_size - 1)/align_size)*align_size;
+                V_table = ManagedArray<Scalar>(N_align, false, 32); // 32byte alignment for AVX
+                F_table = ManagedArray<Scalar>(N_align, false, 32);
+                auto V_py = v["V"].cast<pybind11::array_t<Scalar>>().unchecked<1>();
+                auto F_py = v["F"].cast<pybind11::array_t<Scalar>>().unchecked<1>();
+                for (unsigned int i = 0; i < width; i++)
```
joaander, line 88: You could write this with less code using the C++ standard library copy algorithm: https://www.cplusplus.com/reference/algorithm/copy/

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+                    {
+                    V_table[i] = V_py[i];
+                    F_table[i] = F_py[i];
+                    }
+                if (V_table.size() != F_table.size())
```
joaander, line 93: This error check should occur first and it should check the lengths of the Python arrays the users provides, not the ones we allocate.

Also, why is "width" a separate parameter? You should be able to determine the width by inspecting the length of the arrays provided by the user in v["V"] and v["F"].

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+                    delete[] V;
+                    });
+                pybind11::capsule free_F(F, [](void *f)
+                    {
+                    Scalar *F = reinterpret_cast<Scalar *>(f);
+                    delete[] F;
+                    });
+
+                pybind11::dict v;
+                v["V"] = pybind11::array_t<Scalar>(pybind11::array::ShapeContainer({width,}),
```
joaander, line 119: @b-butler 

Are numpy arrays pickleable? Do we need to make changes here so that table potentials will work with the new pickling infrastructure?

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+            V(r) is continuous at the cutoff
+            \note There is no need to check if rsq < rcutsq in this method.
+            Cutoff tests are performed in PotentialPair.
+
+            \return True if they are evaluated or false if they are not because
+            we are beyond the cutoff
+        */
+        DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& pair_eng, bool energy_shift)
+            {
+            // compute the force divided by r in force_divr
```
joaander, line 179: This code is used in the inner loop, let's not trust the compiler/optimizer. Only call sqrt once for each unique value and then store that in a variable and reuse it.

Also, use fast::sqrt instead of std::sqrt.

```
Scalar rcut = fast::sqrt(rcutsq);
Scalar r = fast::sqrt(rsq);
```

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+            if (rsq < rcutsq && std::sqrt(rsq) >= rmin)
+                {
+                Scalar delta_r = (std::sqrt(rcutsq) - rmin) / width;
+                // precomputed term
+                Scalar value_f = (std::sqrt(rcutsq) - std::sqrt(rsq)) / delta_r;
```
joaander, line 184: Souldn't this be
`value_f = (r - rmin) / delta_r`
?

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+
+                // compute index into the table and read in values
+                unsigned int value_i = (unsigned int)floor(value_f);
+                // unpack the data
+                Scalar V0 = V_table[value_i];
+                Scalar V1 = V_table[value_i + 1];
+                Scalar F0 = F_table[value_i];
+                Scalar F1 = F_table[value_i + 1];
```
joaander, line 192: Handle out of bounds memory accesses:
```suggestion
                Scalar V0 = V_table[value_i];
                Scalar V1 = 0;
                if (value_i + 1 < width)
                    V1 = V_table[value_i + 1];
```

We might also be concerned about underflow or overflow of value_i for some unexpected floating point rounding conditions. We could write this more defensively by always computing `value_f` and `value_i` (as an `int`), then change the `if (rsq ...)` condition to `if (value_i >= 0 && value_i < width)`

hoomd/md/pair/__init__.py
```diff
@@ -23,5 +23,6 @@
     LJ1208,
     Fourier,
     OPP,
-    TWF
+    TWF,
+    Table
```
joaander, line 27:
```suggestion
    Table,
    TWF
```
A comes before W.

joaander (changes requested):
Thanks!

I reviewed the implementation, but not the documentation or tests. Change requests indicated.

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+                Scalar f = value_f - Scalar(value_i);
+
+                // interpolate to get V and F;
+                Scalar V = V0 + f * (V1 - V0);
+                Scalar F = F0 + f * (F1 - F0);
+
+                // convert to standard variables used by the other pair computes in HOOMD-blue
+                if (rsq > Scalar(0.0))
+                    force_divr = F / r;
+                pair_eng = Scalar(0.5) * V;
```
danevans09, line 204: Should this 0.5 be removed? The tests that compare the simulation forces to forces calculated by differentiating the simulation energy are failing by a factor of 1/2

joaander: Yes, remove the 0.5. The code that calls the evaluator multiplies by 0.5 where necessary.

AlainKadar (approved):

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,236 @@
...
+                    delete[] V;
+                    });
+                pybind11::capsule free_F(F, [](void *f)
+                    {
+                    Scalar *F = reinterpret_cast<Scalar *>(f);
+                    delete[] F;
+                    });
+
+                pybind11::dict v;
+                v["V"] = pybind11::array_t<Scalar>(pybind11::array::ShapeContainer({width, 1}),
```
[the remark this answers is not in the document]
b-butler: I think a NumPy array makes sense since that is how I would think to compute a table potential.

hoomd/md/EvaluatorPairTable.h
b-butler: NumPy arrays are pickable.

hoomd/md/pytest/test_potential.py
```diff
@@ -128,10 +128,10 @@ def _make_invalid_param_dict(valid_dict):
         # Set one invalid argument per dictionary
         # Set two invalid arguments per key
         valid_value = invalid_dicts[count][key]
-        if not isinstance(valid_value, list):
+        if not (isinstance(valid_value, (list, np.ndarray))):
```
b-butler, line 131: The parenthesis around `isinstance` are unnecessary.
```suggestion
        if not isinstance(valid_value, (list, np.ndarray)):
```

hoomd/md/pytest/test_potential.py
```diff
@@ -128,10 +128,10 @@ def _make_invalid_param_dict(valid_dict):
...
             invalid_dicts[count][key] = [1, 2]
             invalid_count += 1
-        if not isinstance(valid_value, str):
+        if not (isinstance(valid_value, (str, np.ndarray))):
```
b-butler, line 134:
```suggestion
        if not isinstance(valid_value, (str, np.ndarray)):
```

hoomd/md/pair/pair.py
```diff
@@ -520,6 +521,89 @@ def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
         self._add_typeparam(params)
 
 
+class Table(Pair):
+    """ Tabulated pair potential.
```
b-butler, line 525: Docstring needs to be formatted.

b-butler (changes requested):
I have requested some changes that need to be made.

b-butler:
@danevans09 sorry I hadn't finished making all the changes I was going to so I had compilation errors.

hoomd/md/EvaluatorPairTable.h
```diff
@@ -0,0 +1,232 @@
...
+                const auto V_py = v["V"].cast<pybind11::array_t<Scalar>>().unchecked<1>();
+                const auto F_py = v["F"].cast<pybind11::array_t<Scalar>>().unchecked<1>();
+                if (V_py.size() != F_py.size())
+                    {
+                    throw std::runtime_error("The length of V and F arrays must be equal");
+                    }
+                width = V_py.size();
+                rmin = v["r_min"].cast<Scalar>();
+                V_table = ManagedArray<Scalar>((unsigned int)width, managed, 32);
+                F_table = ManagedArray<Scalar>((unsigned int)width, managed, 32);
```
b-butler, line 86: C++ style casts are more descriptive of the expected type of cast and can catch more programming errors. It is also more clear for come complex expressions (which this is not admittedly).
```suggestion
                V_table = ManagedArray<Scalar>(static_cast<unsigned int>(width), managed, 32);
                F_table = ManagedArray<Scalar>(static_cast<unsigned int>(width), managed, 32);
```

hoomd/md/module-md.cc
```diff
@@ -375,7 +373,6 @@ PYBIND11_MODULE(_md, m)
         m,
         "PotentialSpecialPairCoulombGPU");
     export_BondTablePotentialGPU(m);
-    export_TablePotentialGPU(m);
```
b-butler, line 378: When this is working for the GPU we will need to export that to Python as well.

hoomd/md/PotentialPairGPU.cuh
```diff
@@ -167,13 +168,29 @@ gpu_compute_pair_forces_shared_kernel(Scalar4* d_force,
...
             }
         }
+
+    unsigned int param_size
+        = num_typ_parameters * sizeof(typename evaluator::param_type) / sizeof(int);
+    for (unsigned int cur_offset = 0; cur_offset < param_size; cur_offset += blockDim.x)
+        {
+        if (cur_offset + threadIdx.x < param_size)
+            {
+            ((int*)s_params)[cur_offset + threadIdx.x] = ((int*)d_params)[cur_offset + threadIdx.x];
```
b-butler, line 182:
```suggestion
            reinterpret_cast<int*>(s_params)[cur_offset + threadIdx.x] = reinterpret_cast<int*>(d_params)[cur_offset + threadIdx.x];
```

hoomd/md/PotentialPairGPU.cuh
```diff
@@ -167,13 +168,29 @@ gpu_compute_pair_forces_shared_kernel(Scalar4* d_force,
...
+            }
+        }
     __syncthreads();
 
+    // initialize extra shared mem
+    char* s_extra = (char*)(s_params + typpair_idx.getNumElements());
```
b-butler, line 188:
```suggestion
    auto s_extra = reinterpret_cast<char*>(s_params + typpair_idx.getNumElements());
```

hoomd/md/PotentialPairGPU.cuh
```diff
@@ -421,6 +438,30 @@ struct PairForceComputeKernel
...
+                unsigned int available_bytes = max_extra_bytes;
+                for (unsigned int i = 0; i < typpair_idx.getNumElements(); ++i)
+                    {
+                    params[i].load_shared(ptr, available_bytes);
+                    }
+                extra_bytes = max_extra_bytes - available_bytes;
+                }
+
+            shared_bytes += extra_bytes;
+
```
b-butler, line 464: Use static or reinterpret cast as necessary.

b-butler (commented):
I know it is not done yet, but I went ahead and gave the changes a look and left some comments.

hoomd/md/module-md.cc
danevans09: I believe this is exporting the old `Table `potential implementation

b-butler: It is; I meant that we will need to expose the Table potential on the GPU too, similar to the other pair potentials.

hoomd/md/PotentialPairGPU.cuh
```diff
@@ -421,6 +440,32 @@ struct PairForceComputeKernel
...
+                {
+                // required for memory coherency
+                hipDeviceSynchronize();
+
+                // determine dynamically requested shared memory
+                char* ptr = static_cast<char*>(nullptr);
+                unsigned int available_bytes = max_extra_bytes;
+                for (unsigned int i = 0; i < typpair_idx.getNumElements(); ++i)
+                    {
+                    d_params[i].load_shared(ptr, available_bytes);
```
danevans09, line 462: This line is causing the code to crash, however all of the tests in `test_potential.py` pass when I comment out this line

joaander: OK, I will take a look.

joaander: Look at the way HPMC handles this. It has a separate `allocate_shared` call that doesn't attempt to write to the pointer. I don't know why the aniso potential doesn't, but it should.

vyasr: Did some quick digging here. I introduced the aniso code in #457, which was merged on Sept 16 2019. Jens introduce `allocate_shared` in #408, which was opened well before #457 but didn't get merged until about a month after. It looks like we just failed to synchronize those changes at the time since they appeared orthogonal.

joaander: Looking further, the aniso potential is also using global arrays which can't hold nested data structures like managed arrays. I'm currently working to convert `PotentialPairGPU` to use `std::vector<param_type, managed_allocator<param_type> >` as HPMC does.

vyasr: Due to the size and long-lived nature of a lot of Jens's PRs that made these changes it's not surprising that we ended up out of sync, as I recall you and I were sort of feeling our way through Jens's unified memory handling when we worked on adding shared memory support to anisotropic potentials. Hopefully the changes aren't terribly difficult to make.

joaander (commented):
Thanks! This is ready for another round of review.

I still need to review the user-facing documentation.

joaander (approved):
Thanks! Subject to @b-butler approving the changes in `parameterdicts.py`, this looks good to me. 

I applied my comments directly as changes to the pair.Table documentation.

hoomd/md/pair/pair.py
```diff
@@ -526,6 +527,95 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0.):
...
+          * ``r_min`` (`float`, **required**) - the minimum distance to apply
+            the tabulated potential, corresponding to the first element of the
+            energy and force arrays :math:`[\\mathrm{length}]`.
+
+          * ``V`` ((*N*,) `numpy.ndarray` of ``numpy.float64``, **required**) -
+            the tabulated energy values :math:`[\\mathrm{energy}]`.
+
+          * ``F`` ((*N*,) `numpy.ndarray` of ``numpy.float64``, **required**) -
+            the tabulated force values :math:`[\\mathrm{force}]`. Must have the
+            same length as ``V``.
```
b-butler, line 597:
```suggestion
          * ``V`` ((*N*,) `numpy.ndarray` of `float`, **required**) -
            the tabulated energy values :math:`[\\mathrm{energy}]`.

          * ``F`` ((*N*,) `numpy.ndarray` of `float`, **required**) -
            the tabulated force values :math:`[\\mathrm{force}]`. Must have the
            same length as ``V``.
```
This is equivalent and the way NumPy documents and recommends using for methods/functions such as `astype`.

b-butler (commented):
I have a few comments that I would like addressed, but I don't want to hold this up, so I will just leave it as a comment. Feel free to merge once resolved without another review by me.

b-butler (approved):

hoomd/md/pair/pair.py
```diff
@@ -10,6 +10,7 @@
 from hoomd.md.nlist import NList
 from hoomd.data.parameterdicts import ParameterDict, TypeParameterDict
 from hoomd.data.typeparam import TypeParameter
+import numpy
```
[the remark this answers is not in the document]
joaander: I realize I'm against the bandwagon here, but just because something is recommended in the docs doesn't mean it is a good practice. In this case the proliferation of `import numpy as np` provides a fluid representation that serves no purpose other than to make example code look cute. In contrast with what the linked documentation says, fluid representations make concepts more difficult to understand, not easier. If the authors truly intend for all users to use `np`, then they should have named the library `np`.

I personally find it more useful to `import numpy` and  copy and past method names and arguments that use the full name from the documentation (e.g. `numpy.testing.assert_allclose` documented here: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html). The full names appear in the documentation more often than the short versions. As such, HOOMD has a mix of using numpy as `numpy` and `np` in different files.

PR #972, opened 2021-04-07
Check for invalid typeids

### LINKED ISSUE #838: create_state_from_gsd does not validate type input
## Description

GSD files that reference non-existent type ids parse and load without error.

## Script

```python
import hoomd
import gsd.hoomd

snapshot = gsd.hoomd.Snapshot()
snapshot.particles.N = 2
snapshot.particles.position = [[0,0,0], [1,1,1]]
snapshot.particles.typeid = [1,2]
snapshot.configuration.box = [10, 10, 0, 0, 0, 0]

with gsd.hoomd.open(name='lattice.gsd', mode='wb') as f:
        f.append(snapshot)

sim = hoomd.Simulation(device=hoomd.device.CPU())
sim.create_state_from_gsd(filename='lattice.gsd')

snap = sim.state.snapshot
print(snap.particles.types)
print(snap.particles.typeid)
```

## Output

```
['A']
[1 2]
```

## Expected output

I expect an error.

## Developer

Seeking interested developers. This could possibly be solved in the GSD reader and/or in the ParticleData snapshot initialization method.

joaander:
## Description

On initialization, check for invalid type ids.

## Motivation and context

When the user initializes a system with type ids that are invalid, various portions of HOOMD may access out of bounds memory and run invalid simulations or cause a segmentation fault.

This PR provides users with an understandable error message on initialization. Note that local snapshots provide direct access to underlying data, so it is still possible for users to set invalid type ids.

Resolves #838 

## How has this been tested?

This PR adds a unit test.

## Change log

```
Changed:

- Raise an exception when initializing systems with invalid particle type ids.
```

## Checklist:


hoomd/ParticleData.cc
```diff
@@ -1188,6 +1194,24 @@ void ParticleData::initializeFromSnapshot(const SnapshotParticleData<Real>& snap
...
+    // loops above) to avoid MPI communication deadlocks when only some ranks have invalid types.
+    // As a convenience, broadcast the value so that all ranks throw the exception.
+    #ifdef ENABLE_MPI
+    if (m_decomposition)
+        {
+        bcast(max_typeid, 0, m_exec_conf->getMPICommunicator());
+        }
+    #endif
+
+    if (max_typeid >= m_type_mapping.size())
```
bdice, line 1208: Is it valid to have a system with zero particles and zero types? If so, this would throw an error because `0 >= 0`. That might be undesired.

hoomd/pytest/test_snapshot.py
```diff
@@ -309,3 +309,16 @@ def test_from_gsd_snapshot_populated(s, device):
     gsd_snap = make_gsd_snapshot(s)
     hoomd_snap = Snapshot.from_gsd_snapshot(gsd_snap, device.communicator)
     assert_equivalent_snapshots(gsd_snap, hoomd_snap)
+
+
+def test_invalid_particle_types(simulation_factory, lattice_snapshot_factory):
```
b-butler, line 314: I think we should emphasis the distinction between `types` and `typeids`.
```suggestion
def test_invalid_particle_typeids(simulation_factory, lattice_snapshot_factory):
```

hoomd/pytest/test_snapshot.py
```diff
@@ -309,3 +309,16 @@ def test_from_gsd_snapshot_populated(s, device):
...
+    """Test that using invalid particle types raises an error."""
```
b-butler, line 315:
```suggestion
    """Test that using invalid particle typeids raises an error."""
```

b-butler (approved):
How does a person know what are valid values for `typeid`? I don't see the answer in the API documentation? The implementation seems to work.

joaander:
@cbkerr feel free to review if you like, but I did add you just as an example so don't feel obligated.

hoomd/ParticleData.cc
joaander: It is valid to have a gsd file with 0 particles in a frame. I would need to check and see how HOOMD behaves with 0 particles in the whole system. My guess is most of the code will behave fine - it does need to support 0 particles on an individual MPI rank.

I would guess that most code in HOOMD assumes there is at least one particle type. Do you think it would be valid to run a simulation with 0 particle types? There would be no way to specify pair interaction potentials. I guess you could have 0 particle types but 1 bond type and model gaussian chains... I would have to test and see whether those code paths allow that and whether those that require at least 1 type raise appropriate errors.

hoomd/ParticleData.cc
```diff
@@ -987,6 +988,9 @@ void ParticleData::initializeFromSnapshot(const SnapshotParticleData<Real>& snap
                 inertia_proc[rank].push_back(vec_to_scalar3(snapshot.inertia[snap_idx]));
                 tag_proc[rank].push_back(nglobal++);
                 N_proc[rank]++;
+
+                // determine max typeid on root rank
+                max_typeid = std::max(max_typeid, snapshot.type[snap_idx]);
```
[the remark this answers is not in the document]
joaander: Thanks. This would be a better name, but we can solve that in a larger code refactoring PR.

joaander:
I need to add a corresponding check for `BondedGroup` typeids.

hoomd/ParticleData.cc
mphoward: I regularly run HOOMD with no HOOMD particles and only MPCD particles. I removed a few checks on number of particles when I implemented MPCD to support this, but in general, everything seemed to work fine because ranks could already have zero particles.

I could see it being reasonable to have 0 particle types if and only if you have no HOOMD particles in a system. I don't think you can have a bond without particles, though, since bonds are defined between two particles; these would need to have a type.

joaander: > I could see it being reasonable to have 0 particle types if and only if you have no HOOMD particles in a system. I don't think you can have a bond without particles, though, since bonds are defined between two particles; these would need to have a type.

This makes sense. I will update the check accordingly.

joaander:
I was mistaken, `BondedGroupData.cc` did already have a check for invalid typeids. I cleaned up the error messages to include the full message in the exception reason.

I have addressed all comments, this should be good to merge.

PR #973, opened 2021-04-09
Fix setting operations in simulation when unscheduled

b-butler:
## Description

Fixes a bug where setting the `Simulation.operations` property when the `Operations` object was unscheduled did not do anything.

## Motivation and context

Fixes a lapse in the expected behavior.

## How has this been tested?

Has not yet, tests incoming.

## Change log

```
Fixed: Setting the operations attribute in ``Simulation`` objects in specific circumstances.
```

## Checklist:

### DIFF
--- a/hoomd/pytest/test_simulation.py
+++ b/hoomd/pytest/test_simulation.py
@@ -334,7 +334,20 @@ def test_seed(simulation_factory, lattice_snapshot_factory):
 def test_operations_setting(simulation_factory, lattice_snapshot_factory):
     sim = simulation_factory()
     sim.create_state_from_snapshot(lattice_snapshot_factory())
-    old_operations = sim.operations
+
+    def check_operation_setting(sim, old_operations, new_operations):
+        assert sim.operations is not new_operations
+        assert new_operations._simulation is None
+        assert old_operations._simulation is sim
+        scheduled = old_operations._scheduled
+        sim.operations = new_operations
+        assert sim.operations is new_operations
+        assert new_operations._simulation is sim
+        assert old_operations._simulation is None
+        if scheduled:
+            assert new_operations._scheduled
+            assert not old_operations._scheduled
+
     operations = hoomd.Operations()
     # Add some operations to test the setting
     operations += hoomd.update.BoxResize(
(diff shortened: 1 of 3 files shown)


hoomd/simulation.py
```diff
@@ -236,12 +236,12 @@ def operations(self, operations):
                     "Cannot add `hoomd.Operations` object that belongs to "
                     "another `hoomd.Simulation` object.")
             # Switch out `hoomd.Operations` objects.
-            elif self._operations._scheduled:
+            self._operations._simulation = None
+            operations._simulation = self
+            self._operations = operations
+            if self._operations._scheduled:
                 self._operations._unschedule()
```
joaander, line 243: Do we guarantee that `unschedule` will never reference `self._simulation`?

I would think it safer to:
```
reschedule = False
if self._operations._scheduled:
    self._operations._unschedule()
    reschedule = True

self._operations._simulation = None
operations._simulation = self
self._operations = operations

if reschedule:
    self._operations._schedule()
```

The `if reschedule` check may not be necessary as we don't necessarily need to maintain the scheduled status of the operations after setting.

joaander (commented):
Thanks, this looks like a needed fix.

We should include a test that fails before the fix and passes after.

hoomd/simulation.py
b-butler: In current implementation it does not ever, but we could do that. I am just thinking it may be confusing to users to need a `run(0)` call again to do some things, but I am open to either approach.

hoomd/pytest/test_simulation.py
```diff
@@ -329,3 +329,50 @@ def test_seed(simulation_factory, lattice_snapshot_factory):
...
+
+    assert sim.operations is not new_operations
+    assert sim.operations._simulation is sim
+    assert new_operations._simulation is None
+    sim.operations = new_operations
+    assert sim.operations is new_operations
+    assert new_operations._simulation is sim
+    assert new_operations._scheduled
+    assert not operations._scheduled
+    assert operations._simulation is None
```
tommy-waltmann, line 378: Many of these checks are done earlier in this test as well, perhaps some of these checks could be put in their own helper method? It could be called `_assert_before_after_reassign_operations`, or something like that.

hoomd/simulation.py
```diff
@@ -236,12 +236,17 @@ def operations(self, operations):
...
             # Switch out `hoomd.Operations` objects.
-            elif self._operations._scheduled:
+            reschedule = False
+            if self._operations._scheduled:
                 self._operations._unschedule()
-                self._operations._simulation = None
-                operations._simulation = self
-                operations._schedule()
-                self._operations = operations
+                reschedule = False
```
tommy-waltmann, line 242:
```suggestion
                reschedule = True
```
I think this will fix some of those failing tests.

tommy-waltmann (changes requested):

hoomd/simulation.py
tommy-waltmann, line 242:
```suggestion
                reschedule = True
```
I think this will fix some of those failing tests.

tommy-waltmann (approved):

joaander (approved):
Thanks!

PR #842, opened 2020-11-14
RATTLE integrators and manifold class

SchoeniPhlippsn:
## Description

Adding RATTLE integrators to hoomd version 3 both for CPU and GPU usage. This involves, in particular, the `NVE_Rattle`, `Brownian_Rattle` and `Langevin_Rattle` integrator. The `manifold` class is introduced as well. The following surfaces are implemented:
- Sphere
- Ellipsoid
- Plane
- Cylinder
- Gyroid
- Diamond
- Primitive

## Motivation and context

The RATTLE integrators facilitate to perform MD simulations of particles confined to surfaces. The algorithms are based on Paquay and Kusters, Biophys. J., 110, 6, (2016) [https://doi.org/10.1016/j.bpj.2016.02.017](url). The class of manifolds can easily be extended later. 

This way of confining particles would make `md.constrain.sphere` obsolete!

## How has this been tested?

The code works fine both on CPU and GPU. It runs accurately and also removes degrees of freedom according to the confinement. Additionally, I controlled the documentation with sphinx. Lastly, I added some first pytests, which check the attributes of the new integrators before and after attachment.

## Change log

```
Rattle Integrators and Manifold infrastructure
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

PR #982, opened 2021-04-16
Add style checks to CI tests

joaander:
## Description

Add `flake8`, `yapf`, and `clang-format` checks to the CI tests.

## Motivation and context

Use auto-formatting and auto-style check tools to reduce the number of code style PR comments and iteration cycles.

A corresponding mass-reformatting PR will need to be issued to fix these errors. I want to wait on submitting that until the outstanding PRs with major modifications have been merged:
* #944
* #842

Other open PRs should have resolvable merge conflicts.

## How has this been tested?

The GitHub Actions workflow `style_check.yml` tests this.

## Change log

These are developer focused changes and do not need a change log entry.

## Checklist:


joaander:
@vyasr @bdice You have been the most outspoken in the need for this.

.yapfignore
```diff
@@ -0,0 +1,2 @@
+hoomd/extern
+sphinx-doc/conf.py
```
bdice, line 2: I would go ahead and format this file, no need to make a special exclusion.

run-clang-format.py
```diff
@@ -0,0 +1,392 @@
+#!/usr/bin/env python
```
bdice, line 1: What's the benefit of this script over the standard CLI? pre-commit can handle the parallel parts if that's relevant. In general, I'd encourage the use of pre-commit over manual management of the hooks. I think freud/signac/etc. style checks have gotten a lot easier to maintain since adopting pre-commit. It's extremely automation-friendly.

setup.cfg
```diff
@@ -1,11 +1,32 @@
 [flake8]
-select = E,F,W,N,D
+select = E,F,W,N,D,RST
 ignore = W503,E126,E133,D107
-exclude = .git,__pycache__,sphinx-doc/conf.py,build,dist
+exclude = .git,
+    __pycache__,
+    sphinx-doc/conf.py,
```
bdice, line 6: I'd just make this file compliant instead of deal with an exception.

setup.cfg
```diff
@@ -1,11 +1,32 @@
...
+    run-clang-format.py,
+    build,
+    hoomd/extern,
 max_line_length = 80
```
bdice, line 10: I encourage `max_line_length = 100` or so (but leaving `max_doc_length = 80`).

setup.cfg
```diff
@@ -1,11 +1,32 @@
...
 max_doc_length = 80
 hang_closing = True
 docstring-convention = google
+rst-directives =
+    program,
+    option,
+rst-roles =
+    file,
```
bdice, line 18: Is there an order of priority here? If not, I'd sort them alphabetically.

bdice (commented):
I'm excited to see HOOMD get more automated style checks. I think it's fine to set this up and just let things fail until you can work through it over time. I'm also willing to put in some evening/weekend time on this if you want a hand with pre-commit. From my experience with freud and signac, I'm convinced this is a worthy investment in maintainability and code quality.

Overall, I would make two recommendations:
- Use pre-commit
- Separate the pre-commit hooks that can be installed by pre-commit (yapf, flake8) and can be run on any dev's machine from the hooks that require additional software and should only be run on CI or machines with that software installed (clang-format).

References from freud:
- https://github.com/glotzerlab/freud/blob/master/.pre-commit-config.yaml
- https://github.com/glotzerlab/freud/blob/master/.pre-commit-config-cpp.yaml

.github/workflows/style_check.yml
```diff
@@ -0,0 +1,47 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: clang_format
+            command: ./run-clang-format.py -r .
+          - name: flake8
+            command: flake8
+          - name: yapf
+            command: yapf -d -r .
```
bdice, line 33: If we decide against using pre-commit, I would advocate for combining the Python style checks. Run yapf first, then flake8. That way you ensure that the result of the formatting _pipeline_ is good, instead of having one action (yapf) reformat files and another action (flake8) validate them.

run-clang-format.py
joaander: clang-format does not have built in support for recursive runs or printing diffs. This script (which I copied from somewhere) does. I have no experience with `pre-commit`, so converting the scripts to that would require a large learning curve.

I do not plan to manually manage git hooks, just use the CI to make the checks on PRs and educate users (in `style.rst`) how to run the formatters when needed. Aside from the time to learn a new tool, I have two concerns with pre-commit hooks. 1) Formatting takes a long time - I don't want to wait minutes for checks to run every time I run `git commit`. 2) I've had issues in the past when running merge or rebase operations when intermediate commits fail the hooks.

setup.cfg
joaander: @b-butler strongly advocated for 80 chars early on in the HOOMD v3 development. We have been manually reformatting down to this since then.

setup.cfg
joaander: This is just the order in which they were added.

joaander: This config was copied from GSD by the way.

.github/workflows/style_check.yml
joaander: This invocation of `yapf` does not modify any files (the `-i` option is missing). It only prints the diff of what it would modify.

run-clang-format.py
bdice: I see, that could be helpful for clang-format. If it's useful, we used to use this snippet for freud:
```bash
        for f in cpp/**/*; do
          if ! diff $f <(clang-format-6.0 -style=file $f) > /dev/null; then
            echo "$f does not match clang-format style."
            diff --color $f <(clang-format-6.0 -style=file $f) || true
          fi
        done
```

Pre-commit hooks only run on changed files and execute in parallel by default, so your concern about time to run `git commit` might not be a problem. I never felt any significant slowdown (maybe 1-2 seconds) when using [cuDF's pre-commit hooks](https://github.com/rapidsai/cudf/blob/branch-0.20/.pre-commit-config.yaml) on a similarly large C++ / Python code base. One reason I like pre-commit is that users can opt in for local runs but don't _have_ to, you can always rely on CI to catch problems. However, there's not an easy way to execute these checks with a git hook for developers that prefer that workflow. I leave this up to you, just wanted to suggest it since I've been so happy with the tool.

bdice: Another aspect not already mentioned is versioning: developers using yapf / flake8 locally may not have the same versions as the container, and that can be quite troublesome. Using pre-commit allows much finer control over hook versions, and can be automatically updated for all developers using pre-commit if the configuration changes.

setup.cfg
b-butler: 80 characters is my preference ( I know that is old-school), if most developers feel differently though we can move to something else.

bdice: No need to change the standard at this point if a decision has already been reached. I’m fine with it. 👍

run-clang-format.py
joaander: Feel fee to contribute an equivalent `pre-commit` config to this PR. That would be faster than me attempting to learn how the system works.

Note that the docker image I'm using has the following flake8 plugins installed:
* pep8-naming
* flake8-docstrings
* flake8-rst-docstrings

setup.cfg
joaander: At this point, it's letting an automatic tool reformat the lines so it's not a big deal to change. I prefer 100 and we do use that for our C++ code.

run-clang-format.py
joaander: Also, the clang-format version needs to be relatively recent to support the config I use (>= 10).

vyasr:
Just to let you know, I am planning to have a look at this but have not had a chance yet.

.github/workflows/style_check.yml
```diff
@@ -0,0 +1,47 @@
+name: Style check
+
+on:
+  # Trigger on pull requests.
+  pull_request:
+
+  # Trigger on pushes to the mainline branches and version tags. This prevents building commits
```
vyasr, line 7: Does the version tags bit need to be specified below? Right now I only see the branch trigger.

setup.cfg
vyasr: I have no real preference as long as the choice is consistently enforced by an autoformatter.

setup.cfg
```diff
@@ -1,11 +1,30 @@
 [flake8]
-select = E,F,W,N,D
+select = E,F,W,N,D,RST
 ignore = W503,E126,E133,D107
```
vyasr, line 3: I personally find it quite helpful to add comments on why specific errors are ignored, it saves me having to look these errors up in the relevant tables online. Moreover, while sometimes the reasons are stylistic, other times it's to allow for certain specific edge cases that eventually become irrelevant at which point it's easier to see that the ignores can be removed.

setup.cfg
vyasr: I think I'm at least partially responsible for this ordering in GSD, I recall adding a number of these when rewriting some docs. Maybe the GSD CLI? Not sure.

vyasr (approved):
I second @bdice's vote for pre-commit, but it looks like he's already moved forward on that front with #983 so I'll leave that to him. Please double-check my comment regarding the version tags. Aside from that I have no major concerns and am definitely in favor of this change, it'll make reviews much simpler to not have to think about code style (much).

.clang-format-ignore
```diff
@@ -0,0 +1,12 @@
+# CMake test files
```
bdice, line 1: This can be removed, it is handled in the pre-commit configuration.

.github/workflows/style_check.yml
```diff
@@ -0,0 +1,47 @@
+name: Style check
```
bdice, line 1: This can be removed, it is redundant with the pre-commit configuration. The one piece that might be handy to copy would be the "single requirement for branch merge" job at the end.

run-clang-format.py
bdice: This can be removed, resolved by pre-commit #983.

setup.cfg
```diff
@@ -1,11 +1,30 @@
...
-exclude = .git,__pycache__,sphinx-doc/conf.py,build,dist
+exclude = .git,
+    __pycache__,
+    run-clang-format.py,
```
bdice, line 6: This line can be removed once the file is removed.
```suggestion
```

bdice (commented):
I made some notes about the clean-up tasks for this PR now that #983 has been merged.

.github/workflows/style_check.yml
joaander: This is a miscopied comment. There is no need to run style checks on tags.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,66 @@
...
+  hooks:
+  - id: flake8
+    # This option may be useful after we get HOOMD to 0 flake8 errors
+    # args:
+    # - --show-source
+    additional_dependencies:
+    - pep8-naming
+    - flake8-docstrings
+    - flake8-rst-docstrings
+    # exclude components yet yet ported to hoomd
```
bdice, line 42:
```suggestion
    # Exclude components not yet ported to hoomd v3
```

.yapfignore
```diff
@@ -0,0 +1 @@
+hoomd/extern
```
bdice, line 1: You could delete this file and control the exclusion via pre-commit's config if desired. Just a tradeoff between "works without pre-commit" and "having many files in the repo root."

ARCHITECTURE.md
```diff
@@ -25,8 +25,9 @@ compiler and build configurations, including:
...
 
 Visit the [workflows] page to find recent builds. The pipeline configuration
-files are in [.github/workflows/] which are built from templates in templates in
-[.github/workflows/templates/] using `make_workflows.py`.
+files are in [.github/workflows/] which are built from Jinja templates in
+[.github/workflows/templates/] using `make_workflows.py` which is automatically
+run by `pre-commit`. To make changes to the workflows, edit the templates.
 
 [GitHub Actions]: https://docs.github.com/en/actions
 [workflows]: https://github.com/glotzerlab/hoomd-blue/actions
```
bdice, line 33: The two lines below this have their links flipped, and a typo in `tempaltes`.

setup.cfg
```diff
@@ -1,14 +1,46 @@
...
+    # allow occasional use of uppercase variable and argument names (e.g. N)
+    N806,N803,
+    # match yapf formatting style
+    E126,E133,
+    # do not document __init__ separately from the class
+    D107
+exclude = .git,
+    __pycache__,
+    build,
+    hoomd/extern,
```
bdice, line 20: FYI, running bumpversion will strip all the comments in `setup.cfg`. Not sure if this affects HOOMD.

bdice (approved):
Looks good! I have a few minor comments to address.

setup.cfg
joaander: I use a separate `.bumpversion.cfg` to mitigate this issue. These comments will persist.

.yapfignore
joaander: done.

ARCHITECTURE.md
joaander: fixed.

PR #983, opened 2021-04-16
Pre-commit style checks

bdice:
## Description

Extension of #982 to add pre-commit support, locally and on CI.

### DIFF
--- a/.pre-commit-config-cpp.yaml
+++ b/.pre-commit-config-cpp.yaml
@@ -8,7 +8,7 @@ repos:
         name: clang-format
         entry: clang-format -i
         language: system
-        types: [c++]
+        types_or: [c++, cuda, inc]
         # Exclude specific files.
         exclude: |
           (?x)(
--- a/.pre-commit-config-cpp.yaml
+++ b/.pre-commit-config-cpp.yaml
@@ -1,0 +1,26 @@
+# This file contains pre-commit hooks for checking the C++ code. Since these
+# hooks require tools that may not be trivial to install (unlike the Python
+# hooks like flake8), we keep these in a separate file to be applied manually.
+repos:
+  - repo: local
+    hooks:
+      - id: clang-format
+        name: clang-format
+        entry: clang-format -i
+        language: system
+        types: [c++]
+        # Exclude specific files.
+        exclude: |
+          (?x)(
+            ^CMake/|
+            ^hoomd/extern/|
+            ^hoomd/hpmc/UpdaterClusters\.h|
+            ^hoomd/hpmc/IntegratorHPMCMono\.h|
+            ^doc/dev/
+          )
+        verbose: true
+        # Unclear if --style=file is necessary, it seems like clang-format uses
+        # .clang-format if found by default, at least for newer versions. I'll
+        # specify this here just in case.
+        args:
+        - --style=file
--- a/.pre-commit-config-cpp.yaml
+++ b/.pre-commit-config-cpp.yaml
@@ -19,8 +19,3 @@ repos:
             ^doc/dev/
           )
         verbose: true
-        # Unclear if --style=file is necessary, it seems like clang-format uses
-        # .clang-format if found by default, at least for newer versions. I'll
-        # specify this here just in case.
-        args:
-        - --style=file
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,15 @@
+repos:
+  - repo: https://github.com/google/yapf
+    rev: 'v0.31.0'
+    hooks:
+      - id: yapf
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: '3.9.1'
+    hooks:
+      - id: flake8
+        args:
+        - --show-source
+        additional_dependencies:
+          - pep8-naming
+          - flake8-docstrings
+          - flake8-rst-docstrings
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -20,4 +20,4 @@ jobs:
     - uses: actions/setup-python@v2
     - uses: pre-commit/action@v2.0.2
       with:
-        extra_args: -c .pre-commit-config-cpp.yaml
+        extra_args: --all-files -c .pre-commit-config-cpp.yaml
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -23,4 +23,4 @@ jobs:
     - uses: actions/setup-python@v2
     - uses: pre-commit/action@v2.0.2
       with:
-        extra-args: -c .pre-commit-config-cpp.yaml
+        extra_args: -c .pre-commit-config-cpp.yaml
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,26 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - uses: pre-commit/action@v2.0.2
+
+  pre-commit-cpp:
+    runs-on: ubuntu-latest
+    container:
+      image: glotzerlab/ci:2021.03-clang10_py38
+      options: -u 0
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+    - uses: pre-commit/action@v2.0.2
+      with:
+        extra-args: -c .pre-commit-config-cpp.yaml
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -15,9 +15,6 @@ jobs:
 
   pre-commit-cpp:
     runs-on: ubuntu-latest
-    container:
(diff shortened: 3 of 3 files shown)


bdice:
@joaander @b-butler Here is a pre-commit configuration (with CI) that closely resembles that of #982. The pre-commit CI configuration is somewhat shorter, slightly faster[1], and does not rely on custom containers. A possible challenge I foresee is ensuring the correct version of clang-format if we don't want the version provided in the `ubuntu-latest` image.

[1]:
- clang-format took 54 seconds in the container (#982) and 35 seconds with pre-commit (this PR). The only difference is that it doesn't have to pull the container, and uses pre-commit to execute instead of the script Josh added, `./run-clang-format.py`.
- yapf takes 1m23s and flake8 takes 45s in the container (#982) and 1m46s with pre-commit, both tools combined, in this PR.

bdice:
If we wanted an incremental approach that only runs style checks on changed files, we could adopt the approach in this comment:
https://github.com/pre-commit/action/issues/7#issuecomment-669149713

joaander:
> If we wanted an incremental approach that only runs style checks on changed files, we could adopt the approach in this comment:
> [pre-commit/action#7 (comment)](https://github.com/pre-commit/action/issues/7#issuecomment-669149713)

Would you recommend this configuration?

.pre-commit-config-cpp.yaml
```diff
@@ -0,0 +1,26 @@
...
+            ^hoomd/hpmc/UpdaterClusters\.h|
+            ^hoomd/hpmc/IntegratorHPMCMono\.h|
+            ^doc/dev/
+          )
+        verbose: true
+        # Unclear if --style=file is necessary, it seems like clang-format uses
+        # .clang-format if found by default, at least for newer versions. I'll
+        # specify this here just in case.
+        args:
+        - --style=file
```
joaander, line 26: This option shouldn't be necessary. It is the default.

.pre-commit-config-cpp.yaml
```diff
@@ -0,0 +1,26 @@
...
+# hooks require tools that may not be trivial to install (unlike the Python
+# hooks like flake8), we keep these in a separate file to be applied manually.
+repos:
+  - repo: local
+    hooks:
+      - id: clang-format
+        name: clang-format
+        entry: clang-format -i
+        language: system
+        types: [c++]
```
joaander, line 11: We should run clang-format on `.cu`, `.cuh`, and `.inc` files as well as standard C++ file types.

.pre-commit-config-cpp.yaml
```diff
@@ -0,0 +1,26 @@
+# This file contains pre-commit hooks for checking the C++ code. Since these
+# hooks require tools that may not be trivial to install (unlike the Python
```
joaander, line 2: `pre-commit`'s web page suggests that it handles the installation of even non-trivial tools. Does it not support `clang-format` in that way. It would be nice to have the whole config together.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+    hooks:
+      - id: yapf
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: '3.9.1'
+    hooks:
+      - id: flake8
+        args:
+        - --show-source
+        additional_dependencies:
+          - pep8-naming
```
joaander, line 13: Are there other flake8 plugins that we should be using?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
+repos:
+  - repo: https://github.com/google/yapf
```
joaander, line 2: Are there other formatter tools that we should be using?

joaander (commented):
Thanks! This looks reasonable. I don't have an issue relying on `clang-format` in `ubuntu-latest` for now, they provide several recent versions (9,10,11 - https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md). Which version is the default?

bdice:
> 
> 
> > If we wanted an incremental approach that only runs style checks on changed files, we could adopt the approach in this comment:
> > [pre-commit/action#7 (comment)](https://github.com/pre-commit/action/issues/7#issuecomment-669149713)
> 
> Would you recommend this configuration?

I think it is fine to run CI on all files, instead of only changed files. The only reason to adopt the incremental approach would be if the CI run time was too long. It's currently < 2 minutes, so I think the current configuration (all files) is good enough.

.pre-commit-config-cpp.yaml
bdice: Thanks, I wasn't sure about this. Vyas configured this for freud and I left this argument with the comment. I'll remove it now.

.pre-commit-config-cpp.yaml
bdice: This suggestion won't catch `.cuh` via `types` yet -- I opened https://github.com/pre-commit/identify/pull/195 to address that.
```suggestion
        types_or: [c++, cuda, inc]
```

bdice:
> Are there other flake8 plugins that we should be using?

> Are there other formatter tools that we should be using?

Good questions. From the pre-commit hooks for [coxeter](https://github.com/glotzerlab/coxeter/blob/master/.pre-commit-config.yaml) and [freud](https://github.com/glotzerlab/freud/blob/master/.pre-commit-config.yaml) and [signac](https://github.com/glotzerlab/signac/blob/master/.pre-commit-config.yaml), I would recommend adding these ones that can be automatically installed (no system installations needed like for clang-format):
- [isort](https://github.com/glotzerlab/freud/blob/ff9d561c2b8d719eea8e7167f083da3fbcc296a6/.pre-commit-config.yaml#L21-L24) -- keeps Python import statements sorted and categorized (standard library, 1st and 3rd party libraries, internal imports, etc.)
- [cmake-format](https://github.com/glotzerlab/freud/blob/master/.pre-commit-config.yaml#L35-L38)
- [pyupgrade](https://github.com/glotzerlab/freud/blob/ff9d561c2b8d719eea8e7167f083da3fbcc296a6/.pre-commit-config.yaml#L15-L20) -- replaces Python idioms from older versions with an equivalent newer syntax. e.g. `IOError` was made to behave the same as `OSError` in Python 3.3, so `except (IOError, OSError):` can be replaced with `except OSError:`.
- Some subset of the [pre-commit-hooks](https://github.com/pre-commit/pre-commit-hooks), which has things like removing trailing whitespace and preventing commits with debug statements like `breakpoint()`. Here's the [ones we use for freud](https://github.com/glotzerlab/freud/blob/ff9d561c2b8d719eea8e7167f083da3fbcc296a6/.pre-commit-config.yaml#L1-L14).

However, if this seems like a lot to adopt all at once, you can take it step by step and customize along the way. I got comfortable enough with mass-reformatting that I applied all the hooks at once to signac and some other repos.

.pre-commit-config-cpp.yaml
bdice: I looked at the [list of pre-commit hooks](https://pre-commit.com/hooks.html) that wrap clang-format. I think all of them require clang-format to be installed on the system. There is one version that can use a Docker container, but `docker` isn't any easier to fulfill as a requirement than `clang`. https://gitlab.com/daverona/pre-commit/cpp

.pre-commit-config.yaml
bdice: Addressed in top-level PR comment.

.pre-commit-config.yaml
bdice: Addressed in top-level PR comment.

bdice:
@joaander @b-butler @vyasr This is ready for a second pass of review. There are some pieces of #982 (e.g. the `run-clang-format.py` script) that could be deleted once this is merged into that branch.

.pre-commit-config-cpp.yaml
vyasr: The clang-format version that we were originally working with in freud was 6.0 I think, and it wouldn't search for `.clang-format` without passing this option explicitly. Agreed that this can be removed now.

vyasr:
Will have a look this weekend.

joaander (approved):
Thanks!

This looks like a great start. As you say, we can add additional tools iteratively. 

I'll deal with the additional cleanup necessary on the `test-style` branch after we merge the pre-commit configuration.

vyasr (approved):
This looks great to me, I think it's a good starting point and better to iterate with it in place rather than delay the PR.

I think I culled the list of `flake8` plugins I use in `coxeter` (and have since ported to `rowan`, if not elsewhere) from [this exhaustive list](https://github.com/DmytroLitvinov/awesome-flake8-extensions). In general I think more automated formatting validation is good because it establishes a clear standard of code that looks pretty nice and removes most room for debate, which saves valuable time. I think the set of plugins in this PR is probably fine though, the others are either not relevant (`flake8-black`) or may just provide noise in HOOMD (`flake8-bugbear`). The one exception is `isort`, which I definitely recommend since I've seen unnecessarily long threads in PR review about reordering imports, but that can easily be applied later.

I also added `clang-tidy` and `cppcheck` pre-commit hooks for `freud` and went through the code base to fix them. I think they're both great and good to have in place, but I also think they're also going to be _a ton_ of work for HOOMD, to the point where I'm not sure it's worth it. One option may be to apply them incrementally to one file at a time, maybe as a collection of hackathon tasks. That would also be a good way to teach people about (modern) C++ best practices.

b-butler (approved):
LGTM

Issue #770, opened 2020-09-04
Remove old logging infrastructure


joaander:
## Description

HOOMD v3 implements `Logger` in Python. We should remove the old unused C++ logging infrastructure.

## Motivation and context

This is a code clean up task and not critical for any particular release. The code is pervasive throughout almost every file in HOOMD, so this is best done in the later beta stages when developments are more targeted.

PR #984, opened 2021-04-20
Remove C++ Logging Infrastructure

### LINKED ISSUE #911: Add reservoir energy loggable
## Description

Add the reservoir energy loggable property to the NVT, NPT, and related integrators.

## Motivation and context

These were not implemented in the initial port.

b-butler:
## Description

Removes the logging infrastructure in C++ which has been replaced by a Python model.

## Motivation and context

This is unused now that logging is handled in Python.

Resolves #911, #770

## How has this been tested?

Existing tests pass, and this merely removes unused code.

## Change log
Since this is just a refactor, I don't know if we need an entry but...

```
Removed: C++ logging infrastructure

Added: loggable properties for integrator method reservoir energies.
```

## Checklist:


hoomd/md/special_pair.py
```diff
@@ -40,7 +40,7 @@ def _attach(self):
         else:
             cpp_cls = getattr(_md, self._cpp_class_name + "GPU")
         # TODO remove string name argument
```
joaander, line 42:
```suggestion
```

joaander (approved):
Thanks! This is a very nice clean up.

Any idea why all the gcc unit tests now have linker errors? We will need to solve those before merging.

b-butler:
Yes, I forgot to refactor a test that uses the logging infrastructure to test the class.

hoomd/md/methods.py
```diff
@@ -441,6 +443,20 @@ def thermalize_thermostat_and_barostat_dof(self):
         self._cpp_obj.thermalizeThermostatAndBarostatDOF(
             self._simulation.timestep)
 
+    @hoomd.logging.log
+    def thermostat_energy(self):
+        if not self._attached:
+            return None
+        else:
+            return self._cpp_obj.getThermostatEnergy(self._simulation.timestep)
```
joaander, line 451: Is this a fix for #911?

joaander (changes requested):
The compilation failures are:
```
[ 70%] Linking CXX executable test_external_periodic
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `PyTuple_SetItem'
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `PyObject_Repr'
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `PyInstanceMethod_Type'
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `PyExc_ValueError'
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `PyLong_AsLong'
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `PyDict_GetItemString'
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `PyExc_KeyError'
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `_Py_TrueStruct'
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `PyExc_IndexError'
../../_hoomd.cpython-36m-x86_64-linux-gnu.so: undefined reference to `PyFrame_FastToLocals'
```

I don't know what is causing this. It is not occurring on other branches, yet I don't see any changes to the CMake configuration that would cause this. It seems to only occur with the gcc compiler.

joaander:
@AlainKadar let us know if you have any questions about this PR as you review.

hoomd/md/methods.py
b-butler: Well this only handles the NVT, NPT, and NPH methods, but I think so. The functions were necessary for C++ testing, so I went ahead and added the Python interface.

joaander: Those are the methods that have a reservoir, so this PR will resolve #911.

joaander:
Root cause of the linker error is that cmake is not putting the python library at the correct place in the command line:
```
/usr/bin/g++-7   -Wall -Wconversion -Wno-sign-conversion -Wno-unknown-pragmas -Wno-deprecated-declarations -fdiagnostics-color=always -O3 -DNDEBUG   CMakeFiles/test_external_periodic.dir/test_external_periodic.cc.o  -o test_external_periodic -Wl,-rpath,/root/hoomd/build
/hoomd/md:/root/hoomd/build/hoomd ../_md.cpython-36m-x86_64-linux-gnu.so /usr/lib/x86_64-linux-gnu/libpython3.6m.so ../../_hoomd.cpython-36m-x86_64-linux-gnu.so ../../libquickhull.so
```
I'm not certain why or how this built previously and does not now, or how it continues to work with clang. I'm investigating to see if there is a way to get the python libs at the correct place in the linker order.

joaander:
There is no way to get the correct linker order without adding the python library as a dependency to `_hoomd` which we can't do. I added linker flags to ignore the test build undefined symbols.

joaander (approved):
Thanks!

joaander:
Now that #984 is merged into master, we need to merge master up into this branch and remove the old C++ logging code from the new integrators.

hoomd/md/methods.py
AlainKadar: Is the Langevin/Berendsen energy tracking feature being removed/ do they use a different method?

joaander: Neither Berendsen nor Langevin have equations of motion derived from a Hamiltonian, so there is no conserved quantity in those integrators to report.

b-butler:
@joaander is this last comment meant for a different PR?

PR #985, opened 2021-04-21
Convert SDF to v3 API

### LINKED ISSUE #880: Port analyze.sdf to the v3 API
## Description

Port analyze.sdf to the v3 API and add tests.

## Motivation and context

SDF should be implemented as a Compute and should provide its results as loggable quantities.

danevans09:
## Description
This converts the SDF class to the v3 API while also changing SDF to a `compute`. I created a `ComputeSDF.h` file that contains the code in `AnalyzerSDF.h` that calculates the sdf histogram, however the histogram is now only calculated over one step and not averaged over a number of steps. 

## Motivation and context

The sdf class is still an `analyzer` and has not yet been converted to the v3 API.

Resolves #880 

## How has this been tested?
I tested that the parameter dict works properly before and after attaching, however I did not write any validation tests.

## Change log

```

```

## Checklist:


tommy-waltmann (changes requested):
I would like to see AnalyzerSDF removed from the codebase before this gets merged.

joaander:
I will review this next week.

hoomd/hpmc/compute.py
```diff
@@ -107,3 +107,100 @@ def free_volume(self):
...
+        self._cpp_obj = cpp_cls(self._simulation.state._cpp_sys_def,
+                                integrator._cpp_obj,
+                                self.xmax,
+                                self.dx)
+
+        super()._attach()
+
+    @log
+    def sdf(self):
+        """Free volume available to the test particle."""
```
joaander, line 201:
```suggestion
        """Scale distribution function."""
```

joaander (changes requested):
Thanks! This is a great start. I've suggested some changes and given some next steps.

joaander (commented):
I made a few tweaks to the documentation and loggable method.

Before we merge I want to add a validation test in that ensure the correct pressure is computed. First I'm going to set up the framework for that kind of validation test on another branch.

In the meantime, this is ready for another round of review on the implementation and docs.

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -0,0 +1,35 @@
...
+def test_before_attaching():
+    sdf = hoomd.hpmc.compute.SDF(xmax=0.02, dx=1e-4)
+
+    assert sdf.xmax == 0.02
+    assert sdf.dx == 1e-4
+    assert sdf.sdf is None
+    assert sdf.betaP is None
+
+
+def test_after_attaching(simulation_factory, lattice_snapshot_factory):
```
tommy-waltmann, line 15: I'd like to see this test repeated for every integrator shape, because the python SDF class needs to use a different cpp class depending on which integrator is being used for the simulation.

tommy-waltmann (changes requested):
Good work Dan. There are a few more tweaks to make, and then this should be ready to go.

tommy-waltmann:
One more thing I forgot to mention, please add this class to `sphinx-doc/module-hpmc-compute.rst` so we can see how the documentation renders.

hoomd/hpmc/module_union_faceted_ellipsoid.cc
```diff
@@ -40,7 +40,7 @@ void export_union_faceted_ellipsoid(py::module& m)
     {
     export_IntegratorHPMCMono< ShapeUnion<ShapeFacetedEllipsoid> >(m, "IntegratorHPMCMonoFacetedEllipsoidUnion");
     export_ComputeFreeVolume< ShapeUnion<ShapeFacetedEllipsoid> >(m, "ComputeFreeVolumeFacetedEllipsoidUnion");
-    // export_AnalyzerSDF< ShapeUnion<ShapeFacetedEllipsoid> >(m, "AnalyzerSDFFacetedEllipsoidUnion");
+    // export_ComputeSDF< ShapeUnion<ShapeFacetedEllipsoid> >(m, "ComputeSDFFacetedEllipsoidUnion");
```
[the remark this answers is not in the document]
danevans09: I uncommented all of these and they all compile and attach without any errors except for `FacetedEllipsoid`.

hoomd/hpmc/compute.py
```diff
@@ -106,3 +107,114 @@ def free_volume(self):
...
+    def sdf(self):
+        """Scale distribution function."""
+        if self._attached:
+            self._cpp_obj.compute(self._simulation.timestep)
+            return self._cpp_obj.sdf
+        else:
+            return None
+
+    @log
+    def betaP(self):
```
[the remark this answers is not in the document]
joaander: `pressure` is a misnomer here because the value is not in pressure units. The `betaP` naming choice is consistent with `BoxMC`. We could instead update the documentation equation to use and introduce `\beta`. I don't care for the alternative name of `pressure_over_kT`

hoomd/hpmc/module_union_faceted_ellipsoid.cc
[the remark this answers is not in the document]
joaander: Why does `FacedtedEllipsoid` error when compiling?

hoomd/hpmc/compute.py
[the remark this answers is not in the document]
tommy-waltmann: I am fine with just modifying the docs to include \beta somwhere then.

b-butler:
@joaander @danevans09 I am removing myself from the reviewer list. If there are specific questions, you would like to ask feel free to mention me in them, and add me back if my review is needed.

hoomd/hpmc/module_faceted_ellipsoid.cc
```diff
@@ -42,7 +41,6 @@ void export_faceted_ellipsoid(py::module& m)
     {
     export_IntegratorHPMCMono< ShapeFacetedEllipsoid >(m, "IntegratorHPMCMonoFacetedEllipsoid");
     export_ComputeFreeVolume< ShapeFacetedEllipsoid >(m, "ComputeFreeVolumeFacetedEllipsoid");
-    export_AnalyzerSDF< ShapeFacetedEllipsoid >(m, "AnalyzerSDFFacetedEllipsoid");
     export_UpdaterMuVT< ShapeFacetedEllipsoid >(m, "UpdaterMuVTFacetedEllipsoid");
```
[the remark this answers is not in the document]
danevans09: All of the shapes are now supported

hoomd/hpmc/compute.py
```diff
@@ -102,3 +103,109 @@ def free_volume(self):
...
+        r"""Pressure in NVT simulations.
+
+        Use a polynomial curve fit to estimate :math:`\frac{s(0+)}{2d}` and
+        compute the pressure via:
+        .. math::
+            \frac{P}{kT} = \rho \left(1 + \frac{s(0+)}{2d} \right)
+        """
+        # get the values to fit
+        n_fit = int(numpy.ceil(self.xmax / self.dx))
+        sdf_fit = numpy.asarray(self.sdf[0:n_fit]) / sum(self.sdf[0:n_fit])
```
joaander, line 197:
```suggestion
        sdf_fit = self.sdf[0:n_fit]
```
Dividing by the sum is not the appropriate normalization. See the comments in `ComputeSDF.h` on how to correct `self.sdf` to be a probability density.

joaander (changes requested):
I merged master up into this branch and updated some of the documentation. I found several issues leading to the incorrect pressures you observed @danevans09 . Details in the comments.

joaander:
I tested the disk equation of state simulations offline, and found that SDF and `betaP` work. However, the simulation length needed for the system to equilibrate is too long to include in the CI validation tests. We will need to device a different system for executing very long running validation tests (perhaps another repository using sigac-flow) on HPC resources.

@danevans09 you can remove `test_disk_equation_of_state.py` in this PR. Please port the v2 test in https://github.com/glotzerlab/hoomd-blue/blob/2.9/hoomd/hpmc/test-py/test_sdf.py and add that one.

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -0,0 +1,179 @@
...
+    assert sdf.dx == 1e-4
+    with pytest.raises(hoomd.error.DataAccessError):
+        sdf.sdf
+    with pytest.raises(hoomd.error.DataAccessError):
+        sdf.betaP
+
+
+def test_after_attaching(valid_args, simulation_factory,
+                         lattice_snapshot_factory):
+    snap = lattice_snapshot_factory(particle_types=['A'])
```
joaander, line 22: This test fails on the GPU for two reasons. 1) The lattice snapshot factory makes 3D simulations by default and 2) Some of the particles in `valid_args` have rather large sizes.

@danevans09 can you please fix by adding the dimensionality of the system to `valid_args` and pass it to `lattice_snapshot_factory`. Also update the other HPMC tests that use `lattice_snapshot_factory` and `two_particle_snapshot_factory` to use this parameters. Please also adjust the vertices of the shapes in `valid_args` so that the diameter of the test shapes is close to 1. One of the convex polygon shapes currently has a diameter of 9.65539.

joaander (changes requested):

joaander:
@tommy-waltmann this PR is ready for another full review.

hoomd/hpmc/compute.py
```diff
@@ -106,3 +107,114 @@ def free_volume(self):
...
+
+        sdf = hoomd.hpmc.compute.SDF(xmax=0.02, dx=1e-4)
+
+
+    Attributes:
+        xmax (float): Maximum *x* value at the right hand side of the rightmost
+          bin.
+
+        dx (float): Bin width.
+
```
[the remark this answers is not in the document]
tommy-waltmann: Looking back at this and I can't remember what I meant. Let's just consider this resolved.

hoomd/hpmc/IntegratorHPMCMonoGPU.h
```diff
@@ -872,11 +872,20 @@ template<class Shape> void IntegratorHPMCMonoGPU<Shape>::update(uint64_t timeste
...
-            throw std::runtime_error("Error performing HPMC update");
+
+            oss << "nominal_width = " << this->m_nominal_width << std::endl;
+            if (global_box.getPeriodic().x)
+                oss << "nearest_plane_distance.x=" << nearest_plane_distance.x << std::endl;
+            if (global_box.getPeriodic().y)
+                oss << "nearest_plane_distance.y=" << nearest_plane_distance.y << std::endl;
+            if (this->m_sysdef->getNDimensions() == 3 && global_box.getPeriodic().z)
+                oss << "nearest_plane_distance.z=" << nearest_plane_distance.z << std::endl;
+            throw std::runtime_error(oss.str());
```
tommy-waltmann, line 888: This will now print the error message on every rank, I don't think we want that to happen.

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -0,0 +1,173 @@
...
+import pytest
+import numpy as np
+import hoomd.hpmc.pytest.conftest
+
+
+def test_before_attaching():
+    sdf = hoomd.hpmc.compute.SDF(xmax=0.02, dx=1e-4)
+
+    assert sdf.xmax == 0.02
+    assert sdf.dx == 1e-4
```
tommy-waltmann, line 13: Let's make sure we can set these parameters to some new values and get the new values back.

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -0,0 +1,173 @@
...
+    sim.operations.add(mc)
+
+    sdf = hoomd.hpmc.compute.SDF(xmax=0.02, dx=1e-4)
+
+    sim.operations.add(sdf)
+    assert len(sim.operations.computes) == 1
+    sim.run(0)
+
+    assert sdf.xmax == 0.02
+    assert sdf.dx == 1e-4
```
tommy-waltmann, line 47: Let's make sure we can set these values and get them back after attaching.

tommy-waltmann (changes requested):
This looks almost ready, just a couple small things.

hoomd/hpmc/IntegratorHPMCMonoGPU.h
joaander: It is better than the message not being printed at all. Somehow in the current implementation, the exception and resulting `MPI_Abort` call shut down the process before the stderr stream is flushed so these detailed error messages are lost more often than not. The details in the exception do get shown to the user. Especially in pytest output, the detailed message in the exception is easier to find.

I've bee slowly converting error message in HOOMD to this pattern. See the discussion in https://github.com/glotzerlab/hoomd-blue/issues/811

joaander (approved):
Thanks!

tommy-waltmann (approved):
LGTM

PR #990, opened 2021-04-30
Feature/anisolj

alacour:
## Description

I have started working on porting the anisotropic potential. I have ported over the GJK files and EvaluatorPairALJ. Within EvaluatorPairALJ, I moved the two structs inside the class EvaluatorPairALJ and renamed them type_param and shape_param. 

I also edited hoomd/md/par/aniso.py with the potential, including the documentation. I assumed that the pair potential behave similarly to those in HPMC for convex polyhedron.

## Motivation and context

People would like to use it in hoomd 3.0.

## How has this been tested?

It has not.
## Change log

```
Ported over the GJK files and EvaluatorPairALJ. 
Within EvaluatorPairALJ, moved the two structs inside the class EvaluatorPairALJ and renamed them type_param and shape_param. 
Edited hoomd/md/par/aniso.py with the potential, including the documentation.
```

## Checklist:

- [ x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

### DIFF
--- a/hoomd/VectorMath.h
+++ b/hoomd/VectorMath.h
@@ -95,11 +95,42 @@ struct vec3
         z = tz;
         }
 
+    DEVICE Real& operator[](const unsigned int i)
+        {
+        switch(i) {
+            case 0:
+                return x;
+            case 1:
+                return y;
+            case 2:
+                return z;
+            }
+        // This branch is unreachable, but must include something to avoid
+        // compiler warnings. The return value is chosen to match the non-const
+        // version of the operator.
+        return x;
+        }
+
+    DEVICE const Real operator[](const unsigned int i) const
+        {
+        switch(i) {
+            case 0:
+                return x;
+            case 1:
+                return y;
+            case 2:
+                return z;
+            }
+        // This branch is unreachable, but must include something to avoid
+        // compiler warnings. The return value is chosen to match the non-const
+        // version of the operator.
+        return x;
+        }
+
     Real x; //!< x-component of the vector
     Real y; //!< y-component of the vector
     Real z; //!< z-component of the vector
     };
-
 
 //! Addition of two vec3s
 /*! \param a First vector
--- a/hoomd/md/AllAnisoPairPotentials.h
+++ b/hoomd/md/AllAnisoPairPotentials.h
@@ -9,6 +9,7 @@
 
 #include "AnisoPotentialPair.h"
 
+#include "EvaluatorPairALJ.h"
 #include "EvaluatorPairGB.h"
 #include "EvaluatorPairDipole.h"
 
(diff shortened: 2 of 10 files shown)


vyasr:
Would it be possible to rebase the original branch and push it from the other remote? There are _a lot_ of useful historical commits that will be lost from a direct copy of the files. I'm happy to try and handle the initial merge/rebase if @alacour and @b-butler are fine with doing the follow-up work, I just haven't pushed on it since I had offered to pair with @b-butler on this at some point.

@b-butler I see your commit message regarding `operator []` for `vec3`. The problem actually goes beyond just needing refactoring. The GJK math requires setting those indices dynamically, which means that a lot of those `if` statements that are currently only 1-level deep would have to get converted to sets of nested `if` statements checking different properties. I don't see any way around that, and from what I recall the additional divergence I observed from that was a significant performance hit relative to the current approach (which admittedly does require a lookup in the L1 cache rather since registers can't be indexed for a vector, but that was still significantly faster).

b-butler:
@vyasr I wouldn't mind using a branch rebased on `hoomd-blue/master`; however, I anticipate it will be a lot of work. If you are volunteering to do so, the remaining work will be minimal, so I think we could handle it.

As for `vec3`, I saw when looking at the code the need for dynamic (i.e. using a variable of unknown value at compile time) for indexing which made me decide this was right for now. I thought about having a derived `indexable_vec3` type that supported indexing in addition to the standard vec3 behavior to ensure at the type level that we restrict this kind of behavior to where it is strictly necessary.

As regarding the pair programming, I didn't anticipate the necessary changes to be extensive enough to warrant bothering you with it.

vyasr:
@b-butler Yes I'll do it. I'll let you know when I get around to it. 

The `indexable_vec3` idea would be nice but would run into issues with dynamic polymorphism on the GPU. I had at one point defined an entirely new `gjk_vec3` class with just the stripped down functionality needed here and implemented with an underlying array rather than 3 attributes, but @joaander preferred adding the operator to `vec3` to adding that level of duplicated code. As an aside, this is the kind of commit history I'd like to retain in case for some reason we wanted to walk back that decision.

joaander:
> @b-butler Yes I'll do it. I'll let you know when I get around to it.

@alacour In the meantime, continue your efforts. There are many peeps in the group using ALJ now and it would be great if they could migrate to v3 sooner rather than later.

Regarding the `operator[]`, it has no impact on the data layout of the class and has no effect on the performance of other modules using `vec3` without the new operator. It is cleaner to add it to the existing class.

vyasr:
> @alacour In the meantime, continue your efforts. There are many peeps in the group using ALJ now and it would be great if they could migrate to v3 sooner rather than later.

I don't want to delay anything, so I did the port as best I can. I'm going to close this PR in favor of #992 and assign @alacour there.

> Regarding the `operator[]`, it has no impact on the data layout of the class and has no effect on the performance of other modules using `vec3` without the new operator. It is cleaner to add it to the existing class.

Agreed, it's definitely cleaner to do it this way. I only added the separate classes originally in the interest of not exposing a relatively inefficient operator that would have to be filtered out of future PRs using `vec3` and to squeeze out every bit of performance, but the number of operators that need to get implemented made that solution a pretty blatant DRY violation for very minimal gain.

b-butler:
Just to be clear, I don't like the [] operation because if I type `vec[4]` through incorrect use I get back out the `x` value which seems hard to debug.

vyasr:
Going to message offline to continue this discussion.

PR #992, opened 2021-05-08
Implements an anisotropic Lennard-Jones-like potential

vyasr:
## Description

This PR adds a new anisotropic pair potential based on [this paper](https://aip.scitation.org/doi/pdf/10.1063/5.0019735). It was originally implemented for the 2.x line of HOOMD, and needs to be ported to v3. In addition to any other changes to make this code fit well with v3 (@b-butler probably has a good idea of what's needed there), the following items remain to be addressed:

- [x] Enable the use of managed memory for `shape_params`. The potential is definitely usable without it, but in my recollection use of managed memory was something like a 20% performance bump (don't recall if that was 2D or 3D).
- [x] Add the ability to set default parameter values using a function, rather than a value (maybe already possible?). The `contact_sigma_[ij]` parameters should default to `0.15 * sigma_[ij]`, but unless things have changed recently the type parameter infrastructure only supported constant defaults. If necessary, we could push that logic into C++, but that would probably not be the preferred solution since nontrivial defaults would be useful for other HOOMD features as well.
- [x] Port the logic for constructing the shape parameters from a convex hull. I assume the majority of users want convex shapes, and the translation from vertices to a convex hull with vertices correctly ordered can be nontrivial, so we should maintain that logic if possible.
- [x] Add a citation to the paper. I'm not sure what the final decision was on handling citations in 3.0.
- [x] Investigate cleaner solutions to the default return value of `vec3<Scalar>::operator []`.

## Motivation and context

Replaces #990.

## How has this been tested?

There is a test that passed in 2.x, but needs to be updated to use the 3.x API once the rest of the code has been updated.

## Change log

```
Added:

* ``md.pair.ansio.ALJ`` - an anisotropic Lennard-Jones-like pair potential for polyhedra and ellipsoids.
* ``coxeter`` - optional dependency, needed for some ``ALJ`` methods.
```

## Checklist:


joaander:
> * Add the ability to set default parameter values using a function, rather than a value (maybe already possible?). The `contact_sigma_[ij]` parameters should default to `0.15 * sigma_[ij]`, but unless things have changed recently the type parameter infrastructure only supported constant defaults. If necessary, we could push that logic into C++, but that would probably not be the preferred solution since nontrivial defaults would be useful for other HOOMD features as well.

If we implement this, it should be on a separate PR. 

I do have a few questions about function based defaults. Will we allow users to set functions on parameters as well? How will we pickle them? How will the parameter check routines deal with the dependency tree of which parameters need to be evaluated in what order to provide the inputs to the functions? What are the specific other use-cases where we would want this? 

We could just change contact parameter for ALJ to a value that multiplies sigma and set the default to 0.15. We have many examples in the HOOMD v3 codebase where the parameters we expose to the users are transformed into a different set of parameters for C++.

vyasr:
> If we implement this, it should be on a separate PR.
> 
> I do have a few questions about function based defaults. Will we allow users to set functions on parameters as well? How will we pickle them? How will the parameter check routines deal with the dependency tree of which parameters need to be evaluated in what order to provide the inputs to the functions? What are the specific other use-cases where we would want this?
> 
> We could just change contact parameter for ALJ to a value that multiplies sigma and set the default to 0.15. We have many examples in the HOOMD v3 codebase where the parameters we expose to the users are transformed into a different set of parameters for C++.

All good questions. I'd be fine pushing the conversion down to C++ if that type of transformation is going to remain prevalent in 3.0, I wasn't sure how that was being handled in general with the conversion to the 3.0 type parameters. That is probably the easiest solution and would bypass the complexities of function-based defaults. If you think that all such cases can be generally handled by moving the transformation to C++ in this fashion, then I'd retract my comment that it would be preferable to do in Python.

hoomd/md/pair/aniso.py
```diff
@@ -232,3 +234,157 @@ def type_shapes(self):
...
+                      [5,4,6],
+                      [5,6,7],
+                      [3,2,0],
+                      [3,0,1],
+                      [3,6,2],
+                      [3,7,6],
+                      [3,1,5],
+                      [3,5,7]]
+
+        alj.params["A"] = dict(epsilon=2.0,
```
b-butler, line 343: `alj.params` requires two types as a pair potential something like `[('A', 'B')]`.

joaander:
See the recent changes on #965 needed here. Namely, the use of `allocate_shared` separate from `load_shared` which allows `const` to be used again. You also need to use `std::vector` with the managed allocator to manage data structures with nested managed arrays: See 77a5a2a

b-butler:
@alacour if you are still interested, the compilation is fixed, and this should be ready to test. If you could add the necessary lines for unit testing (see `hoomd/md/pytest/test_potential.py`) this should be nearly ready for merging.

b-butler:
This likely still has more changes, but now they should mostly be clean up and documentation (plus some small bug fixes potentially lets see how tests do), so I am removing the draft status.

sphinx-doc/testing.rst
```diff
@@ -34,6 +34,7 @@ optional requirements are missing.
 - pytest
 - rowan (optional)
 - CuPy (optional)
+- coxeter (optional)
```
joaander, line 37: Based on the currently failing tests, coxeter is not an optional dependency as advertised here. We should either a) hard code output from coxeter so that it is not even an optional dependency or b) modify the test with skips so that tests will run without coxeter installed.

setup.cfg
```diff
@@ -7,7 +7,7 @@ select =
     RST # flake8-rst-docstrings
 ignore =
     # line breaks should occur before the binary operator
-    W503,
+    W503, W504,
```
joaander, line 10:
```suggestion
    W503,
```
Code that triggers W504 needs to be fixed. We cannot allow line breaks both before and after the operator.

hoomd/md/pytest/test_alj_conservation.py
```diff
@@ -0,0 +1,115 @@
...
+    # Create box resize updater
+    packing_fraction = 0.4
+    final_area = hexagon.area * sim.state.N_particles / packing_fraction
+    L_final = np.sqrt(final_area)
+    final_box = hoomd.box.square(L_final)
+
+    n_compression_start = 1e4
+    n_compression_end = 1e5
+    n_compression_total = n_compression_end - n_compression_start
+    n_total = 1e6
```
joaander, line 55: How long does it take to run this test? If more than 2 minutes, we will need to move this test to the separate hoomd validation test suite that needs to be developed.

hoomd/md/pair/aniso.py
```diff
@@ -248,3 +251,206 @@ def type_shapes(self):
...
+    r"""Anistropic LJ potential.
+
+    Args:
+        r_cut (float): Default cutoff radius (in distance units).
+        nlist (:py:mod:`hoomd.md.nlist`): Neighbor list
+        name (str): Name of the force instance.
+        average_simplices (bool): Whether or not to perform simplex averaging
+            (see below for more details).
+
+    :py:class:`alj` computes the LJ potential between anisotropic particles.
```
joaander, line 266:
```suggestion
    :py:class:`ALJ` computes the Lennard-Jones potential between anisotropic particles.
```
Also, add a reference to the methods paper.

sphinx-doc/testing.rst
joaander: Coxeter is also not just a testing a dependency, but an optional runtime dependency - imported when a user invokes get_ordered_vertices. So it needs to be documented as an optional runtime dependency as well.

hoomd/md/pair/aniso.py
```diff
@@ -248,3 +251,206 @@ def type_shapes(self):
...
+
+    Specifying only ``rounding radii`` creates an ellipsoid, while specifying
+    only ``vertices`` creates a convex polytope (set ``vertices`` and ``faces``
+    to empty list to create the ellipsoid). To automate the computation of
+    faces, the convenience class method `get_ordered_vertices` can be used.
+    However, because merging of faces requires applying a numerical threshold to
+    find coplanar faces, in some cases `get_ordered_vertices` may result in not
+    all coplanar faces actually being merged. In such cases, users can
+    precompute the faces and provide them.
+
```
joaander, line 342: Add formulas that explicitly document the potential energy computed, using diagrams if needed. We can check with @vyasr and the publication agreement to see if we are able to use figures from the paper or if we need to create new ones from scratch.

joaander (changes requested):
Excellent, good to see this moving forward. I have some change requests.

hoomd/md/EvaluatorPairALJ.h
```diff
@@ -0,0 +1,1424 @@
...
+              average_simplices(params["average_simplices"].cast<bool>())
+            {
+            }
+
+        pybind11::object toPython()
+            {
+            using namespace pybind11::literals;
+            return pybind11::dict("epsilon"_a = epsilon,
+                                  "sigma_i"_a = sigma_i,
+                                  "sigma_j"_a = sigma_j,
```
tommy-waltmann, line 190: What is this `_a` syntax?

tommy-waltmann (changes requested):
Please also add this potential to `sphinx-doc/module-md-pair-aniso.rst` so we can see how the docs render

hoomd/md/pytest/test_alj_conservation.py
b-butler: We will likely need to move this. I have not run it locally to check.

hoomd/md/EvaluatorPairALJ.h
b-butler: It lets you create a dictionary all at once using a Python like syntax. It is a C++ literal that lets you create a string key for a dictionary.

hoomd/md/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ set(_md_headers ActiveForceComputeGPU.h
                 EvaluatorSpecialPairCoulomb.h
                 EvaluatorExternalElectricField.h
                 EvaluatorExternalPeriodic.h
+		EvaluatorPairALJ.h
```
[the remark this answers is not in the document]
b-butler: I probably did this on a Linux machine and that used tabs.

hoomd/md/pair/aniso.py
b-butler: I am not too familiar with the actual formulas. Honestly if @vyasr could add that section himself, or someone more familiar that would be great. I can eventually, but i don't know if that is the best use of my time.

hoomd/md/pytest/test_alj_conservation.py
b-butler: I have no idea how long this takes to run, TBH. I just ported Vyas's validation test to v3 and "called it good". Likely, this will need to be part of the separate validation suite.

vyasr: IIRC this test is not too slow and took around 2 minutes on V100s. If I'm remembering the correct test, it was definitely sub 3 minutes, but may have been slightly over 2.

hoomd/md/pair/aniso.py
vyasr: Yeah I'm happy to handle documenting this at some point in the next week or so.

vyasr: Well I obviously didn't get around to this in the promised time period, but taking a look again.

@joaander according to the author license [here](https://publishing.aip.org/wp-content/uploads/2019/10/AIPP-Author-License.pdf), we should be OK to reuse the figures from the paper. Specifically, the first clause under "Author Rights and Permitted Uses":
> Reprint portions of the Work (excerpts, figures, tables) in future works created by the Author, in keeping with professional publication ethics.
HOOMD documentation for this feature should count as a work created by an author (either of us), right?

vyasr: @b-butler I've updated the docstrings with the main formula and a bit more description of how it works. I also rendered the docs and they look fine to me except that the `params` and `shape` docs for the Type (the TypeParam) maybe need to indicate what the keys/values of the dict should be? I'm asking whether in a line like this one:
```
Type: `hoomd.data.typeparam.TypeParameter` [`tuple` [``particle_types``, ``particle_types``], `dict`]
```
the `dict` needs to have members in the same way that the `tuple`'s members are documented.

tommy-waltmann:
How was the data for the `aniso_forces_and_energies.json` file generated?

tommy-waltmann:
I would also like to clarify with someone who had worked on this PR previously: Is the fix for the bug that Kwang and Thi found on this branch right now, or is that something that still needs to be incorporated?

joaander:
> I would also like to clarify with someone who had worked on this PR previously: Is the fix for the bug that Kwang and Thi found on this branch right now, or is that something that still needs to be incorporated?

Yes, see the commit ed848a7c1acbdb22bf548c7e9e9875854457da52 bu `jekw`.

vyasr:
What was the issue that was uncovered here? I heard about a bug popping up a couple of times from a couple of the users of this code but I hadn't seen a full explanation of the issue and the fix. From the changeset it looks like an issue with finding the simplexes/faces that interact, was there a numerical problem with using the angles of each center-->vertex vector to determine the faces that is fixed by doing an exhaustive search? Do we know what the performance implications are?

tommy-waltmann:
> What was the issue that was uncovered here? I heard about a bug popping up a couple of times from a couple of the users of this code but I hadn't seen a full explanation of the issue and the fix. From the changeset it looks like an issue with finding the simplexes/faces that interact, was there a numerical problem with using the angles of each center-->vertex vector to determine the faces that is fixed by doing an exhaustive search? Do we know what the performance implications are?

@thiv89 implemented the fix, the exact details of which are unknown to me, but I do believe it had something to do with finding the right faces. You may need to reach out to him on another platform if you want all the information, he may not be very responsive on here,

joaander:
> > What was the issue that was uncovered here? I heard about a bug popping up a couple of times from a couple of the users of this code but I hadn't seen a full explanation of the issue and the fix. From the changeset it looks like an issue with finding the simplexes/faces that interact, was there a numerical problem with using the angles of each center-->vertex vector to determine the faces that is fixed by doing an exhaustive search? Do we know what the performance implications are?
> 
> @thiv89 implemented the fix, the exact details of which are unknown to me, but I do believe it had something to do with finding the right faces. You may need to reach out to him on another platform if you want all the information, he may not be very responsive on here,

@vyasr I worked with Kwang and Thi to troubleshoot a segfault. The simulation was with a rather simple shape (elongated bipyramid if I recall correctly). We tracked the issue to that face searching method which some numerical instability led to it returning the number of faces +1 - causing the seg fault. I asked Kwang to identify the geometry of the configuration that led to the seg fault in hopes of understanding how to avoid the instability. I expected a slight tolerance or change to an if condition. I never heard back until they long after the identified this rewrite as the fix. I haven't seen any performance tests on the changes either. As Tommy says, you would be best to contact them on other channels to learn the details. Let me know what you find out and if you have an alternate fix - I'd like to get this merged soon so we can release it in v3.0.

hoomd/md/pytest/test_alj.py
```diff
@@ -104,7 +109,7 @@ def test_conservation(simulation_factory, lattice_snapshot_factory):
     velocities = []
     total_energies = []
     while sim.timestep < n_total:
-        sim.run(1_000)
+        sim.run(1000)
```
b-butler, line 112: Why the change?

hoomd/md/pytest/aniso_forces_and_energies.json
```diff
@@ -6,16 +6,16 @@
...
       [
-        -5.385019740019003e-05,
+        -5.35167507608467e-05,
         -8.695546850996762e-08
       ],
       [
-        0.08737785466076176,
-        -0.2012594602508328
+        236.71142578125,
+        -0.29260432720184326
```
b-butler, line 18: This is the only one that is drastically different. Is this expected?

b-butler (commented):
A lot of logic (duplicate) has been added to `'findSimplex` and `findFace` would it be possible to create a function that both rely on that computes the indices which `findFace` uses, and in `findSimplex` are used to find the unique vectors?

@joaander @vyasr, and @jekw90 I know this code is designed to be as fast as possible so it this is ill-advised for performance reasons just let me know.

hoomd/md/pytest/test_alj.py
tommy-waltmann: Isn't the above statement syntactically incorrect?

vyasr:
> @vyasr I worked with Kwang and Thi to troubleshoot a segfault. The simulation was with a rather simple shape (elongated bipyramid if I recall correctly). We tracked the issue to that face searching method which some numerical instability led to it returning the number of faces +1 - causing the seg fault. I asked Kwang to identify the geometry of the configuration that led to the seg fault in hopes of understanding how to avoid the instability. I expected a slight tolerance or change to an if condition. I never heard back until they long after the identified this rewrite as the fix. I haven't seen any performance tests on the changes either. As Tommy says, you would be best to contact them on other channels to learn the details. Let me know what you find out and if you have an alternate fix - I'd like to get this merged soon so we can release it in v3.0.

I agree, let's prioritize getting this into a release ASAP. I'll reach out to them and try to get an example and track down where things go wrong to see if there's a simpler solution, and I'll take a look at what the performance implications are, but there's no need to delay this PR on that account alone. I can always patch it back in at a later date if necessary.

hoomd/md/pytest/test_alj.py
vyasr: No, Python allows you to use underscores as separators for readability in numbers as of Python 3.6 (specifically, [PEP 515](https://www.python.org/dev/peps/pep-0515/)).

vyasr:
> @joaander @vyasr, and @jekw90 I know this code is designed to be as fast as possible so it this is ill-advised for performance reasons just let me know.

I'll need to investigate more thoroughly to understand the nature of the problem that was fixed here. @thiv89 and @jekw90 implemented those changes to fix what I suspect is a numerical issue in the old calculation that involved using trig functions to convert dot products into angles. The new code seems to be based on directly computing intersections, which if anything I would expect to be significantly slower, but according to @thiv89 there wasn't much noticeable difference in performance and it prevented the system from exploding with certain geometries. @jekw90 would you be able to run a few tests to confirm that the change didn't slow down the code much? 

That said, I don't think adding a new function like @b-butler suggests would make this any less performant, and it would help clean things up, so I don't see any reason not to do that.

tommy-waltmann:
I have run a simulation of 10,000 ALJ hexagons in 2D with alpha=0 before and after making the bugfix on 1 GPU. The performance before the bugfix was `TPS=731.4067` and the performance after the bugfix was `TPS=728.078`. Timing numbers are the average TPS taken every 1000 time steps over the last 100,000 steps of the simulation run.

tommy-waltmann:
I have just run a simulation of cubes in 3D under all the same parameters as the previous simulations, the average TPS before the bugfix is `TPS=99.763` and after the bugfix is `TPS=115.861`. There seem to be some performance benefits in 3D.

vyasr:
That's awesome if the new version is faster rather than slower. I guess the new algorithm must converge the search faster on average than the old one, making up for any issues with more local state affecting register pressure etc on the GPU.

hoomd/md/pair/aniso.py
```diff
@@ -247,3 +250,248 @@ def type_shapes(self):
...
+                * ``vertices`` (:math:`(N_v, 3)` `numpy.ndarray` of `float`) -
+                  The vertices of the convex hull.
+                * ``faces`` (`list` [`list` [`int`]]) - The indices into the
+                  vertices of vertices defining the faces.
+        """
+        try:
+            import coxeter
+        except ImportError as error:
+            raise RuntimeError(
+                "Method requires coxeter as a dependency.") from error
```
joaander, line 464: Allow Python's `ImportError` to raise to the user's script. The message it provides may help the user understand why `coxeter` cannot be imported.

hoomd/md/pair/aniso.py
```diff
@@ -247,3 +250,248 @@ def type_shapes(self):
...
+          simplices. Defaults to ``True``. See class documentation for more
+          information.
+
+        Type: `hoomd.data.typeparam.TypeParameter` [`tuple` [``particle_types``,
+        ``particle_types``], `dict`]
+
+    .. py:attribute:: shape
+
+        The shape of a given type. The dictionary has the following keys per
+        type.
```
joaander, line 343:
```suggestion
        type:
```

hoomd/md/pair/aniso.py
```diff
@@ -247,3 +250,248 @@ def type_shapes(self):
...
+        self._extend_typeparam((params, shape))
+
+    @staticmethod
+    def get_ordered_vertices(vertices, return_faces=True):
+        """Compute vertices and faces of a convex hull of given vertices.
+
+        Warning:
+            This method requires the ``coxeter`` package which is available on
+            PyPi and conda-forge by that name, and can be found on GitHub under
+            ``glotzerlab/coxeter``.
```
joaander, line 442:
```suggestion
            This method requires the `coxeter <https://coxeter.readthedocs.io/>`_ package.
```
The coxeter documentation provides the single source of truth.

hoomd/md/pair/aniso.py
joaander: Should this be resolved?

hoomd/md/AnisoPotentialPair.h
```diff
@@ -285,12 +293,12 @@ AnisoPotentialPair<aniso_evaluator>::AnisoPotentialPair(std::shared_ptr<SystemDe
                       m_rcutsq.getNumElements() * sizeof(Scalar),
                       cudaMemAdviseSetReadMostly,
                       0);
-        cudaMemAdvise(m_params.get(),
-                      m_params.getNumElements() * sizeof(param_type),
+        cudaMemAdvise(m_params.data(),
```
joaander, line 296: I found this while debugging performance issues with @tcmoore3 on #1138. `m_rcutsq` is in unified memory when `allConcurrentManagedAccess()` returns `true`, but `m_params` and `m_shape_params` always in unified memory whenever running on the GPU.

Adjust this block accordingly:
```
if (isCUDAEnabled)
    {
    memadvise(m_params)
    memadvise(m_shape_params)
    prefetch(m_params)
    prefetch(m_shape_params)
    
    if (allConcurrentManagedAccess)
        {
        memadvise(m_rcutsq)
        prefetch(m_rcutsq)
        }
    }
````

File: hoomd/md/AnisoPotentialPair.h [hunk 15]
```diff
@@ -32,6 +32,8 @@
 
 #include <pybind11/pybind11.h>
 
+using namespace hoomd::detail;
```
joaander, line 35:
```suggestion
```
Add `detail::` where needed. `using` directives in header files will apply to all files that include it.

hoomd/data/typeparam.py
```diff
@@ -208,6 +208,7 @@ def to_base(self):
 
     def to_dict(self):
         """Alias for `to_base`."""
+        return self.to_base()
```
joaander, line 211: Alternately, remove the whole `to_dict` method.

sphinx-doc/testing.rst
joaander: Why was this resolved?

The documentation has not been updated. `coxeter` is a hard requirement to call the method `md.pair.anis.ALJ.get_ordered_vertices`. Therefore, it should be documented in the `Install prerequisites` section of `BUILDING.rst`.

joaander (changes requested):
Thanks, this is almost done. In addition to the inline change requests: fix the failing validation test: https://github.com/glotzerlab/hoomd-blue/runs/4971941936?check_suite_focus=true#step:7:174

hoomd/md/pair/aniso.py
b-butler: from error will point to the original error while adding our explanation.

joaander: I wasn't aware of this. There are many places in HOOMD we could use `from error` to help users and developers understand exceptions, such as the `AttributeError`s that we catch from C++ and re-raise as a different error.

tommy-waltmann:
Conceptual question here: If a validation test is failing solely due to statistics, does that mean I should change the test to run the simulation out longer, or adjust the tolerance on the standard deviation of the values?

joaander:
> Conceptual question here: If a validation test is failing solely due to statistics, does that mean I should change the test to run the simulation out longer, or adjust the tolerance on the standard deviation of the values?

It depends on the tolerance and sampling. One standard deviation corresponds to a ~68% confidence interval, so one should reasonably expect tests using one standard deviation to fail 32% of the time. The minimum tolerance that makes sense to me in a unit testing environment is 2-3 standard deviations. However, this assumes that you achieve good enough sampling that the measured standard deviation attains something close to the true standard deviation of the distribution. If you don't have sufficient sampling, then you will get inconsistent or erroneous test results.

There are a number of validation tests currently in HOOMD that fall into the insufficient sampling category. We need to create a `hoomd-validation` repository and move these tests there so their length may be extended and add scripts to launch the jobs on HPC resources.

hoomd/data/typeparam.py
tommy-waltmann: @b-butler are the `to_dict` and `to_base` methods supposed to be user-facing?

tommy-waltmann:
I've updated the coxeter related documentation so it is listed in the install prerequisites section, and documented it as a hard testing dependency.

hoomd/data/typeparam.py
joaander: Yes, `to_base` is a user-facing method. It replaces the deprecated `to_dict`.

tommy-waltmann:
> > Conceptual question here: If a validation test is failing solely due to statistics, does that mean I should change the test to run the simulation out longer, or adjust the tolerance on the standard deviation of the values?
> 
> It depends on the tolerance and sampling. One standard deviation corresponds to a ~68% confidence interval, so one should reasonably expect tests using one standard deviation to fail 32% of the time. The minimum tolerance that makes sense to me in a unit testing environment is 2-3 standard deviations. However, this assumes that you achieve good enough sampling that the measured standard deviation attains something close to the true standard deviation of the distribution. If you don't have sufficient sampling, then you will get inconsistent or erroneous test results.
> 
> There are a number of validation tests currently in HOOMD that fall into the insufficient sampling category. We need to create a `hoomd-validation` repository and move these tests there so their length may be extended and add scripts to launch the jobs on HPC resources.

How do I determine what the true standard deviations *should* be? The tolerances right now on energy and momentum conservation per particle are set at `1e-4` and `1e-6`, respectively, but I have no idea where those numbers came from.

joaander:
> I've updated the coxeter related documentation so it is listed in the install prerequisites section, and documented it as a hard testing dependency.

Sorry, my explanation wasn't clear. I updated the documentation to match what I recommended. `coxeter` is an optional runtime dependency, therefore does not need to be listed as a test dependency. The test dependencies are those only used by tests and not required to use hoomd at run time.

hoomd/md/pytest/test_alj.py
```diff
@@ -0,0 +1,246 @@
...
+        self._sim = sim
+        self._thermo = thermo
+        self._velocities = []
+        self._energies = []
+
+    def act(self, timestep):
+        with self._sim.state.cpu_local_snapshot as snap:
+            self._energies.append(self._thermo.potential_energy
+                                  + self._thermo.kinetic_energy)
+            self._velocities.append(np.array(snap.particles.velocity,
```
joaander, line 25: Momentum in the local snapshot will not be conserved as particles in the local domain interact with particles in neighboring domains. Compute the momentum for the total system.

`Integrator::computeTotalMomentum` does this, but it is not currently exposed to Python. I will add in another PR.

joaander (commented):
MPI tests are failing. The inline comment explains why.

hoomd/md/pytest/test_alj.py
```diff
@@ -0,0 +1,246 @@
...
+    sim.operations.writers.append(writer)
+    sim.run(1000)
+    total_energies = storer.total_energies
+    velocities = storer.velocities
+
+    # Ensure energy conservation up to the 3 digit per-particle.
+    assert np.std(total_energies) / sim.state.N_particles < 1e-3
+
+    # Test momentum conservation.
+    assert np.std(np.linalg.norm(np.sum(velocities, axis=1), axis=-1)) < 1e-6
```
joaander, line 147: Use `max` instead of `std`. The distribution of momentum magnitudes is positive only. Therefore, any fluctuations are not about 0.

Also, for both energy and momentum you should check E-E[0] and p-p[0] instead of the actual values of E and p.

tommy-waltmann: Are E and P in this case the average values of the energy and momentum magnitude, or the last value of the energy and momentum magnitude?

joaander: They are neither.

hoomd/data/typeparam.py
b-butler: Because the `hoomd.data.collection` classes have `to_base` method and not all of them are mapping types, so `to_base` allows a consistent API for converting to standard Python data types.

joaander (approved):
Thanks!

I pushed a commit that fixes the compile error with clang6.

b-butler (approved):
Thanks @tommy-waltmann for bringing it to completion.

PR #997, opened 2021-05-19
Run style formatters

joaander:
## Description

Run the pre-commit auto-formatters for #982.


joaander:
Note: I removed IMD rather than trying to fix its warnings. I don't think anyone has used this functionality in 10 years.

bdice (approved):
I skimmed some diffs and all looks good to me.

Issue #998, opened 2021-05-20
Error when detaching integration methods


joaander:
## Description

MD integration methods cannot be removed from the methods list after attaching.

## Script

```python
# Include a minimal script that reproduces the problem
import hoomd

snap = hoomd.Snapshot()
snap.particles.N = 2
snap.particles.types = ['A']
sim = hoomd.Simulation(hoomd.device.CPU(), seed=1)
sim.create_state_from_snapshot(snap)

integrator = hoomd.md.Integrator(dt=0.005)
nvt = hoomd.md.methods.Langevin(kT=1.5, filter=hoomd.filter.All())
integrator.methods.append(nvt)
sim.operations.integrator = integrator

sim.run(0)
integrator.methods.pop(0)
```

## Output

```
Traceback (most recent call last):
  File "/scratch/joaander/test/detach-method.py", line 15, in <module>
    integrator.methods.pop(0)
  File "/usr/lib/python3.9/_collections_abc.py", line 1100, in pop
    del self[index]
  File "/scratch/joaander/build/hoomd/hoomd/data/syncedlist.py", line 153, in __delitem__
    self._list[index]._detach()
  File "/scratch/joaander/build/hoomd/hoomd/operation.py", line 233, in _detach
    self._update_param_dict()
  File "/scratch/joaander/build/hoomd/hoomd/operation.py", line 279, in _update_param_dict
    self._param_dict[key] = getattr(self, key)
  File "/scratch/joaander/build/hoomd/hoomd/operation.py", line 46, in __getattr__
    return self._getattr_param(attr)
  File "/scratch/joaander/build/hoomd/hoomd/md/methods.py", line 82, in _getattr_param
    parameter = getattr(self._cpp_obj, attr)
AttributeError: 'hoomd.md._md.TwoStepLangevin' object has no attribute 'tolerance'
```

## Expected output

No output.

## Configuration

Platform:
- Linux
- CPU

Installation method:
- Compiled from source

Versions:
- Python version: 3.9
- **HOOMD-blue** version: master branch

## Developer

@b-butler do you have any ideas?

b-butler:
@joaander, yes I believe I know the case. It is somewhat subtle, but since the rattle methods have tolerances and manifolds, but the non-rattle one's do not. In `_update_param_dict` the Python side is querying for those quantities (since they are defined on the Python side), but they don't exist on the C++ side. A few possible solutions I briefly thought of

[...]

Issue #898, opened 2020-12-18
Move rotational diffusion to an Updater


joaander:
## Description

Move the rotational diffusion component of the active force compute into a separate updater.

## Motivation and context

The ActiveForceCompute class modifies particle orientations as part of the force computation. This breaks a number of assumptions in the HOOMD model that force computes only compute forces and do not modify the system state. 

For instance, the forces need to be recomputed when the particle sort is applied. ActiveForceCompute does not do this currently, so it actually may produce incorrect forces on particles on steps where the particle sort occurs.

PR #1003, opened 2021-05-22
Active forces with manifold constraints

### LINKED ISSUE #869: Remove oneD, sphere, and ellipsoid constraints

SchoeniPhlippsn:
## Description
Overhaul of active forces. I integrated manifolds to allow for active particles on curved surfaces and deleted the original ellipsoidal technique. Therefore, I divided the active force class into two separate classes ( constraint: `ActiveForceConstraintCompute` and unconstrained: `ActiveForceCompute`). All the manifold constraints can be used. 

## Motivation and context

The active force class contained old, unused code from the outdated ellipsodal constraints. By switching to manifold constraint, users can restrict active motion to more manifolds and apply active forces in combination with RATTLE integrators. 

Resolves  #869 and #936

## How has this been tested?

Pytests  have been updated and extended. I also ckecked the documentation.

## Change log

```
Adding manifold constraints to active forces.
```

## Checklist:


hoomd/md/ActiveForceConstraintCompute.h
```diff
@@ -0,0 +1,221 @@
...
+    ArrayHandle<Scalar4> h_orientation(m_pdata->getOrientationArray(),
+                                       access_location::host,
+                                       access_mode::readwrite);
+    ArrayHandle<Scalar4> h_pos(m_pdata->getPositions(), access_location::host, access_mode::read);
+
+    assert(h_f_actVec.data != NULL);
+    assert(h_pos.data != NULL);
+    assert(h_orientation.data != NULL);
+
+    if (!m_manifold.fitsInsideBox(m_pdata->getGlobalBox()))
```
j-proc, line 123: I'm not sue if it's necessary to check this here as I believe it will work out to once per timestep. This might be a very cheap operation, but assuming it is not, you could perhaps check it once when the manifold is set and then again each time the box is changed, assuming the manifold does not change with it. You can do this by subscribing the function to the box change signal.

hoomd/md/ActiveForceConstraintCompute.h
```diff
@@ -0,0 +1,221 @@
...
+        {
+        throw std::runtime_error("Parts of the manifold are outside the box");
```
j-proc, line 125: Most errors in HOOMD are done with the m_exec_conf objects. I believe this is to properly handle a variety of run environments that the std library may not.

hoomd/md/force.py
```diff
@@ -376,6 +363,21 @@ def __init__(self, filter, rotation_diff=0.1):
...
+                return self._param_dict["manifold_constraint"]
+            parameter = getattr(self._cpp_obj, attr)
+            return parameter
+        else:
+            return self._param_dict[attr]
+
+    def _setattr_param(self, attr, value):
+        if attr == "manifold_constraint":
+            raise AttributeError(
+                "Cannot set manifold_constraint after construction.")
```
j-proc, line 378: I wonder if it would make sense/be plausible to remove and recreate the associated objects post construction?

hoomd/md/CMakeLists.txt
```diff
@@ -50,6 +50,8 @@ set(_md_sources module-md.cc
 
 set(_md_headers ActiveForceComputeGPU.h
                 ActiveForceCompute.h
+		ActiveForceConstraintCompute.h
+		ActiveForceConstraintComputeGPU.h
```
j-proc, line 54: I think you may want to add a few more of your files to the various sources.

j-proc (commented):
In general this seems like a major cleanup of active forces while simultaneously generalizing the constraints. I had some minor comments but I think with your feedback on those and the style corrections of pre-commit this looks ready to go.

hoomd/md/ActiveForceConstraintCompute.h
joaander: The use of `std::runtime_error` is correct here.

Exceptions are handled properly even in MPI and provide a cleaner message to the user. There is another big cleanup project in the future to convert `error << msg; throw runtime_error` to `throw runtime_error(msg)` in all of hoomd.

joaander (changes requested):
In addition to some specific change requests, I wonder how these changes impact #898. We do need to switch to using an Updater to schedule the rotational diffusion updates, but with this PR those updates can now depend on the Manifold.

@b-butler what do you think? Should we move the rotational diffusion out into a completely separate class which would need to duplicate some of the manifold setup, or should we make an Updater that links to an ActiveForceCompute and leverages the existing code in these classes to implement the diffusion?

We don't have to implement #898 in this PR - but let's have a plan before we merge this one. We do need to eventually fix the issue because there are a number of issues with calling the rotational diffusion during the force computation.

hoomd/md/force.py
SchoeniPhlippsn: Usually it is not a good idea to switch constraints after initialisation as the forces become discontinuous. The only thing I can imagine, users might be interested in, is starting with a constraint and lifting it later. But this means detatching the restricted active force from the simulationa and attaching the unrestricted one afterwards anyway.

SchoeniPhlippsn:
Sorry that it took some time. However, this is open for review again.

hoomd/md/CMakeLists.txt
SchoeniPhlippsn: Which ones did I miss?

j-proc (approved):
This looks good to go to me. The minor possible changes can always be made enhancement issues but no reason to hold this up anymore.

j-proc:
Merge conflicts actually look a bit worse than I originally anticipated, I had thought I might help out. A rebase might be in order to clean it up, but outside the simple parts where both changes need to be removed, it seems like things @SchoeniPhlippsn should probably handle to make sure it stays correct.

hoomd/md/ActiveForceConstraintCompute.h
```diff
@@ -0,0 +1,248 @@
...
+                           h_f_actVec.data[type].z);
+            quat<Scalar> quati(h_orientation.data[idx]);
+            vec3<Scalar> fi = rotate(quati,
+                                     f); // rotate active force vector from local to global frame
+
+            Scalar dot_prod = dot(fi, norm);
+
+            Scalar dot_perp_prod = slow::rsqrt(1 - dot_prod * dot_prod);
+
+            Scalar phi_half = slow::atan(dot_prod * dot_perp_prod) / 2.0;
```
joaander, line 172:
```suggestion
            Scalar phi = slow::atan(dot_prod * dot_perp_prod);
```
See below.

hoomd/md/ActiveForceConstraintCompute.h
```diff
@@ -0,0 +1,248 @@
...
+
+        vec3<Scalar> norm = normalize(vec3<Scalar>(m_manifold.derivative(current_pos)));
+
+        Scalar delta_theta = hoomd::NormalDistribution<Scalar>(m_rotationConst)(rng);
+
+        Scalar theta
+            = delta_theta
+              / 2.0; // half angle to calculate the quaternion which represents the rotation
+        quat<Scalar> rot_quat(slow::cos(theta),
+                              slow::sin(theta) * norm); // rotational diffusion quaternion
```
joaander, line 127:
```suggestion
        quat<Scalar> rot_quat = quat<Scalar>::fromAxisAngle(norm, theta);
```
Simplify.

hoomd/md/ActiveForceConstraintCompute.h
```diff
@@ -0,0 +1,248 @@
...
+            fi.y -= norm.y * dot_prod;
+            fi.z -= norm.z * dot_prod;
+
+            Scalar new_norm = slow::rsqrt(dot(fi, fi));
+            fi *= new_norm;
+
+            vec3<Scalar> rot_vec = cross(norm, fi);
+            rot_vec *= fast::sin(phi_half);
+
+            quat<Scalar> rot_quat(cos(phi_half), rot_vec);
```
joaander, line 184:
```suggestion
            quat<Scalar> rot_quat = quat<Scalar>::(rot_vec, phi);
```
Simplify.

hoomd/md/ActiveForceConstraintComputeGPU.cuh
```diff
@@ -0,0 +1,219 @@
...
+
+        vec3<Scalar> f(fact.x, fact.y, fact.z);
+        quat<Scalar> quati(__ldg(d_orientation + idx));
+        vec3<Scalar> fi = rotate(quati, f);
+
+        Scalar dot_prod = fi.x * norm.x + fi.y * norm.y + fi.z * norm.z;
+
+        Scalar dot_perp_prod = slow::rsqrt(1 - dot_prod * dot_prod);
+
+        Scalar phi_half = slow::atan(dot_prod * dot_perp_prod) / 2.0;
```
joaander, line 85: Simplify this with the same changes as the CPU code.

hoomd/md/ActiveForceConstraintComputeGPU.cuh
```diff
@@ -0,0 +1,219 @@
...
+    hoomd::RandomGenerator rng(
+        hoomd::Seed(hoomd::RNGIdentifier::ActiveForceCompute, timestep, seed),
+        hoomd::Counter(ptag));
+
+    Scalar3 current_pos = make_scalar3(posidx.x, posidx.y, posidx.z);
+    vec3<Scalar> norm = normalize(vec3<Scalar>(manifold.derivative(current_pos)));
+
+    Scalar delta_theta = hoomd::NormalDistribution<Scalar>(rotationConst)(rng);
+    Scalar theta = delta_theta / 2.0; // angle on plane defining orientation of active force vector
+    quat<Scalar> rot_quat(slow::cos(theta), slow::sin(theta) * norm);
```
joaander, line 148: Simplify this with the same changes as the CPU code.

hoomd/md/ActiveForceConstraintComputeGPU.h
```diff
@@ -0,0 +1,245 @@
...
+    public:
+    //! Constructs the compute
+    ActiveForceConstraintComputeGPU(std::shared_ptr<SystemDefinition> sysdef,
+                                    std::shared_ptr<ParticleGroup> group,
+                                    Scalar rotation_diff,
+                                    Manifold manifold,
+                                    Scalar deltaT);
+
+    protected:
+    unsigned int m_block_size; //!< block size to execute on the GPU
```
joaander, line 43: Use an Autotuner in each of `ActiveForceComputeGPU` and `ActiveForceConstraintComputeGPU<Manifold>`. Different GPUs, system sizes, etc... lead to different optimal block sizes. `Autotuner` finds the fastest block size.

hoomd/md/CMakeLists.txt
```diff
@@ -47,6 +44,8 @@ set(_md_sources module-md.cc
 
 set(_md_headers ActiveForceComputeGPU.h
                 ActiveForceCompute.h
+                ActiveForceConstraintCompute.h
+                ActiveForceConstraintComputeGPU.h
                 AllAnisoPairPotentials.h
```
joaander, line 49: Add `ActiveForceConstraintComputeGPU.cuh` to this list.

hoomd/md/force.py
```diff
@@ -378,24 +377,45 @@ def _add(self, simulation):
...
         super()._add(simulation)
 
     def _attach(self):
+
         # initialize the reflected c++ class
-        if isinstance(self._simulation.device, hoomd.device.CPU):
-            my_class = _md.ActiveForceCompute
+        sim = self._simulation
+
+        if self.manifold_constraint is None:
```
joaander, line 390: Implement the manifold active force compute in a separate class to avoid issues like #998 for the integration methods.

hoomd/md/force.py
```diff
@@ -303,11 +291,10 @@ class Active(Force):
             It is defined per particle type and stays constant during
             the simulation.
         active_torque (tuple): active torque vector in reference to the
-            orientation of a particle
-            :math:`[\mathrm{force} \cdot \mathrm{length}]`.
-            It is defined per particle type and stays constant
-            during the simulation.
-
+            orientation of a particle. It is defined per particle type and stays
```
joaander, line 294:
```suggestion
            orientation of a particle :math:`[\mathrm{force} \cdot \mathrm{length}]`. It is defined per particle type and stays
```
Define units of the `active_torque` parameter. The text in this section will need to be wrapped appropriately after making this change.

joaander (changes requested):
Thanks for the updates. I have a few suggestions to simplify the vector math code and make it more readable by other developers, we need to use block size autotuners to improve performance, there is a missing header in the CMake script, and the active force manifold class needs to be separated from the non-manifold active force.

hoomd/md/force.py
```diff
@@ -381,21 +357,137 @@ def _add(self, simulation):
...
-            0,
+        self._cpp_obj = my_class(sim.state._cpp_sys_def,
+                                 sim.state._get_group(self.filter),
+                                 self.rotation_diff)
+
+        # Attach param_dict and typeparam_dict
+        super()._attach()
+
+
+class ActiveConstraint(Force):
```
joaander, line 377: I'm going to rename this class name to `ActiveOnManifold`. `ActiveConstraint` suggests that it is the constraint itself that is somehow active.

joaander (approved):
Thanks!

PR #1011, opened 2021-05-25
Improve documentation on building documentation

b-butler:
## Description

Adds more extensive documentation on building Python documentation and centralizes it.
Also adds a warning about Doxygen which isn't regularly tested.

## Motivation and context

This was a result of questions asked to me during the Glotzer group Hackathon.

## How has this been tested?

Documentation builds

## Change log

Likely not necessary but,
```
Added: More documenation on building sphinx documentation.
```

## Checklist:

### DIFF
--- a/INSTALLING.rst
+++ b/INSTALLING.rst
@@ -342,7 +342,6 @@ To open in a browser type in the address bar,::
 
     file:///path/to/sphinx-doc/_build/html/index.html
 
-**Common Pitfalls**
-
-- If when building there are errors regarding tutorials, submodules are either not installed or not
-  updated and must be before documentation can build without error.
+.. note::
+    The documentation build process requires git submodules. Either clone with the
+    ``--recursive`` option, or execute ``git submodule update --init`` to fetch the submodules.
--- a/INSTALLING.rst
+++ b/INSTALLING.rst
@@ -119,10 +119,9 @@ Install prerequisites
 
 **To build documentation**
 
-- Doxygen >= 1.8.5
-- Sphinx (>= 4.0 preferred)
-- sphinx_rtd_theme
-- nbsphinx
+- Doxygen >= 1.8.5 for C++ documentation (the C++ code is documented with comments, but Doxygen
+  documentation building and formatting is not regularly tested).
+- For Python see the :ref:`Build the documentation` section.
 
 Install these tools with your system or virtual environment package manager.
 HOOMD developers have had success with
@@ -296,13 +295,54 @@ These options control CUDA compilation via ``nvcc``:
 Build the documentation
 -----------------------
 
+Requirements
+
+- Sphinx (>= 4.0 preferred)
+- sphinx_rtd_theme
+- nbsphinx
+- IPython / Jupyter notebooks
+
+Install these packages using a package manager.
+This can be the same virtual environment as used for building the package.
+Here is an example creating of a virtual environment using Python's ``venv`` and ``pip`` from the
+base HOOMD directory:
+
+.. code-block:: shell
+
+    $ python3 -m venv ~/envs/hoomd-docs-env
+    $ . ~/envs/hoomd-docs-env/bin/activate
+    (hoomd-docs-env) $ python3 -m pip -r doc-requirements.txt
+
+With conda:
+
+.. code-block:: shell
+
+    $ conda create -n hoomd-docs-env python=3.9
+    $ conda activate hoomd-docs-env
+    (hoomd-docs-env) $ conda install -c conda-forge sphinx sphinx_rtd_theme nbsphinx jupyter
+
 Use sphinx to build the documentation::
 
     $ cd /path/to/hoomd-blue
     $ cd sphinx-doc
     $ make html
 
-Then, open the documentation in ``_build/html/index.html``. On some platforms, you can use this
-command to do open the file::
+Then, open the documentation in ``_build/html/index.html``. On some platforms, namely MacOS, you can
+use this command to do open the file::
 
     $ open _build/html/index.html
+
+For Linux,::
+
+    $ xdg-open _build/html/index.html
+
+should work.
+
+To open in a browser type in the address bar,::
+
+    file:///path/to/sphinx-doc/_build/html/index.html
+
+**Common Pitfalls**
+
+- If when building there are errors regarding tutorials, submodules are either not installed or not
+  updated and must be before documentation can build without error.
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -1,0 +1,4 @@
+sphinx >= 4.0
+sphinx_rtd_theme
+nbsphinx
+jupyter
--- a/sphinx-doc/conf.py
+++ b/sphinx-doc/conf.py
@@ -19,7 +19,8 @@ suppress_warnings = ['ref.any']
 extensions = [
     'nbsphinx', 'sphinx.ext.autodoc', 'sphinx.ext.autosummary',
     'sphinx.ext.napoleon', 'sphinx.ext.intersphinx', 'sphinx.ext.mathjax',
-    'sphinx.ext.todo', 'IPython.sphinxext.ipython_console_highlighting'
+    'sphinx.ext.todo', 'IPython.sphinxext.ipython_console_highlighting',
+    'sphinx.ext.autosectionlabel'
 ]
 
 napoleon_include_special_with_doc = True


doc-requirements.txt
```diff
@@ -0,0 +1,4 @@
+sphinx >= 4.0
```
joaander, line 1: Nothing else uses this file so it is likely to get out of date. Can we just list the requirements by pypi package name in the document instead? It is not a long or convoluted list.

I don't want to force devs to use conda, but there is a conda environment file in `sphinx-doc` that readthedocs uses when building the docs, so that one is tested.

sphinx-doc/conf.py
```diff
@@ -19,7 +19,8 @@
 extensions = [
     'nbsphinx', 'sphinx.ext.autodoc', 'sphinx.ext.autosummary',
     'sphinx.ext.napoleon', 'sphinx.ext.intersphinx', 'sphinx.ext.mathjax',
-    'sphinx.ext.todo', 'IPython.sphinxext.ipython_console_highlighting'
+    'sphinx.ext.todo', 'IPython.sphinxext.ipython_console_highlighting',
+    'sphinx.ext.autosectionlabel'
```
joaander, line 23: What does the autosectionlabel extension do?

INSTALLING.rst
```diff
@@ -296,13 +295,54 @@ These options control CUDA compilation via ``nvcc``:
...
+should work.
+
+To open in a browser type in the address bar,::
+
+    file:///path/to/sphinx-doc/_build/html/index.html
+
+**Common Pitfalls**
+
+- If when building there are errors regarding tutorials, submodules are either not installed or not
+  updated and must be before documentation can build without error.
```
joaander, line 348: This is redundant with the note further up in this document. I understand that some readers may want to just build the docs, but we could make a shorter reminder of that here:
```suggestion
.. note::
    The documentation build process requires git submodules. Either clone with the
    ``--recursive`` option, or execute ``git submodule update --init`` to fetch the submodules.
```

joaander (changes requested):

b-butler:
@joaander in lieu of responding to each comment since they are of a general theme, I am going to just respond in this comment for reduced duplication. 

In writing this PR, I was thinking of signac's (https://docs.signac.io/projects/core/en/latest/support.html), signac-flow's (https://docs.signac.io/projects/flow/en/latest/support.html), freud's (https://freud.readthedocs.io/en/latest/gettingstarted/installation.html), NumPy's (https://numpy.org/doc/stable/dev/index.html), and others' developer/contributor documentation. Nearly all show how to install the packages using pip and/or conda. In many cases if you take all the code snippets you could set up a complete development environment combined with a built package, built docs, run tests, and pre-commit installed and run. While I am not trying to appeal the bandwagon fallacy, these do point to a direction our lab's packages and popular external packages are going to encourage and help contributors.

I do generally agree that it is a bit verbose for people looking for a pure reference. Furthermore, HOOMD has many particular differences to many of the packages mentioned that make specific documentation in many parts more difficult than the examples presented above (some packages do try to tackle similar problems see  https://github.com/tonsky/FiraCode/wiki/Linux-instructions#installing-with-a-package-manager). However, new contributors come with various levels of experience, and from my experience many of these things that seem simple or even trivial are ultimately blockers to people contributing effectively to a project or end up being questions that have to be fielded by more experienced developers.

But we should probably field more options since we disagree here, and it would be good to get different options. Ultimately, if you feel strongly against this kind of contributor/developer/installation documentation, I don't mind backing down and being more terse in the documentation.

sphinx-doc/conf.py
b-butler: This means you can reference intra-dcoument sections automatically with their names.

joaander:
Not that long ago, a number of individuals were extremely upset when I made statements that they interpreted as asking them to use a specific tool. Since then, I have endeavored to always phrase such as weak recommendations and otherwise attempt to work with whatever tools users choose. As such, I will respect whatever virtual (or non-virtual) environment users want to use whether it is `apt-get`, `pacman`, `emerge`, `homebrew`, `conda`, `mamba`, `pip`, or anything else. I have written the compilation instructions for hoomd attempting to follow this. Though I do strongly suggest using python `venv` for virtual environments when users need to install into their home directory, I do allow for others.

The problem I personally have with `conda` as a development environment is that at various times over the years, `conda` environments have been completely broken for this purpose for a variety of issues, such as a `conda` provided linker that is not compatible with the standard library provided by the host. The working `conda-forge` build ecosystem that builds packages is rather complex and is not something I would expect developers to attempt to replicate.

The problem I have with providing only `pip install` instructions is that some users will likely run this inside a `conda` environment and break their environment because many pip packages provide embedded libraries that are a different version or have an ABI incompatible with packages provided by conda. `pip install` instructions are also not valid for system package managers like `apt-get`, except in the context of a virtual environment.

So, while instructions that can be copied and pasted are convenient for new developers, they are rigid and only work in the specific platforms or environment or environments that we dictate. We could attempt to circumvent this by breaking out the explicit compilation commands into many sections, one for each platform (e.g. ubuntu, arch linux, homebrew, macports, generic HPC system, ...), and one for each virtual environment (miniconda/miniforge, python venv, etc...). We would have to make the list rather long, especially if we need to include specific instructions for a variety of HPC systems. Also, just for conda, we have to choose whether we dictate the use of only `conda-forge` packages, or also provide working instructions for vanilla miniconda or even anaconda.

I specifically don't mention pre-commit in this document because it is not a prerequisite for compiling HOOMD and end users don't need it. I put only minimum instructions in `style.rst` because I expect devs can read the pre-commit documentation and choose to run it how they prefer. Also, I expect this to be a moot point very soon as everyone will take the lowest energy pathway of allowing `pre-commit.ci` to reformat their PRs.

So what will it be? Do you want to reformat the doc requirements in this PR into the generic "package manager" installation? Or should I instead write individualized instructions for all possible platforms and tools so that new developers can copy and paste the instructions matching their system?

@pepak13 you were auto-assigned to review this. What do you think?

sphinx-doc/conf.py
joaander: ok, well maybe it is the cause of all the sphinx build errors then?

```
/home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/deprecated.rst:5: WARNING: duplicate label v2.x, other instance in /home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/changelog.rst
/home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/migrating.rst:82: WARNING: duplicate label components, other instance in /home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/components.rst
/home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/style.rst:78: WARNING: duplicate label tools, other instance in /home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/style.rst
/home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/style.rst:87: WARNING: duplicate label documentation, other instance in /home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/style.rst
/home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/tutorial/00-Introducing-HOOMD-blue/02-Performing-Hard-Particle-Monte-Carlo-Simulations.ipynb:12: WARNING: duplicate label overview, other instance in /home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/tutorial/00-Introducing-HOOMD-blue/01-The-Simulation-Object.ipynb
/home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/tutorial/00-Introducing-HOOMD-blue/02-Performing-Hard-Particle-Monte-Carlo-Simulations.ipynb:15: WARNING: duplicate label questions, other instance in /home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/tutorial/00-Introducing-HOOMD-blue/01-The-Simulation-Object.ipynb
/home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/tutorial/00-Introducing-HOOMD-blue/02-Performing-Hard-Particle-Monte-Carlo-Simulations.ipynb:21: WARNING: duplicate label objectives, other instance in /home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1011/sphinx-doc/tutorial/00-Introducing-HOOMD-blue/01-The-Simulation-Object.ipynb
...```

We don't use unique header names across the whole codebase.

**Comment by b-butler** [comment]
Sorry, it seems I was misunderstood in part. I wasn't arguing that all these kinds of examples mentioned above should go in this PR or this file, it shouldn't. I was just trying to make a case for having commands that can be directly used by an end-user/contributor for tasks such as building docs from scratch (including installing packages),

I also was not trying to argue that installing HOOMD is as clean cut as it is for many of the packages I link to. For HOOMD's requirements, it would be difficulty to show every possible way to install the packages that makes sense. For instance since I run Kubuntu, I use `apt` and `apt-get` to install most of HOOMD's dependencies as it is the most convenient option for me, and I wouldn't want to bother with conda for that for reasons you mention. We could show how to do this for a few of the most common package managers to keep the list low such as Ubuntu (the most popular), brew (MacOS), and/or Arch (with pacman) with an explicit disclaimer that these are merely examples and users should use the package manager most convenient for them. These package manager commands should mostly be one line if I remember correctly. But maybe not showing specific instructions is the best here (besides the `venv` encouragement for installing hoomd-blue not system wide).

`pre-commit` has no place here, I completely agree. I just mentioned it as an example to support my argument. Although, I would argue keeping the 2-3 line code snippet in the docs is fine and not a maintenance burden.

This PR was designed to help those who may not be looking to contribute code changes but just documentation edits an easy way to do so. As new users of a package often notice the holes and obscurities of documentation an experienced user would gloss over, their help in documentation can be very beneficial to a package. Having these changes be as low energy to implement as possible is a net positive, I think.

Also, just as an aside I was surprised that people would interpret code snippets as a requirement they should use specific systems or tools, but I guess developers can be particular about their tools.

As a second aside, the resolution of this discussion would be a great inclusion in the developer guide to facilitate a constant level of detail. Meta-documenation! :smiley:

**Comment by joaander** [comment]
> Sorry, it seems I was misunderstood in part. I wasn't arguing that all these kinds of examples mentioned above should go in this PR or this file, it shouldn't. I was just trying to make a case for having commands that can be directly used by an end-user/contributor for tasks such as building docs from scratch (including installing packages),

I understood clearly, and I'm not suggesting a complete  rewrite of all the docs as part of this PR - just as a discussion topic for what direction do we want to take this as that decision affects what we do now.

Also, I'm not trying to suggest that HOOMD is any more complicated to build than other tools. I'm sorry if the way I have written the documentation comes across this way. It should be no more complex than installing any other package that needs to be compiled - including packages like freud, embree, tbb, Python. 1. Install prereqs, 2. Get source, 3. Configure, 4. Build. 5 (optional) build docs.

Complete, copyable instructions to build binaries and docs on ubuntu for example would be something like:
```
apt-get install build-essential cmake pybind11 ....
git clone --recursive https://github.com/glotzerlab/hoomd-blue
cmake -B hoomd-build -S hoomd-blue
cmake --build hoomd-build
sphinx-build -b html -d hoomd-blue/sphinx-doc/_build/doctrees  hoomd-blue/sphinx-doc  hoomd-blue/sphinx-doc/_build/html
```

Docs for the virtual environments would need to link to the system docs for details on getting a working build environment.

The problem isn't writing these instructions, it is keeping them up to date and organized.

File: INSTALLING.rst [hunk 4]
```diff
@@ -296,13 +295,54 @@ These options control CUDA compilation via ``nvcc``:
...
+
+For Linux,::
+
+    $ xdg-open _build/html/index.html
+
+should work.
+
+To open in a browser type in the address bar,::
+
+    file:///path/to/sphinx-doc/_build/html/index.html
```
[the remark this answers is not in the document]
vyasr: I agree that this is definitely unnecessary detail. Saying "open _build/html/index.html" should be sufficient.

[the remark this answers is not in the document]
b-butler: I would rather just remove it then, unless this works in Windows too which I don't think it does. I would argue that the browser address might be the most cross-platform way of doing this if we want to keep the information in some capacity.

vyasr:
My 2 cents:
- As the only realistic cross-platform option, I think it's nice to show how to install via `conda`. That said, I think that only extends to `conda install -c conda-forge hoomd` for installing HOOMD directly.
- Don't try to provide specific instructions on how to install HOOMD's non-Python dependencies. There are too many options and too many ways it can go wrong. The exception is something like the `install_prereq_headers.py` script, which is a one-liner maintained in this repo that can be controlled to give a reproducibly good result. If you do choose to show how to use `conda` or something else, some use of neutral language and acknowledging that YMMV would help blunt any criticism that you're making a preferred recommendation. A sentence like the one you have now indicating that "You can install these with your favorite package manager" is probably a preferable option to reduce maintenance overhead.
- For the purpose of docs, since you have to maintain a requirements file anyway for building on RTD I think it's sufficient to show only that. No need to show multiple ways to get doc prereqs.
- Nothing you do will make everyone happy, so it's probably best to start with a more stripped down set of instructions and add detail based on common questions rather than try to provide comprehensive instructions that leave lots of room for things to be wrong or fall out of date and make people unhappy.

joaander:
So @vyasr , should we move the doc build requirements back up to the main list of requirements and include those packages in the generic `your-package-manager` example install?

joaander:
I'm working on a more extensive rewrite of the build instructions now.

sphinx-doc/conf.py
joaander: I don't know how that plugin works. I checked the docs and they are rather sparse.

vyasr:
> So @vyasr , should we move the doc build requirements back up to the main list of requirements and include those packages in the generic `your-package-manager` example install?

Assuming you're only talking about the Sphinx docs, I think that's worth keeping separate. The requirements are almost entirely disjoint from the install requirements, and they are all Python dependencies that can be installed via `pip` or `conda`, so in that case I think it's fine to include the usual one-liners for pip and conda installs, perhaps with a recommendation to use a dedicated environment if you feel that's important for doc builds as well. I don't think users of various Linux distros will have any issues with that. They can use the distribution's package manager if they want but those packages shouldn't run into any ABI issues since you're not building against them.

C++ docs maybe merit their own (very small) subsection saying "install doxygen with your favorite package manager". I don't know anyone who does that for HOOMD though (I haven't in probably 4 years now) so probably not a big deal either way.

joaander:
Replaced with #1023

PR #1012, opened 2021-05-25
Wrap method for snapshot

### LINKED ISSUE #987: Wrap method for snapshots
## Description

Add a Snapshot method that wraps particles into the box, even if they are multiple images outside.

## Proposed solution

```snapshot.particles.wrap_all()```

## Additional context

General tools that create hoomd snapshots, e.g. mbuild, may not place particles inside the box as defined by HOOMD. At the same time, HOOMD expects snapshots to be well-formed with all particles in the box. Tool developers and users could use a convenience method to take an invliad configuration and make it valid. HOOMD-blue should not automatically call this method, so as not to modify user's data without their request.

We could also consider calling this method on the snapshot loaded internally by GSD to support GSD files with particles outside the box. Again, this should be an opt-in process only for users to activate when they knowingly have bad input data. GSD files with particles outside the box are malformed.

## Developer

Seeking developers.

### PR DESCRIPTION
**Pull request by cbkerr**
## Description

Add the ability to wrap particles back into the snapshot box.

During discussion with Josh and Brandon, we changed the API from the original issue #987 and made the wrap function a member of snapshot itself (in SnapshotSystemData as opposed to SnapshotParticleData) so the user doesn't have to specify which box they want to use for wrapping.

The new API is:
```
snapshot.wrap()
```  

## Motivation and context

Resolves #987 

## How has this been tested?

Added the `test_wrap` function in `test_snapshot.py`.

Added a function to generate points to test by adding multiples of lattice vectors to points known to be in the box. At least one of the test points in each case is at the lower bound of the box.

Boxes checked: cubic, triclinic, 2D, tetragonal, orthorhombic, monoclinic.

**Note** The tests might not work in single precision mode because of the specified `atol=1e-12`

## Change log

```
Added ability to wrap snapshot particles back into the box using snapshot.wrap().
```

## Checklist:

### CODE REVIEW

File: hoomd/SnapshotSystemData.cc [hunk 1]
```diff
@@ -39,6 +39,16 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
     pair_data.replicate(n, old_n);
     }
 
+template<class Real> void SnapshotSystemData<Real>::wrap()
+    {
+    for (unsigned int i = 0; i < particle_data.size; i++)
+        {
+        vec3<double> p = particle_data.pos[i];
+        global_box.wrap(p, particle_data.image[i]);
```
SchoeniPhlippsn, line 47:
```suggestion
        global_box.wrap(p, make_int3(0, 0, 0));
```
I'm not sure if wrapping the particles and counting the image can lead to problems. It really depends when the wrap feature is applied. If it is applied to make sure that this snapshot can be used as an initial configuration, I reckon that updating `particle_data.image[i]` messes up mean squared displacement calculations, for example. Or are the images set to zero before the simulation starts? 
For a snapshot after a simultion it makes sense to update the images. But here this feature is probably unneccesary, as HOOMD wraps the particles in the simulations and no particle should be outside the simulation box anyway.

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +125,18 @@ def test_configuration(s):
...
+def test_wrap(s):
+    if s.communicator.rank == 0:
+        s.configuration.box = [1, 1, 1, 0, 0, 0]
+        # set up particles outside of the box
+        s.particles.N = 2
+        s.particles.position[:] = [[1.5, 0, 0], [0.3, 0.2, 0]]
+        # wrap
+        s.wrap()
+        # check the particles are where they should be
+        assert numpy.all(s.particles.position == [[0.5, 0, 0], [0.3, 0.2, 0]])
```
SchoeniPhlippsn, line 137: I would test all periodic directions, just to be on the safe side.

hoomd/snapshot.py
```diff
@@ -310,6 +310,10 @@ def replicate(self, nx, ny, nz):
         """
         self._cpp_obj.replicate(nx, ny, nz)
 
+    def wrap(self):
```
SchoeniPhlippsn, line 313: In my opinion calling this function `wrap` is clear. However, you can also name it `wrap_to_box` which makes it a bit more specific.

SchoeniPhlippsn (dismissed):
I only have one major comment (see below). Otherwise, a nice feature. Thanks!

hoomd/pytest/test_snapshot.py
joaander, line 137:
```suggestion
        numpy.testing.assert_allclose(s.particles.position, [[0.5, 0, 0], [0.3, 0.2, 0]])
```
`assert_allclose` will pretty-print on failure. It also handles cases where floating point numbers may be slightly different due to round off error. I'm surprised this test works as is because 0.3 and 0.2 are not exactly representable in binary floating point.

joaander (changes requested):
Thanks!

I have one change request in the testing. We should also document the limitation that particles will not be moved more than one box image.

hoomd/SnapshotSystemData.cc
joaander: We do need to update the image to preserve the initial unwrapped position of the particle as the user provided it to us. For example, say a user creates molecules where the center of the molecule is in the box but part of the molecule leaves the box. Then they call wrap. After calling wrap, the molecule is broken across the box boundary, but the image flags can be used to unwrap the particles so the molecule is contiguous. 

This is the use case in https://github.com/mosdef-hub/mbuild/pull/871 where the new `snapshot.wrap` method will be used.

If a user desires all images to be reset to 0, then they can do so after calling wrap. We could add a note in the documentation for `wrap` that it modifies both `position` and `image`.

cbkerr:
I was working on a way to enable the "infinite wrap" case like Tim wanted. I pushed new code to do this. Should it be the default or should we include an option to only wrap once?

I discovered that we wrote our test without carefully consulting the docstring for the `BoxDim.wrap` method:
The box is centered at the origin with dimensions going from -1/2*box_length to +1/2*box_length.

Thus when I change our function to work until the wrap stops affecting the point, it was wrapping our points' x-coordinates to -0.5. If we had put in a point with x-coordinate of 0.5, `BoxDim.wrap` would have wrapped it back to -0.5 as well. This suggests we need more careful testing.

hoomd/pytest/test_snapshot.py
cbkerr: Thanks for this suggestion!

cbkerr: This turned out to be a good call :)
Will add those tests

joaander:
I don't think that wrapping only by a single image needs to be an option.

The only reason it exists in the hoomd source code is for performance and because it is the only things that we need for simulations.

hoomd/snapshot.py
bdice: We use `wrap` in freud, and I think it's clear enough.

hoomd/SnapshotSystemData.h
```diff
@@ -66,6 +66,9 @@ template<class Real> struct SnapshotSystemData
      */
     void replicate(unsigned int nx, unsigned int ny, unsigned int nz);
 
+    //! Wrap the snapshot
```
bdice, line 69: Be more explicit about the behavior. This is kind of long but there's probably a way to convey all of this information about expected inputs/outputs.
```suggestion
    //! Wrap the snapshot's particle positions into the box. Particle images are updated based on the number of wrapped images.
```

bdice (changes requested):

bdice:
Responding to @cbkerr's comment:
> Thus when I change our function to work until the wrap stops affecting the point, it was wrapping our points' x-coordinates to -0.5. If we had put in a point with x-coordinate of 0.5, BoxDim.wrap would have wrapped it back to -0.5 as well. This suggests we need more careful testing.

This suggests to me that the box bound documentation should be:

```diff
-                              -m_L.z/2 <= z <= m_L.z/2
+                              -m_L.z/2 <= z < m_L.z/2
-                       -m_L.y/2 + yz*z <= y <= m_L.y/2 + yz*z
+                       -m_L.y/2 + yz*z <= y < m_L.y/2 + yz*z
-        -m_L.x/2 + (xz-xy*yz)*z + xy*y <= x <= m_L.x/2 + (xz-xy*yz)*z + xy*y
+        -m_L.x/2 + (xz-xy*yz)*z + xy*y <= x < m_L.x/2 + (xz-xy*yz)*z + xy*y
```

Tests should make this behavior clear. I believe that freud strictly adheres to the convention of "open upper bound", like `[-L/2, L/2)`, with points on that boundary wrapping to `-L/2`.

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,23 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
...
+        {
+        new_pos = particle_data.pos[i];
+        old_pos = new_pos;
+        global_box.wrap(new_pos, particle_data.image[i]);
+        particle_data.pos[i] = new_pos;
+        }
+        while (new_pos != old_pos);
+        }
+    }
+
```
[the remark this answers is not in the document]
cbkerr: > re-implementing the lower level box wrap method
Perhaps instead this is where we could make a method `BoxDim.wrap_multiple` so that `BoxDim.wrap` stays fast for single wraps only?

joaander:
> Tests should make this behavior clear. I believe that freud strictly adheres to the convention of "open upper bound", like `[-L/2, L/2)`, with points on that boundary wrapping to `-L/2`.

Yes, HOOMD's implementation consistently uses a closed lower bound and an open upper bound. The quoted documentation is out of sync with the implementation.

hoomd/SnapshotSystemData.cc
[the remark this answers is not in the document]
joaander: We could mimic freud's implementation with a call to `makeFraction`, modding by 1, then `makeCoordinates` and updating the image by static_cast<int>(fraction).

joaander:
@cbkerr, @Nipuli-Gunaratne have you made progress on responding to this review?

Also, there have been a number of changes on master recently. I recommend merging master up into your branch regularly to keep up with recent developments.

cbkerr:
I updated the tests to use a method that generates test cases and updated the initial comment to reflect this.

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,40 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
...
+    {
+    vec3<double> frac;
+    vec3<double> wrapped;
+    vec3<double> pos;
+    vec3<double> fin;
+    int3 img;
+
+    for (unsigned int i = 0; i < particle_data.size; i++)
+        {
+        pos = particle_data.pos[i];
```
bdice, line 52: Declare these variables inside the loop, as `auto const`.
```suggestion
        auto const pos = particle_data.pos[i];
```

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,40 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
...
+        frac = global_box.makeFraction(particle_data.pos[i]);
+        wrapped = vec3<double>(
+            fmod(fmod(frac.x, 1.0) + 1.0, 1.0),
+            fmod(fmod(frac.y, 1.0) + 1.0, 1.0),
+            fmod(fmod(frac.z, 1.0) + 1.0, 1.0)
+            );
```
bdice, line 58: 1. Use a lambda to define the math expression only once.
2. Be explicit with the namespace `std::`. Someplace in this code has `using namespace std` but this is [considered bad practice](https://www.geeksforgeeks.org/using-namespace-std-considered-bad-practice/).
3. Use the template parameter `Real` precision unless @joaander disagrees.
```suggestion
        auto modulus_positive = [](auto x){ return std::fmod(std::fmod(x, Real(1.0)) + Real(1.0), Real(1.0)); };
        wrapped = vec3<Real>(
            modulus_positive(frac.x),
            modulus_positive(frac.y),
            modulus_positive(frac.z)
            );
```

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,40 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
...
+        fin = global_box.makeCoordinates(wrapped);
+        particle_data.pos[i] = fin;
```
bdice, line 60: Remove the unnecessary variable `fin`. It's generally better to use long variable names like `final_position` wherever a variable is needed.
```suggestion
        particle_data.pos[i] = global_box.makeCoordinates(wrapped);
```

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,40 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
...
+        img.x = static_cast<int>(frac.x);
+        img.y = static_cast<int>(frac.y);
+        img.z = static_cast<int>(frac.z);
+        particle_data.image[i] = img;
+        // TODO: test image update
+
+        std::cout << "Particle " << '[' << pos.x << ", " << pos.y << ", " << pos.z << "] --> ";
```
bdice, line 67: Remove debugging output.

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +125,104 @@ def test_configuration(s):
         assert s.configuration.dimensions == 3
 
 
+def test_wrap(s):
+
+    def generateOutside(box, inside, multipliers):
```
bdice, line 130: Python typically prefers snake case names.
```suggestion
    def generate_outside(box, inside, multipliers):
```

bdice (changes requested):
Some suggestions for improvement below.

hoomd/SnapshotSystemData.cc
cbkerr: Thanks for catching. Aimed to be temporary while I still had the print statements

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,32 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
...
+            modulus_positive(frac.x),
+            modulus_positive(frac.y),
+            modulus_positive(frac.z)
+            );
+        particle_data.pos[i] = global_box.makeCoordinates(wrapped);
+        img.x = static_cast<int>(frac.x);
+        img.y = static_cast<int>(frac.y);
+        img.z = static_cast<int>(frac.z);
+        particle_data.image[i] = img;
+        // TODO: test image update
```
cbkerr, line 64: How can we test this?

joaander: Well, as your test constructs the configuration byoving particles by box vectors, the image should be equal to the number of box vectors moved after wrapping.

hoomd/SnapshotSystemData.cc
cbkerr: I think I've implemented this now

hoomd/SnapshotSystemData.cc
cbkerr: Ah I didn't see that we do expose these values through `s.particles`

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,28 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
     pair_data.replicate(n, old_n);
     }
 
+template<class Real> void SnapshotSystemData<Real>::wrap()
+    {
+    for (unsigned int i = 0; i < particle_data.size; i++)
+        {
+        auto const pos = particle_data.pos[i];
+        auto const frac = global_box.makeFraction(particle_data.pos[i]);
```
bdice, line 47:
```suggestion
        auto const frac = global_box.makeFraction(pos);
```

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,28 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
...
+        auto const wrapped = vec3<Real>(
+            modulus_positive(frac.x),
+            modulus_positive(frac.y),
+            modulus_positive(frac.z)
+            );
+        particle_data.pos[i] = global_box.makeCoordinates(wrapped);
+        int3 img;
+        img.x = static_cast<int>(frac.x);
+        img.y = static_cast<int>(frac.y);
+        img.z = static_cast<int>(frac.z);
```
bdice, line 58: Use this function from `extern/cudacpu_vector_functions.h` to initialize the thing you want, rather than creating an object and then assigning it values.
```suggestion
        auto const img = make_int3(
            static_cast<int>(frac.x),
            static_cast<int>(frac.y),
            static_cast<int>(frac.z));
```

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,27 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
...
+        auto const wrapped = vec3<Real>(
+            modulus_positive(frac.x),
+            modulus_positive(frac.y),
+            modulus_positive(frac.z)
+            );
+        particle_data.pos[i] = global_box.makeCoordinates(wrapped);
+        auto const img = make_int3(
+            std::floor(frac.x),
+            std::floor(frac.y),
+            std::floor(frac.z)
```
joaander, line 57:
```suggestion
            static_cast<int>(std::floor(frac.x)),
            static_cast<int>(std::floor(frac.y)),
            static_cast<int>(std::floor(frac.z))
```
HOOMD should compile with no warnings, so you need the explicit cast here to tell the compiler it is OK to convert a Real to an int. Also, I'd suggest a comment explaining why floor is the right operation here. It's not intuitive why a simple static cast to an int doesn't work.

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +125,116 @@ def test_configuration(s):
...
+        box = [1, 1, 1, 0, 0, 0]
+        inside = [[0, 0, 0], [-0.5, 0.0, -0.2], [0.0, 0.3, -0.1],
+                  [0.3, 0.2, -0.1], [-0.5, 0.2, -0.2]]
+        outs, ins, mults = generate_outside(box, inside, multiples)
+        s.configuration.box = box
+        s.particles.N = len(ins)
+        s.particles.position[:] = outs
+        s.wrap()
+        numpy.testing.assert_allclose(s.particles.position, ins, atol=1e-12)
+        numpy.testing.assert_allclose(s.particles.image, mults)
```
joaander, line 173: You can use `assert_array_equal` here. The images are integers and do not need a tolerance check.

joaander (changes requested):
A few more tweaks to make.

joaander:
@cbkerr Can you address my latest review? I'd like to merge this PR soon.

joaander:
I needed to move one of the test points off the corner of the triclinic box.  6.7 is not exactly representable as a binary floating point value (the closest is 6.70000000000000017763568394002504646778106689453125), so it is not possible to place a particle exactly on corner of the box.

joaander (approved):
Thanks! This looks good.

joaander:
Reviewers, please take a look at this soon. This pull request is blocking https://github.com/mosdef-hub/mbuild/pull/871

cbkerr:
> I needed to move one of the test points off the corner of the triclinic box. 6.7 is not exactly representable as a binary floating point value (the closest is 6.70000000000000017763568394002504646778106689453125), so it is not possible to place a particle exactly on corner of the box.

As long as this was the only point failing that explained the few failing tests, I think it's okay to merge.

hoomd/SnapshotSystemData.cc
```diff
@@ -39,6 +39,24 @@ void SnapshotSystemData<Real>::replicate(unsigned int nx, unsigned int ny, unsig
...
+        auto modulus_positive
+            = [](Real x) { return std::fmod(std::fmod(x, Real(1.0)) + Real(1.0), Real(1.0)); };
+        auto const wrapped = vec3<Real>(modulus_positive(static_cast<Real>(frac.x)),
+                                        modulus_positive(static_cast<Real>(frac.y)),
+                                        modulus_positive(static_cast<Real>(frac.z)));
+        particle_data.pos[i] = global_box.makeCoordinates(wrapped);
+        auto const img = make_int3(static_cast<int>(std::floor(frac.x)),
+                                   static_cast<int>(std::floor(frac.y)),
+                                   static_cast<int>(std::floor(frac.z)));
+        particle_data.image[i] = img;
```
bdice, line 56: Do we need to consider wrapping particles that already have a nonzero image, and handling the image as an offset? I would expect that the `wrap` method is idempotent for particles inside the box, but this would reset their images to [0, 0, 0] instead of keeping the existing image flags.

This case should be tested, if you agree.
```suggestion
        particle_data.image[i] += img;
```

bdice (changes requested):
I think there's a case we're not considering: I think particles' existing image flags should be preserved (and added to) when calling `wrap`. I think this behavior would be important for users who may want to use the image flags, e.g., to compute an MSD.

@cbkerr It would also help if you can go back and click "Resolve conversation" on old comments if they've been addressed.

hoomd/SnapshotSystemData.cc
joaander: Yes, this would be good to add and test.

cbkerr: Good catch, will implement this weekend

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +125,128 @@ def test_configuration(s):
...
 
 
+def test_wrap(s):
+
+    def generate_outside(box, interior_points, multipliers, initial_images):
+        """Generate test cases from interior points by adding box vectors."""
+        # construct unit cell from box vectors
+        # see
+        # hoomd-blue.readthedocs.io/en/latest/...
+        # ...package-hoomd.html#hoomd.Box.from_matrix
```
bdice, line 135: Don't split URLs like this, it makes them too difficult to copy-paste. If the line is too long for flake8's standards, you can add a "noqa" flag to this comment. See here for an example: https://stackoverflow.com/a/47876831

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +125,128 @@ def test_configuration(s):
         assert s.configuration.dimensions == 3
 
 
+def test_wrap(s):
+
+    def generate_outside(box, interior_points, multipliers, initial_images):
```
bdice, line 130: @cbkerr An improved method of organization would be to write these helper functions outside of the `test_wrap` function, and then have multiple `test_*` functions that call those for each case being tested.

So like:

```python
def generate_outside(...):
    # Implementation here
    pass

def run_box_type(...):
    # Implementation here
    pass

def test_cubic_box(s):
    run_box_type(...)

def test_triclinic_box(s):
    run_box_type(...)
```

This way each test appears as a separate entry in the pytest output.

bdice (changes requested):
pytest code needs separate functions for each test case.

hoomd/pytest/test_snapshot.py
cbkerr: implemented this with keeping "wrap" in the name for running all those together `-k wrap`

bdice (changes requested):
Requesting test improvements.

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +125,134 @@ def test_configuration(s):
...
+
+def test_wrap_cubic(s):
+    if s.communicator.rank == 0:
+        run_box_type(s,
+                     box=[1, 1, 1, 0, 0, 0],
+                     interior_points=[[0, 0, 0], [-0.5, 0.0, -0.2],
+                                      [0.0, 0.3, -0.1], [0.3, 0.2, -0.1],
+                                      [-0.5, 0.2, -0.2]],
+                     multiples=multiples,
+                     initial_images=test_images)
```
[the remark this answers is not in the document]
cbkerr: I intended to generate test points as a cartesian product of the given initial points, multiples, and images, so the number of points checked (ie the length of the `input_*` and `check_*` arrays) is len(interior_points) * len(multiples) * len(test_images). 

Would it help to rewrite the for loop in `generate_outside` using `itertools.product`?

I named `multiples` and `test_images` separately in case there are certain images one wanted to check that are different from multiples.

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +125,134 @@ def test_configuration(s):
...
 
 
+def generate_outside(box, interior_points, multipliers, initial_images):
+    """Generate test cases from interior points by adding box vectors."""
+    # construct unit cell from box vectors
+    # see
+    # hoomd-blue.readthedocs.io/en/latest/package-hoomd.html#hoomd.Box.from_matrix # noqa
+    a = numpy.array([box[0], 0, 0])
+    b = numpy.array([box[1] * box[3], box[1], 0])
+    c = numpy.array([box[2] * box[4], box[2] * box[5], box[2]])
```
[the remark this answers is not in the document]
cbkerr: I started out doing everything manually to avoid using code used in the wrap itself, but this is just at the python level so good to simplify.

`f` was a shorter form of `factor` to help avoid as much line-wrapping as possible

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +126,131 @@ def test_configuration(s):
         assert s.configuration.dimensions == 3
 
 
+def generate_outside(box, interior_points, multipliers, initial_images):
```
bdice, line 129: There's some inconsistent naming here between `multiples` and `multipliers`. But neither name is clear that it relates to a number of lattice vectors to unwrap. Based on my experience with freud's naming conventions, I would suggest `unwrap_images` instead of `multiples` and `multipliers`, which also parallels the name `initial_images`. Examples of "unwrap":
- https://freud.readthedocs.io/en/latest/modules/box.html#freud.box.Box.unwrap
- https://hoomd-blue.readthedocs.io/en/latest/module-hoomd-write.html?highlight=unwrap#hoomd.write.DCD.unwrap_full
```suggestion
def generate_outside(box, interior_points, unwrap_images, initial_images):
```
...and likewise below. (`factor` would become `unwrap_image`, to parallel `initial_image`.)

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +126,131 @@ def test_configuration(s):
...
+    """Generate test cases from interior points by adding box vectors."""
+    box = Box.from_box(box)
+    lattice_vectors = box.lattice_vectors
```
bdice, line 132: I looked into the difference between `lattice_vectors` and `matrix` and filed an issue (#1070). Since this is using the transpose, it can be replaced by the matrix.
```suggestion
    matrix = box.matrix
```

hoomd/pytest/test_snapshot.py
```diff
@@ -125,6 +126,131 @@ def test_configuration(s):
...
+    input_points = numpy.zeros(
+        (len(interior_points), len(multipliers), len(initial_images), 3))
+    check_points = numpy.zeros_like(input_points)
+    input_images = numpy.zeros_like(input_points, dtype=int)
+    check_images = numpy.zeros_like(input_points, dtype=int)
+    for i, inside_point in enumerate(interior_points):
+        for j, factor in enumerate(multipliers):
+            for k, image in enumerate(initial_images):
+                input_points[i, j, k, :] = lattice_vectors.T @ numpy.array(
```
bdice, line 141:
```suggestion
                input_points[i, j, k, :] = matrix @ numpy.array(
```

hoomd/pytest/test_snapshot.py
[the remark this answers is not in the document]
bdice: See comment above -- this might be clearer if `multiples` were named `unwrap_images`.

bdice (changes requested):

bdice (approved):
Nice work @cbkerr!

PR #1013, opened 2021-05-25
Port external potentials to v3

tcmoore3:
## Description

Updates external potentials (electric field and periodic) to version 3.

## Motivation and context

v3

## How has this been tested?

We have basic tests now, but plan to add more extensive tests.

## Change log

```
Added
`md.external.Periodic` and `md.external.ElectricField` - external fields
```

## Checklist:


tommy-waltmann:
Downgrading this to a draft PR for now because there are some things I would still like to change, and I know after the hackathon we likely have our plates full with other PR reviews.

inhyukcho (commented):
This is looking good to me. It might not be the most detailed review considering my lack of knowledge on hoomd, but it seems reasonable and pasts tests, so great work!

hoomd/md/pytest/test_external.py
```diff
@@ -0,0 +1,125 @@
...
+        sim.run(10)
+
+        # test energies
+        new_snap = sim.state.snapshot
+        forces = sim.operations.integrator.forces[0].forces
+        energies = sim.operations.integrator.forces[0].energies
+        if new_snap.communicator.rank == 0:
+            F, E = evaluator(new_snap, param_dict)
+            np.testing.assert_allclose(F, forces)
+            np.testing.assert_allclose(E, energies)
```
b-butler, line 125: I would prefer longer more descriptive names like `expected_energies` and `expected_forces`.

b-butler (changes requested):
I have left some comments to address the biggest one is the possibility to change the electric external potential to use a tuple instead of a dict for specifying its per type dependency. We do this in `hoomd.md.pair.aniso.Dipole` with `mu` instead of `shape`. This seems to me to be another case where we should have the parameter name `E` replace `params` and use a tuple instead of a dict as the base entry in the type parameter.###

hoomd/md/external.py
```diff
@@ -18,352 +18,123 @@
...
 
-        # create the c++ mirror class
-        if not hoomd.context.current.device.cpp_exec_conf.isCUDAEnabled():
-            self.cpp_force = _md.PotentialExternalElectricField(
-                hoomd.context.current.system_definition, self.name)
-        else:
-            self.cpp_force = _md.PotentialExternalElectricFieldGPU(
-                hoomd.context.current.system_definition, self.name)
+        * ``E`` (`tuple` [`float`, `float`, `float`], **required**) -
+            The electric field vector :math:`\\vec{E}`.
```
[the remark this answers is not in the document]
tommy-waltmann: I would prefer the first version `obj.E['A'] = (x, y, z)` because it has the parameter name `E` in it.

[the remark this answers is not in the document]
tommy-waltmann: The version we have now (`obj.params['A'] = dict(E=(x, y, z))`) is too verbose, I agree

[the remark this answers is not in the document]
tommy-waltmann: @joaander any thoughts here?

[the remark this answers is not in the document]
joaander: Use `obj.E['A'] = (x, y, z)` here.

I think `params` only makes sense when using a dictionary that contains the named parameters.

b-butler (changes requested):
Thanks for the changes, I have a few more comments, but nothing major hopefully.

Also, I think I remember us deciding on having a `field` submodule with `external` and `wall` as part of that. @joaander is that what we decided on? If so, we should change this before merging.

joaander:
> Also, I think I remember us deciding on having a `field` submodule with `external` and `wall` as part of that. @joaander is that what we decided on? If so, we should change this before merging.

Or was it an `external` module with `field` and `wall` as members? I remember the discussion, but I don't remember the outcome.

hoomd/md/EvaluatorExternalElectricField.h
```diff
@@ -41,9 +41,35 @@ class EvaluatorExternalElectricField
...
+
+        param_type(pybind11::object params)
+            {
+            pybind11::tuple py_E(params);
+            E.x = pybind11::cast<Scalar>(py_E[0]);
+            E.y = pybind11::cast<Scalar>(py_E[1]);
+            E.z = pybind11::cast<Scalar>(py_E[2]);
+            }
+
+        pybind11::object asDict()
```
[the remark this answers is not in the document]
tommy-waltmann: This can be done for all external potentials (i.e. `PotentialExternalElectricField` and `PotentialExternalPeriodic`), converting all evaluators throughout the codebase might be out of scope for this PR.

tommy-waltmann:
Did we ever decide how we were going to refer to units? We talked about either doing something like [energy \cdot length^{-1}] or [energy][length]^{-1}. Did we ever come to a conclusion on that?

joaander:
Yes, #1019 sets the new style, and looks like it is ready for a final review:
```
:math:`[\\mathrm{velocity} \\cdot \\mathrm{time}^{-1}]`
```

hoomd/md/PotentialExternal.h
```diff
@@ -209,7 +239,10 @@ template<class T> void export_PotentialExternal(pybind11::module& m, const std::
     {
     pybind11::class_<T, ForceCompute, std::shared_ptr<T>>(m, name.c_str())
         .def(pybind11::init<std::shared_ptr<SystemDefinition>>())
-        .def("setParams", &T::setParams)
+        .def("setParams", &T::setParamsPython)
+        .def("getParams", &T::getParams)
+        .def("setE", &T::setParamsPython)
+        .def("getE", &T::getParams)
```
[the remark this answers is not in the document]
tommy-waltmann: I like this idea, but I have one suggestion. Instead of putting the template specializations in `module-md.cc`, I think it may be a better idea to put them in the header where the general template is defined for better organization. For example, that means I would put my template specialization in `PotentialExternal.h`, and we could move your Dipole specialization to `PotentialPairAniso.h`.

hoomd/md/external.py
```diff
@@ -5,9 +5,9 @@
 
 R""" External forces.
```
joaander, line 6: Replace this with just 
```"""External forces for molecular dynamics."""

File: hoomd/md/external.py [hunk 6]
```diff
@@ -18,352 +18,119 @@
...
-        i = coeff['i']
-        w = coeff['w']
-        p = coeff['p']
+        The `Periodic` external potential parameters. The dictionary has the
+        following keys:
 
-        return _hoomd.make_scalar4(_hoomd.int_as_scalar(i), A, w,
-                                   _hoomd.int_as_scalar(p))
+        * ``A`` (`float`, **required**) - Ordering parameter :math:`A`
+            (in energy units).
```
joaander, line 70: Match the units format in #1019

hoomd/md/module-md.cc
```diff
@@ -279,13 +295,13 @@ PYBIND11_MODULE(_md, m)
...
     export_PotentialExternal<PotentialExternalPeriodic>(m, "PotentialExternalPeriodic");
     export_PotentialExternal<PotentialExternalElectricField>(m, "PotentialExternalElectricField");
-    export_PotentialExternalWall<EvaluatorPairLJ>(m, "WallsPotentialLJ");
-    export_PotentialExternalWall<EvaluatorPairYukawa>(m, "WallsPotentialYukawa");
-    export_PotentialExternalWall<EvaluatorPairSLJ>(m, "WallsPotentialSLJ");
-    export_PotentialExternalWall<EvaluatorPairForceShiftedLJ>(m, "WallsPotentialForceShiftedLJ");
-    export_PotentialExternalWall<EvaluatorPairMie>(m, "WallsPotentialMie");
-    export_PotentialExternalWall<EvaluatorPairGauss>(m, "WallsPotentialGauss");
-    export_PotentialExternalWall<EvaluatorPairMorse>(m, "WallsPotentialMorse");
+    // export_PotentialExternalWall<EvaluatorPairLJ>(m, "WallsPotentialLJ");
```
joaander, line 298:
```suggestion
    // TODO: Port walls to HOOMD v3
    // export_PotentialExternalWall<EvaluatorPairLJ>(m, "WallsPotentialLJ");
```
Let's document why we are commenting this out.

joaander (changes requested):
In addition to the things we talked about, here are a few more change requests. 

Also, remove `external.py` from the ignore list in `setup.cfg` and `pre-commit-config.yaml` so that the pre-commit checks will run on this file.

hoomd/md/external.py
tommy-waltmann: I don't understand what this is supposed to to mean. Are you saying get rid of the line that says External forces, or are you saying get rid of the empty line below that one?

joaander: Make it consistent with the other packages that use just a one line description, like
```
"""Pair potentials."""
```
in `pair.py`.

There is no need for tutorial-like descriptions in the reference documentation. In any case `Integrator` documents the types of objects that can be added to the forces list. Check the `Integrator` docs to make sure that the external module is mentioned.

tommy-waltmann:
This is ready for another round of reviews, when you make your review, you'll have to build the docs locally, since for some reason github doesn't want to rebuild the docs based on the latest commits.

hoomd/md/external/field.py
```diff
@@ -0,0 +1,129 @@
...
+        * ``A`` (`float`, **required**) - Ordering parameter :math:`A` \
+            :math:`[\\mathrm{energy}]`.
+        * ``i`` (`int`, **required**) - :math:`\\vec{b}_i`, :math:`i=0, 1, 2`,
+            is the simulation box's reciprocal lattice vector in the :math:`i`
+            direction :math:`[\\mathrm{dimensionless}]`.
+        * ``w`` (`float`, **required**) - The interface width :math:`w`
+            relative to the distance :math:`2\\pi/|\\mathbf{b_i}|` between
+            planes in the :math:`i`-direction :math:`[\\mathrm{dimensionless}]`.
+        * ``p`` (`int`, **required**) - The periodicity :math:`p` of the
+            modulation :math:`[\\mathrm{dimensionless}]`.
```
bdice: Something in this section isn't rendering right. Perhaps the backslash escape in the first line?
https://hoomd-blue--1013.org.readthedocs.build/en/1013/module-md-external-field.html#hoomd.md.external.field.Periodic
image

hoomd/md/external/field.py
```diff
@@ -0,0 +1,129 @@
...
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeparam import TypeParameter
+
+
+class Field(force.Force):
+    """Common external field potential documentation.
+
+    Users should not invoke `Field` directly. Documentation common to all
+    external field potentials is located here. External potentials represent
+    forces which are applied to all particle in the simulation by an external
```
bdice, line 22:
```suggestion
    forces which are applied to all particles in the simulation by an external
```

tommy-waltmann:
> 
> 
> This is ready for another round of reviews, when you make your review, you'll have to build the docs locally, since for some reason github doesn't want to rebuild the docs based on the latest commits.

Nevermind, it appears the issue was on my end.

hoomd/md/external/field.py
```diff
@@ -0,0 +1,129 @@
...
+
+    where :math:`q_i` is the particle charge and :math:`\\vec{E}` is the field
+    vector. The field vector :math:`\\vec{E}` must be set per unique particle
+    types.
+
+    .. py:attribute:: E
+
+        The electric field vector, :math:`E`, as a tuple (i.e.
+        :math:`(E_x, E_y, E_z)`) \
+        :math:`[\\mathrm{energy} \\cdot \\mathrm{length^{-1}}]`.
```
joaander, line 112:
```suggestion
        :math:`[\\mathrm{energy} \\cdot \\mathrm{charge}^{-1} \\cdot \\mathrm{length^{-1}}]`.
```
Isn't there a charge^{-1} in the units of E? This would cancel out the units of charge from the `q` multiplier.

joaander (approved):
Thanks, this looks great!

I have one small comment on units that I think needs to be address. Other than that, I don't need to see this again.

hoomd/md/wall.py
```diff
@@ -479,7 +479,7 @@ def __repr__(self):
 #           *** Potentials ***
 
 
-class wallpotential(external._external_force):
+class wallpotential(external.field.Field):
```
b-butler, line 482: I think wall will inherit from a different class then `Field`, like an `External` base class in `hoomd/md/external/external.py`. Since this is not ported it is fine for now.

hoomd/md/external/field.py
tommy-waltmann: yes, good catch.

hoomd/md/external/field.py
```diff
@@ -0,0 +1,129 @@
...
+
+    The external potential :math:`V(\\vec{r})` is implemented using the
+    following formula:
+
+    .. math::
+
+       V(\\vec{r}) = A * \\tanh\\left[\\frac{1}{2 \\pi p w} \\cos\\left(
+       p \\vec{b}_i\\cdot\\vec{r}\\right)\\right]
+
+    The coefficients above must be set per unique particle type.
```
b-butler, line 54: Shouldn't this be below?

b-butler (commented):
I just have a couple comments, but this should be ready to merge soon.

hoomd/md/external/field.py
tommy-waltmann: I'm not sure what you mean here, below where?

b-butler: I was thinking the description of the attributes, but that doesn't make sense. I was more concerned with specifying the coefficients must be set referring to the equation, but that is probably too nit-picky.

b-butler (approved):
LGTM

PR #1014, opened 2021-05-25
Port the HPMC remove_drift updater to v.3.0 API

### LINKED ISSUE #514: Move HPMC remove_drift updater to v.3.0 API
## Description
Refactor the HPMC remove_drift updater to the HOOMD v.3.0 API. The template can be found in the `feature/new-object-API` branch's `Clusters` updater in `hoomd/hpmc/update.py`.

## Proposed solution
The solution involves following the template mentioned above.

## Additional context
Any pull request should merge into `feature/new-object-API` **not** `master`.

lyrivera:
## Description

This PR ports the HPMC remove_drift updater to v.3.0 API

## Motivation and context

Resolves #514 

## How has this been tested?

Not tested yet.

## Checklist:


hoomd/hpmc/UpdaterRemoveDrift.h
```diff
@@ -43,10 +42,46 @@ template<class Shape> class RemoveDriftUpdater : public Updater
...
+                       std::shared_ptr<IntegratorHPMCMono<Shape>> mc,
+                       pybind11::list ref_positions)
+        : Updater(sysdef), m_mc(mc)
         {
+        m_ref_positions.resize(m_pdata->getN());
+        setRefPositions(ref_positions);
+        }
+
+    //! Get reference positions as a python list of 3-tuples
+    pybind11::list getRefPositions()
```
b-butler, line 54: Given that this is `N` particles long. I think accepting a NumPy array would be preferable here, as that is typically how we store large numerical data (roughly the size of the simulation) versus smaller amounts of data like a shape definition or pair parameter parametrization.

hoomd/hpmc/update.py
```diff
@@ -490,72 +490,56 @@ def N(self):  # noqa: N802 - allow N as a function name
...
 
     Args:
-        mc (:py:mod:`hoomd.hpmc.integrate`): MC integrator.
-        external_lattice (:py:class:`hoomd.hpmc.field.lattice_field`): lattice
-          field where the lattice is defined.
-        period (int): the period to call the updater
-
-    The command hpmc.update.remove_drift sets up an updater that removes the
-    center of mass drift of a system every period timesteps,
+        ref_positions (`list` [`tuple` [`float`, `float`]], **required**) - the
```
b-butler, line 497: This presumes 2D currently since the list says two floats. I would prefer to see this make a NumPy array.

hoomd/hpmc/update.py
```diff
@@ -490,72 +490,56 @@ def N(self):  # noqa: N802 - allow N as a function name
...
-            else:
-                hoomd.context.current.device.cpp_msg.error(
-                    "update.remove_drift: Unsupported integrator.\n")
-                raise RuntimeError("Error initializing update.remove_drift")
-        else:
-            raise RuntimeError(
-                "update.remove_drift: Error! GPU not implemented.")
+    def __init__(self, ref_positions, trigger=1):
+        super().__init__(trigger)
+        _param_dict = dict(ref_positions=list(ref_positions), trigger=trigger)
```
b-butler, line 513: `trigger` is not needed in the dict here. Also, no validation of the `ref_positions` is being done here.

hoomd/hpmc/update.py
```diff
@@ -490,72 +490,56 @@ def N(self):  # noqa: N802 - allow N as a function name
...
-                               external_lattice.cpp_compute, mc.cpp_integrator)
-        self.setupUpdater(period)
+    def _add(self, simulation):
+        """Add the operation to a simulation.
+
+        HPMC uses RNGs. Warn the user if they did not set the seed.
+        """
+        if simulation is not None:
+            simulation._warn_if_seed_unset()
+
```
b-butler, line 523: This particular operation does not use the seed as far as I am aware though, so this warning is not needed.

hoomd/hpmc/update.py
```diff
@@ -490,72 +490,56 @@ def N(self):  # noqa: N802 - allow N as a function name
...
+        integrator = self._simulation.operations.integrator
+        if not isinstance(integrator, integrate.HPMCIntegrator):
+            raise RuntimeError("The integrator must be a HPMC integrator.")
+
+        if not integrator._attached:
+            raise RuntimeError("Integrator is not attached yet.")
+
+        if isinstance(self._simulation.device, hoomd.device.GPU):
+            raise RuntimeError("GPU not supported.")
+
```
b-butler, line 536: We can just update on the CPU with a `UserWarning` about performance right?

hoomd/hpmc/pytest/test_remove_drift.py
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) 2009-2021 The Regents of the University of Michigan
```
b-butler, line 1: Every test is restricted to serial. Does the `UpdaterRemoveDrift` not work in MPI simulations?

b-butler (changes requested):
I have a few comments. Currently, I am not sure why we restrict all our test to serial runs of pytest, and I would like to see ref_positions` be moved to a NumPy array, and be called `reference_positions`. We should probably return the array by copy and document it as such to prevent users from causing memory errors and having unexpected effects.

hoomd/hpmc/UpdaterRemoveDrift.h
lyrivera: Makes sense but I'm not sure how to handle Numpy arrays with ParameterDict. Is this done in some other place that I can take as an example?

b-butler: No, I don't believe we have had an example of this yet, but I can whip up a type validator for `numpy.ndarray` fairly easily, and will ping you when done.

lyrivera: Great, thanks.

joaander:
@lyrivera have you made progress on responding to this review?

Also, there have been a number of changes on master recently. I recommend merging master up into your branch regularly to keep up with recent developments.

lyrivera:
> @lyrivera have you made progress on responding to this review?
> 
> Also, there have been a number of changes on master recently. I recommend merging master up into your branch regularly to keep up with recent developments.

Yes. Sorry for the delay, I will have it ready for another review by the end of the day.

lyrivera:
@b-butler, @yuanzhou0827  I believe this is ready for another review. The pickling test is still failing because the element-wise  comparison that the `==` operator performs on Numpy arrays. I believe this can be fixed by overloading the `__eq__` method of the `ParameterDict` class but I'm not sure.

b-butler (changes requested):
I have a few comments to address, and you are right, we need to modify the `ParameterDict.__eq__` to handle array quantities correctly.

hoomd/hpmc/update.py
```diff
@@ -466,72 +467,51 @@ def N(self):  # noqa: N802 - allow N as a function name
...
 
     Args:
-        mc (:py:mod:`hoomd.hpmc.integrate`): MC integrator.
-        external_lattice (:py:class:`hoomd.hpmc.field.lattice_field`): lattice
-          field where the lattice is defined.
-        period (int): the period to call the updater
-
-    The command hpmc.update.remove_drift sets up an updater that removes the
-    center of mass drift of a system every period timesteps,
+        reference_positions ((*N*, 3) `numpy.ndarray` of ``numpy.float32``): the
```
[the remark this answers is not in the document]
joaander: The API should not depend on the build configuration. Don't use `Scalar` for datatypes exposed to the API, use `double`.

hoomd/hpmc/update.py
```diff
@@ -466,72 +467,51 @@ def N(self):  # noqa: N802 - allow N as a function name
...
+        integrator = self._simulation.operations.integrator
+        if not isinstance(integrator, integrate.HPMCIntegrator):
+            raise RuntimeError("The integrator must be a HPMC integrator.")
+
+        if not integrator._attached:
+            raise RuntimeError("Integrator is not attached yet.")
+
+        if isinstance(self._simulation.device, hoomd.device.GPU):
+            self._simulation.device._cpp_msg.warning(
+                "Falling back on CPU. No GPU implementation available.\n")
```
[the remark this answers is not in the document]
joaander: We have one case where we use `warnings.warn`: the deprecation of `Snapshot.exists`. I don't know that we have any concrete plans to replace the Messenger class.

hoomd/hpmc/update.py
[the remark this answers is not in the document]
lyrivera: @joaander Do you mean that the `reference_positions` should also be stored internally as `double` instead of `Scalar`?  My confusion comes from the observation that there's other cases of exported properties to the python API that are stored internally as `Scalar`. Or perhaps  I'm misunderstanding?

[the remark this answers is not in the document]
joaander: Does the param dict reference the array by copy or by reference?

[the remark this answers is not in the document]
lyrivera: I'm not entirely sure. @b-butler is better suited to answer this.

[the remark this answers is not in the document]
joaander: OK, I looked at the code. `getReferencePositions` returns a copy. So you can modify `setReferencePositions` and `getReferencePositions` to work with `double` explicitly even while the internal data array is stored as a `Scalar` (floats can be cast to double with no data loss). This will keep the API consistent going forward when we enable mixed precision computations and the internal buffer will also be in double - but you don't need to make that change now.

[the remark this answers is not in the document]
lyrivera: Ah, I see what you mean now. Thanks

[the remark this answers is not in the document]
joaander: Then you'll also need to update the documentation to `float64` as the data type.

hoomd/hpmc/update.py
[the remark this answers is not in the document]
lyrivera: Leaving as is for now until a concrete plan is devised .

lyrivera:
@b-butler @yuanzhou0827 @joaander This is ready for a final review. All test should pass now.

hoomd/hpmc/update.py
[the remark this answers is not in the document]
b-butler: I was thinking of how we deal with exceptions and error messages. We have moved the error message to the exception in multiple cases. We can leave this here though if that is what is decided. It is just not catch-able by users in Python then.

lyrivera:
@b-butler There is still one last failing test that I haven't been able to figure out. Any help would much appreciated.

b-butler:
@lyrivera could you add a test that tests the behavior of `UpdaterRemoveDrift::update`?

lyrivera:
> @lyrivera could you add a test that tests the behavior of `UpdaterRemoveDrift::update`?

Sure.

hoomd/hpmc/UpdaterRemoveDrift.h
```diff
@@ -43,10 +42,49 @@ template<class Shape> class RemoveDriftUpdater : public Updater
...
         {
+        setReferencePositions(ref_positions);
+        }
+
+    //! Set reference positions from a Nx3 numpy array
+    void setReferencePositions(const pybind11::array_t<double> ref_pos)
+        {
+        const size_t N = ref_pos.request().shape[0];
+        const size_t dim = ref_pos.request().shape[1];
+        if (N != this->m_pdata->getN() || dim != 3)
```
lyrivera, line 57: @b-butler I believe we need `getNGlobal()` here or am I missing something?

hoomd/hpmc/UpdaterRemoveDrift.h
```diff
@@ -43,10 +42,49 @@ template<class Shape> class RemoveDriftUpdater : public Updater
...
+            {
+            throw std::runtime_error("The array must be of shape (N_particles, 3).");
+            }
+        const double* rawdata = static_cast<double*>(ref_pos.request().ptr);
+        m_ref_positions.resize(m_pdata->getN());
```
lyrivera, line 62: @b-butler same here

hoomd/hpmc/UpdaterRemoveDrift.h
b-butler: Likely so, unless we want to deal with particle migration.

hoomd/hpmc/UpdaterRemoveDrift.h
b-butler: Yes

hoomd/hpmc/update.py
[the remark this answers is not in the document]
joaander: I was trying to say that we haven't decided anything yet, so this is fine as it is in this PR. We can bring up the topic of how to better offer warnings to user in a future conversation and update them in bulk in another PR.

hoomd/data/parameterdicts.py
```diff
@@ -345,6 +346,13 @@ def __len__(self):
         """int: The number of keys."""
         return len(self._dict)
 
+    def __eq__(self, other):
+        """Equality between ParameterDict objects."""
+        if not isinstance(other, ParameterDict):
+            return NotImplemented
+        return (set(self.keys()) == set(other.keys()) and np.all(
+            np.all(self[key] == other[key]) for key in self.keys()))
```
joaander, line 354: I have the same question on this PR as in #965. Is it valid to use numpy for this check on keys that may not be numpy arrays?

hoomd/hpmc/UpdaterRemoveDrift.h
```diff
@@ -43,10 +42,49 @@ template<class Shape> class RemoveDriftUpdater : public Updater
...
+        {
+        std::vector<size_t> dims(2);
+        dims[0] = this->m_ref_positions.size();
+        dims[1] = 3;
+        // the cast from vec3<Scalar>* to Scalar* is safe since vec3 is tightly packed without any
+        // padding. This also makes a copy so, modifications of this array do not effect the
+        // original reference positions.
+        const auto reference_array = pybind11::array_t<double>(
+            dims,
+            reinterpret_cast<const double*>(&(this->m_ref_positions[0])));
```
joaander, line 83:
```suggestion
            reinterpret_cast<const double*>(this->m_ref_positions.data()));
```
Cleaner access to the data pointer.

hoomd/hpmc/UpdaterRemoveDrift.h
```diff
@@ -43,10 +42,49 @@ template<class Shape> class RemoveDriftUpdater : public Updater
...
+        const double* rawdata = static_cast<double*>(ref_pos.request().ptr);
+        m_ref_positions.resize(m_pdata->getNGlobal());
+        for (size_t i = 0; i < N; i++)
+            {
+            const size_t array_index = i * 3;
+            this->m_ref_positions[i] = vec3<Scalar>(rawdata[array_index],
+                                                    rawdata[array_index + 1],
+                                                    rawdata[array_index + 2]);
+            }
+        }
```
joaander, line 70: In MPI builds, broadcast m_ref_positions from the root rank to all ranks and ignore values set on non-root ranks.

hoomd/hpmc/pytest/test_remove_drift.py
```diff
@@ -0,0 +1,189 @@
...
+        trigger=hoomd.trigger.Periodic(1),
+        reference_positions=reference_positions)
+    sim.operations.updaters.append(remove_drift)
+    sim.run(1)
+
+    s = sim.state.snapshot
+    if s.communicator.rank == 0:
+        reference_com = np.mean(reference_positions, axis=0)
+        new_com = np.mean(s.particles.position, axis=0)
+        assert np.allclose(reference_com, new_com, atol=0.05)
```
joaander, line 174: This test is failing. Given the periodic boundary conditions, I don't think that the remove drift updater leads to a conservation of the center of mass of the system.

To test, you need to compute the drift (you can avoid periodic boundary conditions by unwrapping positions) from the initial lattice position and ensure that the drift is close to 0.

hoomd/hpmc/pytest/test_remove_drift.py
```diff
@@ -0,0 +1,189 @@
...
+    sim.operations.integrator = mc
+
+    # use initial lattice configuration as reference
+    comm = MPI.COMM_WORLD
+    s = sim.state.snapshot
+    if s.communicator.rank == 0:
+        reference_positions = s.particles.position
+    else:
+        reference_positions = None
+    reference_positions = comm.bcast(reference_positions, root=0)
```
joaander, line 158: Implement the broadcast in C++. Users should not have to use mpi4py to ensure that they specify the same lattice positions on all ranks.

hoomd/hpmc/update.py
```diff
@@ -466,72 +467,54 @@ def N(self):  # noqa: N802 - allow N as a function name
...
     Args:
-        mc (:py:mod:`hoomd.hpmc.integrate`): MC integrator.
-        external_lattice (:py:class:`hoomd.hpmc.field.lattice_field`): lattice
-          field where the lattice is defined.
-        period (int): the period to call the updater
-
-    The command hpmc.update.remove_drift sets up an updater that removes the
-    center of mass drift of a system every period timesteps,
+        reference_positions ((*N*, 3) `numpy.ndarray` of ``numpy.float64``): the
+            reference positions of the lattice.
```
joaander, line 475:
```suggestion
            reference positions of the lattice :math:`[\\mathrm{length}]`.
```

joaander (changes requested):
Thanks, this needs a few more changes before it is ready to merge.

hoomd/data/parameterdicts.py
b-butler: Yes, this should be valid. When passed a Boolean returns that Boolean value, and when passed a sequence computes the n-wise conjunction as expected.

hoomd/hpmc/pytest/test_remove_drift.py
lyrivera: This was just a workaround to be used in this test only since I'm not setting the reference positions explicitly on python. Instead, I'm using whatever initial positions are set by the `lattice_snapshot_factory`, which I can only get through the snapshot API. I can avoid it by setting the positions explicitly, I believe.

hoomd/hpmc/pytest/test_remove_drift.py
lyrivera: I see. I thought it was related to the particle tags.

hoomd/hpmc/pytest/test_remove_drift.py
joaander: I still think the array should be broadcast in c++ to ensure consistency. Using the snapshot to get the initial lattice is likely a common use-case. Even if users don't use the snapshot, they should not be forced to duplicate the data on all ranks to ensure the simulation is valid.

hoomd/hpmc/pytest/test_remove_drift.py
lyrivera: @joaander I changed the test to compute the drift from the original configuration (after unwrapping) and check if its close to zero but it keeps failing. I've checked the source code for any obvious issues but nothing stands out to me.

joaander: To understand why this test is failing, I would suggest inspecting the before and after remove drift configurations in a visualization tool.

hoomd/hpmc/pytest/test_remove_drift.py
joaander: Now that you've added the broadcast in c++, you can remove the broadcast in the python test.

hoomd/hpmc/pytest/test_remove_drift.py
joaander: This is not resolved, one test still is marked `@skip_mpi`.

joaander (commented):
This is not quite ready to merge. There are several unresolved conversations and a unit test is failing.

hoomd/hpmc/update.py
```diff
@@ -466,72 +467,54 @@ def N(self):  # noqa: N802 - allow N as a function name
...
-
-    The command hpmc.update.remove_drift sets up an updater that removes the
-    center of mass drift of a system every period timesteps,
+        reference_positions ((*N*, 3) `numpy.ndarray` of ``numpy.float64``): the
+            reference positions of the lattice :math:`[\\mathrm{length}]`.
+        trigger (`hoomd.trigger.Trigger`): Select the timesteps to remove drift.
 
-    Example::
+    During the time steps specified by *trigger*, particle positions are
+    modified such that the their center of mass coincides with that of
```
joaander, line 479: Correct the phrasing here. The center of mass does not enter into this operation.

Additionally, following the guidance of #1078, please add LaTeX math that describes the computation of the drift. Use the same notation as #1079 for consistency (here, `\vec{r}_i` for particle positions and `\mathrm{N_particles}` for the number of particles). I haven't introduced notation for the box minimum image yet, one idea would be `\mathrm{min_image}(\vec{a})`.

joaander (changes requested):
@lyrivera what change led to the test passing? I don't see anything obvious.

Regardless, I made many suggestions to simplify the test, remove unnecessary code, and improve the documentation.

hoomd/hpmc/UpdaterRemoveDrift.h
```diff
@@ -113,8 +155,8 @@ template<class Shape> class RemoveDriftUpdater : public Updater
         }
 
     protected:
-    std::shared_ptr<ExternalFieldLattice<Shape>> m_externalLattice;
     std::shared_ptr<IntegratorHPMCMono<Shape>> m_mc;
+    std::vector<vec3<Scalar>> m_ref_positions;
```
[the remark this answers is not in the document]
lyrivera: I agree. In that case, should we consider removing the Shape template altogether and, perhaps, moving the updater to a different namespace, like `hoomd.update.RemoveDrift`?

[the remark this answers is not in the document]
joaander: @b-butler what do you think?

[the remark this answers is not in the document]
b-butler: I would say if there is any utility to the updater in MD then use the more general namespace (I am not sure there is), and if not leave it in HPMC. If we were to add a new MC move type that did not directly rely on the integrator, we would still put that under `hoomd.hpmc` (I think), so I feel like other updaters should be the same.

However, if MD could use the updater productively in some cases, then the more general namespace should be used for visibility since many people who use `hoomd.md` may never use `hoomd.hpmc`.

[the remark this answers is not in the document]
joaander: This method is to improve the accuracy of FrenkelLladd calculations. Those can be performed in MD as well as MC, so there is at least one case where a user may want to use it outside of HPMC.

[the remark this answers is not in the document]
lyrivera: Since it doesn't require integrator information and there is at least one use case, I believe it should be placed a more general namespace. However, I'll wait for consensus before proceeding. On another note: I have already removed the shape template and have addressed a bunch of the open comments locally.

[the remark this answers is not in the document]
b-butler: Okay, then we should move to the more general namespace.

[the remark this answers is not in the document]
lyrivera: Sounds good. I'll work on it.

lyrivera:
@joaander @b-butler I believe this is ready for another pass.

hoomd/update/remove_drift.py
```diff
@@ -0,0 +1,52 @@
...
+from hoomd import _hoomd
+import numpy as np
+
+
+class RemoveDrift(Updater):
+    r"""Remove the average drift from a system restrained on a lattice.
+
+    Args:
+        reference_positions ((*N_particles*, 3) `numpy.ndarray` of
+            ``numpy.float64``): the reference positions of the
```
joaander, line 19:
```suggestion
            `float`): the reference positions of the
```
@b-butler is recommending using `float` for numpy array data types that store double precision floating point values.

sphinx-doc/module-hoomd-update.rst
```diff
@@ -9,11 +9,12 @@ hoomd.update
...
     BoxResize
+    RemoveDrift
     CustomUpdater
 
 .. rubric:: Details
 
 .. automodule:: hoomd.update
     :synopsis: Modify the system state periodically.
-    :members: BoxResize, CustomUpdater
+    :members: BoxResize, RemoveDrift, CustomUpdater
```
joaander, line 19:
```suggestion
    :members: BoxResize, CustomUpdater, RemoveDrift
```
Alphabetize.

sphinx-doc/module-hoomd-update.rst
```diff
@@ -9,11 +9,12 @@ hoomd.update
     :nosignatures:
 
     BoxResize
+    RemoveDrift
     CustomUpdater
```
joaander, line 13:
```suggestion
    CustomUpdater
    RemoveDrift
```
Alphabetize.

joaander (changes requested):
Thanks! This is looking a lot better now. I made a few suggestions to make the documentation consistent with the rest of HOOMD-blue. 

@yuanzhou0827, @b-butler, do you have any other suggestions?

b-butler (changes requested):
I have one comment about the copying of reference positions to expose to Python that could cause errors in single precision mode.

hoomd/UpdaterRemoveDrift.h
```diff
@@ -0,0 +1,153 @@
...
+        {
+        std::vector<size_t> dims(2);
+        dims[0] = this->m_ref_positions.size();
+        dims[1] = 3;
+        // the cast from vec3<Scalar>* to Scalar* is safe since vec3 is tightly packed without any
+        // padding. This also makes a copy so, modifications of this array do not effect the
+        // original reference positions.
+        const auto reference_array = pybind11::array_t<double>(
+            dims,
+            reinterpret_cast<const double*>(this->m_ref_positions.data()));
```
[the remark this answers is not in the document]
joaander: Good catch. Store then as `Scalar` and don't cast or copy on return. The data types in the documentation naming double will be consistent with the default and supported build types.

Later, when we implement mixed precision, `Scalar` will be changed to the type that matches the particle positions.

joaander:
@lyrivera Please merge `master` into this branch and address the change that @b-butler requested. Then this is ready for a final review.

hoomd/UpdaterRemoveDrift.h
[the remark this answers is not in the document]
lyrivera: @joaander I tried removing the `reinterpret_cast` (f that what you meant) but I was getting compilation errors due to a mismatch of types in the `array_t` constructor. I couldn't figure it out (tbh didn't spend too much time in it) so I'm simply casting const Scalar * as @b-butler  suggested.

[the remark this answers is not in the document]
joaander: This is expected. `m_ref_positions` is an array of `vec3<Scalar>`. It must be cast to be used in `array_t`.

joaander (approved):

b-butler (approved):
LGTM! Thanks @lyrivera!

PR #1016, opened 2021-05-25
Tests for angle, bond and dihedral potentials

Charlottez112:
@danevans09 and I added some tests for the angle, bond and dihedral potentials. The tests for dihedral potentials are not passing and we are not sure why.

## Checklist:


b-butler:
The tests still do not work for dihedral, but I fixed a mostly documentation bug about `n` for the `Harmonic` potential that I noticed when trying to review to offer comments on what to do. I also simplified the dihedral snapshot factory. I am not sure now why the energies are off. I also do not know what was done to generate the reference forces.

Charlottez112:
> The tests still do not work for dihedral, but I fixed a mostly documentation bug about `n` for the `Harmonic` potential that I noticed when trying to review to offer comments on what to do. I also simplified the dihedral snapshot factory. I am not sure now why the energies are off. 

I think it has to do with how the four particles are connected (the order of the particle coordinates), which may affect the value of the dihedral angle. (Four particles cannot define two planes.) I updated the positions of the four particles in the `test_dihedral_harmonic.py`  to make the test pass, but it seems like it only works for pi/4 now.

> I also do not know what was done to generate the reference forces.

We used Matlab/ Mathematica to manully compute the potential energies/ forces for the initial configuration.

danevans09:
The `OPLS` forces and energies all pass now, however the `Harmonic` forces and energies do not. The `Harmonic` tests pass when n*phi is equal to pi

Charlottez112:
> The `OPLS` forces and energies all pass now, however the `Harmonic` forces and energies do not. The `Harmonic` tests pass when n*phi is equal to pi

I only updated the values in `test_dihedral_harmonic.py` so the old values in `test_dihedral.py` were wrong. To avoid confusion, I deleted `test_dihedral_harmonic.py` and updated the correct values. I also updated the OPLS forces. They should be negative. 

The entire test for the dihedral potential passed on my local device and the "real" dihedral angle used in the calculation is `-pi/4`.

I think it would be very helpful if we could add some documentation on how this angle is determined. Now the explanation is a little vague. I had to guess the real dihedral angle by trying different orders of the four particles. What are your thoughts @b-butler @joaander

joaander:
> I think it would be very helpful if we could add some documentation on how this angle is determined. Now the explanation is a little vague. I had to guess the real dihedral angle by trying different orders of the four particles. What are your thoughts @b-butler @joaander

Please create an issue that requests this documentation.

hoomd/md/OPLSDihedralForceCompute.cc
```diff
@@ -92,10 +92,10 @@ pybind11::dict OPLSDihedralForceCompute::getParams(std::string type)
     ArrayHandle<Scalar4> h_params(m_params, access_location::host, access_mode::read);
     auto val = h_params.data[typ];
     pybind11::dict params;
```
joaander, line 94:
```suggestion
    pybind11::dict params;
    // note: the values stored in params are precomputed k/2 values
```
This * 2 is correct, but the class documentation doesn't make it very clear how this is stored. Add this as a reminder.

joaander (approved):
Thanks! 

I have one suggestion to document a change, but other than that this is great.

@b-butler do you have any comments?

hoomd/md/pytest/test_angle.py
```diff
@@ -0,0 +1,185 @@
...
+                box[2] = 0
+                s.particles.position[:] = [[
+                    -d * np.sin(theta_rad / 2), d * np.cos(theta_rad / 2) + 0.1,
+                    0.0
+                ], [0.0, 0.1, 0.0],
+                                           [
+                                               d * np.sin(theta_rad / 2),
+                                               d * np.cos(theta_rad / 2) + 0.1,
+                                               0.0
+                                           ]]
```
b-butler, line 89:
```suggestion
            base_positions = np.array([
                [-d * np.sin(theta_rad / 2), d * np.cos(theta_rad / 2), 0.0],
                [0.0, 0.0, 0.0],
                [d * np.sin(theta_rad / 2), d * np.cos(theta_rad / 2), 0.0]
            ]
            # move particles slightly in direction of MPI decomposition which
            # varies by simulation dimension
            nudge_dimension = 2 if dimensions == 3 else 1
            base_positions[:, nudge_dimension] += 0.1
            s.particles.position[:] = base_positions
            s.particles.types = particle_types
```

Make this kind of change in other snapshot factory methods created in this PR as well please

hoomd/md/pytest/test_angle.py
```diff
@@ -0,0 +1,185 @@
...
+        return s
+
+    return make_snapshot
+
+
+@pytest.mark.parametrize("angle_and_args", get_angle_and_args())
+def test_before_attaching(angle_and_args):
```
b-butler, line 96:
```suggestion
@pytest.mark.parametrize("angle_cls, potential_kwargs", get_angle_and_args())
def test_before_attaching(angle_cls, potential_kwargs):
```
Make this change in the function and all other functions that use this and similar fixtures in the tests.

hoomd/md/pytest/test_angle.py
```diff
@@ -0,0 +1,185 @@
...
+
+@pytest.mark.parametrize("angle_and_args", get_angle_and_args())
+def test_after_attaching(triplet_snapshot_factory, simulation_factory,
+                         angle_and_args):
+    snap = triplet_snapshot_factory()
+    if snap.exists:
+        snap.angles.N = 1
+        snap.angles.types = ['backbone']
+        snap.angles.typeid[0] = 0
+        snap.angles.group[0] = (0, 1, 2)
```
b-butler, line 114: Is `triplet_snapshot_factory` used anywhere else other than `test_angle.py`? If not this this should be in that factory function.

hoomd/md/pytest/test_angle.py
```diff
@@ -0,0 +1,185 @@
...
+    integrator.forces.append(angle_potential)
+
+    langevin = hoomd.md.methods.Langevin(kT=1,
+                                         filter=hoomd.filter.All(),
+                                         alpha=0.1)
+    integrator.methods.append(langevin)
+    sim.operations.integrator = integrator
+
+    sim.run(0)
+    for key in args.keys():
```
b-butler, line 132:
```suggestion
    for key in args:
```

hoomd/md/pytest/test_angle.py
```diff
@@ -0,0 +1,185 @@
...
+
+    return make_snapshot
+
+
+@pytest.mark.parametrize("angle_and_args", get_angle_and_args())
+def test_before_attaching(angle_and_args):
+    angle, args = angle_and_args
+    angle_potential = angle()
+    angle_potential.params['backbone'] = args
+    for key in args.keys():
```
b-butler, line 100:
```suggestion
    for key in args:

File: hoomd/md/pytest/test_angle.py [hunk 7]
```diff
@@ -0,0 +1,185 @@
...
+
+    sim.run(0)
+    for key in args.keys():
+        np.testing.assert_allclose(angle_potential.params['backbone'][key],
+                                   args[key],
+                                   rtol=1e-6)
+
+
+@pytest.mark.parametrize("angle_args_force_and_energy",
+                         get_angle_args_forces_and_energies())
```
b-butler, line 139: split up the tuple here as suggested above.

b-butler (changes requested):
Thanks for the work. :smiley: I have left a few comments, these generally apply to all three or two of the three test files, so please make the changes in all files to resolve a comment.

vyasr:
Sorry I lost track of being requested here. I won't be able to look at this until early next week, but please ping me if I haven't reviewed by next Tuesday or so (if all other reviewers are happy with the state of this code before then I'm also fine if my review request is removed).

b-butler (approved):
LGTM

PR #1018, opened 2021-05-25
Raise exceptions for unattached simulations

### LINKED ISSUE #812: Property exceptions for unattached simulations
## Description

While writing my first script using the HOOMD v3 API, I was tripped up a few times by objects returning `None` for a property when accessing information that isn't available until the simulation is attached. For example, `mc.type_shapes` isn't available until after calling `sim.run(timesteps)`. Though the `type_shapes` data is formally "defined" as soon as the shape definitions have been provided, it requires attachment and a round trip through C++ before that information can be accessed. I noticed that `None` is returned by many objects, such as other properties of an HPMC integrator like `mc.overlaps` but also things like `sim.state` before a state is created. I found this confusing, and would have expected errors instead of `None`.

For example, I expected this to work:
```python
mc = hoomd.hpmc.integrate.Ellipsoid(seed=42, d=0.1, a=0.1)
mc.shape['A'] = {'a': 3., 'b': 1., 'c': 1.}
print(mc.type_shapes)  # returns None
```

## Proposed solution

Raise an exception instead of returning `None` for properties that require C++ attachment. I suggested `AttributeError` but @joaander indicated there are reasons to avoid this (something to do with how Python interprets the failure?). @b-butler proposed creating a custom error. I would recommend creating an exception derived from `RuntimeError` (despite my claim in #811 that we sometimes overuse `RuntimeError` 😄) because I couldn't find a better option. Perhaps the class could have some standard boilerplate message with only minimal specification for the current class attribute name.

```python
class AttachedDataError(RuntimeError):
    """Raised when data is inaccessible until the simulation is attached."""
    def __init__(self, data_name):
        self.data_name = data_name

    def __str__(self):
        return f"The property {self.data_name} is unavailable until the simulation runs for 0 or more steps."

# Then use it like this:
@property
def property_name(self):
    if self._attached:
        return self._cpp_obj.getProperty()
    else:
        raise AttachedDataError('property_name')
```

From discussion with @joaander and @b-butler, the design decision to return `None` is partly informed by the logging infrastructure. The loggers would need to detect whatever exception is raised and log `None` instead.

## Additional context

@joaander noted that this pattern can be found easily by searching for `self._attached` in the code base.

SchoeniPhlippsn:
## Description

Trying to access object properties now raises an exception `AttachedDataError` when simulation is not attached yet. It returned `None` before.

We added a new error class `AttachedDataError` and raise it if simulation is not attached yet for all properties in hpmc and md.

## Motivation and context

We want to avoid confusion about the properties which are technically defined before simulation attachment but where no values have been asssigned yet.

Resolves #812

## How has this been tested?

Updated all pytest.

## Change log

```
Raise AttachedDataError for object properties before simulation attachment 
```

## Checklist:


hoomd/logging.py
```diff
@@ -464,7 +465,11 @@ def from_tuple(cls, entry):
         return cls(entry[0], method, category)
 
     def __call__(self):
-        attr = getattr(self.obj, self.attr)
+        try:
+            attr = getattr(self.obj, self.attr)
+        except AttachedDataError:
+            attr = None
```
AlainKadar, line 471: Why necessary to assign ``None`` to ``attr`` as opposed to leaving unassigned, as in the other cases?

AlainKadar (commented):
Will changes eventually be made to all test files to ensure calling an unattached property throws ``AttachedDataError`` ? Eg changing ``hoomd/pytest/test_logging.py`` to include test which checks calling _LoggerEntry without attached attributes throws ``AttachedDataError`` ?

hoomd/logging.py
joaander: Because a number of facilities, like `write_debug_data` are designed assuming that a Logger can be evaluated even if the simulation is not attached. The previous behavior was to log `None` in this case, so this maintains that.

SchoeniPhlippsn:
> Will changes eventually be made to all test files to ensure calling an unattached property throws `AttachedDataError` ? Eg changing `hoomd/pytest/test_logging.py` to include test which checks calling _LoggerEntry without attached attributes throws `AttachedDataError` ?

At the moment the logger does not throw an `AttachedDataError`. The only time the data error is raised is when the user tries to access a property of an object directly and that object is not attached yet. So we don't have to update the logger tests. I think everywhere else I added or updated the `AttachedDataError` tests.

hoomd/simulation.py
```diff
@@ -213,7 +214,10 @@ def create_state_from_snapshot(self, snapshot):
     @property
     def state(self):
         """hoomd.State: The current simulation state."""
-        return self._state
+        if self._state is None:
+            raise AttachedDataError("state")
```
b-butler, line 218: This is a different sort of error than the others. Internally the state can exist without operations being attached.

b-butler (changes requested):
I think we should try to avoid the word attached here since we don't user that word in the context of C++ initialization (which is the reason for the error) in user documentation/API (or shouldn't). We do use it for custom actions, but it is more of a setup step, and maybe we should actually rename that to reflect that...

----

I added an optional argument to `hoomd.logging.log` -- `requires_attach` -- that lets you automatically raise an error when not attached without needed the boiler plate. I would also recommend the default error message go there and not in the error class to enable other error messages (assuming we rename the error as I suggest). Either way, this should be used in all locations where we raise the error for a loggable quantity.

----

I would prefer a `hoomd.error` module so if/when we decide on other useful derived error types we can add them to a central module. In this case we could move `hoomd.data.typecoverter.TypeConversionError`.

On the other hand, this could be defined in `operation.py` since that is where the idea of attaching is presented.

hoomd/logging.py
```diff
@@ -378,6 +383,8 @@ def log(func=None, *, is_property=True, category='scalar', default=True):
             quantities even when logging other quantities of that type. The
             default category allows for these to be pass over by
             `hoomd.logging.Logger` objects by default. Argument keyword only.
+        requires_attach (`bool`, optional): Whether this property is
+            accessible before attaching.
```
AlainKadar, line 387: The name of the argument seems to be the opposite of its own description (if a property requires attachment then it _is not_ accessible before attaching)? Change description to "Whether this property requires attachment before access"?

hoomd/simulation.py
SchoeniPhlippsn: This is true. Shall I just add another type of error here or just pass `None`?

SchoeniPhlippsn:
I think this is ready for another review. 
The only issue I face is that the changes added by Brandon lead to a pytest error. 
To be more precise by testing `test_write_debug_data`, I get the TypeError : `wrapped_with_exception() missing 1 required positional argument: 'self'`

hoomd/simulation.py
joaander: I think we should just return `None` here. `None` is a clear indication that the state does not yet exist because it has not been created.

hoomd/error.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2009-2021 The Regents of the University of Michigan
+# This file is part of the HOOMD-blue project, released under the BSD 3-Clause
+# License.
+
+"""HOOMD Errors."""
```
joaander, line 5: Import `error` in `__init__.py` so that it is available in user namespaces when users `import hoomd`.

Also, add a `module-hoomd-error.rst` that generates the documentation for this module and add it to the proper place in the package index toctree.

hoomd/simulation.py
```diff
@@ -213,7 +214,10 @@ def create_state_from_snapshot(self, snapshot):
     @property
     def state(self):
         """hoomd.State: The current simulation state."""
-        return self._state
+        if self._state is None:
+            raise DataAccessError("state")
+        else:
+            return self._state
```
joaander, line 220:
```suggestion
        return self._state
```
I think a `None` return value here makes sense. The state has not yet been initialized.

joaander (changes requested):
Thanks! I have a few suggestions for improvement.

hoomd/simulation.py
b-butler: I think `None` is the correct solution here; we are making a statement that the state is null here.

b-butler:
@SchoeniPhlippsn I fixed the bug with the `@log` decorator for the `requires_attach` argument.

hoomd/hpmc/compute.py
```diff
@@ -102,4 +103,4 @@ def free_volume(self):
             self._cpp_obj.compute(self._simulation.timestep)
             return self._cpp_obj.free_volume
         else:
-            return None
+            raise DataAccessError("free_volume")
```
b-butler, line 106: Use the `requires_attach` argument.

hoomd/hpmc/integrate.py
```diff
@@ -208,7 +209,7 @@ def type_shapes(self):
 
     def _return_type_shapes(self):
         if not self._attached:
-            return None
+            raise DataAccessError("type_shapes")
```
b-butler, line 212: Use `log(requires_attach=True)` on the `type_shapes` loggable quantity for subclasses.

hoomd/logging.py
```diff
@@ -355,7 +356,12 @@ def _add_loggable_docstring_info(cls, new_cls, attr, category, default):
             getattr(new_cls, attr).__doc__ += str_msg.format(' ' * indent)
 
 
-def log(func=None, *, is_property=True, category='scalar', default=True):
+def log(func=None,
+        *,
+        is_property=True,
+        category='scalar',
+        default=True,
+        requires_attach=False):
```
b-butler, line 364:
```suggestion
        requires_run=False):
```

hoomd/logging.py
```diff
@@ -377,6 +383,8 @@ def log(func=None, *, is_property=True, category='scalar', default=True):
             quantities even when logging other quantities of that type. The
             default category allows for these to be pass over by
             `hoomd.logging.Logger` objects by default. Argument keyword only.
+        requires_attach (`bool`, optional): Whether this property requires
+            attachment before being accessible.
```
b-butler, line 387:
```suggestion
        requires_run (`bool`, optional): Whether this property requires
            the simulation to run before being accessible.
```

hoomd/logging.py
```diff
@@ -396,6 +404,19 @@ def helper(func):
                 "Multiple loggable quantities named {}.".format(name))
         Loggable._meta_export_dict[name] = _LoggableEntry(
             LoggerCategories[category], default)
+        if requires_attach:
```
b-butler, line 407:
```suggestion
        if requires_run:
```

hoomd/md/force.py
```diff
@@ -44,14 +45,14 @@ class Force(_HOOMDBaseObject):
     Initializes some loggable quantities.
     """
 
-    @log
+    @log(requires_attach=True)
```
b-butler, line 48:
```suggestion
    @log(requires_run=True)
```

b-butler (changes requested):
None of the properties have been converted to use `log(requires_attach)` which I have suggested to rename `requires_run` since `log` is public (I realized that naming it `requires_attach` exposed attaching to the user).

AlainKadar (approved):

b-butler (approved):
Thanks LGTM!

joaander (approved):
Looks great, thanks!

We'll need to watch check that other PRs merged after this switch to the same mechanism.

PR #1019, opened 2021-05-25
Reformat units in the hoomd documentation

### LINKED ISSUE #801: Reformat units in the documentation
## Description

As I update the docs, I'm finding the units documentation for each parameter cumbersome. For example:
```
sigma (in energy units)
gamma (in units of force/velocity)
```
We could instead use the square bracket notation used by wolfram alpha for basic units:
```
sigma [energy]
gamma [force/velocity]
```

## Motivation and context

I'm open to other suggestions.

jekw90:
## Description

We have reformatted units of variables and parameters.

## Motivation and context

We believe such reformatting can make hoomd-blue documents more human-readable.

Resolves #801 

## How has this been tested?

Whenever we reformatted units, we made sphinx documents and check that any changes to
documentation display properly.

## Change log

```

```

## Checklist:

- [V] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [V] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.


tommy-waltmann (approved):

sphinx-doc/index.rst
```diff
@@ -96,3 +96,4 @@ workflows all with Python code in your job.
    license
    credits
    indices
+   units
```
joaander, line 99: Make `units` the first item in the reference section.

joaander (changes requested):
Thanks, this is great!

I have a few change requests. 
1. You need to wrap lines at 80 characters. See the pre-commit check where all the lines that need to be wrapped are listed.
2. You will see that flake8 also lists `\cdot` is also invalid in a docstrings. We need to fix that as well.

There are two options to fixing the the `\c` errors. First, we could keep the formatting as is and replace with `\` with `\\` everywhere you added it in a `"""` docstring (but not in `r"""` docstrings). In this case, we would also need to add D301 to the flake8 ignore list in `setup.cfg`. 

The second option to fix this is to follow wolfram alphas formatting of units e.g. (https://www.wolframalpha.com/input/?i=foot+pound) which puts each unit in its own square brackets (e.g. `[mass] [length]^2 [time]^{-2}`) at which point LaTeX `\` commands are not needed.

Which style do you like @tommy-waltmann @b-butler ?

tommy-waltmann:
I like the version we currently have a little better, but don't have a huge preference. I like that the brackets separate the units part of the description from the rest of the parameter description, which I think makes the docs somewhat clearer and is somewhat lost when using the Wolfram alpha style.

jekw90:
Dear developers and reviewers.

I have made changes to this branch based on "change request" comment. 

If you find any error or wrong change please tell me any time. Thank you

hoomd/hpmc/update.py
```diff
@@ -20,8 +20,8 @@ class BoxMC(Updater):
 
     Args:
         betaP (`float` or :py:mod:`hoomd.variant.Variant`):
-            :math:`\frac{p}{k_{\mathrm{B}}T}` (units of inverse area in 2D or
-            inverse volume in 3D).
+            :math:`\frac{p}{k_{\mathrm{B}}T}` :math:`[\mathrm{area}^{-1}]`
+            in 2D or :math:`[\mathrm{volume}^{-1}]` in 3D.
         trigger (hoomd.trigger.Trigger): Select the timesteps to perform box
             trial moves.
```
tommy-waltmann, line 27: The delta key in the volume dict for this class has written that the units are area and volume, but we should put that in the new style of writing units.

hoomd/md/angle.py
```diff
@@ -53,10 +53,10 @@ class Harmonic(Angle):
             The dictionary has the following keys:
 
             * ``k`` (`float`, **required**) - potential constant
-              (in units of energy/radians^2)
+              :math:`[\mathrm{energy} \cdot \mathrm{radians}^{-2}]`
```
tommy-waltmann, line 56: We document radians as an explicit unit here, but that is not done in any of hpmc (specifically for rotational move size arguments). Should we start documenting this everywhere?

hoomd/md/compute.py
```diff
@@ -30,8 +30,8 @@ class ThermodynamicQuantities(_Thermo):
     :py:class:`ThermodynamicQuantities` acts on a given group of particles and
```
tommy-waltmann, line 30: Let's document the volume property of this class as being in volume units.

hoomd/md/manifold.py
```diff
@@ -66,9 +66,10 @@ class Cylinder(Manifold):
     r"""Cylinder manifold.
 
     Args:
-        r (`float`): radius of the cylinder constraint (in distance units).
-        P (`tuple` [`float`, `float`, `float`]): point defining position of the
-            cylinder axis (default origin).
+        r (`float`): radius of the cylinder constraint.
+          :math:`[\mathrm{length}]`
+        P (`tuple` [`float`, `float`, `float`]): point defining position of
+            the cylinder axis (default origin).
```
tommy-waltmann, line 72:
```suggestion
            the cylinder axis (default origin). :math:`[\mathrm{length}]`
```

hoomd/md/manifold.py
```diff
@@ -170,13 +171,13 @@ class Ellipsoid(Manifold):
...
-        c (`float`): length of the c-axis of the ellipsoidal constraint (in
-            distance units).
-        P (`tuple` [`float`, `float`, `float`]): center of the ellipsoidal
+        a (`float`): length of the a-axis of the ellipsoidal constraint.
+            :math:`[\mathrm{length}]`
+        b (`float`): length of the b-axis of the ellipsoidal constraint.
+            :math:`[\mathrm{length}]`
+        c (`float`): length of the c-axis of the ellipsoidal constraint.
+            :math:`[\mathrm{length}]`
+        P (`tuple` [`float`, `float`, `float`]): center of the ellipsoidal제
```
tommy-waltmann, line 180: What is the character at the end of this line?

hoomd/md/manifold.py
```diff
@@ -170,13 +171,13 @@ class Ellipsoid(Manifold):
...
             constraint (default origin).
```
tommy-waltmann, line 181:
```suggestion
            constraint (default origin). :math:`[\mathrm{length}]`.
```

hoomd/md/manifold.py
```diff
@@ -369,8 +370,8 @@ class Sphere(Manifold):
     """Sphere manifold.
 
     Args:
-        r (`float`): raduis of the a-axis of the spherical constraint (in
-            distance units).
+        r (`float`): radius of the a-axis of the spherical constraint.
+          :math:`[\\mathrm{length}]`
         P (`tuple` [`float`, `float`, `float`] ): center of the spherical
             constraint (default origin).
```
tommy-waltmann, line 376:
```suggestion
            constraint (default origin). :math:`[\\mathrm{length}]`.
```

hoomd/md/many_body.py
```diff
@@ -47,7 +47,7 @@ class Triplet(Force):
 
     .. py:attribute:: r_cut
 
-        *r_cut* (in distance units), *optional*: defaults to the value
+        *r_cut* :math:`[length]`, *optional*: defaults to the value
```
tommy-waltmann, line 50:
```suggestion
        *r_cut* :math:`[\mathrm{length}]`, *optional*: defaults to the value
```

hoomd/md/many_body.py
```diff
@@ -180,40 +180,41 @@ class Tersoff(Triplet):
...
     .. py:attribute:: params
 
         The Tersoff potential parameters. The dictionary has the following
         keys:
 
         * ``magnitudes`` (tuple[`float`, `float`]) - :math:`(A_1, A_2)` -
-          Magnitudes of the repulsive and attractive terms (dimensionless,
-          *default*: (1.0, 1.0))
+          Magnitudes of the repulsive and attractive
+          terms (*default*: (1.0, 1.0)) :math:`[\mathrm{dimensionless}]`
```
tommy-waltmann, line 193:
```suggestion
          terms (*default*: (1.0, 1.0)) :math:`[\mathrm{energy}]`
```

tommy-waltmann (changes requested):
Great work so far, I found a couple of things that were overlooked and had some other comments as I did my review. Also, this PR does not touch `hpmc/compute.py`, and I think it should.

I have only reviewed the first 14 files in this PR, so expected another round of comments later today at some point.

hoomd/md/methods.py
```diff
@@ -367,18 +371,21 @@ class NPT(Method):
...
-            point for the barostat (in pressure units).
+            point for the barostat.
             In Voigt notation,
-            :math:`[S_{xx}, S_{yy}, S_{zz}, S_{yz}, S_{xz}, S_{xy}]`. Stress can
-            be reset after method object is created. For example, An isoropic
-            pressure can be set by ``npt.S = 4.``
+            :math:`[S_{xx}, S_{yy}, S_{zz}, S_{yz}, S_{xz}, S_{xy}]`. Stress
+            can be reset after method object is created. For example, An
+            isoropic pressure can be set by ``npt.S = 4.``
+            :math:`[\mathrm{pressure}]`
```
joaander, line 385:
```suggestion
            :math:`[S_{xx}, S_{yy}, S_{zz}, S_{yz}, S_{xz}, S_{xy}]` 
            :math:`[\mathrm{pressure}]`. Stress can be reset after the method
            object is created. For example, An isoropic pressure can be set by
            ``npt.S = 4.``
```

hoomd/md/methods.py
```diff
@@ -522,14 +531,15 @@ class NPH(Method):
         filter (hoomd.filter.ParticleFilter): Subset of particles on which to
             apply this method.
 
-        S: Stress components set point for the barostat (in pressure units).
+        S: Stress components set point for the barostat.
            In Voigt notation:
-           :math:`[S_{xx}, S_{yy}, S_{zz}, S_{yz}, S_{xz}, S_{xy}]`.  In case of
-           isotropic pressure P (:math:`[p, p, p, 0, 0, 0]`), use ``S = p``.
+           :math:`[S_{xx}, S_{yy}, S_{zz}, S_{yz}, S_{xz}, S_{xy}]`. In case
```
joaander, line 536: Same change as above here. Put the units right after the quantity.

hoomd/md/methods.py
```diff
@@ -1137,11 +1158,13 @@ class Brownian(MethodRATTLE):
             coefficient can be directly set instead of the ratio of particle
             diameter (:math:`\gamma = \alpha d_i`). The type of ``gamma``
             parameter is either positive float or zero.
+            :math:`[\mathrm{mass} \cdot \mathrm{time}^{-1}]`
 
         gamma_r (TypeParameter[``particle type``, [`float`, `float`, `float`]]):
             The rotational drag coefficient can be set. The type of ``gamma_r``
             parameter is a tuple of three float. The type of each element of
             tuple is either positive float or zero.
+            :math:`[\mathrm{mass} \cdot \mathrm{time}^{-1}]`
```
joaander, line 1167: `gamma_r` multiplies an angular velocity which results in a torque. Therefore, its units should be force * distance * radians^{-1} * time^{-1}.

hoomd/md/pair/pair.py
```diff
@@ -583,9 +583,9 @@ class DPD(Pair):
     Args:
         nlist (`hoomd.md.nlist.NList`): Neighbor list
         kT (`hoomd.variant` or `float`): Temperature of
-          thermostat (in energy units).
-        r_cut (float): Default cutoff radius (in distance units).
-        r_on (float): Default turn-on radius (in distance units).
+          thermostat :math:`[energy]`.
```
joaander, line 586:
```suggestion
          thermostat :math:`[\mathrm{energy}]`.
```

joaander (changes requested):
Thanks for the great work. Just a few changes to make and this will be really clean.

hoomd/md/angle.py
joaander: In 2D, the rotation move is in radians but I don't think it is in 3D. We could document it as arbitrary units for 3D.

hoomd/md/compute.py
```diff
@@ -53,7 +53,9 @@ def _attach(self):
 
     @log(requires_run=True)
     def kinetic_temperature(self):
-        r""":math:`kT_k`, instantaneous thermal energy of the group [energy].
+        r""":math:`kT_k`, instantaneous thermal energy of the group.
+
+        :math:`[\mathrm{energy}]`
```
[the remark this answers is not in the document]
jekw90: Thank you for this suggestion! I have one quick question. When I made changes following this suggestion, and git-pushed the changes, I got following pre-commit error. 

D205 1 blank line required between summary line and description
D415 First line should end with a period, question mark, or exclamation point

If it is okay with you can you help me how to solve this issue?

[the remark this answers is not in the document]
joaander: With the GitHub suggestion interface, it is hard to tell what blank lines might remain after the suggestion. Add a period after
```
:math:`[\mathrm{energy}]`
```
to solve the second error. Then add a blank line to solve the first.

joaander (changes requested):
Here is an example of how to change the raw strings to work with line continuation indicators.

hoomd/md/angle.py
jekw90: Thank you for this comment. I have one quick question related to this comment. 

Would I set unit as 

* ``k`` (`float`, **required**) - potential constant
:math:`[\mathrm{energy} \cdot \mathrm{radians}^{-2}]` for 2D and 
:math:`[\mathrm{energy} \cdot \mathrm{arbitrary angle unit}^{-2}]` for 3D

* ``t0`` (`float`, **required**) -  rest angle
:math:`[\mathrm{radians}]` for 2D and 
:math:`[\mathrm{arbitrary angle unit}]` for 3D

Or do you have any other suggestion or preference for unit style?

hoomd/md/force.py
```diff
@@ -292,13 +297,14 @@ class Active(Force):
         filter (:py:mod:`hoomd.filter`): Subset of particles on which to apply
             active forces.
         rotation_diff (float): rotational diffusion constant, :math:`D_r`, for
-            all particles in the group.
+            all particles in the group. :math:`[\mathrm{time}^{-1}]`
```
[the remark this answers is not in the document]
jekw90: Thank you for this comment. 
I talked with Philipp and found that the unit for rotational diffusion constant is radians^{2} * time^{-1}.

I have one quick question. Similar to the comment above, I think the unit will depend on spatial dimension (2D or 3D)

I think
in 2D, the unit will be {radians}^{2} \cdot time^{-1}
in 3D, the unit may be {arbitrary angle unit}^{2} \cdot time^{-1} (please correct me if I am wrong)

Would you want me to state units in 2D and 3D in this way?  Or do you prefer any other styles for the units?

hoomd/md/angle.py
joaander: No, don't change the units for the angle potentials. I was referring to the units of the `a` value in the HPMC integrator that @tommy-waltmann asked about.

hoomd/hpmc/integrate.py
```diff
@@ -78,16 +78,18 @@ class HPMCIntegrator(BaseIntegrator):
 
     Attributes:
         a (`TypeParameter` [``particle type``, `float`]):
-            Maximum size of rotation trial moves.
+            Maximum size of rotation trial moves
+            :math:`[\\mathrm{arbitrary \\ unit}]`.
```
tommy-waltmann, line 82: I think explicitly writing arbitrary unit here is confusing. @joaander I am not that familiar with hpmc moves, is there something more specific we can put here? I would be fine with dimensionless, but I think writing arbitrary unit could be confusing to users.

tommy-waltmann (commented):
I just have one comment left that I think may need to be addressed.

hoomd/hpmc/integrate.py
joaander: The comment in the implementation for 3D is:
```
        // Frenkel and Smit reference Allen and Tildesley, referencing Vesley(1982), referencing
        // Marsaglia(1972). Generate a random unit quaternion. Scale it to a small rotation and
        // apply.
```

It is not obvious to me what units the value is in from the implementation - it is used as a scale factor that multiples a unit quaternion. I don't have the time right now to trace down the original references for the algorithm or determine the proper units. Feel free to read those references and suggest the proper value.

tommy-waltmann: In that case I think we can just leave it as dimensionless.

joaander (approved):
Thanks for all your hard work @jekw90 !

hoomd/hpmc/integrate.py
joaander: OK, I made that change. Good to go?

jekw90: Thank you!!

hoomd/hpmc/integrate.py
```diff
@@ -1036,9 +1047,10 @@ class FacetedEllipsoid(HPMCIntegrator):
 
     Args:
         d (float): Default maximum size of displacement trial moves
-            (distance units).
+            :math:`[\mathrm{length}]`.
 
-        a (float): Default maximum size of rotation trial moves.
+        a (float): Default maximum size of rotation trial moves
+            :math:`[\mathrm{arbitrary \ unit}]`.
```
tommy-waltmann, line 1053:
```suggestion
        a (float): Default maximum size of rotation trial moves
            :math:`[\mathrm{dimensionless}]`.
```

tommy-waltmann (commented):
I found one last place we need to write dimensionless instead of arbitrary unit

tommy-waltmann (approved):

sphinx-doc/units.rst
```diff
@@ -0,0 +1,26 @@
+Unit Conversion
+===================
```
b-butler, line 2:
```suggestion
Units
+++++

HOOMD-blue follow self-consistent units. This means that users are free to define the units (e.g. meters versus centimeters for length) of HOOMD-blue's base units, and units of all derived quantities can be determined from the user selected units. For instance if the base length is chosen to be 1 meter, energy 1 Joule, and mass 1 kilogram, then velocity is in meters per second, but if one nanometer is chosen for length, kilojoules per mole for energy, and 1 amu for mass then velocity is in nanometers per picosecond.

Base Units
============

Base units in HOOMD-blue are
- energy
- length
- mass

Unit Conversion
===================
```

b-butler:
@tommy-waltmann I added some more content to the units restructured text file. I would appreciate a review to ensure that it sounds okay before merging. Thanks!

joaander:
There is some good language for that in the tutorial: https://hoomd-blue.readthedocs.io/en/latest/tutorial/00-Introducing-HOOMD-blue/03-Initializing-the-System-State.html#Units

sphinx-doc/units.rst
```diff
@@ -0,0 +1,39 @@
+Units
++++++
+
+HOOMD-blue follow self-consistent units. This means that users are free to define the units (e.g. meters versus centimeters for length) of HOOMD-blue's base units, and units of all derived quantities can be determined from the user selected units. For instance if the base length is chosen to be 1 meter, energy 1 Joule, and mass 1 kilogram, then velocity is in meters per second, but if one nanometer is chosen for length, kilojoules per mole for energy, and 1 amu for mass then velocity is in nanometers per picosecond.
```
tommy-waltmann, line 4:
```suggestion
HOOMD-blue does not adopt any particular system of units, but instead follows a self-consistent system of units. Users are free to define HOOMD-Blue's base units (e.g. meters versus centimeters for length), and units of all derived quantities can be determined from the base units. For instance if the base units are chosen to be 1 meter, 1 Joule, and 1 kilogram, then velocity is in meters per second, but if one choses nanometers for length, kilojoules per mole for energy, and 1 amu for mass then velocity is in nanometers per picosecond.
```

tommy-waltmann (changes requested):
I reworded the text that was there, there is more info in the tutorial (referenced by Josh) explaining how the units work. If we want to leave it that way we can, otherwise we can add more to this section.

I am fine with leaving the more detailed info out from this section and just having it in the tutorial.

b-butler:
@tommy-waltmann, I think reference documentation in general should be sufficient in itself, i.e. it is possible to understand the API without reference to the tutorials. The tutorials are more novice friendly, but shouldn't contain information not found in some form in the reference documentation itself.

tommy-waltmann:
In that case I would also include the information about HPMC units from the tutorial in the reference documentation, since they must be understood slightly differently than MD units.

tommy-waltmann (approved):
LGTM!

PR #1020, opened 2021-05-25
Add MD walls

### LINKED ISSUE #551: Unify Walls python API for HPMC and MD and enable deferred initialization
At some point we will have to make this have a deferred initialization for the walls. Since we want to be able to create walls without having a mc integrator instance at the time of creation. They need the same attach concept as well.

_Originally posted by @b-butler in https://github.com/glotzerlab/hoomd-blue/pull/536_

There MD python walls currently allows for deferred initialization via `wall.group`. Would be likely easier to utilize it for both HPMC and MD than keeping separate implementations.

j-proc:
## Description

Walls has shared geometry definitions that can be unified. Additionally we can define some processing of arbitrary lists of walls into useful vectors. Thus each user/owner of walls will have a variable based on this type.

## Motivation and context

Resolves #551 

## How has this been tested?

Too many pieces are left to do to test currently.

## Change log

```
Unify the hoomd walls API
```

## TODOs:
- [X] Define new unified structure in hoomd.wall for geometry
- [X] Create meta-list class to handle arbitrary wall lists being passed to geometry specialized vectors
- [ ] Specialize synced lists for hpmc/md attachment of the associated cpp objects.
- [ ] Rewrite md cpp objects to accept std::vectors instead of group objects.
- [ ] Documentation
- [ ] Testing

## Checklist:


hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
+# Copyright (c) 2009-2021 The Regents of the University of Michigan
+# This file is part of the HOOMD-blue project, released under the BSD 3-Clause
+# License.
+
+r"""Wall potentials.
+
+Wall potentials add forces to any particles within a certain distance,
+:math:`r_{\mathrm{cut}}`, of each wall. In the extrapolated mode, all particles
+deemed outside of the wall boundary are included as well.
```
joaander, line 9:
```suggestion
outside of the wall boundary are included as well.
```
We don't judge or consider whether particles are outside the wall. They either are or the are not.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+    r"""Generic wall potential.
+
+    Warning:
+        `WallPotential` should not be used directly.  It is a base class that
+        provides features and documentation common to all standard wall
+        potentials.
+
+    All wall potential commands specify that a given potential energy and
+    potential be computed on all particles in the system within a cutoff
+    distance, :math:`r_{\mathrm{cut}}`, from each wall in the given wall group.
```
joaander, line 64:
```suggestion
    distance, :math:`r_{\mathrm{cut}}`, from each wall.
```

joaander: Also, reword to account for the extrapolated mode which you mention correctly in the module level docs.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+    pointing from the particle to the wall or half-space boundary and
+    :math:`V_{\mathrm{pair}}(r)` is the pair potential specified by subclasses
+    of `WallPotential`.  Wall forces are implemented with the concept of
+    half-spaces in mind. There are two modes which are allowed currently in wall
+    potentials: standard and extrapolated.
+
+    .. rubric:: Standard Mode.
+
+    In the standard mode, when :math:`r_{\mathrm{extrap}} \le 0`, the potential
+    energy is only applied to the half-space specified in the wall group.
```
joaander, line 75:
```suggestion
    energy is only applied to the half-space specified by each wall.
```
The "wall group" language no longer applies. The new API defines a list of walls and potentials are evaluated for each wall in the list.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+    :math:`V(r)` is evaluated in the same manner as when the mode is shift for
+    the analogous :py:mod:`pair <hoomd.md.pair>` potentials within the
+    boundaries of the half-space.
```
joaander, line 78: Add a figure demonstrating the definition of `\vec{r}` for the inside and outside cases for an example wall (e.g. a sphere wall and/or plane wall in 2D). See https://github.com/glotzerlab/hoomd-blue/tree/master/sphinx-doc/figures where I store notebooks to generate documentation figures with a consistent style.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+
+        \begin{eqnarray*}
+        \vec{F}  = & -\nabla V(r) & 0 < r < r_{\mathrm{cut}} \\
+                 = & 0 & r \ge r_{\mathrm{cut}} \\
+                 = & 0 & r \le 0
+        \end{eqnarray*}
+
+    .. rubric: Extrapolated Mode:
+
+    The wall potential can be linearly extrapolated beyond a minimum separation
```
joaander, line 108: The figure that explains the extrapolated mode should be shown here instead as it applies to all wall potentials. It would be great to also recreate the figure in the new consistent documentation style.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+        \end{eqnarray*}
+
+    In other words, if :math:`r_{\rm extrap}` is chosen so that the pair force
+    would point into the active half-space, the extrapolated potential will push
+    all particles into the active half-space. See `LJ` for a pictorial example.
+
+    To use extrapolated mode ``r_extrap`` must be set per particle type.
+
+    .. attention::
+        The current wall force implementation does not support NPT integrators.
```
joaander, line 158:
```suggestion
        Walls are fixed in space and do not adjust with the box size. For example, NPT simulations may not behave as expected.
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+    Note:
+        - The virial due to walls is computed, but the pressure and reported by
+          ``hoomd.md.compute.ThermodynamicQuantities`` is not well defined. The
+          volume (area) of the box enters into the pressure computation, which
+          is not correct in a confined system. It may not even be possible to
+          define an appropriate volume with soft walls.
+        - An effective use of wall forces **requires** considering the geometry
+          of the system. Each wall is only evaluated in one simulation box and
+          thus is not periodic. Forces will be evaluated and added to all
+          particles from all walls in the wall group. Additionally there are no
```
joaander, line 169:
```suggestion
          particles from all walls. Additionally there are no
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+            to use for the potential.
+
+    Wall force evaluated using the Lennard-Jones potential.  See
+    `hoomd.md.pair.LJ` for force details and base parameters and `WallPotential`
+    for generalized wall potential implementation.
+
+    Example::
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        lj = hoomd.md.wall.LJ(walls)
```
joaander, line 237:
```suggestion
        lj = hoomd.md.external.wall.LJ(walls=walls)
```
Fix `external` module use and encourage users to use keyword arguments.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+
+    Wall force evaluated using the Gaussian potential.  See
+    `hoomd.md.pair.Gauss` for force details and base parameters and
+    `WallPotential` for generalized wall potential implementation
+
+    Example::
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        # add walls to interact with
+        gaussian_wall=hoomd.md.wall.Gauss(walls)
```
joaander, line 300:
```suggestion
        gaussian_wall=hoomd.md.external.wall.Gauss(walls=walls)
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+            "params", "particle_types",
+            hoomd.data.parameterdicts.TypeParameterDict(epsilon=float,
+                                                        sigma=float,
+                                                        r_cut=float,
+                                                        r_extrap=0.0,
+                                                        len_keys=1))
+        self._add_typeparam(params)
+
+
+class SLJ(WallPotential):
```
joaander, line 344: Remove SLJ. It will soon be replaced by ExpandedLJ.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+
+    Wall force evaluated using the Yukawa potential.  See `hoomd.md.pair.Yukawa`
+    for force details and base parameters and `WallPotential` for generalized
+    wall potential implementation
+
+    Example::
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        # add walls to interact with
+        yukawa_wall = hoomd.md.wall.Yukawa(walls)
```
joaander, line 425:
```suggestion
        yukawa_wall = hoomd.md.external.wall.Yukawa(walls=walls)
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+    Wall force evaluated using the Morse potential.  See
+    :py:class:`hoomd.md.pair.Morse` for force details and base parameters and
+    :py:class:`WallPotential` for generalized wall potential implementation
+
+    Example::
+
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        # add walls to interact with
+        morse_wall=hoomd.md.wall.Morse(walls)
```
joaander, line 487:
```suggestion
        morse_wall=hoomd.md.external.wall.Morse(walls=walls)
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+    Wall force evaluated using the Force-shifted Lennard-Jones potential.  See
+    `hoomd.md.pair.ForceShiftedLJ` for force details and base parameters and
+    `WallPotential` for generalized wall potential implementation.
+
+    Example::
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        # add walls to interact with
+        shifted_lj_wall=hoomd.md.wall.ForceShiftedLJ(
+            walls)
```
joaander, line 549:
```suggestion
        shifted_lj_wall=hoomd.md.external.wall.ForceShiftedLJ(
            walls=walls)
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+
+    Wall force evaluated using the Mie potential.  See `hoomd.md.pair.Mie` for
+    force details and base parameters and `WallPotential` for generalized wall
+    potential implementation
+
+    Example::
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        # add walls to interact with
+        mie_wall=hoomd.md.wall.Mie(walls, default_r_cut=3.0)
```
joaander, line 609:
```suggestion
        mie_wall=hoomd.md.external.wall.Mie(walls=walls)
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+        * ``r_extrap`` (`float`, **optional**) -
+          The distance to extrapolate the potential, defaults to 0
+          :math:`[\mathrm{length}]`
+
+        Type: `TypeParameter` [``particle_types``, `dict`]
+    """
+
+    _cpp_class_name = "WallsPotentialMie"
+
+    def __init__(self, walls=None):
```
joaander, line 639:
```suggestion
    def __init__(self, walls):
```
What use-case would users have for initializing the class with no walls?

This comment applies to all the wall classes.

hoomd/wall.py
```diff
@@ -0,0 +1,396 @@
...
+    Args:
+        origin (`tuple` [`float`,`float`,`float`], optional):
+            A point that lies on the plane used with ``normal`` to fully specify
+            the plane.
+        normal (`tuple` [`float`,`float`,`float`], optional):
+            The normal vector to the plane.
+        inside (`bool`, optional):
+            Whether particles are restricted to the space inside or outside the
+            cylinder. Inside is the side of the plane the normal points to, and
+            outside constitutes the other side.
```
joaander, line 170: Plane walls should not have an `inside` parameter. Note that the "inside" of a plane is the side that the normal points towards.

Perhaps we should consider inverting this logic in the code? Computer graphics literature and software typically define outward facing normals.

joaander (changes requested):
Thanks, @b-butler and @j-proc!

The overall architecture is excellent and the user-facing API is clear. As a bonus, there will be some level of interoperability between MD and HPMC walls, nice!

I have some suggestions for improvement.

joaander:
Also:
* remove `wall.py` from `setup.cfg` and `.pre-commit-config.yaml` as it no longer exists.
* Fix the failing GPU builds in CI.

CMakeLists.txt
```diff
@@ -56,8 +56,8 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     "MinSizeRel" "RelWithDebInfo")
 endif()
 
-# enable c++14
-set(CMAKE_CXX_STANDARD 14)
+# enable c++17
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CUDA_STANDARD 14)
```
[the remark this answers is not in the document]
b-butler: This lead to compilation errors on the GPU, but I forget exactly why.

hoomd/md/EvaluatorExternalElectricField.h
```diff
@@ -69,7 +69,7 @@ class EvaluatorExternalElectricField
 #endif // ifndef __HIPCC__
         } __attribute__((aligned(16)));
 
-    typedef Scalar3 field_type;
```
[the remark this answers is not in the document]
b-butler: While I am not sure this is useful, this does change the behavior to not allow per type electric fields.

hoomd/md/external/wall.py
b-butler: To set the walls later, so they don't have to be defined at construction, but we can require it at construction.

hoomd/wall.py
b-butler: That is true, we just need to be clear in our documentation which half-space the normal points to. Here I don't know if it makes much difference since it is about which direction the wall is active in which is not a typical use case.

hoomd/md/EvaluatorExternalElectricField.h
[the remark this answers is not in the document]
joaander: The electric force is `\vec{F}_i = q_i \vec{E}`, where `q_i` is a per-**particle** quantity. This is already more specific than a per-type field.

CMakeLists.txt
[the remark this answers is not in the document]
joaander: Then we can fix the errors.

hoomd/wall.py
joaander: We should remove barriers to interoperation so that building walls in 3D graphics tools (e.g. blender) can be a typical use case.

tcmoore3 (commented):
Nice work. Most of my comments are smaller edits. However, my big suggestions are:

1. (Echoing Josh's comment) Add a figure to explain what is meant by the inside and outside of walls, including how forces are computed for extrapolated and non-extrapolated mode.
2. Add more comments around the tests, especially the semi-validation tests, so that you can quickly interpret what is being tested.
3. Not necessary to merge this PR, but a tutorial would be super helpful

hoomd/wall.py
```diff
@@ -0,0 +1,396 @@
...
+        radius (float):
+            The radius of the circle faces of the cylinder.
+        origin (`tuple` [`float`,`float`,`float`]):
+            The origin of the cylinder defined as the center of the bisecting
+            circle along the cylinder's axis.
+        axis (`tuple` [`float`,`float`,`float`]):
+            A vector perpendicular to the circular faces.
+        inside (bool):
+            Whether particles are restricted to the space inside or outside the
+            cylinder.
```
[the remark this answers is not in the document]
j-proc: You could actually generate ellipses this way although I'd have to think about if the projection of the force to 2D would work out correctly. Or I should say the person who wants to use it that way can think through that math as all of the details of implementation are documented. 

A more obviously working case is a cylinder on its side being used for a channel. While this is obviously the same as two plane walls in this case, I'm sure someone might prefer it.

hoomd/wall.py
```diff
@@ -0,0 +1,396 @@
...
+    def to_dict(self):
+        """Return a dictionary specifying the sphere."""
+        return {
+            "radius": self.radius,
+            "origin": self.origin,
+            "inside": self.inside
+        }
+
+
+class Cylinder(WallGeometry):
```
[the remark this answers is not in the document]
j-proc: Their length is determined by the box dimensions (barring being intentionally capped by a larger sphere or planes). Like all wall objects they exist in bounds of the simulation box. Making sure that a geometry configuration is commensurate with the box is not trivial for most cases.
This is of course without the added complications of a floppy box and which constraints would keep things commensurate. That and the pressure definition of HOOMD being incompatible with walls blocking off amounts of volume are the main things blocking NPT walls. 

So basically the wall geometry setup can come with a note to user that it is up to them to get right. But clarifying that the geometry is specified within the simulation box can be stated for clarity if you think that would help.

CMakeLists.txt
```diff
@@ -56,8 +56,8 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     "MinSizeRel" "RelWithDebInfo")
 endif()
 
-# enable c++14
-set(CMAKE_CXX_STANDARD 14)
+# enable c++17
```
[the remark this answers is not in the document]
j-proc: Josh and Brandon are already discussing the solution that doesn't involve this change a few lines down. Marking resolved just to clean up the overview.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,653 @@
...
+
+Wall potentials add forces to any particles within a certain distance,
+:math:`r_{\mathrm{cut}}`, of each wall. In the extrapolated mode, all particles
+outside of the wall boundary are included as well.
+
+Wall geometries (`hoomd.wall`) are used to specify half-spaces. There are two
+half spaces for each of the possible geometries included and each can be
+selected using the inside parameter. In order to fully specify space, it is
+necessary that one half space be closed and one open. Setting *inside=True* for
+closed half-spaces and *inside=False* for open ones. See `hoomd.wall` for more
```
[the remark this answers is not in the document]
j-proc: This makes more sense for spheres and cylinders than it does planes, but yes for cases when you want to keep particles out of a region of space or the periodic boundaries make it simpler or more efficient to specify a geometry this way.
I personally felt leaving it for planes kept things simpler by consistency while also enabling users to quickly copy a geometry and flip the inside parameter to make something more like field geometries. It's of particular use in my opinion when using something like a Gaussian potential that actually needs consistency at its core/zero distance.

hoomd/wall.py
j-proc: My only counter argument to inverting the logic would be that in computer graphics it makes sense for the normals to be outward facing because you're typically thinking of the polygons that describe a discrete object that interacts with the world/light simulations outwardly. I feel like it's really a matter of perspective here that we're usually thinking of a confinement situation where we'd be inside of that object. 

However obviously it's just a case of documenting the choice we make. If @joaander is envisioning interoperability with general 3D renders as a common use with walls then I'd tend to agree going with their inside definition makes more sense.

j-proc: @b-butler I actually make the argument here https://github.com/glotzerlab/hoomd-blue/pull/1020#discussion_r748693794 that it is a common use case for some potentials. So I would argue it is important to keep and clearly define.

hoomd/md/external/wall.py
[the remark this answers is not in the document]
j-proc: @tcmoore3 I'm also rereading this and wanted to clarify you mean geometrically/spatially closed and not like mathematical interval closed and open, correct?

hoomd/wall.py
b-butler: I should clarify, I meant there isn't a standard direction for the normal in the case of confining walls in software that I know of since this is much more niche than graphics not that the ability to use each half-space is not useful.

hoomd/md/pytest/test_wall_potential.py
```diff
@@ -0,0 +1,242 @@
...
+            p["r_extrap"] = _params.rng.random() * 5
+        else:
+            p["r_extrap"] = r_extrap
+    return base
+
+
+_params.rng = np.random.default_rng(26456)
+
+
+@pytest.mark.parametrize("cls, params", zip(_potential_cls, _params()))
```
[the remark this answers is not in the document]
b-butler: We can but that would do exactly what `zip` is doing here, and we use `_params` with different values so then the list would need to be constructed multiple times.

hoomd/wall.py
```diff
@@ -0,0 +1,396 @@
...
+
+    def __init__(self, type, index=0):
+        self.index = index
+        self.type = type
+
+    def __repr__(self):
+        return f"_MetaListIndex(type={self.type}, index={self.index})"
+
+
+class _WallsMetaList(MutableSequence):
```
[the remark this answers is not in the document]
b-butler: Yes in `hoomd/md/pytest/test_meta_wall_list.py`

hoomd/md/WallData.h
```diff
@@ -178,7 +176,7 @@ vecPtToWall(const PlaneWall& wall, const vec3<Scalar>& position, bool& inside)
     {
     vec3<Scalar> t = position;
     Scalar d = dot(wall.normal, t) - dot(wall.normal, wall.origin);
-    inside = (((d >= 0.0) && wall.inside) || ((d < 0.0) && !(wall.inside))) ? true : false;
```
j-proc, line 177: @joaander @b-butler 
I'm afraid I'm going to have to argue we actually don't do this. While the other day I was thinking of it as a matter of convenience, when actually seeing this bit of code getting removed it occurred to me that it introduces a real issue.

Removing the inside parameter will actually mean that there's no way to make our space open vs closed. If you simply flip the normal, you're going to have the 0 point be in both wall sections.
I keep using a wall with a Gaussian potential as my test case and for this instance you can imagine that while the forces would both evaluate to zero and not cause a problem, the energy would be twice what it should be for that particle.
I don't feel like the user should have to think about this edge case and try to figure out what will work out in the floating point math to move the origin just enough to not trigger this. If we keep this parameter internally but don't like the vagueness of "inside" for the user we could rename it at the python level to "open"/"closed" or something like that.

joaander: What use-case would require that the user be able to configure open vs closed spaces? In either case, we should make the plane geometry consistent with the sphere and cylinder geometries. Either we include the surface of the shape on the inside or we don't. In practical simulations, it doesn't make any difference because 1e-16 is infinitesimally far away from 0.

With any potential that is finite at r=0 (and `r_extrap=0`), the particles can pass through the wall into the outside space where they will move freely. There will be a large jump in the force and energy as the particle crosses the wall why may cause the integration to explode. This is not a configuration users will likely wish to choose. They will either use a force that goes infinite at r=0 or set `r_extrap > 0` so that particles that move outside the wall will be pushed back in.

I still think that having two variables (the normal direction and the `inside` flag) to determine which side of the plane the particles interact with is confusing to users. There should be one way to set this.

In your example are you imagining two walls place with opposing normals that the particle is supposed to pass through without a jump in energy and force? What research use case would that be used for? Wouldn't it be cleaner to have a `two_sided` flag that evaluates one wall potential on the absolute value of the signed distance?

j-proc: > 
> 
> What use-case would require that the user be able to configure open vs closed spaces? In either case, we should make the plane geometry consistent with the sphere and cylinder geometries. Either we include the surface of the shape on the inside or we don't. In practical simulations, it doesn't make any difference because 1e-16 is infinitesimally far away from 0.
> 
There's a few use cases I suppose we were trying to cover by making these wall geometries very general and flexible. Perhaps the name wall is what is most confusing about this however. I think while these options could be confusing to some users on a first read-through the documentation, those who are looking for more complicated uses would find a complete implementation that allows them to construct a consistent field geometry. I'll illustrate this more through the rest of my response.

Wall as a name obviously makes the most sense when you're talking about potentials which go infinite at r=0. For these use case whether the space is open or closed makes *no difference* as you pointed out. We made a choice that perhaps feels arbitrary now that inside would be the closed space, although I think this is still the choice I'd make for two reasons:
1) It can have an easy geometric interpretation for 2/3 geometries
2) Without introducing a new keyword we also can make it clear that the space is open or closed (i.e. inside is always closed/includes the bounds of the geometry)
If for some reason you think that it makes more sense for inside to always be open, I don't have a problem with it, but it might perhaps make sense to take a quick poll since I would then say the intuitive answer is subjective and we should default to a majority opinion and obviously still document it.

> With any potential that is finite at r=0 (and `r_extrap=0`), the particles can pass through the wall into the outside space where they will move freely. There will be a large jump in the force and energy as the particle crosses the wall why may cause the integration to explode. This is not a configuration users will likely wish to choose. They will either use a force that goes infinite at r=0 or set `r_extrap > 0` so that particles that move outside the wall will be pushed back in.
> 
I perhaps need to go back and double check that the implementation hasn't changed. But my original intent of the `r_extrap` functionality was explicitly to extend a potential interaction through the wall smoothly. So in this case, the particle should pass through the wall and still be assigned the force evaluated at `r_extrap`. So it should not move freely unless V(`r_extrap`)/dr = 0. And I believe the potential energy of the particle was also assigned it's value in this extended picture. This is more a case that covers one wall potential being "two-sided". The whole point was essentially to avoid cases where you could end up with a large jump and to do so in as consistent of a way as possible. This certainly breaks down a bit as the particle crosses the box boundary, but that is unavoidable and hopefully possible to anticipate and either avoid or track via the ParticleData image.

I wrote the following and then realized that I think I implemented this wrong for this functionality! It was documented how it was coded, but to cover this use case, we would need to make a minor change covering this exact scenario. So glad we had this discussion and glad things are getting better testing. It's clearly my own logic that had a mismatch when coding one part and another.

>> This does bring up one case that I think I should perhaps code a test case for to be sure it works (assuming we agree on how it should work) regarding the combination of a potential that is finite at r=0 (and `r_extrap=0`), I'll list the components:
>> - A Gauss wall potential applied to *any* geometry
>> - `r_extrap=0`
>> - `inside=True`
>> - At any point on the wall geometry we should then have the exact value `epsilon` for the potential energy and 0 force.
>> - At any point "outside" the wall geometry ^^
>> However for `inside=False` I'm honestly not positive if there would be an issue. I don't think so because...
>> 
This is where I double checked myself before saying,  "I believe `r_extrap=0` is meant to still allow the whole space being open or closed to be overridden." What I coded is `r_extrap<=0` but I'm now making the argument for `r_extrap<0`. I think this is another case of it not making a difference for the vast majority of wall potentials, but I've clearly constructed an edge case that I'd argue could be used so I think it should be coded differently. This will of course need to cover the python side handling converting False to -1. rather than 0 which is probably why I didn't do this correctly the first time... a bug masquerading as simplicity.

> I still think that having two variables (the normal direction and the `inside` flag) to determine which side of the plane the particles interact with is confusing to users. There should be one way to set this.
> 
I'm definitely not going to say you're wrong on this. But I think if we set a default of `inside=True` (or False if you think it make us more compatible with computer graphics conventions) for planes users that don't need this functionality would ignore it and users that do could still easily find, understand, and use it.

> In your example are you imagining two walls place with opposing normals that the particle is supposed to pass through without a jump in energy and force? What research use case would that be used for? Wouldn't it be cleaner to have a `two_sided` flag that evaluates one wall potential on the absolute value of the signed distance?
>
Revisiting my above example of test case. If we did want this Gauss wall potential to be two sided we could only correctly construct it by having access to control over the spaces being open and closed. And I think this is the clearest case for why these functionalities should be combined. The two_sided potential is the only time when you'd want to have an open and a closed potential, but you'd also be combining the same geometry in a way that that needs to be named and I believe it's intuitively to refer to it as "inside and outside" which combined  makes the whole space defined.

joaander:
TODO: Once the `fix-broken-sphinx-refs` branch (#1175) is merged, restore the ``hoomd.md.wall`` reference `md/integrate.py`.

hoomd/wall.py
```diff
@@ -0,0 +1,411 @@
...
+    Whether the wall is interpreted as a sphere or circle is dependent on the
+    dimension of the system the wall is applied to. For 2D systems the
+    z-component of the origin should be zero.
+
+    Note:
+        `Sphere` objects are immutable.
+
+    Args:
+        radius (`float`):
+            The radius of the sphere.
```
joaander, line 46: Add units for all args and attributes.

joaander (changes requested):
Thanks! There are some changes needed before this can be merged.

We also need to solve the failing MPI tests. There appears to be a deadlock causing the test to time out.

joaander:
One more request: For each wall time, add an equation that shows how the signed distance to a point is computed from the wall parameters #1078.

hoomd/wall.py
```diff
@@ -0,0 +1,411 @@
...
+            "axis": self.axis,
+            "inside": self.inside,
+            "open": self.open
+        }
+
+
+class Plane(WallGeometry):
+    """Define a Plane in 3D Euclidean space.
+
+    The normal points to the active half space.
```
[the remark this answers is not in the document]
b-butler: We have, but I forgot to update the documentation.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,597 @@
...
+    energy is only applied to the half-space specified by each wall.
+    :math:`V(r)` is evaluated in the same manner as when the mode is shift for
+    the analogous :py:mod:`pair <hoomd.md.pair>` potentials within the
+    boundaries of the half-space.
+
+    .. math::
+
+        V(r)  = V_{\mathrm{pair}}(r) - V_{\mathrm{pair}}(r_{\mathrm{cut}})
+
+    For ``inside=True`` (closed) half-spaces:
```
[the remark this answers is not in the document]
b-butler: Yes, I need to update documentation here.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,597 @@
...
+    .. math::
+        :nowrap:
+
+        \begin{eqnarray*}
+        \vec{F}  = & -\nabla V(r) & 0 \le r < r_{\mathrm{cut}} \\
+                 = & 0 & r \ge r_{\mathrm{cut}} \\
+                 = & 0 & r < 0
+        \end{eqnarray*}
+
+    For ``inside=False`` (open) half-spaces:
```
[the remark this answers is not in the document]
b-butler: Yes need to update documentation

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,597 @@
...
+
+    All wall potential commands specify that a given potential energy and
+    force be computed on all particles in the system within a cutoff
+    distance, :math:`r_{\mathrm{cut}}`, from each wall.
+    The force :math:`\vec{F}` is in the direction of :math:`\vec{r}`, the vector
+    pointing from the particle to the wall or half-space boundary and
+    :math:`V_{\mathrm{pair}}(r)` is the pair potential specified by subclasses
+    of `WallPotential`.  Wall forces are implemented with the concept of
+    half-spaces in mind. There are two modes which are allowed currently in wall
+    potentials: standard and extrapolated.
```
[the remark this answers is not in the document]
b-butler: We compute the distance from the particle to the wall in C++ for the wall Geometry and then this is negated for some reason when actually computed rather than just computing the negated version first.

joaander:
The new latex math equations aren't rendering properly on readthedocs.

joaander (changes requested):
I will review again when the LaTeX math equations render correctly. See https://hoomd-blue--1020.org.readthedocs.build/en/1020/module-hoomd-wall.html

joaander (approved):
I applied my final change suggestions directly to the branch.
Looking forward to a review by @tcmoore3.

hoomd/wall.py
```diff
@@ -420,7 +427,7 @@ def _unsync(self):
         for wall_list in self._backend_lists.values():
             wall_list._unsync()
 
-    def _get_obj_backend_index(self, frontend_index, new_type, old_type=None):
+    def _get_0bj_backend_index(self, frontend_index, new_type, old_type=None):
```
b-butler, line 430: Why the name change?

b-butler:
@joaander thanks for the documentation fix.

hoomd/wall.py
joaander: This was an errant find and replace. Will fix.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+
+    Wall force evaluated using the Gaussian potential.  See
+    `hoomd.md.pair.Gauss` for force details and base parameters and
+    `WallPotential` for generalized wall potential implementation.
+
+    Example::
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        # add walls to interact with
+        gaussian_wall=hoomd.md.external.wall.Gauss(walls=walls)
```
tcmoore3, line 294:
```suggestion
        gaussian_wall = hoomd.md.external.wall.Gauss(walls=walls)
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+            to use for the potential.
+
+    Wall force evaluated using the Force-shifted Lennard-Jones potential.  See
+    `hoomd.md.pair.ForceShiftedLJ` for force details and base parameters and
+    `WallPotential` for generalized wall potential implementation.
+
+    Example::
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        shifted_lj_wall=hoomd.md.external.wall.ForceShiftedLJ(
```
tcmoore3, line 473:
```suggestion
        shifted_lj_wall = hoomd.md.external.wall.ForceShiftedLJ(
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+
+    Wall force evaluated using the Morse potential.  See
+    :py:class:`hoomd.md.pair.Morse` for force details and base parameters and
+    :py:class:`WallPotential` for generalized wall potential implementation.
+
+    Example::
+
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        morse_wall=hoomd.md.external.wall.Morse(walls=walls)
```
tcmoore3, line 413:
```suggestion
        morse_wall = hoomd.md.external.wall.Morse(walls=walls)
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+            to use for the potential.
+
+    Wall force evaluated using the Mie potential.  See `hoomd.md.pair.Mie` for
+    force details and base parameters and `WallPotential` for generalized wall
+    potential implementation.
+
+    Example::
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        mie_wall=hoomd.md.external.wall.Mie(walls=walls)
```
tcmoore3, line 533:
```suggestion
        mie_wall = hoomd.md.external.wall.Mie(walls=walls)
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+            to use for the potential.
+
+    Wall force evaluated using the Gaussian potential.  See
+    `hoomd.md.pair.Gauss` for force details and base parameters and
+    `WallPotential` for generalized wall potential implementation.
+
+    Example::
+
+        walls = [hoomd.wall.Sphere(radius=4.0)]
+        # add walls to interact with
```
tcmoore3, line 293:
```suggestion
        # add walls
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+            hoomd.data.parameterdicts.TypeParameterDict(epsilon=float,
+                                                        sigma=float,
+                                                        r_cut=float,
+                                                        r_extrap=0.0,
+                                                        len_keys=1))
+        self._add_typeparam(params)
+
+
+class Mie(WallPotential):
+    r"""Mie potential wall potential.
```
tcmoore3, line 520:
```suggestion
    r"""Mie wall potential.
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+class WallPotential(force.Force):
+    r"""Generic wall potential.
+
+    Warning:
+        `WallPotential` should not be used directly.  It is a base class that
+        provides features and documentation common to all standard wall
+        potentials.
+
+    All wall potential classes specify that a given potential energy and
+    force be computed on all particles in the system when the signed cutoff
```
tcmoore3, line 54: This sentence is pretty unclear. What does "when the signed cutoff distance is near the wall surface" mean?

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+    distance is near the wall surface: :math:`r < r_{\mathrm{cut}}`.
+    The force :math:`\vec{F}` is in the direction of :math:`\vec{r}`, the vector
```
tcmoore3, line 56: I think the directions are reversed here. This says that a particle that is very close to a LJ wall (e.g., in the repulsive region) would have a very large force acting on it that points toward the wall.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+
+        V(r)  = V_{\mathrm{pair}}(r) - V_{\mathrm{pair}}(r_{\mathrm{cut}})
+
+    For ``open=True`` spaces:
+
+    .. math::
+        :nowrap:
+
+        \begin{eqnarray*}
+        \vec{F}  = & -\nabla V(r) & 0 < r < r_{\mathrm{cut}} \\
```
tcmoore3, line 83:
```suggestion
        \vec{F}  = & -\nabla V(r) \,\,\,& 0 < r < r_{\mathrm{cut}} \\
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+    pointing from the particle to closest point on the wall's surface and
+    :math:`V_{\mathrm{pair}}(r)` is the pair potential specified by subclasses
+    of `WallPotential`.  Walls are two-sided surfaces with positive signed
+    distances to points on the active side of the wall and negative signed
+    distances to points on the inactive side. Additionally, the wall's mode
+    controls how forces and energies are computed for particles on or near the
+    inactive side. The ``inside`` flag (or ``normal`` in the case of
+    `hoomd.wall.Plane`) determines which side of the surface is active.
+
+    .. rubric:: Standard Mode.
```
tcmoore3, line 66:
```suggestion
    .. rubric:: Standard Mode
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+
+    In the standard mode, when :math:`r_{\mathrm{extrap}} \le 0`, the potential
+    energy is only computed on the active side. :math:`V(r)` is evaluated in the
+    same manner as when the mode is shift for the analogous :py:mod:`pair
+    <hoomd.md.pair>` potentials within the boundaries of the active space.
```
tcmoore3, line 71:
```suggestion
    <hoomd.md.pair>` potentials within the boundaries of the active space:
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+    inactive side. This can be useful to move particles from the inactive side
+    to the active side.
+
+    The extrapolated potential has the following form:
+
+    .. math::
+        :nowrap:
+
+        \begin{eqnarray*}
+        V_{\mathrm{extrap}}(r) =& V(r) &, r > r_{\rm extrap} \\
```
tcmoore3, line 122:
```suggestion
        V_{\mathrm{extrap}}(r) =& V(r) & r > r_{\rm extrap} \\
```

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+             =& V(r_{\rm extrap})
+                + (r_{\rm extrap}-r)\vec{F}(r_{\rm extrap})
+                \cdot \vec{n}&, r \le r_{\rm extrap}
```
tcmoore3, line 125:
```suggestion
                \cdot \vec{n}\,\,\,& r \le r_{\rm extrap}
```

Same for the equations below.

hoomd/md/external/wall.py
```diff
@@ -0,0 +1,577 @@
...
+
+
+    To use extrapolated mode ``r_extrap`` must be set per particle type.
+
+    .. attention::
+        Walls are fixed in space and do not adjust with the box size. For
+        example, NPT simulations may not behave as expected.
+
+    Note:
+        - The virial due to walls is computed, but the pressure and reported by
```
tcmoore3, line 164:
```suggestion
        - The virial due to walls is computed, but the pressure computed and reported by
```

hoomd/wall.py
```diff
@@ -0,0 +1,469 @@
...
+            "radius": self.radius,
+            "origin": self.origin,
+            "axis": self.axis,
+            "inside": self.inside,
+            "open": self.open
+        }
+
+
+class Plane(WallGeometry):
+    r"""Define a Plane in 3D Euclidean space.
```
tcmoore3, line 213:
```suggestion
    r"""Define a plane in 3D Euclidean space.
```

hoomd/wall.py
```diff
@@ -0,0 +1,469 @@
...
+            The origin of the cylinder defined as the center of the bisecting
+            circle along the cylinder's axis :math:`[\mathrm{length}]`.
+        inside (`bool`, optional):
+            Whether particles are restricted to the space inside or outside the
+            cylinder.
+        open (`bool`, optional):
+            Whether to include the surface of the cylinder in the space.
+            ``True`` means do not include the surface, defaults to ``True``.
+
+    Cylinder in HOOMD span the simulation box in the direction given by the
```
tcmoore3, line 134:
```suggestion
    Cylinder walls in HOOMD span the simulation box in the direction given by the
```

hoomd/wall.py
```diff
@@ -0,0 +1,469 @@
...
+from hoomd.data.syncedlist import identity, SyncedList
+
+from hoomd.operation import _HOOMDGetSetAttrBase
+from hoomd.data.parameterdicts import ParameterDict
+
+
+class WallGeometry(ABC, _HOOMDGetSetAttrBase):
+    """Abstract base class for a HOOMD wall geometry.
+
+    Walls are used in both HPMC and MD subpackages. Subclass of `WallGeometry`
```
tcmoore3, line 15:
```suggestion
    Walls are used in both HPMC and MD subpackages. Subclasses of `WallGeometry`
```

tcmoore3 (approved):
Looks good. Most suggestions are little things to clear up the docs.

tcmoore3 (changes requested):
I prematurely approved. Upon testing, I can't get a system with walls to actually run. My script is attached. The error message I get is:

```
**ERROR**: The required virial terms are not defined for the current setup.
  File "test-md-walls.py", line 75, in <module>
    sim.run(10000)
  File "/Users/tcmoore3/software/hoomd-builds/wall_api/lib/python3.7/site-packages/hoomd/simulation.py", line 436, in run
RuntimeError: NPT is not supported for requested features
```

I do not have an NPT integrator active in my script. It is possible that I'm missing something obvious, but if that's the case, then a lot of users will probably run into the same issue.

tcmoore3:
> I prematurely approved. Upon testing, I can't get a system with walls to actually run. My script is attached. The error message I get is:
> 
> ```
> **ERROR**: The required virial terms are not defined for the current setup.
> Traceback (most recent call last):
>   File "test-md-walls.py", line 75, in <module>
>     sim.run(10000)
>   File "/Users/tcmoore3/software/hoomd-builds/wall_api/lib/python3.7/site-packages/hoomd/simulation.py", line 436, in run
>     self._cpp_sys.run(steps_int, write_at_start)
> RuntimeError: NPT is not supported for requested features
> ```
> 
> I do not have an NPT integrator active in my script. It is possible that I'm missing something obvious, but if that's the case, then a lot of users will probably run into the same issue.

Script:
```python
import sys
import gsd.hoomd
import itertools
import math
import numpy as np
import os

import hoomd

# set up integrator potential and device
cpu = hoomd.device.CPU()
sim = hoomd.Simulation(device=cpu, seed=0)
cell = hoomd.md.nlist.Cell(buffer=0.4)
lj = hoomd.md.pair.LJ(nlist=cell)
lj.params[('A', 'A')] = dict(epsilon=1, sigma=1)
lj.r_cut[('A', 'A')] = 2.5
integrator.forces.append(lj)
nvt = hoomd.md.methods.NVT(kT=1.5, filter=hoomd.filter.All(), tau=1.0)
integrator.methods.append(nvt)

# make the system
m = 10
N_particles = m**3
spacing = 2**(1/6)
K = math.ceil(N_particles**(1 / 3))
L = K * spacing
x = np.linspace(-L / 2, L / 2, K, endpoint=False)
position = np.array(list((itertools.product(x, repeat=3))))
position = position - np.mean(position, axis=0)
position = position[0:N_particles]
orientation = [(1, 0, 0, 0)] * N_particles
snapshot = gsd.hoomd.Snapshot()
snapshot.particles.N = N_particles
snapshot.particles.position = position
snapshot.particles.orientation = orientation
snapshot.particles.typeid = [0] * N_particles
snapshot.configuration.box = [3*L, 3*L, 3*L, 0, 0, 0]
with gsd.hoomd.open(name='lattice.gsd', mode='wb') as f:
    f.append(snapshot)
sim.create_state_from_gsd(filename='lattice.gsd')
sim.state.thermalize_particle_momenta(filter=hoomd.filter.All(), kT=1.5)

# write to screen
logger = hoomd.logging.Logger(categories=['scalar'])
logger.add(sim, quantities=['timestep', 'tps'])
table_writer = hoomd.write.Table(
        trigger=hoomd.trigger.Periodic(300),
        logger=logger
)
sim.operations.writers.append(table_writer)

# gsd writer
gsd_shape_logger = hoomd.logging.Logger()
gsd_traj = hoomd.write.GSD(filename='trajectory.gsd',
                      trigger=hoomd.trigger.Periodic(100),
                      mode='wb')
sim.operations.writers.append(gsd_traj)

# add spherical wall
radius = 8.8
walls = [hoomd.wall.Sphere(radius=radius, inside=True)]
lj_walls = hoomd.md.external.wall.LJ(walls=walls)
lj_walls.params['A'] = {"sigma": 1.0, "epsilon": 1.0, "r_cut": 2.5, 'r_extrap': 0}
sim.operations.integrator.forces.append(lj_walls)

# run
sim.run(10000)
```
```

File: hoomd/wall.py [hunk 49]
```diff
@@ -0,0 +1,469 @@
...
+        return {
+            "radius": self.radius,
+            "origin": self.origin,
+            "axis": self.axis,
+            "inside": self.inside,
+            "open": self.open
+        }
+
+
+class Plane(WallGeometry):
```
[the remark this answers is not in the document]
joaander: The shapes are defined in 3D. While the signed distance is general and works in any dimension, the partial derivatives for the force are takin in 3 dimensions. This means that planes work as lines in 2D, provided that the normal's z is 0. Cylinders work as circles in 2D (redundant with spheres), provided that the axis is in the z direction. Without these conditions, the force will have a z component even in 2D simulations.

hoomd/md/external/wall.py
joaander: I suggest removing the comment entirely. The comment is irrelevant to the construction of a wall potential class.

hoomd/md/external/wall.py
joaander: How about : "where the signed distance :math:`d` from the particle to the wall is less than the potential's cutoff: :math:`d \le r_\mathrm{cut}`.

Also, let's change "specify" language. That is a HOOMD v1 relic when these were imperative commands. I've started converting these to the form "Wall potential classes compute the potential energy and force on ....."

hoomd/md/external/wall.py
joaander: Forces have signs, so stating "in the direction" here is ambiguous. I will push a change to correct this wording to be unambiguous. We also need to reference the vector from the wall to the point as `\vec{d}` because `\vec{r}` is used for particle positions.

hoomd/wall.py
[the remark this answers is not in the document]
joaander: On second thought, it isn't just the partial derivative, it is geometry. If a cylinder is tilted (axis z != 0) the shortest vector from the cylinder to point on the xy plane has a non-zero z component.

joaander:
@b-butler force pushing to branches makes it challenging for other developers to contribute.

hoomd/md/external/wall.py
joaander: What is this supposed to do? It doesn't seem to change anything in the rendered output.

hoomd/md/external/wall.py
joaander: Let me know what you think of the reworded overview.

joaander:
> I prematurely approved. Upon testing, I can't get a system with walls to actually run. My script is attached. The error message I get is:
> 
> ```
> **ERROR**: The required virial terms are not defined for the current setup.
> Traceback (most recent call last):
>   File "test-md-walls.py", line 75, in <module>
>     sim.run(10000)
>   File "/Users/tcmoore3/software/hoomd-builds/wall_api/lib/python3.7/site-packages/hoomd/simulation.py", line 436, in run
>     self._cpp_sys.run(steps_int, write_at_start)
> RuntimeError: NPT is not supported for requested features
> ```
> 
> I do not have an NPT integrator active in my script. It is possible that I'm missing something obvious, but if that's the case, then a lot of users will probably run into the same issue.

@tcmoore3 The loggers request the computation of the virial which triggers this error. This would have been present in HOOMD v2 as well - strange that no one reported it before. I removed the error as we have a clear warning in the docs.

Your script fails at a new point now because you place particles outside or very close to the wall, leading to very high forces.

hoomd/md/external/wall.py
tcmoore3: Much better.

hoomd/md/external/wall.py
tcmoore3: It adds a little bit of white space between the functional forms and valid domains. It was all mashed together and just looked like `V(r)0` without those spaces.

tcmoore3:
> > I prematurely approved. Upon testing, I can't get a system with walls to actually run. My script is attached. The error message I get is:
> > ```
> > **ERROR**: The required virial terms are not defined for the current setup.
> > Traceback (most recent call last):
> >   File "test-md-walls.py", line 75, in <module>
> >     sim.run(10000)
> >   File "/Users/tcmoore3/software/hoomd-builds/wall_api/lib/python3.7/site-packages/hoomd/simulation.py", line 436, in run
> >     self._cpp_sys.run(steps_int, write_at_start)
> > RuntimeError: NPT is not supported for requested features
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I do not have an NPT integrator active in my script. It is possible that I'm missing something obvious, but if that's the case, then a lot of users will probably run into the same issue.
> 
> @tcmoore3 The loggers request the computation of the virial which triggers this error. This would have been present in HOOMD v2 as well - strange that no one reported it before. I removed the error as we have a clear warning in the docs.
> 
> Your script fails at a new point now because you place particles outside or very close to the wall, leading to very high forces.

Thanks. That fix (and using a larger radius for the sphere wall) makes it run.

tcmoore3 (approved):
My only remaining question is about the [tests that don't do anything](https://github.com/glotzerlab/hoomd-blue/pull/1020/files#r779080241). But I'm not concerned enough about those to not approve the PR. If there's a reason for them to be there, cool, if not, just remove them.

b-butler:
@joaander sorry, I was trying to fix a commit I had just pushed I am surprised it required a force pull for you.

@tcmoore3 The blank tests are to avoid errors. In #1178, I added a full generic test suite for `list`, `tuple`, and `dict` like objects. I use that for the tests here, but since `list` like objects should have `__contains__` and `remove` , I need to nullify these tests.

joaander:
> @joaander sorry, I was trying to fix a commit I had just pushed I am surprised it required a force pull for you.

No worries. I needed a force pull on my `wall_api` checkout. The changes merged cleanly into another branch I have based on this one though, so no harm done.

PR #1021, opened 2021-05-26
Add md.pair.TWF to the documentation

joaander:
## Description

Add md.pair.TWF to the documentation

## Motivation and context

Implemented pair potentials should appear in the documentation.

I needed to make unique paper links because `paper link` appeared twice with different URLs which is invalid restructured text.

## How has this been tested?

I verified that the sphinx documentation included TWF locally.

## Change log

```
- Add ``md.pair.TWF`` to the documentation.
```

## Checklist:

### DIFF
--- a/hoomd/md/pair/pair.py
+++ b/hoomd/md/pair/pair.py
@@ -1602,10 +1602,10 @@ class OPP(Pair):
     coefficients. The coefficients must be set per unique pair of particle
     types.
 
-    The potential comes from Marek Mihalkovič and C. L. Henley 2012
-    `paper link`_.
+    The potential comes from `Marek Mihalkovič and C. L. Henley 2012`_.
 
-    .. _paper link: https://dx.doi.org/10.1103/PhysRevB.85.092102
+    .. _Marek Mihalkovič and C. L. Henley 2012:
+       https://dx.doi.org/10.1103/PhysRevB.85.092102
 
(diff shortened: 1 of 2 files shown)


hoomd/md/pair/pair.py
```diff
@@ -1602,10 +1602,10 @@ class OPP(Pair):
     coefficients. The coefficients must be set per unique pair of particle
     types.
 
-    The potential comes from Marek Mihalkovič and C. L. Henley 2012
-    `paper link`_.
+    The potential comes from `Marek Mihalkovič and C. L. Henley 2012`_.
 
-    .. _paper link: https://dx.doi.org/10.1103/PhysRevB.85.092102
```
b-butler, line 1608: multiple uses of `paper link` are fine you just need anonymous links which use two underscores instead of one.

b-butler (approved):
This looks good to  me.

hoomd/md/pair/pair.py
joaander: Yes, but descriptive lines are more accessible: https://usability.yale.edu/web-accessibility/articles/links

b-butler: Thanks for the link! I did not think of this. I made this change elsewhere for the line constraint, but we should revert those changes in another PR.

PR #1022, opened 2021-05-27
Import optional modules based on build configuration

joaander:
## Description

Use the CMake build configuration for `BUILD_HPMC`, `BUILD_MD`, etc.. to determine when to import these packages.

## Motivation and context

Recently, I helped @tommy-waltmann find an `ImportError` that was masked by the current `try/except ImportError` block. With the new code, that error would no longer be masked.

This also fixes another problem I have had at times. When I build HOOMD with all packages enabled, then change the build settings to disable a module, the old built package is still left in the build directory. Sometimes this leads to errors when the old package is not ABI compatible with the rest of the build, but `__init__.py` still tried to import it because the files are there.

## How has this been tested?

I manually tested each of the `BUILD_` settings in the ON and OFF configuration locally. We have no CI tests for this and I do not want to add that many additional CI configurations just to test all the possible ON/OFF combinations.

The new docs for the variables appear in the documentation.

## Change log

```
*Added*

- ``dem_built``, ``hpmc_built``, ``md_built``, and ``mpcd_built`` to ``hoomd.version`` - flags that indicate when optional submodules have been built.
```

## Checklist:

### DIFF
--- a/hoomd/CMakeLists.txt
+++ b/hoomd/CMakeLists.txt
@@ -32,6 +32,12 @@ else()
     set(_dem_built "False")
 endif()
 
+if (BUILD_METAL)
+    set(_metal_built "True")
+else()
+    set(_metal_built "False")
+endif()
+
 if (BUILD_MPCD)
     set(_mpcd_built "True")
 else()
--- a/hoomd/CMakeLists.txt
+++ b/hoomd/CMakeLists.txt
@@ -7,6 +7,37 @@ install(FILES ${HOOMD_BINARY_DIR}/hoomd/include/HOOMDVersion.h
         DESTINATION ${PYTHON_SITE_INSTALL_DIR}/include
         )
 
+# translate cmake true/false to Python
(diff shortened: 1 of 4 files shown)


tommy-waltmann (commented):
There are other build options like `BUILD_METAL` and `BUILD_JIT` which we may also want to define variables for. Those aren't imported in `hoomd/__init__.py` by default, but they could be useful in the future.

joaander:
> There are other build options like `BUILD_METAL` and `BUILD_JIT` which we may also want to define variables for. Those aren't imported in `hoomd/__init__.py` by default, but they could be useful in the future.

Good catch. I added `BUILD_METAL`. 

#937 refactors JIT so that it is no longer a separate component. If not already done, we need to remove `BUILD_JIT` and replace it with `ENABLE_LLVM` on that branch so I'll skip `BUILD_JIT` on this PR.

tommy-waltmann (approved):

PR #1023, opened 2021-05-27
Rewrite compilation docs

joaander:
## Description

Here is a rewrite of the compilation docs following discussions in #1011. Per some offline discussions, I also converted `CONTRIBUTING.md` to `CONTRIBUTING.rst` and include it in the documentation. Unfortunately, this does need to be imported from the root because github stipulates the directory names where it looks for this file, and `sphinx-doc` is not one of them: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors#adding-a-contributing-file

In the rewrite, I use the "bottom line up front" style - where the most important and general information is listed first and details are further expanded on later.

## Motivation and context

#1011 outlines some of the reasons. I've also had offline discussions where a number of people call hoomd's build system complicated. It is no more complicated than the median open source tool that requires compilation, and the docs should make that clear. HOOMD can be compiled with a standard install prereqs, get source, configure, and build steps.

I use the "bottom line up front" style thinking that this will help. The existing docs bury the needed build steps in the middle of dozens of details that don't apply to everyone, which can make it hard to parse. The rewritten docs give the basic outline of the process at the top and then readers can go down to the more detailed sections to find tips and options that apply to them.

I also keep the discussion on installing prerequisites very general because it is impossible to provide snippets that can be copied and pasted for every package manager that readers might be interested in using. I try to use the canonical names of these tools, but various package managers name them all differently - and some (like apt-get) break a number of the packages up (e.g. you need to install python3, python3-distutils, and python3-dev). It's not an ideal set up because the reader needs to do some work, but we can't play favorites choose one specific package manager.

## How has this been tested?

I considered testing this by observing someone completely new to building code follow the instructions, but I don't know how practical that is.

## Checklist:


CONTRIBUTING.rst
```diff
@@ -0,0 +1,114 @@
...
+`versionadded, versionchanged, and deprecated Sphinx
+directives <https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded>`__
+so that users will be aware of how functionality changes from version to
+version.
+
+Add developer to the credits
+____________________________
+
+Update the credits documentation to reference what each developer
+contributed to the code.
```
b-butler, line 108: Do we still do this? I think we just list contributors now.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,114 @@
...
+Document public-facing API with Python docstrings in Google style.
+
+Document version status
+_______________________
+
+Each user-facing Python class, method, etc. with a docstring should have
+`versionadded, versionchanged, and deprecated Sphinx
+directives <https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded>`__
+so that users will be aware of how functionality changes from version to
+version.
```
b-butler, line 102: Have we been following this? If this is expected then we need to add version 3.0 to a lot of docstrings if not remove.

INSTALLING.rst
```diff
@@ -1,308 +1,46 @@
...
+    conflicts with conda-forge_ packages.
 
-    $ cd /path/to/hoomd-blue
-    $ cd sphinx-doc
-    $ make html
+.. tip::
 
-Then, open the documentation in ``_build/html/index.html``. On some platforms, you can use this
-command to do open the file::
+    miniforge_ environments automatically add the ``conda-forge`` channel.
```
b-butler, line 43:
```suggestion
    miniforge_ environments automatically adds the ``conda-forge`` channel as the top priority channel.
```

sphinx-doc/components.rst
```diff
@@ -11,7 +11,7 @@ Components can be compiled and installed as **built-in** components or as
 **external** components. The build process compiles HOOMD-blue and all built-in
 components together, requiring one ``cmake`` and ``make``. External components
 compile and link against a HOOMD-blue installation with a separate invocation of
-``cmake`` and ``make install``. A properly configured component may be compiled
+``cmake`` and ``cmake --install``. A properly configured component may be compiled
```
b-butler, line 14: Is this to ensure whatever build system selected at configuration is used?

sphinx-doc/testing.rst
```diff
@@ -32,8 +32,8 @@ Execute the following commands to run the tests:
 * ``python3 -m pytest hoomd``
 
 **ctest** must be run in the **build** directory. When you run **pytest**
-outside of the **build** directory, it will test the HOOMD installed by ``make
-install``.
+outside of the **build** directory, it will the ``hoomd`` package that Python
+imports.
```
b-butler, line 36:
```suggestion
outside of the **build** directory, it will test the ``hoomd`` package that Python
imports (which may not be the version just built).
```

b-butler (changes requested):
I like the overview a lot. I think it does a good job of being concise while showing that from a fresh start to compiled binary is only a few steps.

Also, while not new to this myself, I think the tips are very beginner friendly.

I have some general comments/suggestions, but nothing major.

CONTRIBUTING.rst
joaander: You are correct. I will update this.

CONTRIBUTING.rst
joaander: I will update this. The unwritten guideline we have been following is that everything is implicitly new in the most recent major release and all `versionadded` and `versionchanged` directives should be removed in 3.0.0. We will start following this with new/changed features in 3.1.

sphinx-doc/components.rst
joaander: Yes. `cmake --build` and `cmake --install` are cross-platform and independent of the build tool. CMake forwards these commands onto the build tool selected at the configure step. I removed references to `make install` since they are specific to the make tool. Users that are familiar with their tools are of course welcome to use tool specific commands.

joaander: I reworded this section to match the new suggested general syntax in `BUILDING.rst`

joaander:
This is ready for another review. @seoulfood as a new HOOMD user, I'd appreciate your thoughts on the new documentation.

sphinx-doc/style.rst
```diff
@@ -65,14 +65,14 @@ C++/CUDA
...
-    * Member variables: ``m_`` prefix followed by lowercase with words
-      separated by underscores ``m_member_variable``
-    * Constants: all upper-case with words separated by underscores
-      ``SOME_CONSTANT``
-    * Functions: ``lowerCamelCase``
+  * Namespaces: All lowercase ``somenamespace``
+  * Class names: ``UpperCamelCase``
+  * Methods: ``lowerCamelCase``
+  * Member variables: ``m_`` prefix followed by lowercase with words
+    separated by underscores ``m_member_variable``
```
b-butler, line 72:
```suggestion
  * Member variables: ``m_`` prefix with snake case (e.g. ``m_member_variable``)
```

sphinx-doc/style.rst
```diff
@@ -65,14 +65,14 @@ C++/CUDA
...
+  * Constants: all upper-case with words separated by underscores
+    ``SOME_CONSTANT``
```
b-butler, line 74:
```suggestion
  * Constants: ``UPPER_SNAKE_CASE``
```

b-butler (approved):
I left some suggestions which you can add if you like otherwise, LGTM.

Issue #882, opened 2020-12-04
Port getar to the v3 API


joaander:
## Description

Port getar (dump and init) to the v3 API.

## Motivation and context

Does anyone use getar any longer? Should we remove it instead?

mphoward:
I would vote to remove since it eliminates a dependency, unless someone has a strong need for it.

joaander:
@klarh are you you aware of anyone currently relying on getar files with HOOMD? Would you be interested or know someone would would be interested in porting it to the v3 API?

klarh:
@joaander I still produce them regularly and some members of Julia's group may also use them, although I haven't kept track of who exactly is using them. I'm willing to port the code, assuming that APIs and data structures haven't changed too drastically, though I'm attending a conference this week so I won't be able to start for at least another week. Is there a written summary of hoomd v2->v3 changes that would be relevant to porting somewhere?

joaander:
No worries on the timing, the v3 transition progress will be continuing for some time. The main changes needed are at the interface between the python and C++ code. There are a few changes in the python API relevant for enabling getar. 

1) Initialization of the state is now performed in `Simulation.create_state_from_*` methods - we'll need a `create_state_from_getar`. 

2) Parameter setting/getting is now handled by `_HOOMDBaseObject`. You can see the docs of this and its parent classes or ask @b-butler for details. The short version is that we now match property names for parameters on the Python and C++ levels and defer the initialization of the C++ object until `_attach` is called. Before attaching, `_HOOMDBaseObject` maintains parameters in Python only and after attaching it passes through all parameter get/set to the C++ object. Take a look at the GSD writer code changes between the 2.9 and master branches to see a good example of what is needed.

3) Unit tests are now in pytest. See the various files in hoomd/pytest/ for examples.

joaander:
One other change to consider is that HOOMD v3 now uses cmake to find dependencies (pybind11, eigen, etc..) instead of building embedded copies. We could switch to depending on an external sqlite and maybe even an external libgetar. We could consider making libgetar a soft dependency in this case and disable those features when the library is not available.

PR #1026, opened 2021-06-03
Update GTAR format I/O for v3

klarh:
## Description

This PR-in-progress updates the GTAR file format readers and writers for the new hoomd v3 API. To-do items are listed in the checklist below.

I'm considering adding support to filter which particles are dumped while I'm updating things, but I haven't decided if it's worth the effort yet.

## Motivation and context

Resolves issue #882 

## How has this been tested?

Tests will be added/ported as the PR progresses.

## Change log

```
* Add the ability to initialize a simulation from a getar-format file (`Simulation.create_state_from_getar`)
* Add the ability to dump getar-format-files (`hoomd.write.GTAR`)
```

## Checklist:

- [ ] Add/update unit tests
- [ ] Properly thread parameters through `_HOOMDBaseObject`'s getting/setting mechanism

### DIFF
--- a/hoomd/GetarDumpWriter.cc
+++ b/hoomd/GetarDumpWriter.cc
@@ -291,11 +291,12 @@ const bool& NeedSnapshots::operator[](NeedSnapshotIdx index) const
 
 GetarDumpWriter::GetarDumpWriter(std::shared_ptr<SystemDefinition> sysdef,
                                  const std::string& filename,
+                                 std::shared_ptr<ParticleGroup> group,
                                  GetarDumpMode operationMode,
                                  unsigned int offset)
     : Analyzer(sysdef), m_archive(), m_periods(), m_offset(offset), m_staticRecords(),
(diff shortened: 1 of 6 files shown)


joaander:
Thanks for getting started on this, @klarh. Would it be possible to remove the `libgetar` submodule and instead find and link to the installed package? We could make this an optional dependency.

@b-butler can answer any questions you might have on `_HOOMDBaseObject`.

joaander:
From a philosophical perspective, I've switched to external libraries for those that are readily available, like `cereal`, `eigein`, and `pybind`, but still use submodules for less common ones like `random123`. Here `libgetar` is in a middle ground as it is available on `pypi`, but not via other channels. I could be convinced either way on this point.

I think I asked about making this change on the issue when I had recently enabled `-Werror` on the CI builds. The embedded sqlite code generated a number of warnings that took some source file properties to quiet. It would have been convenient to not be building sqlite embedded in HOOMD, but it is done. Going forward, we may run into issues where the embedded sqlite code needs to be updated to build with newer distributions or compilers.

PR #1028, opened 2021-06-04
New potential with Mie parameters and SLJ parameters

AlainKadar:
## Description
A new modified LJ potential whereby the 6-12 powers are variable parameters (in the manner of the `Mie` potential). The new potential also takes a parameter to radially shift the potential energy. Radial shifting by a user-specified parameter is different to how it is enabled in the `SLJ` potential (`SLJ` shifts by an amount calculated from particle diameters; it is not directly parametrizable https://hoomd-blue.readthedocs.io/en/latest/module-md-pair.html#hoomd.md.pair.SLJ ).

## Motivation and context
The behavior of many colloidal/nanoscale particulate systems differs from the point charge behavior for which LJ was derived. Better agreement between experimental and computational studies is ensured by parameterizing the potential's exponents. Furthermore, systems where the particles are much larger than points (commonly observed in experiment) require radial shifting so that strong repulsion between particles is felt before centre-centre contact. Finally, polydispersity requires that this shifting be a user-defined function of particle pair sizes (which is different to the fixed function used by `SLJ`. However, user specified radial shifting will become the new method for shifting in future HOOMD releases).

## How has this been tested?
Agreement between hand calculations and `ExpandedMie.compute_energy` results observed for a variety of different parameter values.

## Change log

```
Expanded Mie pair potential added
```

## Checklist:


hoomd/md/CMakeLists.txt
```diff
@@ -227,6 +228,7 @@ set(_md_cu_sources ActiveForceComputeGPU.cu
                       GaussDriverPotentialPairGPU.cu
                       LJDriverPotentialPairGPU.cu
                       MieDriverPotentialPairGPU.cu
+		      SMieDriverPotentialPairGPU.cu
```
syjlee, line 231: indentation not aligned?

hoomd/md/pair/pair.py
```diff
@@ -1421,6 +1421,72 @@ def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
...
+            of energy)
+
+          * ``sigma`` (`float`, **required**) - :math:`\\sigma` (in distance
+            units)
+
+          * ``n`` (`float`, **required**) - :math:`n` (unitless)
+
+          * ``m`` (`float`, **required**) - :math:`m` (unitless)
+          
+          * ``Delta`` (`float`, **required**) - :math:`\\Delta` (in distance units)
```
syjlee, line 1469: I'm wondering if delta should be all lowercase as a keyword.

syjlee:
Just to make sure the purpose of this PR, you're introducing delta parameter explicitly to let the users pick delta, instead of getting delta from particle diameter property?

hoomd/md/pair/pair.py
AlainKadar: I kept it capitalized because it is a capital Delta in the equation

AlainKadar:
> Just to make sure the purpose of this PR, you're introducing delta parameter explicitly to let the users pick delta, instead of getting delta from particle diameter property?

Yes but it will also allow users to use a delta parameter for the generic Mie potential (just now, radial shifting is only enabled for LJ potential, not Mie)

hoomd/md/EvaluatorPairSMie.h
```diff
@@ -0,0 +1,211 @@
...
+                sm4 = v["m"].cast<Scalar>();
+
+                auto sigma(v["sigma"].cast<Scalar>());
+                auto epsilon(v["epsilon"].cast<Scalar>());
+
+                Scalar outFront = (sm3/(sm3-sm4)) * fast::pow(sm3/sm4, sm4/(sm3-sm4));
+                sm1 = outFront * epsilon * fast::pow(sigma, sm3);
+                sm2 = outFront * epsilon * fast::pow(sigma, sm4);
+
+                sm5 = v["Delta"].cast<Scalar>();
```
joaander, line 89:
```suggestion
                delta = v["delta"].cast<Scalar>();
```
You are right that this is a capital delta from a language perspective, from a code style perspective users will expect the parameter dictionary keys to be named like variables which are all lowercase with underscores separating words. There is no lowercase delta to confuse this with in the potential, so it is not a problem.

Also, feel free to name struct members with more meaningful names than `smN`.

hoomd/md/EvaluatorPairSMie.h
```diff
@@ -0,0 +1,211 @@
...
+                v["Delta"] = sm5;
+                return v;
+                }
+            #endif
+            }
+            #ifdef SINGLE_PRECISION
+            __attribute__((aligned(8)));
+            #else
+            __attribute__((aligned(16)));
+            #endif
```
joaander, line 113:
```suggestion
            __attribute__((aligned(16)));
```
This struct has 5 Scalar quantities which is most efficiently read with 16-byte wide memory transactions.

hoomd/md/EvaluatorPairSMie.h
```diff
@@ -0,0 +1,211 @@
...
+        DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& pair_eng, bool energy_shift)
+            {
+            // precompute some quantities
+            Scalar rcutinv = fast::rsqrt(rcutsq);
+            Scalar rcut = Scalar(1.0) / rcutinv;
+            Scalar r = fast::sqrt(rsq);
+            Scalar rinv = fast::rsqrt(rsq);
+
+            // compute the force divided by r in force_divr
+            if (r < (rcut + sm5) && sm1 != 0)
```
joaander, line 162:
```suggestion
            if (rsq < (rcutsq) && sm1 != 0)
```
The neighbor list is unaware of `\delta` so the r_cut condition cannot be modified here. If the user wishes to modify `r_cut` based on `\delta`, that is up to them when they set the parameter.

I think after making this change `rcutinv` and `rcut` are no longer needed.

hoomd/md/pair/pair.py
```diff
@@ -1165,6 +1165,82 @@ def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
         self._add_typeparam(params)
 
 
+class SMie(Pair):
+    r"""Shifted Mie pair potential.
```
joaander, line 1169: TODO: Later I'll have some comments on how to update these docs to be consistent with recently merged and soon to be merged PRs that update this style.

joaander (changes requested):
Thanks!

The main comment I have is that the name of the potential should be `ExpandedMie`. I also have a few other change requests below.

Also, this branch is out of date with `master`. There are a number of changes on `master` due to the pre-commit style guidelines. Let me know if you need help merging `master` into this branch. We can do that before or after you make the requested changes.

joaander:
> Just to make sure the purpose of this PR, you're introducing delta parameter explicitly to let the users pick delta, instead of getting delta from particle diameter property?

Yes. And I plan to make the same change for LJ in the future. This PR is setting the new standard.

hoomd/md/EvaluatorPairSMie.h
AlainKadar: Ok yes, I don't think either `rcut` or `rcutinv` are required. I removed them, with tests passing.

hoomd/md/AllExternalPotentials.h
```diff
@@ -30,26 +30,27 @@ typedef PotentialExternal<EvaluatorExternalPeriodic> PotentialExternalPeriodic;
...
 typedef PotentialExternal<EvaluatorExternalElectricField> PotentialExternalElectricField;
 
-typedef PotentialExternal<EvaluatorWalls<EvaluatorPairLJ>> WallsPotentialLJ;
-typedef PotentialExternal<EvaluatorWalls<EvaluatorPairSLJ>> WallsPotentialSLJ;
-typedef PotentialExternal<EvaluatorWalls<EvaluatorPairForceShiftedLJ>> WallsPotentialForceShiftedLJ;
-typedef PotentialExternal<EvaluatorWalls<EvaluatorPairMie>> WallsPotentialMie;
-typedef PotentialExternal<EvaluatorWalls<EvaluatorPairGauss>> WallsPotentialGauss;
-typedef PotentialExternal<EvaluatorWalls<EvaluatorPairYukawa>> WallsPotentialYukawa;
-typedef PotentialExternal<EvaluatorWalls<EvaluatorPairMorse>> WallsPotentialMorse;
+typedef PotentialExternal<EvaluatorWalls<EvaluatorPairLJ> > WallsPotentialLJ;
```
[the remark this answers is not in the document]
joaander: I don't know. `git blame` indicates that you didn't change these lines and yet here they are. In any case, `pre-commit` is going to change them back when we run clang-format on this branch: https://github.com/glotzerlab/hoomd-blue/pull/1028/checks?check_run_id=2785679872#step:4:101

hoomd/md/pair/pair.py
```diff
@@ -1165,6 +1165,83 @@ def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
         self._add_typeparam(params)
 
 
+class ExpandedMie(Pair):
+    r"""Expanded Mie pair potential.
+
+    Args:
+        nlist (`hoomd.md.nlist.NList`): Neighbor list
+        r_cut (float): Default cutoff radius (in distance units).
+        r_on (float): Default turn-on radius (in distance units).
```
joaander, line 1174:
```suggestion
        r_cut (float): Default cutoff radius :math:`[\\mathrm{length}]`.
        r_on (float): Default turn-on radius :math:`[\\mathrm{length}]`.
```
Use the new-style units formatting from #1019

joaander (changes requested):
Here are the promised documentation change comments.

Also, add `ExpandedMie` to `module-md-pair.rst` so it will show up in the documentation. Note that the class lists are alphabetized.

hoomd/md/pair/pair.py
joaander: Apply the same styling to epsilon, sigma, delta, n, and m below.

joaander:
I unfortunately will not be able to review until Jun 28 at the earliest. @b-butler maybe you could take a look at this while I'm out.

hoomd/md/EvaluatorPairExpandedMie.h
```diff
@@ -0,0 +1,218 @@
...
+                \left( \frac{\sigma}{r-\Delta} \right)^{m} \right] r \l (r_{\mathrm{cut}} +
+   \Delta)\\
+                = & 0 & r \ge (r_{\mathrm{cut}} + \Delta) \\
+    \f}
+    where  \f$ \Delta is specified by the user, conventionally calcualted as, \f$ \Delta = (d_i +
+   d_j)/2 - \sigma \f$ and \f$ d_i \f$ is the diameter of particle \f$ i \f$.
+
+    The ExpandedMie potential needs neither charge nor diameter. Two parameters are specified and
+   stored in a Scalar5. \a rep \a att \a npow \a mpow \a delta are stored in \a params.x \a params.y
+   \a params.z and \a params.w \a params.v respectively.
```
b-butler, line 49: This seems incorrect.

hoomd/md/EvaluatorPairExpandedMie.h
```diff
@@ -0,0 +1,218 @@
...
+        Scalar rinv = fast::rsqrt(rsq);
+
+        // compute the force divided by r in force_divr
+        if (rsq < rcutsq && rep != 0)
+            {
+            Scalar rmd = r - delta;
+            Scalar rmdinv = Scalar(1.0) / rmd;
+            Scalar rmd2inv = rmdinv * rmdinv;
+            Scalar rmdninv = fast::pow(rmd2inv, npow / Scalar(2.0));
+            Scalar rmdminv = fast::pow(rmd2inv, mpow / Scalar(2.0));
```
b-butler, line 175:
```suggestion
            Scalar rmdninv = fast::pow(rmd2inv, npow / Scalar(2.0));
```
These lines are the same.

hoomd/md/EvaluatorPairExpandedMie.h
```diff
@@ -0,0 +1,218 @@
...
+   \a params.z and \a params.w \a params.v respectively.
+
+
+    These are related to the standard lj parameters sigma and epsilon and the variable exponents n
+   and m by:
+    - \a rep = epsilon * pow(sigma,n) * (n/(n-m)) * power(n/m,m/(n-m))
+    - \a att = epsilon * pow(sigma,m) * (n/(n-m)) * power(n/m,m/(n-m))
+    - \a npow = n
+    - \a mpow = m
+    - \a delta = delta
```
b-butler, line 58: I would suggest using `repulsive`, `atractive`, `n_pow`, and  `m_pow` as it is more clear what each represents at a glance.

hoomd/md/EvaluatorPairExpandedMie.h
```diff
@@ -0,0 +1,218 @@
...
+            {
+            npow = v["n"].cast<Scalar>();
+            mpow = v["m"].cast<Scalar>();
+
+            auto sigma(v["sigma"].cast<Scalar>());
+            auto epsilon(v["epsilon"].cast<Scalar>());
+
+            Scalar outFront = (npow / (npow - mpow)) * fast::pow(npow / mpow, mpow / (npow - mpow));
+            rep = outFront * epsilon * fast::pow(sigma, npow);
+            att = outFront * epsilon * fast::pow(sigma, mpow);
```
b-butler, line 93: Class, method, and function names are camel case, variables are snake case typically. Also, `prefactor` may be a more apt name.

hoomd/md/EvaluatorPairExpandedMie.h
```diff
@@ -0,0 +1,218 @@
...
+#endif
+
+    protected:
+    Scalar rsq;    //!< Stored rsq from the constructor
+    Scalar rcutsq; //!< Stored rcutsq from the constructor
+    Scalar rep;    //!< mie1 parameter extracted from the params passed to the constructor
+    Scalar att;    //!< mie2 parameter extracted from the params passed to the constructor
+    Scalar npow;   //!< mie3 parameter extracted from the params passed to the constructor
+    Scalar mpow;   //!< mie4 parameter extracted from the params passed to the constructor
+    Scalar delta;  //!< delta parameter extracted from the call to setDiameter
```
b-butler, line 215: These don't actually describe the member variables.

hoomd/md/pair/pair.py
```diff
@@ -1165,6 +1165,78 @@ def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
         self._add_typeparam(params)
 
 
+class ExpandedMie(Pair):
+    """Expanded Mie pair potential.
+
+    Args:
+        nlist (`hoomd.md.nlist.NList`): Neighbor list
+        r_cut (float): Default cutoff radius :math:`[\\mathrm{length}]`.
+        r_on (float): Default turn-on radius :math:`[\\mathrm{length}]`.
```
b-butler, line 1174: These are changing to `default_r_on` and `default_r_cut` now. Please go ahead with this change for this potential (no need to modify others since they are changed on another branch).

hoomd/md/pair/pair.py
```diff
@@ -1165,6 +1165,78 @@ def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
...
+        dict(epsilon=1.0, sigma=1.0, n=12, m=6, delta=0.5)
+        expanded_mie.r_cut[('A', 'B')] = 2**(1.0/6.0)
+        expanded_mie.r_on[('A', 'B')] = 2.0
+        expanded_mie.params[(['A', 'B'], ['C', 'D'])] = \
+        dict(epsilon=1.5, sigma=2.0, n=12, m=6, delta=0.5))
+    """
+    _cpp_class_name = "PotentialPairExpandedMie"
+
+    def __init__(self, nlist, r_cut=None, r_on=0., mode='none'):
+
```
b-butler, line 1226:
```suggestion
```

hoomd/md/pytest/test_potential.py
```diff
@@ -248,6 +248,19 @@ def _invalid_params():
     invalid_params_list.extend(
         _make_invalid_params(slj_invalid_dicts, md.pair.SLJ, {}))
 
+    expanded_mie_valid_dict = {
```
b-butler, line 251: You need to add some lines to 567 as well to enable all parametrized tests.

hoomd/md/EvaluatorPairExpandedMie.h
```diff
@@ -0,0 +1,218 @@
...
+
+#ifndef ENABLE_HIP
+        //! set CUDA memory hints
+        void set_memory_hint() const { }
+#endif
+
+#ifndef __HIPCC__
+        param_type() : rep(0), att(0), npow(0), mpow(0), delta(0) { }
+
+        param_type(pybind11::dict v)
```
b-butler, line 83:
```suggestion
        param_type(const pybind11::dict v)
```

hoomd/md/EvaluatorPairExpandedMie.h
```diff
@@ -0,0 +1,218 @@
...
+            auto epsilon(v["epsilon"].cast<Scalar>());
+
+            Scalar outFront = (npow / (npow - mpow)) * fast::pow(npow / mpow, mpow / (npow - mpow));
+            rep = outFront * epsilon * fast::pow(sigma, npow);
+            att = outFront * epsilon * fast::pow(sigma, mpow);
+
+            delta = v["delta"].cast<Scalar>();
+            }
+
+        pybind11::dict asDict()
```
b-butler, line 98:
```suggestion
        pybind11::dict asDict() const
```

hoomd/md/EvaluatorPairExpandedMie.h
```diff
@@ -0,0 +1,218 @@
...
+    //! Evaluate the force and energy
+    /*! \param force_divr Output parameter to write the computed force divided by r.
+        \param pair_eng Output parameter to write the computed pair energy
+        \param energy_shift If true, the potential must be shifted so that V(r) is continuous at the
+       cutoff \note There is no need to check if rsq < rcutsq in this method. Cutoff tests are
+       performed in PotentialPair.
+
+        \return True if they are evaluated or false if they are not because we are beyond the cutoff
+    */
+    DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& pair_eng, bool energy_shift)
```
b-butler, line 162:
```suggestion
    DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& pair_eng, bool energy_shift) const
```

hoomd/md/EvaluatorPairExpandedMie.h
```diff
@@ -0,0 +1,218 @@
...
+        {
+        // precompute some quantities
+        Scalar r = fast::sqrt(rsq);
+        Scalar rinv = fast::rsqrt(rsq);
```
b-butler, line 166:
```suggestion
        const Scalar r = fast::sqrt(rsq);
        const Scalar rinv = fast::rsqrt(rsq);
```

b-butler (changes requested):
The only comments I have are style/naming for the most part. I recommend using `const` in a few places, there are more that can be found in `EvaluatorPairExpandedMie.h`, but I didn't enumerate them. While, we are not consistent about this we should get more in the habit of marking variables, arguments, and methods as `const` when appropriate.

AlainKadar:
I’ve applied the changes and added `const` suffix to function declarations where appropriate. My understanding is that this should be applied whenever a function doesn’t change variables that are both within the Evaluator class but outside of the function’s own list of arguments and variables (i.e. almost every function). Please correct me if this is wrong.

AlainKadar:
I’ve applied the changes and added `const` suffix to function declarations where appropriate. My understanding is that this should be applied whenever a function doesn’t change variables that are both within the Evaluator class but outside of the function’s own list of arguments and variables (i.e. almost every function). Please correct me if this is wrong.

hoomd/md/pair/pair.py
AlainKadar: For pytest checks to pass, I think the 2 branches will have to be merged given these changes?

b-butler: Yes, this is correct. I just merged master into this branch.

b-butler:
@AlainKadar the only thing left is to get the force and energy tests passing. I am not sure what was used to generate the numbers in the JSON file for the `ExtendedMie`, but either the reference values or implementation are wrong.

AlainKadar:
> @AlainKadar the only thing left is to get the force and energy tests passing. I am not sure what was used to generate the numbers in the JSON file for the `ExtendedMie`, but either the reference values or implementation are wrong.

Thanks for all the changes and comments! They were passing but I think your `Fix typo` commit changed something that was not a typo! Each `rmd` term is raised to a different power.

b-butler:
Hmm, my bad then :sweat_smile: . I didn't mean to change semantics. I was getting a compile error about a variable name that might have been where the error was added. Thanks for the fix @AlainKadar.

b-butler (approved):
LGTM

syjlee (approved):

b-butler:
@joaander, Do you plan to review this? You requested changes, so the merge is block until you review or unassign yourself I believe.

hoomd/md/pair/pair.py
```diff
@@ -1157,6 +1157,80 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0., mode='none'):
...
+    .. math::
+        :nowrap:
+
+        \\begin{eqnarray*}
+        V_{\\mathrm{mie}}(r)
+          = & \\left( \\frac{n}{n-m} \\right) {\\left( \\frac{n}{m}
+          \\right)}^{\\frac{m}{n-m}} \\varepsilon \\left[ \\left(
+          \\frac{\\sigma}{r} \\right)^{n} - \\left( \\frac{\\sigma}{r}
+          \\right)^{m} \\right] & r < r_{\\mathrm{cut}} \\\\
+          = & 0 & r \\ge r_{\\mathrm{cut}} \\\\
```
joaander, line 1181:
```suggestion
          = & \\left( \\frac{n}{n-m} \\right) {\\left( \\frac{n}{m}
          \\right)}^{\\frac{m}{n-m}} \\varepsilon \\left[ \\left(
          \\frac{\\sigma}{r-\Delta} \\right)^{n} - \\left( \\frac{\\sigma}{r-\Delta}
          \\right)^{m} \\right] & r < r_{\\mathrm{cut}} \\\\
          = & 0 & r \\ge r_{\\mathrm{cut}} \\\\
```
Missing the expanded part in the function.

hoomd/md/pair/pair.py
```diff
@@ -1157,6 +1157,80 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0., mode='none'):
...
+          * ``epsilon`` (`float`, **required**) - :math:`\\varepsilon`
+            :math:`[\\mathrm{energy}]`
+          * ``sigma`` (`float`, **required**) - :math:`\\sigma`
+            :math:`[\\mathrm{length}]`.
+          * ``n`` (`float`, **required**) -
+            :math:`n` :math:`[\\mathrm{dimensionless}]`.
+          * ``m`` (`float`, **required**) -
+            :math:`m` :math:`[\\mathrm{dimensionless}]`.
+          * ``delta`` (`float`, **required**) -
+            :math:`\\delta :math:`[\\mathrm{length}]`.
```
joaander, line 1200:
```suggestion
            :math:`\\Delta :math:`[\\mathrm{length}]`.
```

joaander: Most codes refer to this shift as as a capital greek letter delta.

joaander (changes requested):
Two small documentation fixes and this is ready to merge.

AlainKadar:
@joaander Thanks, to clarify, the style check only fails because I am not using raw strings?

vyasr:
@AlainKadar yes, you will need to use raw strings for the docstrings that contain all of the math. However, once you switch to raw strings you will also need to remove all the double escapes (`\\Delta`->`\Delta`) because the double backslash is specifically so that normal strings don't interpret `\D` as a control character. I would recommend building the documentation locally ([here's how](https://hoomd-blue.readthedocs.io/en/latest/building.html#build-the-documentation)) and verifying that you're actually rendering the math that you think you are after switching to raw strings.

AlainKadar:
> @AlainKadar yes, you will need to use raw strings for the docstrings that contain all of the math. However, once you switch to raw strings you will also need to remove all the double escapes (`\\Delta`->`\Delta`) because the double backslash is specifically so that normal strings don't interpret `\D` as a control character. I would recommend building the documentation locally ([here's how](https://hoomd-blue.readthedocs.io/en/latest/building.html#build-the-documentation)) and verifying that you're actually rendering the math that you think you are after switching to raw strings.

Great, thanks, @vyasr

hoomd/md/pair/pair.py
```diff
@@ -1157,6 +1157,82 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0., mode='none'):
...
+
+    .. math::
+        :nowrap:
+
+        \\begin{eqnarray*}
+        V_{\\mathrm{mie}}(r)
+          = & \\left( \\frac{n}{n-m} \\right) {\\left( \\frac{n}{m}
+          \\right)}^{\\frac{m}{n-m}} \\varepsilon \\left[ \\left(
+          \\frac{\\sigma}{r-\Delta} \\right)^{n} - \\left( \\frac
+          {\\sigma}{r-\Delta}
```
joaander, line 1180:
```suggestion
          \\frac{\\sigma}{r-\\Delta} \\right)^{n} - \\left( \\frac
          {\\sigma}{r-\\Delta}
```

hoomd/md/pair/pair.py
```diff
@@ -1157,6 +1157,82 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0., mode='none'):
...
+
+    .. py:attribute:: params
+
+        The Expanded Mie potential parameters.
+        The dictionary has the following keys:
+
+        * ``epsilon`` (`float`, **required**) -
+          :math:`\epsilon` :math:`[\\mathrm{energy}]`.
+        * ``sigma`` (`float`, **required**) -
+          :math:`\sigma` :math:`[\\mathrm{length}]`.
```
joaander, line 1196:
```suggestion
          :math:`\\epsilon` :math:`[\\mathrm{energy}]`.
        * ``sigma`` (`float`, **required**) -
          :math:`\\sigma` :math:`[\\mathrm{length}]`.
```

hoomd/md/pair/pair.py
```diff
@@ -1157,6 +1157,82 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0., mode='none'):
...
+        * ``n`` (`float`, **required**) -
+          :math:`n` :math:`[\\mathrm{dimensionless}]`.
+        * ``m`` (`float`, **required**) -
+          :math:`m` :math:`[\\mathrm{dimensionless}]`.
+        * ``delta`` (`float`, **required**) -
+          :math:`\Delta` :math:`[\\mathrm{length}]`.
```
joaander, line 1202:
```suggestion
          :math:`\\Delta` :math:`[\\mathrm{length}]`.
```

joaander (changes requested):
Thanks @AlainKadar . I think I suggest changes for all the remaining changes needed in this review.

I have no plans to ignore `W605 invalid escape sequence`. You can run [pre-commit](https://pre-commit.com/) locally to check for these kinds of errors before committing.

AlainKadar:
> Thanks @AlainKadar . I think I suggest changes for all the remaining changes needed in this review.
> 
> I have no plans to ignore `W605 invalid escape sequence`. You can run [pre-commit](https://pre-commit.com/) locally to check for these kinds of errors before committing.

Ok @joaander thanks for the clarification

joaander (approved):
Thanks!

PR #1044, opened 2021-06-16
Fix bug in domain decomposition for 2D systems

### LINKED ISSUE #1038: Domain decomposition does not divide the box efficiently in 2D simulations
## Description

The DomainDecomposition does not handle 2D simulation boxes properly. It doesn't divide the box in the x direction. This is caused by the surface area calculation when `L.z` is 0. https://github.com/glotzerlab/hoomd-blue/blob/cce72bf6b1c0d7e6d2c69797a6c0472b1cdaa97a/hoomd/DomainDecomposition.cc#L374-L376

After fixing, we need to remove the assert: https://github.com/glotzerlab/hoomd-blue/blob/cce72bf6b1c0d7e6d2c69797a6c0472b1cdaa97a/hoomd/DomainDecomposition.cc#L343

## Script

```python
import hoomd

device = hoomd.device.CPU()
s = hoomd.Snapshot(device.communicator)
N = 2
L = 20

if s.communicator.rank == 0:
    box = [L, L, 0, 0, 0, 0]
    s.configuration.box = box
    s.particles.N = N
    s.particles.position[:] = [[5, 0, 0], [-5, 0, 0]]
    s.particles.types = ['A']

sim = hoomd.Simulation(device=device)
sim.create_state_from_snapshot(s)
```

`mpirun -n 8 test.py`

## Output

```
HOOMD-blue is using domain decomposition: n_x = 1 n_y = 8 n_z = 1.
```

## Expected output

HOOMD-blue is using domain decomposition: n_x = 2 n_y = 4 n_z = 1 or HOOMD-blue is using domain decomposition: n_x = 4 n_y = 2 n_z = 1.

## Configuration

Platform:
- Mac
- Linux
- GPU
- CPU

Installation method:
- Conda package
- Compiled from source

Versions:
- Python version: 3.9
- **HOOMD-blue** version: master

tcmoore3:
Due to the new convention in v3 where systems are considered 2D if `Lz =
0`, the domain decomposition is broken for 2D systems.
This commit addresses that change.
In 3D, the domain decomposition is chosen as the one that minimizes the
surface area between the domains.
In 2D, we choose the decomposition that minimizes the linear perimeter
between domains.

## Description

Modifies the generation of the domain decomposition to properly handle 2D boxes now that the 2D-ness is implied by having a box with `Lz = 0`.

## Motivation and context

Inefficient domain decomposition != good.

Resolves #1038 

## How has this been tested?

I tested with the script posted in #1038, and it generates a `sqrt(n) x sqrt(n) x 1` decomposition instead of an `1 x n x 1` decomposition.

## Change log

```
- Fixes domain decomposition for 2D systems
```

## Checklist:

### DIFF
--- a/hoomd/DomainDecomposition.cc
+++ b/hoomd/DomainDecomposition.cc
@@ -346,9 +346,13 @@ bool DomainDecomposition::findDecomposition(unsigned int nranks,
     bool is2D = L.z == 0.0;
     double min_surface_area; // surface area in 3D, perimeter length in 2D
     if (is2D)
+        {
         min_surface_area = L.x * (double)(nranks - 1);
+        }
     else
+        {
         min_surface_area = L.x * L.z * (double)(nranks - 1);
+        }
 
     unsigned int nx_in = nx;
     unsigned int ny_in = ny;
@@ -358,8 +362,16 @@ bool DomainDecomposition::findDecomposition(unsigned int nranks,
 
     // initial guess
     nx = 1;
-    ny = nranks;
-    nz = 1;
+    if (is2D)
+        {
+        ny = nranks;
+        nz = 1;
+        }
+    else
+        {
+        ny = 1;
+        nz = nranks;
+        }
 
     for (unsigned int nx_try = 1; nx_try <= nranks; nx_try++)
         {
@@ -379,13 +391,14 @@ bool DomainDecomposition::findDecomposition(unsigned int nranks,
                     continue;
                 double surface_area;
                 if (is2D)
+                    {
                     surface_area = L.x * (ny_try - 1) + L.y * (nx_try - 1);
+                    }
                 else
                     {
                     surface_area = L.x * L.y * (double)(nz_try - 1)
                                    + L.x * L.z * (double)(ny_try - 1)
                                    + L.y * L.z * (double)(nx_try - 1);
-
                     }
                 if (surface_area < min_surface_area || !found_decomposition)
                     {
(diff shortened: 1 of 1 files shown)


hoomd/DomainDecomposition.cc
```diff
@@ -354,8 +358,8 @@ bool DomainDecomposition::findDecomposition(unsigned int nranks,
 
     // initial guess
     nx = 1;
-    ny = 1;
-    nz = nranks;
+    ny = nranks;
+    nz = 1;
```
joaander, line 362: Make the initial guess conditional on `is2D` to keep the existing behavior for 3D boxes where the z direction is divided first.

hoomd/DomainDecomposition.cc
```diff
@@ -371,9 +375,18 @@ bool DomainDecomposition::findDecomposition(unsigned int nranks,
...
                 if (nx_try * ny_try * nz_try != nranks)
                     continue;
-                double surface_area = L.x * L.y * (double)(nz_try - 1)
-                                      + L.x * L.z * (double)(ny_try - 1)
-                                      + L.y * L.z * (double)(nx_try - 1);
+                if (is2D && nz_try > 1)
+                    continue;
+                double surface_area;
+                if (is2D)
+                    surface_area = L.x * (ny_try - 1) + L.y * (nx_try - 1);
```
joaander, line 382:
```suggestion
                    {
                    surface_area = L.x * (ny_try - 1) + L.y * (nx_try - 1);
                    }
```
`clang-format` doesn't catch this one, but we should use curly braces even on small if blocks.

joaander (changes requested):
Thanks! This just needs a few tweaks.

joaander (approved):
Thanks!

PR #1045, opened 2021-06-17
Warn users on source import

joaander:
## Description

* Issue a warning when users attempt to import hoomd from the source directory.
* Make the testing documentation consistent with the build documentation - in that it doesn't assume the user has changed directories unless they are required to do so (for ctest).
* Document additional pytest invocation options that users and developers may want to be aware of.

## Motivation and context

A number of developers recently reported being confused at the behavior of pytest in hoomd. This at least provides a helpful warning when this occurs. HOOMD is a package with many compiled components, so pytest cannot run with the default discovery in a clean source directory as the authors designed. Due to our use of pytest plugins in conftest.py, `pytest` also cannot run with default test discovery in the build directory. The user must specify some option (either a path to the build directory or `--pyargs hoomd`) to indicate where to find the hoomd package to test.

I was unable to determine a configuration that allowed a bare `pytest` to execute tests either from the source or build directories. The closest I got was to add a `pytest.ini` with the contents:
```
[pytest]
addopts = hoomd
```
but adding this file broke several other invocation methods, including the ones that allow you to execute the tests from the source dir by specifying a full path to the build dir.

## How has this been tested?

I tested all the documented pytest invocation methods locally.

## Change log

```
*Changed*

- Warn users when hoomd is imported from the source directory.
```

## Checklist:

### DIFF
--- a/hoomd/__init__.py
+++ b/hoomd/__init__.py
@@ -19,9 +19,10 @@ if ((pathlib.Path(__file__).parent / 'CMakeLists.txt').exists()
     print("""Compile the package and import from the build directory or install
 the package and import from the Python environment.
 
-To run pytest, compile and execute `python3 -m pytest <build-directory>/hoomd`
-or install and execute, ensure your current working directory is outside the
-hoomd source directory and run `python3 -m pytest --pyargs hoomd`.
+To run pytest, either:
+(1) compile then execute `python3 -m pytest <build-directory>/hoomd` or
+(2) compile and install. Then, ensuring your current working directory is outside the
+hoomd source directory, execute `python3 -m pytest --pyargs hoomd`.
 """,
           file=sys.stderr)
 
--- a/hoomd/__init__.py
+++ b/hoomd/__init__.py
@@ -8,7 +8,22 @@
 simulations using HOOMD.
 """
 import sys
+import pathlib
 import os
+
+if ((pathlib.Path(__file__).parent / 'CMakeLists.txt').exists()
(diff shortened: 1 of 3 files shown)


hoomd/__init__.py
```diff
@@ -8,8 +8,23 @@
...
+    print("It appears that hoomd is being imported from the source directory:")
+    print(pathlib.Path(__file__).parent)
+    print()
+    print("""Compile the package and import from the build directory or install
+the package and import from the Python environment.
+
+To run pytest, compile and execute `python3 -m pytest <build-directory>/hoomd`
+or install and execute, ensure your current working directory is outside the
+hoomd source directory and run `python3 -m pytest --pyargs hoomd`.
+""",
```
cbkerr, line 25: Having trouble parsing the second option in this sentence. After trying some different ways, does this capture it?
```suggestion
To run pytest, either:
(1) compile then execute `python3 -m pytest <build-directory>/hoomd` or
(2) compile and install. Then, ensuring your current working directory is outside the
hoomd source directory, execute `python3 -m pytest --pyargs hoomd`.
""",
```

cbkerr (approved):
I think this would resolve most of my confusion. I tested the various combinations.

One mistake people might make is assume the build directory means "hoomd-source/build" rather than "build/hoomd" following the default instructions. We could add a hint, but it would also clutter up the instructions.

I tried to clarify some of the wording in the warning message but looks good otherwise.

PR #1054, opened 2021-07-13
New bond potential/Tethering potential

syjlee:
## Description
A new bond potential is added, where each end of bond is allowed to be distanced within a range, such that it behaves as if tethered by a string. Mathematically, this tethering bond potential is a sum of attractive potential and repulsive potential which are radially distanced, and giving you a deep well of potential between the minimum and the maximum bond length which you specify.
energy
force

## Motivation and context
This tethering bond potential is used to compose three dimensional triangulated mesh membrane, where each vertex of triangle mesh is interacting its bonded vertex under this potential.

## How has this been tested?

Validation will be tested through pytest in test_bond.py

## Change log

```
EvaluatorBondTether.h is added as a new file and other files are edited based on this.
```

## Checklist:


syjlee:
This is not ready to be reviewed. Need to compile and add test as well, also need further discussion on the definition of parameters.

hoomd/md/EvaluatorBondTether.h
```diff
@@ -0,0 +1,181 @@
...
+                       / (r - l_min) / (r - l_min));
+            }
+        if (k_b != Scalar(0.0))
+            {
+            force_divr = (F_att + F_rep) / r;
+
+            bond_eng = U_att + U_rep;
+            }
+
+        return true;
```
joaander, line 159: You can return `false` when the particle goes outside the valid range for the bond to get an error message.

hoomd/md/bond.py
```diff
@@ -381,3 +381,79 @@ def set_from_file(self, bondname, filename):  # noqa
...
                             rmax=rmax_table,
                             coeff=dict(V=V_table, F=F_table, width=self.width))
+
+
+class Tether(Bond):
+    r"""Tethering bond potential.
+
+    :py:class:`Tether` specifies a Tethering potential energy between two
+    particles in each defined bond.
+
```
joaander, line 391: Please include a citation to the paper that developed or first used this potential.

joaander (changes requested):
Thanks! 

Here is a record of the two changes we discussed. Other than that, this pull request looks great!

joaander (approved):
Thanks! This looks good to me.

@SchoeniPhlippsn what do you think?

hoomd/md/bond.py
```diff
@@ -381,3 +381,86 @@ def set_from_file(self, bondname, filename):  # noqa
...
+        \begin{eqnarray*}
+        V_{\mathrm{rep}}(r)  = & k_b \frac{exp(1/(r-l_{c1})}{r-l_{min}}
+                                & r < l_{c1}\\
+                                = & 0
+                                & r \ge l_{c1}
+        \end{eqnarray*}
+
+    .. math::
+        l_{min} < l_{c1} < l_{c0} < l_{max}
```
SchoeniPhlippsn, line 426: I think it is a good idea to raise an error if this requirement is not met.

SchoeniPhlippsn (commented):
Looks good to me as well. I have only one small comment below.

hoomd/md/bond.py
syjlee: Sounds great. Could you let me know how I can work around TypeParameter before arguments are assigned to cpp?

SchoeniPhlippsn: I would suggest to raise a `ValueError` by setting up a postprocess function. Something similar as in the `ReversePerturbationFlow` class in `update.py`.

syjlee: Since individual parameter in the dict cannot be validated unless validated as a whole, I added the ValueError in EvaluatorBondTether.h under pybind11 dict. I tested with an invalid set of parameters and got the ValueError with inserted warning message.
If this is not the most elegant way to throw an error and better be addressed in bond.py, I can make the change later in separate branch.

syjlee: If there's a better way to use if statement to check the order of params, please let me know!

hoomd/md/EvaluatorBondTether.h
```diff
@@ -0,0 +1,196 @@
...
+
+    tether_params(pybind11::dict v)
+        {
+        k_b = v["k_b"].cast<Scalar>();
+        l_min = v["l_min"].cast<Scalar>();
+        l_c1 = v["l_c1"].cast<Scalar>();
+        l_c0 = v["l_c0"].cast<Scalar>();
+        l_max = v["l_max"].cast<Scalar>();
+
+        if (l_min < l_c1 && l_min < l_c0 && l_min < l_max && l_c1 < l_c0 && l_c1 < l_max
```
syjlee, line 58: @SchoeniPhlippsn This is the added blob for order checking.

hoomd/md/EvaluatorBondTether.h
```diff
@@ -0,0 +1,196 @@
...
+            && l_c0 < l_max)
+            { /* deliberately empty to pass*/
+            }
+        else
+            {
+            throw std::invalid_argument("The input parameters should follow l_min < l_c1 < l_c0"
+                                        " < l_max");
+            }
```
joaander, line 66:
```suggestion
        if (!(l_min < l_c1 && l_min < l_c0 && l_min < l_max && l_c1 < l_c0 && l_c1 < l_max
            && l_c0 < l_max))
            {
            throw std::invalid_argument("The input parameters should follow l_min < l_c1 < l_c0"
                                        " < l_max");
            }
```
This is more readable.

joaander (changes requested):
Thanks! I have one suggestion to make the code more readable.

hoomd/md/EvaluatorBondTether.h
syjlee: Oh yeah much readable! I'll commit suggestion right away.

joaander (approved):
Thanks!

SchoeniPhlippsn (approved):
Looks good to me now!

PR #1059, opened 2021-07-20
Fix debug mode tests

### LINKED ISSUE #1029: Check all C++ asserts triggered by pytests
## Description

Run `python -m pytest hoomd` with a Debug build and fix failing asserts. 

## Motivation and context

I was running test with a Debug compiled build and found that [this assert](https://github.com/glotzerlab/hoomd-blue/blob/a7585d01293166cd86665de63ee75c0f22138877/hoomd/md/NeighborList.cc#L1489) fails for `test_write_gsd_dynamic`. I'm not clear on if it's a bad assert or another issue, but I thought we should sort it and any other failing asserts out before v3 is finalized. We can also consider if these calls point out important edge cases that are missed by our ctests.

joaander:
## Description

- Regenerate the exclusion list data arrays when the number of particles changes
- Allow setting the gpu error checking mode in debug builds.

## Motivation and context

These changes fix asserts thrown when running `pytest hoomd` in a Debug build.

Resolves #1029

## How has this been tested?

I tested this locally with a Debug build of HOOMD.

## Change log

```
- Update neighbor list exclusions after the number of particles changes.
```

## Checklist:

### DIFF
--- a/hoomd/md/NeighborList.cc
+++ b/hoomd/md/NeighborList.cc
@@ -366,6 +366,19 @@ void NeighborList::compute(uint64_t timestep)
     if (m_prof)
         m_prof->push("Neighbor");
 
+    // when the number of particles in the system changes, rebuild the exclusion list
+    if (m_need_reallocate_exlist)
+        {
+        clearExclusions();
+        std::set<std::string> exclusions_copy(m_exclusions);
+        for (const std::string& exclusion : exclusions_copy)
+            {
+            setSingleExclusion(exclusion);
+            }
+
+        m_need_reallocate_exlist = false;
+        }
+
     // take care of some updates if things have changed since construction
     if (m_force_update)
         {
@@ -1838,7 +1851,6 @@ void export_NeighborList(py::module& m)
         .def("getSmallestRebuild", &NeighborList::getSmallestRebuild)
         .def("getNumUpdates", &NeighborList::getNumUpdates)
         .def("getNumExclusions", &NeighborList::getNumExclusions)
-        .def("wantExclusions", &NeighborList::wantExclusions)
 #ifdef ENABLE_MPI
         .def("setCommunicator", &NeighborList::setCommunicator)
 #endif
--- a/hoomd/md/NeighborList.cc
+++ b/hoomd/md/NeighborList.cc
@@ -52,7 +52,7 @@ NeighborList::NeighborList(std::shared_ptr<SystemDefinition> sysdef, Scalar r_bu
     m_rebuild_check_delay = 0;
     m_exclusions_set = false;
 
-    m_need_reallocate_exlist = false;
+    m_n_particles_changed = false;
 
     // initialize box length at last update
     m_last_L = m_pdata->getGlobalBox().getNearestPlaneDistance();
@@ -198,7 +198,7 @@ NeighborList::NeighborList(std::shared_ptr<SystemDefinition> sysdef, Scalar r_bu
     TAG_ALLOCATION(m_ex_list_idx);
 
     // reset exclusions
-    clearExclusions();
+    resizeAndClearExclusions();
 
     m_ex_list_indexer = Index2D((unsigned int)m_ex_list_idx.getPitch(), 1);
     m_ex_list_indexer_tag = Index2D((unsigned int)m_ex_list_tag.getPitch(), 1);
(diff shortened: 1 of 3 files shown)


hoomd/md/NeighborList.cc
```diff
@@ -366,6 +366,19 @@ void NeighborList::compute(uint64_t timestep)
     if (m_prof)
         m_prof->push("Neighbor");
 
+    // when the number of particles in the system changes, rebuild the exclusion list
+    if (m_need_reallocate_exlist)
+        {
+        clearExclusions();
+        std::set<std::string> exclusions_copy(m_exclusions);
+        for (const std::string& exclusion : exclusions_copy)
```
b-butler, line 374: Since we are not editing exclusions is there a reason we aren't iterating over `m_exclusions` directly?

b-butler (approved):
The changes are fine; I am curious at one point, but the code looks sound.

hoomd/md/NeighborList.cc
joaander: `setSingleExclusion` inserts `exclusion` into `m_exclusions`. I do not know whether inserting an existing key into a `set` invalidates iterators, so I iterate over a copy for safety.

hoomd/md/NeighborList.cc
```diff
@@ -366,6 +366,19 @@ void NeighborList::compute(uint64_t timestep)
...
+            {
+            setSingleExclusion(exclusion);
+            }
+
+        m_need_reallocate_exlist = false;
```
vyasr, line 379: I think this is probably just because I'm not very familiar with this part of the code, but I find something vaguely unsettling about this block right now. Perhaps you could clarify a few things for me:

It looks like this flag is being set by the particle number change signal handler. That seems sensible, but it looks like `clearExclusions` also uses that check for _part_ of its logic and unsets `m_need_reallocate_exlist` at the end (my understanding is that it clears the list unconditionally but only reallocates if this flag is set, indicating that e.g. the size of the exclusion list has to change to accommodate a change in the number of particles). It makes sense to condition this block on `if (m_need_reallocate_exlist)` since in this case you don't want to clear the exclusions unconditionally, but that raises the following questions for me:
1. Is there any situation where the code could hit `clearExclusions` after the particle number changes, but before `compute` is called, such that the exclusion list is cleared but `setSingleExclusion` is not called in this loop?
2. The two differences I see between this look and the `setExclusions` method are that `setFilterBody(false)` call and the fact that this flag is not being set in that method. Is the former difference meaningful? Should `setExclusions` be checking that the argument list length matches the current exclusion list length, or is that method purely an internal-facing method for modifying without resizing the exclusion list?
3. Could `setExclusions` be modified to accept an iterator rather than a `pybind::list` and then reused here? 
4. Setting this variable to false here is superfluous since `clearExclusions` does so as well. I assume that was just a small missed redundancy, but just want to double-check in light of my other questions: is there any reason we wouldn't want to rely on that behavior of `clearExclusions`?

Sorry to drag out review of a PR that generally looks fine and is very small, I just want to make sure I understand what's happening here.

hoomd/md/NeighborList.cc
vyasr: Insertion invalidates iterators for sequential containers but not for associative containers according to [the standard, 23.2.4/9](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4296.pdf) (I picked C++14, but I believe it holds even in C++03). That said, this code looks fine to me; it's not performance critical and I see no reason to rely on that property of the containers.

hoomd/md/NeighborList.cc
joaander: @vyasr The C++ code was originally written to interface with the v2 python API. In v2, in the setup phase of every call to `run`, Python would call `clearExclusions` and then call the relevant `addExclusionsFrom*` methods. Thus, the number of particles was only allowed to change in between calls to `run`.

In v3, the number of particles can change during a run as well, and there is no setup phase to the `run` method. When we converted the neighborlist over to the v3 API, we failed to handle updates to the exclusion list. This failure resulted in a debug assertion which was reported in #1029.

1. No, there is no case where `clearExclusions` would be called first. We could rename the function to `resizeAndClearExclusions`, make it private, and rename `m_need_reallocate_exlist` to make the meaning of the variable more clear. The intention is to delay the reallocation  (and subsequent reinitialization) of the data structure until the last possible moment just before it is needed.

2. When the list is reset due to a size change, the body filtering boolean is not changed. If the user changes that flag, it will trigger a new call to `setExclusions` which will update the flag accordingly.

3. I don't know.

4. See (1).

vyasr: Got it thanks for the explanations. I think this is good to go then.

vyasr (approved):

PR #1062, opened 2021-07-21
Implement constrain.Distance

### LINKED ISSUE #870: Port constrain.distance to the v3 API
## Description

Port constrain.distance to the v3 API and add tests.

### LINKED ISSUE #1039: Apply pre-commit to constrain.py
## Description

We merged the rigid bodies PR, but `constrain.py` is still ignored by pre-commit. Remove `constrain.py` from the ignores in `setup.cfg` and `pre-commit-config.yaml` and solve the resulting errors in pre-commit.

## Motivation and context

We need to apply the pre-commit style formatters and guidelines to `constrain.py`. `constrain.distance` is not ported yet, but we could add a few `#noqa` comments where needed.

joaander:
## Description

Port `md.constrain.distance` to the v3 API.

## Motivation and context

Add support for a feature present in v2.

resolves #870, resolves #1039

## How has this been tested?

This pull request adds a unit test. A trajectory frame generated by the simulation in the unit test looks like this:
polymers

## Change log

```
- ``md.constrain.Distance`` - fix distances between pairs of particles.
```

## Checklist:


hoomd/md/constrain.py
```diff
@@ -44,73 +44,55 @@ def _attach(self):
         super()._attach()
 
 
-class distance(Constraint):
-    R"""Constrain pairwise particle distances.
+class Distance(Constraint):
+    """Constrain pairwise particle distances.
```
tommy-waltmann, line 48: I think an example of how this is done would be good to add since using this operation involves setting some things in the snapshot in addition to using this class. That could be done either in this docstring or in a tutorial, but I think this operation will be confusing to figure out how to use without an example.

hoomd/md/constrain.py
```diff
@@ -44,73 +44,55 @@ def _attach(self):
...
 
-    :py:class:`distance` specifies that forces will be applied to all particles
-    pairs for which constraints have been defined.
+    Args:
+        tolerance (float): Relative tolerance for constraint violation warnings.
 
-    The constraint algorithm implemented is described in:
+    `Distance` applies constraint forces between particles to satisfy
+    user-specified distance constraints. The algorithm implemented is described
```
tommy-waltmann, line 54: Do the distance constraints fix a distance between particles, or do they prevent the distance between particles from exceeding a maximum value? This could be clarified somewhere in this docstring to make this class's purpose clearer.

tommy-waltmann (changes requested):
Just a couple of comments that I think could improve documentation

hoomd/md/constrain.py
joaander: Explaining this is far too much to put in a docstring. The docstring states all of the needed information on where the constraint data is specified and how to apply the constraint forces. The point you bring up is general to bonds, angles, and dihedrals as well and would be good to explain in a tutorial: https://github.com/glotzerlab/hoomd-examples/issues/38

hoomd/md/constrain.py
joaander: This implements a set of equality constraints using Lagrange multipliers. I have reworded the documentation to clarify.

Constraints based on inequalities are much more complex to solve. Many MD engines provide fixed distance constraints as some force fields specify this instead of using harmonic bonds.

tommy-waltmann (approved):

b-butler (approved):
LGTM!

PR #1067, opened 2021-07-25
Alchemical MD integration methods

j-proc:
## Description
This pull request adds an `alchemy` module to `hoomd.md` to enable molecular dynamics digital alchemy simulations.

## Motivation and context

AlchMD code has previously been private, following a significant recoding effort we prepare to release it to the public and support it as part of the HOOMD package.

## How has this been tested?

We added a suite of unit tests for the alchemy classes. I also attempted to reproduce the results of the square lattice toy system from the SI of the [original digital alchemy paper](https://pubs.acs.org/doi/10.1021/acsnano.5b04181). The results are consistent with those from the paper, in that the location of the second minimum in the LJGauss potential oscillates around ~1.4 in the validation tests. The original paper found a minimum in the alchemical potential around a value of 1.38. Exact quantitative agreement should not be expected between the validation tests and the paper: we use springs to tether the particles in place and measure the steady-state value of the alchemical parameter whereas the paper measures the constitutive relation between the value of the alchemical parameter and the alchemical potential and estimates the minimum using ocular detection.

## Change log

Add digital alchemy methods to molecular dynamics package in `hoomd.md.alchemy`

## TODO:

- [x] Move alchemy Python code to `hoomd.md.alchemy` package
- [x] Remove NVE alchemy.
- [x] Decide whether or not to support pickling
- [x] Review and revise API
- [x] Unit tests
- [x] Add unit tests for LJ gauss potential
- [x] Validate: square lattice from the original alchemy paper
- [x] Document that MPI and GPU are not supported
- [x] Document that normalized alchemy is only valid for single type systems with a single interaction
- [x] Add links to alchemy papers in sphinx docs
- [x] Document that alchemical methods cannot be pickled (but that the potentials can be)
- [x] Write/improve docstrings for sphinx
- [x] Write top-level alchemy documentation, including some theory
- [x] Revisit logging of param_dict values


hoomd/AlchemyData.h
```diff
@@ -0,0 +1,160 @@
...
+    Scalar2 mass
+        = make_scalar2(1.0, 1.0); // mass (x) and it's inverse (y) (don't have to recompute constantly)
+    Scalar mu = 0.;          //!< the alchemical potential of the particle
+    GlobalArray<Scalar> m_alchemical_derivatives; //!< Per particle alchemical forces
+    protected:
+    // the timestep the net force was computed and the netforce
+    std::pair<uint64_t, Scalar> m_timestepNetForce;
+    };
+
+class AlchemicalPairParticle : public AlchemicalMDParticle
```
j-proc, line 131: This is the only particle I'm actually using, however I'm trying to lay down some basic information that would be required for any alchemical method to potentially be used elsewhere when needed

hoomd/md/AlchemicalPotentialPair.h
```diff
@@ -0,0 +1,385 @@
...
+
+/*! \file AlchemicalPotentialPair.h
+    \brief Defines the template class for alchemical pair potentials
+*/
+
+#ifdef __HIPCC__
+#error This header cannot be compiled by nvcc
+#endif
+
+template<class evaluator> struct AlchemyPackage
```
j-proc, line 37: Holds all the alchemical information to pass throughout the force compute

hoomd/md/AlchemicalPotentialPair.h
```diff
@@ -0,0 +1,385 @@
...
+    // Extra steps to insert
+    inline AlchemyPackage<evaluator> pkgInitialize(const uint64_t& timestep) override;
+    inline void pkgPerNeighbor(const unsigned int& i,
+                               const unsigned int& j,
+                               const unsigned int& typei,
+                               const unsigned int& typej,
+                               const bool& in_rcut,
+                               evaluator& eval,
+                               AlchemyPackage<evaluator>&) override;
+    inline void pkgFinalize(AlchemyPackage<evaluator>&) override;
```
j-proc, line 163: The code to be injected at various points related to the alchemical information. Hopefully general enough it could be applied to other use cases if needed in the future. Unless it prevents a GPU implementation.

hoomd/md/AlchemicalPotentialPair.h
```diff
@@ -0,0 +1,385 @@
...
+        for (unsigned int j = 0; j < evaluator::num_alchemical_parameters; j++)
+            if (pkg.compute_mask[i][j])
+                {
+                if (m_normalized)
+                    m_alchemical_particles[j * m_alchemy_index.getNumElements() + i]->setNetForce(
+                        pkg.normalizationValues[i]);
+                else
+                    m_alchemical_particles[j * m_alchemy_index.getNumElements() + i]->setNetForce();
+                }
+    }
```
j-proc, line 362: This is mostly required for the way I implemented normalization (an important part of adding constraints for digital alchemy). It could likely be made a destructor method for the package instead.

hoomd/md/EvaluatorNormalizer.h
```diff
@@ -0,0 +1,26 @@
+// Maintainer: jproc
+#ifndef __PAIR_EVALUATOR_NORMALIZER_H__
+#define __PAIR_EVALUATOR_NORMALIZER_H__
+
+#include "hoomd/HOOMDMath.h"
+
+template<class evaluator> class Normalized : public evaluator
```
j-proc, line 7: After a lot of experimentation this was the way I went about applying a scalar normalization hopefully efficiently. It could alternatively be done by splitting into two perNeighbor calls (pre and post eval) but I decided this seemed cleaner.

hoomd/md/EvaluatorPairOPP.h
```diff
@@ -167,6 +170,30 @@ class EvaluatorPairOPP
...
+
+            alchemical_derivatives[0]
+                = -sin(params.k * alphas[0] * (r - Scalar(1.0)) + params.phi * alphas[1]) * params.k
+                  * (r - Scalar(1.0)) * r3inv;
+            alchemical_derivatives[1]
+                = -sin(params.k * alphas[0] * (r - Scalar(1.0)) + params.phi * alphas[1])
+                  * params.phi * r3inv;
+            }
+        }
+
```
j-proc, line 196: This code is remaining from an old implementation that only used k and phi as its params. It does not cover the current OPP methods and must be replaced.

hoomd/md/IntegratorTwoStep.cc
```diff
@@ -114,8 +114,11 @@ void IntegratorTwoStep::update(uint64_t timestep)
         m_prof->push("Integrate");
 
     // perform the second step of the integration on all groups
-    for (auto& method : m_methods)
+    // reversed for integrators so that the half steps will be performed symmetrically
+    // TODO: make sure this is valid for non-alchemical integrators
+    for (auto method_ptr = m_methods.rbegin(); method_ptr != m_methods.rend(); method_ptr++)
         {
+        auto method = (*method_ptr);
```
j-proc, line 121: This is probably the most concerning thing that I did. But is required for any method utilizing a Trotter factorization to separate timescales. Their halfsteps must be like bookends with the steps of the shorter timescale occurring between. This also means that our methods with this longer timescale must be first in the m_methods vector.

hoomd/md/PotentialPair.h
```diff
@@ -83,7 +83,8 @@
     now, that will be queried from the evaluator.
     \sa export_PotentialPair()
 */
-template<class evaluator> class PotentialPair : public ForceCompute
+template<class evaluator, typename extra_pkg = std::nullptr_t>
+class PotentialPair : public ForceCompute
```
j-proc, line 87: These are the minimum changes I felt were possible for the polymorphism/code injection I was attempting to achieve to make maintenance easier for such specialized pair interactions.

j-proc: A template parameter specifying the type of the information that will be created, utilized once per pair, and destroyed.

hoomd/md/pair/alch.py
```diff
@@ -0,0 +1,220 @@
...
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeconverter import SetOnce
+from hoomd.md.pair import pair
+
+# alternative access? hoomd.alch.md.pair would be nice to somehow link here
+# without making overly complicated
+# I just like alch(MD/MC) as alchemy specification aliases
+
+# TODO: remove this variable when the log names are finalized
+_api = ['value', 'mass', 'mu', 'avalue', 'amomentum', 'net_aforce', 'aforces']
```
j-proc, line 18: I'm simply leaving this here because I'm already using the code and if I need to modify analysis code in the future, this will be useful.

hoomd/md/pair/alch.py
```diff
@@ -0,0 +1,220 @@
...
+
+from hoomd.logging import log, Loggable
+from hoomd.operation import _HOOMDBaseObject
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeconverter import SetOnce
+from hoomd.md.pair import pair
+
+# alternative access? hoomd.alch.md.pair would be nice to somehow link here
+# without making overly complicated
+# I just like alch(MD/MC) as alchemy specification aliases
```
j-proc, line 15: I think we'd want to leave things in specialized locations unless something like `import hoomd.alch` was called to make easier access.

hoomd/md/pair/alch.py
```diff
@@ -0,0 +1,220 @@
...
+class LJGauss(pair.LJGauss, metaclass=_AlchemicalPairPotential):
+    _alchemical_parameters = ['epsilon', 'sigma2', 'r0']
+
+    def __init__(self,
+                 nlist,
+                 default_r_cut=None,
+                 default_r_on=0.0,
+                 mode='none'):
+        _AlchemicalMethods.__init__(self)
+        super().__init__(nlist, default_r_cut, default_r_on, mode)
```
j-proc, line 186: I attempted to keep this minimal because it will likely be repeated with little variation for any new alchemical pair potentials.

hoomd/md/pair/alch.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def __new__(cls,
+                force: _AlchemicalPairPotential,
+                name: str = '',
+                typepair: tuple = None,
+                mass: float = 1.0):
+        typepair = tuple(sorted(typepair))
+        if (typepair, name) in force.alchemical_particles:
+            return force.alchemical_particles[typepair, name]
+        return super().__new__(cls)
```
j-proc, line 48: These objects are only defined relative to the pair potential they are in relation too. They're singletons in this regard. So here's a messy way to mask them that at least works. This doesn't allow delayed instantiation if the the force doesn't exist, but it doesn't have to be attached. Perhaps would be better to hide these methods?

hoomd/md/pair/alch.py
```diff
@@ -0,0 +1,220 @@
...
+    @log(default=False, requires_run=True, category='particle')
+    def aforces(self):
+        return self._cpp_obj.forces
+
+    @log(requires_run=True)
+    def net_aforce(self):
+        return self._cpp_obj.net_force()
+
+    def _enable(self):
+        self.force._cpp_obj.enableAlchemicalPairParticle(self._cpp_obj)
```
j-proc, line 116: This should only be called when we're adding the alchemical particle to an alchemostat method. And removing it should probably disable it and set the parameter value in the dict to it's current value. That hasn't been implemented yet however.

hoomd/md/pair/alch.py
```diff
@@ -0,0 +1,220 @@
...
+
+        def __setitem__(self, key, val):
+            raise NotImplementedError
+
+        def __contains__(self, key):
+            key = self._validate_and_split_alchem(key)
+            return key in self._dict.keys()
+
+        def keys(self):
+            return self._dict.keys()
```
j-proc, line 169: My somewhat convoluted attempt to make accessing/creating the AlchMDParticles of an alchemical pair potential easy.

j-proc:
```python
#some standard system setup...

# this part I have a decent v3 API for although I'm sure it could be improved
from hoomd.md.pair.alch import LJGauss
ljg=LJGauss(cell,default_r_cut=3.0)
ljg.params[('A','A')]=dict(epsilon=1., sigma2=0.02, r0=1.8)
integrator.forces.append(ljg)
ar0=ljg.alchemical_particles[('A','A'),'r0']

# The alchemostat methods are without a v3 api and have several oddities
sim.run(0) #make sure our system is attached for the next call
anvt = hoomd.md._md.TwoStepNVTAlchemy(sim.state._cpp_sys_def,hoomd.variant.Constant(1)) #variant is alchemical kT
#this must be first in the list, see hoomd/md/IntegratorTwoStep.cc changes
integrator.methods._list.insert(0,anvt)
integrator.methods._synced_list.insert(0,anvt)

anvt.setAlchemTimeFactor(10)
anvt.setNextAlchemicalTimestep(sim.timestep)
# this is the most important functionality, each alchemical particle should only be integrated by one alchemostat and 
# adding it should enable the particle, removing should disable and set the parameter value to the current value to 
# prevent unexpected behavior
ar0._enable()
anvt.addAlchemicalParticle(ar0._cpp_obj)

# continue as normal
sim.run(100)

```

hoomd/AlchemyData.h
```diff
@@ -0,0 +1,173 @@
...
+        : AlchemicalMDParticle(exec_conf), m_type_pair_param(type_pair_param) {};
+    int3 m_type_pair_param;
+    };
+
+inline void export_AlchemicalMDParticle(pybind11::module& m)
+    {
+    pybind11::class_<AlchemicalMDParticle, std::shared_ptr<AlchemicalMDParticle>>(
+        m,
+        "AlchemicalMDParticle")
+        .def_property("mass", &AlchemicalMDParticle::getMass, &AlchemicalMDParticle::setMass)
```
danevans09, line 157: Also define properties (possibly read-only) for the `force`, `typepair`, and `name` python properties. The `typepair` and `name` getter and setter methods should convert between integers and strings

hoomd/md/AlchemicalPotentialPair.h
joaander: I will rewrite as a subclass. I would prefer minimal code duplication in a subclass over unused template parameters for the majority of PotentialPair.

At this time, there are no plans for a GPU implementation.

hoomd/md/IntegratorTwoStep.cc
joaander: In MD, the methods operate on disjoint sets of particles. The order of method application does not change the results.

hoomd/md/pair/alch.py
joaander: We will move all alchemy related potentials and methods to a new `hoomd.md.alchemy` namespace.

hoomd/md/EvaluatorPairOPP.h
joaander: OPP is not exposed in Python, not used in publications (as far as I know), and would therefore require significant effort to enable. I am deleting this code for now, future work could enable this functionality when needed.

j-proc:
Sorry I haven't been available for this. I think the main thing to mention is that I think we probably shouldn't release the NVE alchemostat. While the equations of motion are what you'd expect, the idea of NVE is violated by the changing the the potential. I was experimenting with it with regards to normalization but I absolutely do not consider it production ready without a huge amount of warnings to the user.

I can strip it out unless either of you think it's better to leave it in or find it easier to do so yourself while you're poking around.

joaander:
> While the equations of motion are what you'd expect, the idea of NVE is violated by the changing the the potential. I was experimenting with it with regards to normalization but I absolutely do not consider it production ready without a huge amount of warnings to the user.

Then we will remove it.

joaander:
`param_dict` entries are not loggable. On second review, `value` is the most useful loggable here. If users have a strong need to log `alpha`, they can write a custom entry, or we can revisit loggable param_dict entires in a future pull request.

sphinx-doc/module-md-alchemy.rst
```diff
@@ -0,0 +1,18 @@
+.. Copyright (c) 2009-2022 The Regents of the University of Michigan.
+.. Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+md.alchemy
+--------------
```
b-butler:
```suggestion
md.alchemy
----------
```

hoomd/md/EvaluatorPairLJGauss.h
```diff
@@ -0,0 +1,242 @@
...
+
+#ifndef __HIPCC__
+#include <string>
+#endif
+
+#include "hoomd/HOOMDMath.h"
+
+/*! \file EvaluatorPairLJGauss.h
+    \brief Defines the pair evaluator class for Lennard Jones Gaussian potentials
+    \details .....
```
b-butler:
```suggestion
```

hoomd/md/EvaluatorPairLJGauss.h
```diff
@@ -0,0 +1,242 @@
...
+                Scalar rcut6inv = rcut2inv * rcut2inv * rcut2inv;
+                pair_eng
+                    -= rcut6inv * (rcut6inv - Scalar(2.0))
+                       - (epsilon * fast::exp(-Scalar(1.0) / Scalar(2.0) * (rcutsq - r0) / sigma2));
+                }
+            return true;
+            }
+        else
+            return false;
+        }
```
b-butler:
```suggestion
        if (rsq >= rcutsq)
            {
            return false;
            }
        Scalar r = fast::sqrt(rsq);
        Scalar rdiff = r - r0;
        Scalar rdiff_sigma2 = rdiff / sigma2;
        Scalar exp_val = fast::exp(-Scalar(0.5) * rdiff_sigma2 * rdiff);
        Scalar r2inv = Scalar(1.0) / rsq;
        Scalar r6inv = r2inv * r2inv * r2inv;

        force_divr = (r2inv * r6inv * Scalar(12.0) * (r6inv - Scalar(1.0)))
                     - (exp_val * epsilon * rdiff_sigma2 / r);
        pair_eng = r6inv * (r6inv - Scalar(2.0)) - exp_val * epsilon;

        if (energy_shift)
            {
            Scalar rcut2inv = Scalar(1.0) / rcutsq;
            Scalar rcut6inv = rcut2inv * rcut2inv * rcut2inv;
            pair_eng
                -= rcut6inv * (rcut6inv - Scalar(2.0))
                   - (epsilon * fast::exp(-Scalar(1.0) / Scalar(2.0) * (rcutsq - r0) / sigma2));
            }
        }
```

hoomd/md/EvaluatorPairLJGauss.h
```diff
@@ -0,0 +1,242 @@
...
+
+    DEVICE Scalar evalPressureLRCIntegral()
+        {
```
b-butler, line 168: It would be helpful to document the purpose of these new functions for alchemy.

hoomd/md/EvaluatorPairLJGauss.h
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+    DEVICE Scalar evalEnergyLRCIntegral()
+        {
+        return 0;
+        }
+
+#ifndef __HIPCC__
+
+    DEVICE void alchemParams(const std::array<Scalar, num_alchemical_parameters>& alphas)
```
b-butler: It seems that `updateAlchemyParams` would make a better name.

hoomd/md/EvaluatorPairLJGauss.h
```diff
@@ -0,0 +1,242 @@
...
+        return params;
+        }
+
+    DEVICE void
+    evalAlchDerivatives(std::array<Scalar, num_alchemical_parameters>& alchemical_derivatives,
+                        const std::array<Scalar, num_alchemical_parameters>& alphas)
+        {
+            {
+            Scalar r = fast::sqrt(rsq);
+            // Scalar sigma = fast::sqrt(sigma2);
```
b-butler:
```suggestion
```

hoomd/md/EvaluatorPairLJGauss.h
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+    //! Get the name of this potential
+    /*! \returns The potential name. Must be short and all lowercase, as this is the name energies
+       will be logged as via analyze.log.
+    */
+    static std::string getName()
+        {
+        return std::string("lj_gauss");
+        }
```
b-butler:
```suggestion
    //! Get the name of this potential
    /*! \returns The potential name. Must be short and all lowercase, used for warnings messages,
        and autotuners.
    */
    static std::string getName()
        {
        return std::string("lj_gauss");
        }
```

hoomd/ParticleGroup.h
```diff
@@ -144,9 +144,6 @@ class PYBIND11_EXPORT ParticleGroup
     //! \name Initialization methods
     // @{
 
-    //! Constructs an empty particle group
-    ParticleGroup() : m_num_local_members(0) {};
```
b-butler, line 148: Is this being deleted because it is unused? If so just resolve the comment, if not why?

hoomd/md/AlchemyData.h
```diff
@@ -0,0 +1,191 @@
...
+
+    Scalar momentum = 0.; // the momentum of the particle
+    Scalar2 mass
+        = make_scalar2(1.0,
+                       1.0); // mass (x) and it's inverse (y) (don't have to recompute constantly)
+    Scalar mu = 0.;          //!< the alchemical potential of the particle
+    GlobalArray<Scalar> m_alchemical_derivatives; //!< Per particle alchemical forces
+    protected:
+    // the timestep the net force was computed and the netforce
+    std::pair<uint64_t, Scalar> m_timestep_net_force;
```
b-butler, line 132: It seems odd that the method that saves the second element of this is different from the first, especially since they are stored as a pair.

hoomd/md/AlchemyData.h
```diff
@@ -0,0 +1,191 @@
...
+        .def_readwrite("momentum", &AlchemicalMDParticle::momentum)
+        .def_property_readonly("forces", &AlchemicalMDParticle::getDAlphas)
+        .def_property_readonly("net_force",
+                               pybind11::overload_cast<>(&AlchemicalMDParticle::getNetForce))
+        .def("notifyDetach", &AlchemicalMDParticle::notifyDetach);
+
+    pybind11::class_<AlchemicalPairParticle,
+                     AlchemicalMDParticle,
+                     std::shared_ptr<AlchemicalPairParticle>>
+        nameThatShouldNeverAppear(m, "AlchemicalPairParticle");
```
b-butler: I don't understand this `nameThatShouldNeverAppear`.

hoomd/md/alchemy/_alchemical_methods.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # FIXME: breaks delayed instantiation by validating types
+        def _validate_and_split_alchem(self, key):
+            if isinstance(key, tuple) and len(key) == self._len_keys:
+                for param, typepair in zip(key, reversed(key)):
+                    if set(self.outer._alchemical_parameters).issuperset(
+                            self._validate_and_split_len_one(param)) and set(
+                                self.outer._simulation.state.particle_types
+                            ).issuperset(
+                                self._validate_and_split_len_one(typepair)):
```
b-butler:
```suggestion
                    # Check that types specified exist in the simulation.
                    if set(self.outer._alchemical_parameters).issuperset(
                            self._validate_and_split_len_one(param)) and set(
                                self.outer._simulation.state.particle_types
                            ).issuperset(
                                self._validate_and_split_len_one(typepair)):
```

hoomd/md/alchemy/_alchemical_methods.py
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) 2009-2022 The Regents of the University of Michigan.
+# Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeconverter import SetOnce
+from hoomd.operation import _HOOMDBaseObject
+
+
+class _AlchemicalMethods(_HOOMDBaseObject):
+
```
b-butler:
```suggestion

      # class's Alchemical degree of freedom type (ie. normalized or not)
    _dof_type = AlchemicalDOF
```

hoomd/md/alchemy/_alchemical_methods.py
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) 2009-2022 The Regents of the University of Michigan.
+# Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeconverter import SetOnce
+from hoomd.operation import _HOOMDBaseObject
+
+
+class _AlchemicalMethods(_HOOMDBaseObject):
```
b-butler: Some brief documentation on these classes would be helpful.

b-butler: The name is misleading as we have integrator method which this is not I think since it applies to `LJGauss`.

hoomd/md/alchemy/_alchemical_methods.py
```diff
@@ -0,0 +1,64 @@
...
+                        break
+                else:
+                    raise KeyError(
+                        key, """Not a valid combination of particle types and
+                        alchemical parameter keys.""")
+                return tuple(self._yield_keys(typepair)), tuple(
+                    self._validate_and_split_len_one(param))
+
+        def __getitem__(self, key):
+            typepair, param = self._validate_and_split_alchem(key)
```
b-butler:
```suggestion
            typepairs, params = self._validate_and_split_alchem(key)
```

hoomd/md/alchemy/pair.py
```diff
@@ -0,0 +1,333 @@
...
+            return force._alchemical_dof[typepair, name]
+        return super().__new__(cls)
+
+    def __init__(self,
+                 force: _AlchemicalPairPotential,
+                 name: str = '',
+                 typepair: tuple = None,
+                 alpha: float = 1.0,
+                 mass: float = 1.0,
+                 mu: float = 0.0):
```
b-butler: The init signature is different from new. This means you can't actually pass any of these arguments to init, and will error if you try to do so.

hoomd/md/alchemy/pair.py
```diff
@@ -0,0 +1,333 @@
...
+
+        alchemical_momentum (float): The momentum of the alchemical parameter.
+
+    """
+
+    def __new__(cls,
+                force: _AlchemicalPairPotential,
+                name: str = '',
+                typepair: tuple = None,
+                mass: float = 1.0):
```
b-butler:
```suggestion
    def __new__(cls,
                force: _AlchemicalPairPotential,
                name: str = '',
                typepair: tuple = None,
                *args,
                **kwargs):
```

hoomd/md/alchemy/pair.py
```diff
@@ -0,0 +1,333 @@
...
+        param_dict['mass'] = mass
+        param_dict['mu'] = mu
+        param_dict['alpha'] = alpha
+        param_dict['alchemical_momentum'] = 0.0
+
+        # set defaults
+        self._param_dict.update(param_dict)
+
+    def _attach(self):
+        assert self.force._attached
```
b-butler: Either raise an `RuntimeError` or remove the assert.

hoomd/md/alchemy/pair.py
```diff
@@ -0,0 +1,333 @@
...
+
+        .. math::
+            F_{\\mathrm{alchemical}} = \\sum_{i=0}^{N_{\\mathrm{paricles}}-1}
+            F_{\\mathrm{alchemical},i}
+
+        """
+        return self._cpp_obj.net_force
+
+    def _enable(self):
+        assert self._attached
```
b-butler: See other comment about asserts.

hoomd/md/alchemy/_alchemical_methods.py
```diff
@@ -0,0 +1,64 @@
...
+    class AlchemicalParticleAccess(TypeParameterDict):
+
+        def __init__(self, outer):
+            self.outer = outer
+            super().__init__(SetOnce(outer._dof_type), len_keys=2)
+
+        # FIXME: breaks delayed instantiation by validating types
+        def _validate_and_split_alchem(self, key):
+            if isinstance(key, tuple) and len(key) == self._len_keys:
+                for param, typepair in zip(key, reversed(key)):
```
b-butler: I am confused on how one of these is the param, but the other through reversing is the typepair, shouldn't they both either be types or alchemical parameters?

hoomd/md/alchemy/_alchemical_methods.py
```diff
@@ -0,0 +1,64 @@
...
+from hoomd.data.typeconverter import SetOnce
+from hoomd.operation import _HOOMDBaseObject
+
+
+class _AlchemicalMethods(_HOOMDBaseObject):
+
+    def __init__(self):
+        self._alchemical_dof = self.AlchemicalParticleAccess(self)
+
+    class AlchemicalParticleAccess(TypeParameterDict):
```
b-butler: Does the idea of multiple typeparam dicts for each degree of freedom not work here? We could overwrite `_setattr_type_paramdict` to correctly handle alchemical degrees of freedom, and then we don't need to create this class.

hoomd/md/alchemy/pair.py
```diff
@@ -0,0 +1,333 @@
...
+
+from hoomd.logging import log, Loggable
+from hoomd.operation import _HOOMDBaseObject
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.md.pair import LJGauss as BaseLJGauss
+
+from hoomd.md.alchemy._alchemical_methods import _AlchemicalMethods
+
+
+class _AlchemicalPairPotential(Loggable):
```
b-butler: This could easily be accomplished via a decorator without appealing to metaclasses and would be simpler and easier to maintain. We can just use direct inheritance for the `_AlchemicalMethod` for any classes.

hoomd/md/alchemy/methods.py
```diff
@@ -0,0 +1,126 @@
...
+    def _update(self):
+        if self._attached:
+            self._cpp_obj.alchemical_dof = self.alchemical_dof._synced_list
+
+    def _attach(self):
+        super()._attach()
+        self.alchemical_dof._sync(None, [])
+        self._update()
+
+    class _OwnedAlchemicalParticles(syncedlist.SyncedList):
```
b-butler: `SyncedList` should already error if the same item is added to another list, so there shouldn't be a need for this `_owned` API.

Also, shouldn't the `SyncedList` being synced, handle the update part?

hoomd/md/IntegratorTwoStep.cc
```diff
@@ -104,8 +104,10 @@ void IntegratorTwoStep::update(uint64_t timestep)
         }
 
     // perform the second step of the integration on all groups
-    for (auto& method : m_methods)
+    // reversed for integrators so that the half steps will be performed symmetrically
+    for (auto method_ptr = m_methods.rbegin(); method_ptr != m_methods.rend(); method_ptr++)
         {
+        auto method = (*method_ptr);
```
b-butler: Are the parenthesis necessary here?

hoomd/md/module-md.cc
```diff
@@ -212,6 +216,16 @@ PYBIND11_MODULE(_md, m)
     export_PotentialPair<PotentialPairFourier>(m, "PotentialPairFourier");
     export_PotentialPair<PotentialPairOPP>(m, "PotentialPairOPP");
     export_PotentialPair<PotentialPairTWF>(m, "PotentialPairTWF");
+    export_PotentialPair<PotentialPairLJGauss>(m, "PotentialPairLJGauss");
+    export_AlchemicalMDParticles(m);
```
b-butler: I think it makes sense to but the this next to the alchemical potentials.

hoomd/md/TwoStepNVTAlchemy.cc
```diff
@@ -0,0 +1,149 @@
...
+
+    m_exec_conf->msg->notice(10) << "TwoStepNVTAlchemy: 1st Alchemcial Half Step" << std::endl;
+
+    m_nextAlchemTimeStep += m_nTimeFactor;
+
+    m_alchem_KE = Scalar(0);
+
+    Scalar dUextdalpha = Scalar(0);
+
+    m_validState = false;
```
b-butler: Is this necessary as this is only set in the `integrateStep{One,Two}`.

hoomd/md/TwoStepNVTAlchemy.cc
```diff
@@ -0,0 +1,149 @@
...
+
+void export_TwoStepNVTAlchemy(py::module& m)
+    {
+    // export_AlchemostatTwoStep<TwoStepNVTAlchemy>(m,"AlchemostatTwoStepBase")
+    py::class_<TwoStepNVTAlchemy, AlchemostatTwoStep, std::shared_ptr<TwoStepNVTAlchemy>>(
+        m,
+        "TwoStepNVTAlchemy")
+        .def(py::init<std::shared_ptr<SystemDefinition>, unsigned int, std::shared_ptr<Variant>>())
+        .def("setT", &TwoStepNVTAlchemy::setT)
+        .def("setQ", &TwoStepNVTAlchemy::setQ)
```
b-butler: Unnecessary?

hoomd/md/PotentialPairAlchemical.h
```diff
@@ -0,0 +1,719 @@
...
+            }
+        return alpha_p;
+        }
+
+    void enableAlchemicalPairParticle(std::shared_ptr<alpha_particle_type> alpha_p)
+        {
+        m_alchemy_mask[m_alchemy_index(alpha_p->m_type_pair_param.x, alpha_p->m_type_pair_param.y)]
+                      [alpha_p->m_type_pair_param.z]
+            = true;
+        alpha_p->resizeForces(m_pdata->getN());
```
b-butler, line 82: I am not sure why this line is here.

hoomd/md/PotentialPairAlchemical.h
```diff
@@ -0,0 +1,719 @@
...
+        }
+    pkgFinalize(pkg);
+
+    computeTailCorrection();
+    }
+
+template<class evaluator, typename extra_pkg, typename alpha_particle_type>
+template<class InputIterator>
+inline void
+PotentialPairAlchemical<evaluator, extra_pkg, alpha_particle_type>::computeEnergyBetweenSets(
```
b-butler: Is there a reason `computeEnergyBetweenSets` includes the alchemical forces? Because if not we could delete this and rely on the implementation in `PotentialPair.h`.

b-butler (changes requested):
This was obviously a lot of work. I have left some comments to address or answer. Some of them are clarification questions as I didn't follow some parts; if they don't make sense feel free to say so and resolve the comment.

I am a bit concerned at the complexity this adds in Python, and I hope my comments may help reduce that. If not I would be happy to meet to discuss possible solutions if there are any.

hoomd/ParticleGroup.h
joaander: An earlier version of this branch constructed an empty particle group through this constructor. Upon debugging, I found that these default constructed particle groups are very broken (cause seg faults). Rather than fixing this constructor, I removed it to find all the places that used it and to prevent such bugs from occurring again. 

An empty ParticleGroup should be initialized with the Null particle selector.

hoomd/md/PotentialPairAlchemical.h
joaander: I don't know why the original author put that there. We can probably remove the method. I think that the new method likely isn't called as the subclass has different template arguments than the base and there is no non-templated base to define the virtual method signature.

hoomd/md/IntegratorTwoStep.cc
joaander: No.

hoomd/md/AlchemyData.h
joaander: The original code author designed the alchemical integration methods to operate with periods greater than 1. The `uint64_t` tracks the timestep the force was computed for use with an assert to ensure that the alchemical derivatives were computed on the correct timestep.

I don't understand it either as I think that `PotentialPairAlchemical` computes the alchemical forces on all timesteps...

hoomd/md/EvaluatorPairLJGauss.h
joaander: Yes.

hoomd/md/AlchemyData.h
j-proc: It has been a while... but Josh is correct about the time keeping portion and asserts because of the different timestep scales. It's pretty important to get right with the Trotter factorization because obviously if you get to your next start 1/2 timestep and didn't just wrap up your end 1/2 timestep... you've got a problem. It could perhaps be rewritten into a test rather than this tag-along variable. However I intended it to be possible to add some warnings or checks to make sure that an alchemical particle isn't disowned by an integrator before it's completed a full alchemical step.

hoomd/md/AlchemyData.h
j-proc: Me neither... but pretty sure it's included here to avoid a pybind11 error. Probably because I implemented the particles in a likely overly complicated way ... at least at the moment. I did it with the idea that they can be used in the future without a huge rewrite to enable many things that have been talked about but not implemented here.

hoomd/md/AlchemyData.h
j-proc: @joaander as to your second point. `PotentialPairAlchemical` shouldn't compute much for each timestep if the masking aspect is working properly and the Trotter factorization power is greater than 2.

hoomd/md/alchemy/_alchemical_methods.py
j-proc: It is probably a bad name. I couldn't think of a better one at the time of writing. Not method as in the integration methods but methods as in coding methods that only apply to these alchemical pair potentials. So the alchemical methods that are added to these functions. Obviously not attached to that naming, just at a loss for an idea.

hoomd/md/PotentialPairAlchemical.h
joaander: Removed.

hoomd/md/PotentialPairAlchemical.h
joaander: It appears that `AlchemicalMDParticle` has 0-length arrays on construction. This call is needed to initialize the arrays.

hoomd/md/TwoStepNVTAlchemy.cc
joaander: Removed.

hoomd/md/TwoStepNVTAlchemy.cc
joaander: I don't know what the original author intended with this. IntegratorTwoStep clearly defines the two step protocol where it calls step one, computes forces, then calls step two and only calls them once per timestep. `m_validState` appears to be guarding against the impossibility that step two is called twice on the same step. 

I removed it.

hoomd/md/EvaluatorPairLJGauss.h
joaander: Done.

hoomd/md/AlchemyData.h
joaander: > @joaander as to your second point. `PotentialPairAlchemical` shouldn't compute much for each timestep if the masking aspect is working properly and the Trotter factorization power is greater than 2.

OK, I see it now.

joaander: Resolving this. If you feel strongly that it should be written a different way, feel free to fix it yourself.

hoomd/md/AlchemyData.h
joaander: It is not needed.

hoomd/md/alchemy/_alchemical_methods.py
joaander: I renamed it to _AlchemicalPairForce.

b-butler:
I have made the Python changes I suggested.
1. Alchemostats use `_SyncedList` for alchemical particle storage and it handles the enabling and disabling of the objects directly.
2. Removed the use of metaclasses and `__new__`.
3. To get an alchemical particle, you now do something like `particles = ljg.epsilon[("A", ["A", "B"])]`.
    * This includes the removal of custom `TypeParameterDict` subclass in favor of a simple `Mapping` subclass.
    * A new `_SmartTypeIndexer` class has been abstracted out of `_ValidatedDefaultDict`.

b-butler:
I have not tested the new features extensively beyond the existing test which cover most of the desired behavior, but I don't know if I will have time to add them before we need to merge.

b-butler (approved):

hoomd/md/EvaluatorPairLJGauss.h
```diff
@@ -0,0 +1,275 @@
...
+            return v;
+            }
+#endif
+        } __attribute__((aligned(16)));
+
+    // static const std::map<unsigned int, std::string> param_order;
+
+    //! Constructs the pair potential evaluator
+    /*! \param _rsq Squared distance beteen the particles
+        \param _rcutsq Sqauared distance at which the potential goes to 0
```
tcmoore3:
```suggestion
        \param _rcutsq Squared distance at which the potential goes to 0
```

hoomd/md/alchemy/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+mechanical ensemble that includes alchemical degrees of freedom describing
+particle attributes as thermodynamic variables.
+
+Example::
+
+    nvt = hoomd.md.methods.NVT(...)
+    integrator.methods.append(nvt)
+    ljg = hoomd.md.alchemy.pair.LJGauss(...)
+    integrator.forces.append(ljg)
+    ar0 = ljg.r0[('A', 'A')]
```
tcmoore3:
```suggestion
    r0_alchemical_dof = ljg.r0[('A', 'A')]
```

hoomd/md/alchemy/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    alchemostat = hoomd.md.alchemy.methods.NVT(
+        period=period,
+        alchemical_dof=[ar0],
```
tcmoore3:
```suggestion
        alchemical_dof=[r0_alchemical_dof],
```

hoomd/md/alchemy/methods.py
```diff
@@ -0,0 +1,107 @@
...
+        return self._alchemical_particles
+
+    @alchemical_particles.setter
+    def alchemical_particles(self, new_particles):
+        self._alchemical_particles.clear()
+        self._alchemical_particles.extend(new_particles)
+
+
+class NVT(Alchemostat):
+    r"""Alchemical NVT Integration.
```
tcmoore3:
```suggestion
    r"""Alchemical NVT integration.
```

hoomd/md/alchemy/methods.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Attention:
+        `hoomd.md.alchemy.methods.NVT` does not support execution on GPUs.
+
+    Attention:
+        `hoomd.md.alchemy.methods.NVT` does not support MPI parallel
+        simulations.
+
+    Attention:
+        `hoomd.md.alchemy.methods.NVT` objects are not picklable.
```
tcmoore3, line 74: Is this still true after @b-butler's latest updates?

hoomd/md/alchemy/methods.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Note:
+        :py:class:`Alchemostat` is the base class for all alchemical integration
+        methods. Users should use the subclasses and not instantiate
+        `Alchemostat` directly.
+
+    """
+
+    def __init__(self, alchemical_particles):
+        self._alchemical_particles = syncedlist.SyncedList(
```
tcmoore3: When did we switch back to calling this `alchemical_particles`? `alchemical_dof` is more descriptive.

hoomd/md/alchemy/pair.py
```diff
@@ -0,0 +1,367 @@
...
+
+    .. py:attribute:: r0
+
+        Store/mapping of alchemical particles for alchemical degrees of freedom
+        for the potential parameter :math:`r_0`.
+
+        Type: `TypeParameter` [`tuple` [``particle_type``, ``particle_type``],
+        `AlchemicalDOF`])
+    """
+    _alchemical_dofs = ['epsilon', 'sigma2', 'r0']
```
tcmoore3: This is just `sigma` now

tcmoore3 (commented):
We are very close on this. Somewhere along the way during @b-butler's updates, we started using the `alchemical_particle` terminology again, which I still think is more confusing than `alchemical_dof`.

hoomd/md/alchemy/methods.py
tcmoore3: Yes, I believe it is still true. I will let @b-butler or @joaander confirm and mark this as resolved. Note: I did check that a `pickling_check` on an alchemostat fails locally.

hoomd/md/alchemy/methods.py
tcmoore3: I switched it back. If we have a strong reason to keep calling it `alchemical_particles`, them I'd be OK with it. Otherwise, mark as resolved.

hoomd/md/alchemy/pair.py
```diff
@@ -0,0 +1,367 @@
...
+
+        Store/mapping of alchemical particles for alchemical degrees of freedom
+        for the potential parameter :math:`\sigma`.
+
+        Type: `TypeParameter` [`tuple` [``particle_type``, ``particle_type``],
+        `AlchemicalDOF`])
+
+    .. py:attribute:: r0
+
+        Store/mapping of alchemical particles for alchemical degrees of freedom
```
tcmoore3: The phrasing "store/mapping of alchemical particles for alchemical degrees of freedom..." here bothers me, but I can't come up with a better description. I think we need to drop the "alchemical particles" syntax altogether.

hoomd/md/alchemy/pair.py
```diff
@@ -0,0 +1,367 @@
...
+    specific pair force.
+
+    `AlchemicalDOF` represents an alchemical degree of freedom to be
+    numerically integrated via an `alchemical integration method
+    <hoomd.md.alchemy.methods>`.
+
+    Note:
+        To access an alchemical particle for a particular degree of freedom
+        query the corresponding attribute in the class instance with alchemical
+        degrees of freedom.
```
tcmoore3: This note is more confusing than helpful as written. Maybe something like this would be better: 

```suggestion
    Note:
        To access an alchemical degree of freedom for a particular type pair,
        query the corresponding attribute in the alchemical pair force instance.
```

tcmoore3: @joaander feel free to accept suggestion or modify as you see fit and resolve.

tcmoore3 (approved):
Once we address my question/comments on documentation, this looks good.

hoomd/md/alchemy/methods.py
joaander: If you checked it and found the pickling check to fail, then the warning is still required.

hoomd/md/alchemy/pair.py
```diff
@@ -0,0 +1,368 @@
...
+        cls._dof_cls = AlchemicalNormalizedDOF
+    else:
+        cls.normalized = False
+        cls._dof_cls = AlchemicalDOF
+    cls._cpp_class_name = ''.join(new_cpp_name)
+    cls._accepted_modes = ('none', 'shift')
+    return cls
+
+
+class AlchemicalPairDOFStore(Mapping):
```
joaander: This class is not directly used by users. I will make it private unless there are objections.

hoomd/md/alchemy/pair.py
joaander: Indeed, I just removed "Store/mapping of alchemical particles for" from each of these.

joaander (approved):
Thanks all! I made a few final tweaks as noted in the comments. Will merge later today if there are no objections.

hoomd/md/alchemy/pair.py
tcmoore3: No objections here.

b-butler: It is used by users, every time `ljg.r0` is accessed the user uses it. We can make it private but we have made `TypeParameter` public which is its equivalent.

b-butler:
@tcmoore3 I liked the particle language because we use fictitious particles to equilibriate the system along those degrees of freedom. The DOF also contain quantities that only make sense in the particle language like `momentum`. Also, in HOOMD you don't add DOF to a method you add particles (via filters). I think any confusion would simply be due to not documenting the concepts in the proper language.

However, if you and Josh like the DOF language better I am not going to disagree beyond my above message.

joaander:
> @tcmoore3 I liked the particle language because we use fictitious particles to equilibriate the system along those degrees of freedom. The DOF also contain quantities that only make sense in the particle language like `momentum`. Also, in HOOMD you don't add DOF to a method you add particles (via filters). I think any confusion would simply be due to not documenting the concepts in the proper language.
> 
> However, if you and Josh like the DOF language better I am not going to disagree beyond my above message.

I prefer the DOF language. In HOOMD, we refer to a "particle" as a point in space that potentially has up to 6 degrees of freedom (3 spatial and 3 rotational), along with mass, charge, etc... What is the position of the "alchemical particle"? What is its orientation, moment of inertia? Can it exist in a neighbor list and interact with other particles? It cannot. It is just an additional degree of freedom added to the Hamiltonian, much like the thermostat and barostat in NVT and NPT. In the Hamiltonian formalism, generalized degrees of freedom have conjugate momenta.

When we add a particle to a method via a filter, it implicitly adds all the available DOF of that particle to the equations of motion solved by that method.

joaander:
Thanks all!

PR #1069, opened 2021-07-26
update cmake-modules to support git worktree

j-proc:
Update the previously used GetGitRevisionDescription calls in cmake using a more recent implementation supporting new features.

## Motivation and context
I've tried to use git worktrees with hoomd before and always run into an issue when running cmake that it can't describe. I finally looked into it and the change was as trivial as getting updated files from https://github.com/rpavlik/cmake-modules. 

## How has this been tested?
Compiling seems to work just as before but now can be done in worktrees as well (the .git will be a file not a directory which apparently also allows hoomd to be a submodule although I don't have a use case for that). I expect the pipeline builds will reveal any issues.


joaander (approved):
Thanks!

PR #1072, opened 2021-07-26
More explicit inlcludes

j-proc:
## Description
A very minor PR.
Adds some explicit includes in files which have caused issues for me when compiling. Any others that contributors have run into are welcome to be added here.

## Motivation and context

The <limits> standard library is not included automatically as of GCC11 https://www.gnu.org/software/gcc/gcc-11/porting_to.html#header-dep-changes. The same seems to be true of clang but I don't know at what version that changed. This portion could likely be backported into the maint branch, so I split it into a separate commit. 

The other minor include relates to the changing of param values storage for pair potentials changing to a std::vector rather than GlobalArray. If there are no other errors in the file, it doesn't seem to make a difference including it or not, but if there are errors it generates many which obfuscates the actual cause.

## How has this been tested?
It's just extra includes to help the compiler. No testing needed.

## Checklist:

### DIFF
--- a/hoomd/Analyzer.h
+++ b/hoomd/Analyzer.h
@@ -21,9 +21,9 @@
 #include "SharedSignal.h"
 #include "SystemDefinition.h"
 
+#include <limits>
 #include <memory>
 #include <typeinfo>
-#include <limits>
 
 /*! \ingroup hoomd_lib
     @{
--- a/hoomd/Analyzer.h
+++ b/hoomd/Analyzer.h
@@ -21,7 +21,6 @@
 #include "SharedSignal.h"
 #include "SystemDefinition.h"
 
-#include <limits>
 #include <memory>
 #include <typeinfo>
 
--- a/hoomd/Analyzer.h
+++ b/hoomd/Analyzer.h
@@ -23,6 +23,7 @@
 
 #include <memory>
 #include <typeinfo>
+#include <limits>
 
 /*! \ingroup hoomd_lib
     @{
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,7 @@ jobs:
     name: Build release tarball
     runs-on: ubuntu-latest
     container:
-      image: glotzerlab/ci:2021.05-ubuntu20.04
+      image: glotzerlab/ci:2021.07-ubuntu20.04
 
     steps:
       - name: Checkout
--- a/.github/workflows/templates/configurations.yml
+++ b/.github/workflows/templates/configurations.yml
@@ -4,7 +4,7 @@
 # - The build options are `mpi`, `tbb`, and `jit`.
 unit_test_configurations:
 - config: "[clang11_py39, mpi, tbb, jit]"
-- config: "[gcc10_py39]"
+- config: "[gcc11_py39]"
 - config: "[cuda11_gcc9_py38, mpi]"
 - config: "[cuda11_gcc9_py38]"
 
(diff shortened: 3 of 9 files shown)


hoomd/Analyzer.h
```diff
@@ -21,6 +21,7 @@
 #include "SharedSignal.h"
 #include "SystemDefinition.h"
 
+#include <limits>
```
joaander, line 24: Does the compile error for the missing `limits` header occur in `Analyzer.h`? I don't see where it would.

Please move the include to only those `.cc` and `.h` files that use symbols from `limits` to reduce compilation time.

joaander (changes requested):

hoomd/Analyzer.h
j-proc: Done. Sorry I looked for a root file thinking that was the right choice.

joaander (approved):
Thanks!

joaander:
The changelog entry for this one is:
```
Fixed
- Compile errors with gcc11
```

j-proc:
I think the unit test that keeps failing is just related to the default parameters in the constructors from [this test](https://github.com/glotzerlab/hoomd-blue/blob/181720f67eabf4c499b123309df8b74b89b79163/hoomd/mpcd/test/communicator_mpi_test.cc#L987).

I think the exec config instantiation can just follow the same pattern as the other tests, but I'm not sure.

@joaander The only other places where I didn't run into issues but I don't think I can guarantee won't be hit for GCC11 support would be in these 3 files.
   - hoomd/jit/jitify.hpp
   - hoomd/md/NeighborListGPUTree.cu
     - I think this is handled by the `hip_runtime.h` but I'm not positive based on some macro craziness I didn't understand
   - hoomd/RandomNumbers.h

joaander:
> I think the unit test that keeps failing is just related to the default parameters in the constructors from [this test](https://github.com/glotzerlab/hoomd-blue/blob/181720f67eabf4c499b123309df8b74b89b79163/hoomd/mpcd/test/communicator_mpi_test.cc#L987).
> I think the exec config instantiation can just follow the same pattern as the other tests, but I'm not sure.

This test isn't failing in CI. Does it fail locally for you?

> @joaander The only other places where I didn't run into issues but I don't think I can guarantee won't be hit for GCC11 support would be in these 3 files.
> 
> * hoomd/jit/jitify.hpp

This code performs transformations of code that includes `<limits>`, it shouldn't include `<limits>` itself.

> * hoomd/md/NeighborListGPUTree.cu
>   
>   * I think this is handled by the `hip_runtime.h` but I'm not positive based on some macro craziness I didn't understand

I don't know on this one. There is currently no CUDA version that supports GCC11 so I cannot test this.

> * hoomd/RandomNumbers.h

This does appear to need the include, though compilation doesn't fail. We'll leave it alone for now. CI test are passing when building HOOMD with gcc11.

j-proc:
> 
> 
> > I think the unit test that keeps failing is just related to the default parameters in the constructors from [this test](https://github.com/glotzerlab/hoomd-blue/blob/181720f67eabf4c499b123309df8b74b89b79163/hoomd/mpcd/test/communicator_mpi_test.cc#L987).
> > I think the exec config instantiation can just follow the same pattern as the other tests, but I'm not sure.
> 
> This test isn't failing in CI. Does it fail locally for you?
> 
With MPI and validation testing on yes. It actually fails to compile. Given how little info there was on the unit test fails I was thinking it was perhaps this. Though this is obviously unrelated to this PR.

> > * hoomd/RandomNumbers.h
> 
> This does appear to need the include, though compilation doesn't fail. We'll leave it alone for now. CI test are passing when building HOOMD with gcc11.

I felt the same. I tried to backtrack a bit but didn't find where it would be pulled in and yet it builds. So I think I'll consider the note is here that it will be easy to include if it causes an issue at another time.

joaander:
> > > I think the unit test that keeps failing is just related to the default parameters in the constructors from [this test](https://github.com/glotzerlab/hoomd-blue/blob/181720f67eabf4c499b123309df8b74b89b79163/hoomd/mpcd/test/communicator_mpi_test.cc#L987).
> > > I think the exec config instantiation can just follow the same pattern as the other tests, but I'm not sure.
> > 
> > 
> > This test isn't failing in CI. Does it fail locally for you?
> 
> With MPI and validation testing on yes. It actually fails to compile. Given how little info there was on the unit test fails I was thinking it was perhaps this. Though this is obviously unrelated to this PR.

The CMake option `BUILD_VALIDATON` is deprecated and will be removed. I don't test locally or on CI with `BUILD_VALIDATION=on`. That particular test is part of MPCD which has not yet been updated for HOOMD v3 so it has been unmaintained for some time. The compile error has likely been present for a while.

PR #1073, opened 2021-07-26
Fire v3

### LINKED ISSUE #588: CUDA memory errors in FIRE
## Description

FIRE energy minimzer results in CUDA illegal memory error.

## Script

```python
import hoomd
from hoomd import md
import numpy as np

hoomd.context.initialize("--mode=gpu --gpu=0 --gpu_error_checking")

system = hoomd.init.create_lattice(unitcell=hoomd.lattice.sc(a=0.9), n=40) #a=2.0 does not give error

dt = 0.0025
NeighborsListLJ = md.nlist.cell()
lj = hoomd.md.pair.lj(r_cut=2.5, nlist=NeighborsListLJ)
lj.pair_coeff.set('A', 'A', epsilon=1.0, sigma=1.0)
md.integrate.mode_standard(dt=dt)
integrator_nve = md.integrate.nve(group=hoomd.group.all())
for batch in range(100):
    #run NVE
    hoomd.run(1000)
    snap = system.take_snapshot()
    #run fire
    fire=hoomd.md.integrate.mode_minimize_fire(dt=dt, ftol=1e-5, Etol=1e-10)
    hoomd.run(100,quiet=True) #we don't need to fully minimize to get the error
    system.restore_snapshot(snap)
    md.integrate.mode_standard(dt=dt)

```

## Output

```test.py:020  |      fire=hoomd.md.integrate.mode_minimize_fire(dt=dt, ftol=1e-5, Etol=1e-10)
**ERROR**: an illegal memory access was encountered before /hoomd/md/FIREEnergyMinimizerGPU.cc:178
  File "test.py", line 21, in <module>
```

## Expected output

The script should run without producing an error.

## Configuration

collins, Titan V (gpu 0)

Versions:
- Python version: 3.8.1
- **HOOMD-blue** version: 2.9.0

## Developer

Any ideas, @jglaser ?

### LINKED ISSUE #863: Port FIRE energy minimization to the v3 API
## Description

Port FIRE to v3 and include unit tests.

## Motivation and context

While porting, fix the bugs found in #588

tommy-waltmann:
## Description

This code ports the FIRE energy minimizer to the v3 API, as well as fixes a few bugs that were present in v2.

## Motivation and context

Moving things to v3, and some bugs were reported in #588

Resolves #863
Resolves #588

## How has this been tested?

I have tests for typical changes that are needed when porting code, but I still have to write some more which is why this PR is still a draft.

## Change log

```
port FIRE energy minimizer to v3 API and fix bugs
```

## Checklist:


tommy-waltmann:
At this point, I haven't been able to reproduce the errors, but I have fixed the issues @joaander mentioned in #588 . That may or may not have fixed the problem.

Also, block sizes for the kernels are hard coded even though there is an `m_block_size` variable defined by the GPU class, not sure if we want to keep that going forward.

joaander:
> Also, block sizes for the kernels are hard coded even though there is an `m_block_size` variable defined by the GPU class, not sure if we want to keep that going forward.
It is best to use the Autotuner class to choose the block size dynamically and optimize performance.

hoomd/md/FIREEnergyMinimizerGPU.cc
```diff
@@ -74,6 +80,8 @@ void FIREEnergyMinimizerGPU::update(uint64_t timestep)
     Scalar tnorm(0.0);
     Scalar wnorm(0.0);
 
+    initializePartialSumArrays();
```
joaander, line 83: Yes, this should fix all the issues I identified in #588. However, reallocating on every timestep is expensive. Make a boolean flag that sets when this needs to be updated and only update on those steps where the flag is true. Use `resize` on each of the partial sum arrays instead of new allocations so that memory is only reallocated when needed.

tommy-waltmann: I may be wrong, but after looking through the implementation, I believe `resize` will always allocate new memory, even if the new size is the same as the old size.

joaander: You could use `GPUVector` instead of `GPUArray` or implement your own conditional to check to resize only when needed.

joaander: Also, ignore what I said about a boolean flag. I considered using signals and slots to track when reallocation is needed, but it would be much cleaner to just check:
```
if (needed size != current size)
    {
    initializePartialSumArrays();
    }
```

tommy-waltmann:
Adding @b-butler as a reviewer because I need to do some more specific type checking in the `ParameterDict` and don't know how to do that.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,191 @@
...
+                 rigid=None):
+
+        super().__init__(forces, constraints, methods, rigid)
+
+        self._param_dict.update(
+            ParameterDict(
+                dt=float(dt),
+                aniso=OnlyFrom(['true', 'false', 'auto'],
+                               preprocess=_preprocess_aniso),
+                min_steps_adapt=int,  # TODO force positive integers
```
b-butler, line 147: Use `hoomd.data.typeconverter.OnlyTypes(int, postprocessing=raise_if_nonpositive)` where `raise_if_nonpositive` does what it says. We could develop some shorthand for the positive integers and reals if desired, but for now we just have a hard coded function do the error raising.

b-butler (commented):
I left comments on the `_param_dict` entries with comments as asked.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,191 @@
...
+                finc_dt=float,
+                fdec_dt=float,
+                alpha_start=float,  # TODO prevent user from resetting value
```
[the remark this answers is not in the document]
tommy-waltmann: What I really want to avoid here is someone resetting this value in the middle of a simulation run. Maybe it would be better to make it a parameter of the constructor and then override `_setattr_param`.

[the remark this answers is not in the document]
tommy-waltmann: After thinking more, I've reconsidered. This value should be allowed to change.

hoomd/md/FIREEnergyMinimizer.h
```diff
@@ -66,18 +66,49 @@ class PYBIND11_EXPORT FIREEnergyMinimizer : public IntegratorTwoStep
...
+    //! get the fractional increase in the timestep
+    Scalar getFinc()
+        {
+        return m_finc;
+        }
+
     //! Set the fractional increase in the timestep upon a valid search direction
     void setFdec(Scalar fdec);
 
+    //! Set the fractional increase in the timestep upon a valid search direction
```
b-butler, line 88: This is incorrect.

hoomd/md/FIREEnergyMinimizer.h
```diff
@@ -66,18 +66,49 @@ class PYBIND11_EXPORT FIREEnergyMinimizer : public IntegratorTwoStep
...
+    //! bad before finding a new direction
+    unsigned int getNmin()
+        {
+        return m_nmin;
+        }
+
     //! Set the fractional increase in the timestep upon a valid search direction
     void setFinc(Scalar finc);
 
+    //! get the fractional increase in the timestep
```
b-butler, line 79:
```suggestion
    //! get the fractional increase in the timestep upon a valid search direction
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,198 @@
...
+            A minimum number of attempts before convergence criteria are
+            considered
+        aniso (bool):
+            Whether to integrate rotational degrees of freedom (bool), default
+            None (autodetect).
+
+    Added in version 2.2
+
+    .. versionadded:: 2.1
+    .. versionchanged:: 2.2
```
b-butler, line 63: We have been removing these in the conversion to version 3 since this notice is not meaningful in version 3.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,198 @@
...
+        """Returns the energy after the last iteration of the minimizer."""
+        return self._cpp_obj.getEnergy()
+
+    def has_converged(self):
+        """Test if the energy minimizer has converged.
+
+        Returns:
+            True when the minimizer has converged. Otherwise, return False.
+        """
+        return self._cpp_obj.hasConverged()
```
b-butler, line 194: Make properties.

They should be loggable too, `converged` shouldn't be by default though.

hoomd/md/pytest/test_minimize_fire.py
```diff
@@ -0,0 +1,117 @@
...
+def test_run_minimization(lattice_snapshot_factory, simulation_factory):
+    """Run a short minimization simulation."""
+    snap = lattice_snapshot_factory(a=0.9, n=5)
+    sim = simulation_factory(snap)
+
+    lj = md.pair.LJ(default_r_cut=2.5, nlist=md.nlist.Cell())
+    lj.params[('A', 'A')] = dict(sigma=1.0, epsilon=1.0)
+    nve = md.methods.NVE(hoomd.filter.All())
+
+    fire = md.minimize.FIRE(dt=0.0025)
```
b-butler, line 99:
```suggestion
    fire = md.minimize.FIRE(dt=0.0025, methods=[nve], forces=[lj])
```
Currently, the LJ force is not used in the optimization.

hoomd/md/pytest/test_minimize_fire.py
```diff
@@ -0,0 +1,117 @@
...
+    while not fire.has_converged():
+        sim.run(1)
+        steps_to_converge += 1
+
+    assert initial_energy >= fire.get_energy()
+    assert steps_to_converge >= fire.min_steps_conv
+
+    fire.reset()
+
+    # TODO maybe a pickling test
```
b-butler, line 117: Yes, please.

hoomd/test/test_gpu_array.cc
```diff
@@ -410,5 +410,33 @@ UP_TEST(GPUVector_basic_tests)
...
+    UP_ASSERT_EQUAL((unsigned int)v[3], (unsigned int)789);
+    UP_ASSERT_EQUAL((unsigned int)v[4], (unsigned int)321);
+    UP_ASSERT_EQUAL((unsigned int)v[5], (unsigned int)432);
+    UP_ASSERT_EQUAL((unsigned int)v[6], (unsigned int)543);
+    UP_ASSERT_EQUAL((unsigned int)v[7], (unsigned int)678);
+    UP_ASSERT_EQUAL((unsigned int)v[8], (unsigned int)987);
+    UP_ASSERT_EQUAL((unsigned int)v[9], (unsigned int)890);
+    UP_ASSERT(static_cast<bool>(v.getExecutionConfiguration()));
+
+    // test move assignment
```
b-butler, line 430: Does this test move assignment? This seems to just test a constructor method.

Don't you want to test `auto new_vec = std::move(old_vect)`?

b-butler (changes requested):
The PR is looking good, I have left some comments to address from a quick overview.

hoomd/test/test_gpu_array.cc
tommy-waltmann: What I believe happens here is the constructor gets called to create a temporary object, then the move assignment operator gets called with the temporary object and the `v` vector as arguments. I haven't stepped over the code to make sure, but based on what I had seen when debugging the GPUVector code earlier in this PR, that is the sequence of operations that are done.

hoomd/md/pytest/test_minimize_fire.py
tommy-waltmann: That brings up a good point: Since FIRE is only supposed to work with NVE/NPH integrators, I should add some sort of check to make sure those are the only methods in the methods list when attaching to a simulation.

hoomd/test/test_gpu_array.cc
joaander: The `v = vec` above tests assignment.

The `v = GPUVector<unsigned int>(
        5,
        2,
        std::make_shared<ExecutionConfiguration>(ExecutionConfiguration::GPU));`
test below does not appear to test move assignment as @b-butler indicated. Use 
`v = std::move(vec)` instead to the move constructor.

joaander: Checking above, move assignment and the move constructor are two different operations. I'm not sure off the top of my head how to trigger move assignment - if possible we should test that as well with a 3rd test.

hoomd/md/pytest/test_minimize_fire.py
b-butler: Look at using `hoomd.data.typeconverter.OnlyTypes([md.methods.NVE, md.methods.NPH, md.methods.NVERattle])`. Although RATTLE integrators have yet to be merged as separate classes. (This would go into a new `_methods` attribute similar to the existing one.

hoomd/test/test_gpu_array.cc
tommy-waltmann: Since `std::move` just marks its argument as an rvalue reference and doesn't choose which of move assignment/move constructor to use, I would think that something like
```
GPUVector<unsigned int> v;
v = std::move(vec);
```
would test move assignment, not move constructor, and something like
```
GPUVector<unsigned int> v = std::move(vec);
```
would test the move constructor.

joaander (commented):
The overall structure of this pull request is fine. Please request a detailed review from me when tests are passing and you have completed all of your own todo items.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,203 @@
...
+    stays the same), the velocity of the particles is set to 0, :math:`\\alpha
+    \\rightarrow \\ \\alpha_{start}` and :math:`\\delta t \\rightarrow \\delta t
+    \\cdot fdec_{\\alpha}`. Convergence is determined by both the force per
+    particle and the change in energy per particle dropping below `angmom_tol`
+    and `energy_tol`, respectively or
+
+    .. math::
+
+        \\frac{\\sum |F|}{N*\\sqrt{N_{dof}}} < \\text{`angmom_tol`} \\;\\;
+        and \\;\\ \\Delta \\frac{\\sum|E|}{N} < \\text{`energy_tol`}
```
tommy-waltmann, line 88: Does anyone know how to get the \\text{\`angmom_tol\`} and \\text{\`energy_tol\`} in these lines to render like `angmom_tol` and `energy_tol`?

joaander: `\\mathrm{angmom_tol}`

hoomd/md/minimize/fire.py
```diff
@@ -168,25 +171,33 @@ def __init__(self,
...
+    @log(requires_run=True)
+    def energy(self):
+        """float: Get the energy after the last iteration of the minimizer."""
+        return self._cpp_obj.energy
 
-        Returns:
-            True when the minimizer has converged. Otherwise, return False.
-        """
-        return self._cpp_obj.hasConverged()
+    @log(default=False, category='object')
```
b-butler, line 197:
```suggestion
    @log(default=False)
```
`bool` will errorlessly convert to a scalar 1 for `True` 0 for `False`. This will allow loggers that only support scalars to handle Boolean values as well, by treating them as numbers by default (of course any backend is free to handle them as desired).

hoomd/md/pytest/test_minimize_fire.py
```diff
@@ -100,18 +101,59 @@ def test_run_minimization(lattice_snapshot_factory, simulation_factory):
...
+    sim = simulation_factory(snap)
+
+    nve = md.methods.NVE(hoomd.filter.All())
+
+    fire = md.minimize.FIRE(dt=0.0025, methods=[nve])
+
+    pickling_check(fire)
+    sim.operations.integrator = fire
+    sim.run(0)
+    pickling_check(fire)
```
b-butler, line 130: This is done already by `conftest.operation_pickling_check`.

hoomd/md/pytest/test_minimize_fire.py
```diff
@@ -100,18 +101,59 @@ def test_run_minimization(lattice_snapshot_factory, simulation_factory):
...
+
+
+def _try_attach_to_fire(sim, method, should_error=False):
```
b-butler, line 133: Update check to work with immediate error on appending.

b-butler (changes requested):
I still have some comments to address, but the PR is almost there.

hoomd/GPUVector.h
```diff
@@ -213,14 +217,33 @@ GPUVectorBase<T, Array>::GPUVectorBase(const GPUVectorBase& from) : Array(from),
     {
     }
 
+template<class T, class Array>
+GPUVectorBase<T, Array>::GPUVectorBase(GPUVectorBase&& other) : Array(std::move(other))
+    {
+    m_size = std::move(other.m_size);
+    }
```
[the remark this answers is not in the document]
tommy-waltmann: not really, I can add it.

b-butler:
@tommy-waltmann I think there is an attribute that is causing the equality check not to view the unpickled FIRE integrator as the same. To fix this you would need to find the attribute may sure it shouldn't be there or equal (for instance if it only exists when attached or added) and add it to the `_skip_for _equality` class level variable (make sure to include the previous items in the set as well.

joaander:
It looks like the only remaining TODO here is to fix the failing pickling test. You will also need to merge in the changes from the master branch. Let's get this completed.

tcmoore3 (approved):

hoomd/md/NeighborListGPU.cc
```diff
@@ -185,8 +185,6 @@ void NeighborListGPU::filterNlist()
 //! Update the exclusion list on the GPU
 void NeighborListGPU::updateExListIdx()
     {
-    assert(!m_n_particles_changed);
-
```
joaander, line 189: Add this assert back to the code. This method assumes that the number of particles allocated has not changed. #1059 and #1116 fix the Neighbor list code so that it updates these arrays properly and runs without asserts in debug mode.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.data import syncedlist
+from hoomd.data.typeconverter import OnlyFrom, OnlyTypes
+from hoomd.logging import log
+from hoomd.md import _md
+from hoomd.md.integrate import _DynamicIntegrator, _preprocess_aniso
+
+
+def _raise_if_nonpositive(value):
```
joaander, line 22: Use one of the existing validators, such as `nonnegative_real` or `positive_real`.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class FIRE(_DynamicIntegrator):
+    """Energy Minimizer (FIRE).
+
+    Args:
+        dt (float):
+            This is the maximum step size the minimizer is permitted to use.
+            Consider the stability of the system when setting.
+            math:`[\\mathrm{time}]`
```
joaander, line 35:
```suggestion
            This is the maximum step size the minimizer is permitted to use :math:`[\\mathrm{time}]`.
            Consider the stability of the system when setting.
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        min_steps_adapt (int):
+            Number of steps energy change is negative before allowing
+            :math:`\\alpha` and :math:`\\delta t` to adapt.
+        finc_dt (float):
+            Factor to increase :math:`\\delta t` by
```
joaander, line 40:
```suggestion
            Factor to increase :math:`\\delta t` by  :math:`[\\mathrm{dimensionless}]`.
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        fdec_dt (float):
+            Factor to decrease :math:`\\delta t` by
```
joaander, line 42:
```suggestion
            Factor to decrease :math:`\\delta t` by :math:`[\\mathrm{dimensionless}].
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        alpha_start (float):
+            Initial (and maximum) :math:`\\alpha`
```
joaander, line 44:
```suggestion
            Initial (and maximum) :math:`\\alpha [\\mathrm{dimensionless}]`.
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        fdec_alpha (float):
+            Factor to decrease :math:`\\alpha t` by
```
joaander, line 46:
```suggestion
            Factor to decrease :math:`\\alpha t` by :math:`[\\mathrm{dimensionless}].
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        force_tol (float):
+            force convergence criteria
+            math:`[\\mathrm{force} / \\mathrm{mass}]`
```
joaander, line 49:
```suggestion
            Force convergence criteria
            math:`[\\mathrm{force} / \\mathrm{mass}]`.
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        angmom_tol (float):
+            angular momentum convergence criteria
+            math:`[\\mathrm{energy} * \\mathrm{time}]`
```
joaander, line 52:
```suggestion
            Angular momentum convergence criteria
            math:`[\\mathrm{energy} * \\mathrm{time}]`.
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        energy_tol (float):
+            energy convergence criteria math:`[\\mathrm{energy}]`
```
joaander, line 54:
```suggestion
            Energy convergence criteria math:`[\\mathrm{energy}]`.
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        min_steps_conv (int):
+            A minimum number of attempts before convergence criteria are
+            considered
```
joaander, line 57:
```suggestion
            considered.
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        aniso (bool):
+            Whether to integrate rotational degrees of freedom (bool), default
+            None (autodetect).
+
+    `minimize.FIRE` uses the Fast Inertial Relaxation Engine (FIRE) algorithm
+    to minimize the energy for a group of particles while keeping all other
```
joaander, line 63:
```suggestion
    to minimize the potential energy for a group of particles while keeping all other
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+    particles fixed.  This method is published in `Bitzek, et. al., PRL, 2006
+    <http://dx.doi.org/10.1103/PhysRevLett.97.170201>`_.
+
+    At each time step, :math:`\\delta t`, the algorithm uses the NVE Integrator
+    to generate a x, v, and F, and then adjusts v according to
```
joaander, line 68:
```suggestion
    At each time step, :math:`\\delta t`, the algorithm uses the supplied integration methods
    to generate a x, v, and F, and then adjusts v according to
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+    quantities.  While a current search has been lowering the energy of system
+    for more than :math:`N_{min}` steps, :math:`\\alpha` is decreased by
+    :math:`\\alpha \\rightarrow \\alpha fdec_{\\alpha}` and :math:`\\delta t`
+    is increased by :math:`\\delta t \\rightarrow max(\\delta t \\cdot
+    finc_{dt}, \\ \\delta t_{max})`. If the energy of the system increases (or
+    stays the same), the velocity of the particles is set to 0, :math:`\\alpha
+    \\rightarrow \\ \\alpha_{start}` and :math:`\\delta t \\rightarrow \\delta t
+    \\cdot fdec_{\\alpha}`. Convergence is determined by both the force per
+    particle and the change in energy per particle dropping below `angmom_tol`
+    and `energy_tol`, respectively or
```
joaander, line 84:
```suggestion
    \\cdot fdec_{\\alpha}`. The method converges when the force per
    particle is below `force_tol`, the angular momentum is below `angmom_tol`
    and the change in potential energy from one step to the next is below 
    `energy_tol`:
```
You also need to update the equation.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+
+    .. math::
+
+        \\frac{\\sum |F|}{N*\\sqrt{N_{dof}}} < \\mathrm{\\text{angmom_tol}}
+        \\;\\; and \\;\\ \\Delta \\frac{\\sum|E|}{N} <
+        \\mathrm{\\text{energy_tol}}
+
+    where N is the number of particles the minimization is acting over (i.e.
+    the group size) Either of the two criterion can be effectively turned off
+    by setting the tolerance to a large number.
```
joaander, line 94:
```suggestion
    where :math:`N_{\\mathrm{dof}}` is the number of degrees of freedom the minimization is acting over. 
    Any of the criterion can be effectively disabled
    by setting the tolerance to a large number.
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+
+    If the minimization is acted over a subset of all the particles in the
+    system, the "other" particles will be kept frozen but will still interact
+    with the particles being moved.
```
joaander, line 98:
```suggestion
    If the minimization acs on a subset of all the particles in the
    system, the other particles will be kept frozen but will still interact
    with the particles being moved.
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+def _raise_if_nonpositive(value):
+    if value <= 0:
+        raise ValueError("This value cannot be negative")
+    return value
+
+
+class FIRE(_DynamicIntegrator):
+    """Energy Minimizer (FIRE).
+
+    Args:
```
joaander, line 31: As written, this section describes the `Attributes:`, not the arguments to `__init__`. After making the corrections below, rename this to `Attributes:` and move it to the end of the docstring. Then add an `Args:` section that describes the arguments to `__init__`, including  forces, constraints, methods, and rigid. You can probably copy the documentation from md.Integrator. 

Most HOOMD operations make all simple parameters arguments to `__init__` and give them default argument values when they have reasonable defaults. Unfortunately, this means some duplication in the docstring, but it allows users to explicitly see the defaults and set the parameters in a one line instantiation.

Is there a particular reason we force users to accept only the defaults on construction here? If not, add all the parameters to `__init__`.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        minimized is in any case the energy (not the enthalpy or any other
+        quantity).
+
+    Note:
+        As a default setting, the algorithm will start with a :math:`\\delta t
+        = \\frac{1}{10} \\delta t_{max}` and attempts at least 10 search steps.
+        In practice, it was found that this prevents the simulation from making
+        too aggressive a first step, but also from quitting before having found
+        a good search direction. The minimum number of attempts can be set by
+        the user.
```
joaander, line 132: With the defaults are documented in the function signature, you can change this to read:
```suggestion
        In practice, the default parameters prevents the simulation from making
        too aggressive a first step, but also from quitting before having found
        a good search direction. Adjust the parameters as needed for your simulations.
```

hoomd/md/pytest/test_minimize_fire.py
```diff
@@ -0,0 +1,156 @@
...
+    _assert_correct_params(fire, new_params)
+
+    _set_and_check_new_params(fire)
+
+    _assert_error_if_nonpositive(fire)
+
+
+def test_run_minimization(lattice_snapshot_factory, simulation_factory):
+    """Run a short minimization simulation."""
+    snap = lattice_snapshot_factory(a=1.5, n=5)
```
joaander, line 93: The default settings for `lattice_snapshot_factory` create a simulation that will execute with 2 MPI ranks in the CI test environment. You change those parameters here, so the test fails when run with MPI: https://github.com/glotzerlab/hoomd-blue/pull/1073/checks?check_run_id=3507259035#step:7:3782

Either use the default parameters, or increase `n` until the simulation can run in MPI.

hoomd/minimize/fire.py
```diff
@@ -1,217 +0,0 @@
-# coding: utf-8
```
joaander, line 1: Remove the `hoomd/minimize` ignore lines from `setup.cfg` and `.pre-commit-config.yaml` as well.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,206 @@
...
+        energy_tol (float):
+            energy convergence criteria math:`[\\mathrm{energy}]`
+        min_steps_conv (int):
+            A minimum number of attempts before convergence criteria are
+            considered
+        aniso (bool):
+            Whether to integrate rotational degrees of freedom (bool), default
+            None (autodetect).
+
+    `minimize.FIRE` uses the Fast Inertial Relaxation Engine (FIRE) algorithm
```
joaander, line 62:
```suggestion
    `minimize.FIRE` is an `Integrator` that uses the Fast Inertial Relaxation Engine (FIRE) algorithm
```
Looking at the docs, it is not obvious that FIRE is an integrator to be used in place of the `md.Integrator`. In addition to this change, add a `Note:` block that states this to make it more obvious.

joaander (changes requested):
Thanks! Here are some suggested edits to improve the documentation and user-facing API.

joaander:
Also, fix the sphinx documentation rendering. Some of the units are not displaying properly (e.g. [mathrm{force} / mathrm{mass}]) is not formatted in math mode.

hoomd/md/minimize/fire.py
tommy-waltmann: Which part of which equation needs to be updated?

joaander: The equation just below this comment:
```
        \\frac{\\sum |F|}{N*\\sqrt{N_{dof}}} < \\mathrm{\\text{angmom_tol}}
        \\;\\; and \\;\\ \\Delta \\frac{\\sum|E|}{N} <
        \\mathrm{\\text{energy_tol}}
```
It incorrectly compares forces to `angmom_tol` and only lists 2 of the 3 conditions.

tommy-waltmann: Also, should this suggestion say "the angular momentum per particle is below `angmom_tol` and the change ..."?

tommy-waltmann:
the docs build with 1 warning, which causes the build to fail, but there is nothing in the output that can help me figure out what might have gone wrong.

b-butler:
@tommy-waltmann have you looked at the local build? That often helps me debug documentation builds.

joaander:
The warning is listed in the readthedocs output. Scroll down and to the right and you will find it.
```
/home/docs/checkouts/readthedocs.org/user_builds/hoomd-blue/checkouts/1073/hoomd/md/minimize/fire.py:docstring of hoomd.md.minimize.fire.FIRE:56: WARNING: more than one target found for 'any' cross-reference 'Integrator': could be :py:class:`hoomd.md.Integrator` or :py:class:`hoomd.md.integrate.Integrator
```
As @b-butler said, you can also see this when you build docs locally.

hoomd/md/minimize/fire.py
```diff
@@ -12,59 +12,79 @@
...
-    if value <= 0:
-        raise ValueError("This value cannot be negative")
-    return value
-
-
 class FIRE(_DynamicIntegrator):
     """Energy Minimizer (FIRE).
 
     Args:
         dt (float):
```
b-butler, line 26: Please order this in the same order as `__init__` expects.

hoomd/md/minimize/fire.py
```diff
@@ -123,12 +150,44 @@ class FIRE(_DynamicIntegrator):
...
-        = \\frac{1}{10} \\delta t_{max}` and attempts at least 10 search steps.
-        In practice, it was found that this prevents the simulation from making
+        In practice, the default parameters prevents the simulation from making
         too aggressive a first step, but also from quitting before having found
-        a good search direction. The minimum number of attempts can be set by
-        the user.
+        a good search direction. Adjust the parameters as needed for your
+        simulations.
+
+    Attributes:
```
b-butler, line 158: `methods`, `forces`, `constraints`, and `rigid` don't show on the documentation.

hoomd/md/minimize/fire.py
```diff
@@ -139,49 +198,59 @@ def __init__(self,
...
+                              "min_steps_adapt": min_steps_adapt,
+                              "finc_dt": finc_dt,
+                              "fdec_dt": fdec_dt,
+                              "alpha_start": alpha_start,
+                              "fdec_alpha": fdec_alpha,
+                              "force_tol": force_tol,
+                              "angmom_tol": angmom_tol,
+                              "energy_tol": energy_tol,
+                              "min_steps_conv": min_steps_conv
+                          }))
```
b-butler, line 240: This will bypass any check any validation.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,273 @@
...
+
+        fire = md.minimize.FIRE(dt=0.05)
+        fire.force_tol = 1e-2
+        fire.energy_tol = 1e-7
+        fire.methods.append(methods.NVE(filter.All()))
+        sim.operations.integrator = fire
+        while not(fire.converged):
+           sim.run(100)
+
+    Examples::
```
b-butler, line 131: We have two `Examples` blocks now which doesn't make much sense, both examples should be in the block.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,273 @@
...
+
+        fire = md.minimize.FIRE(dt=0.05)
+        fire.force_tol = 1e-2
+        fire.energy_tol = 1e-7
+        fire.methods.append(methods.NVE(filter.All()))
+        sim.operations.integrator = fire
+        while not(fire.converged):
+           sim.run(100)
```
b-butler, line 129:
```suggestion
        fire = md.minimize.FIRE(dt=0.05)
        fire.force_tol = 1e-2
        fire.energy_tol = 1e-7
        fire.methods.append(md.methods.NVE(hoomd.filter.All()))
        sim.operations.integrator = fire
        while not(fire.converged):
           sim.run(100)
```

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,273 @@
...
+
+    Examples::
+
+        fire = md.minimize.FIRE(dt=0.05)
+        fire.methods.append(methods.NPH(filter.All(), S=1, tauS=1,
+                                        couple='none'))
+        sim.operations.integrator = fire
+        while not(fire.converged):
+           sim.run(100)
```
b-butler, line 138:
```suggestion
        fire = md.minimize.FIRE(dt=0.05)
        fire.methods.append(md.methods.NPH(
            hoomd.filter.All(), S=1, tauS=1, couple='none'))
        sim.operations.integrator = fire
        while not(fire.converged):
           sim.run(100)
```

b-butler (changes requested):
I still have some comments to address mostly relating to documentation, but then this PR should be good.

hoomd/md/minimize/fire.py
tommy-waltmann: How do I get the aniso value to be checked upon construction, since it is an `OnlyTypes` object?

b-butler: You will need to construct the `ParameterDict` then update it with the passed in values. There is no way to validate on construction.

hoomd/md/minimize/fire.py
```diff
@@ -0,0 +1,283 @@
...
+    to
+
+    .. math::
+
+        \\vec{v} = (1-\\alpha)\\vec{v} + \\alpha \\hat{F}|\\vec{v}|
+
+    where :math:`\\alpha` and :math:`\\delta t` are dynamically adaptive
+    quantities.  While a current search has been lowering the energy of system
+    for more than :math:`N_{min}` steps, :math:`\\alpha` is decreased by
+    :math:`\\alpha \\rightarrow \\alpha fdec_{\\alpha}` and :math:`\\delta t`
```
joaander, line 93:
```suggestion
    :math:`\\alpha \\rightarrow \\alpha \\mathrm{fdec}_{\\alpha}` and :math:`\\delta t`
```
And similarly where multi-character variables are used (fdec, finc, etc...). Without mathrm, these appear as the multiplication of f, d, e, and c.

hoomd/md/pytest/test_minimize_fire.py
```diff
@@ -0,0 +1,168 @@
...
+    sim.operations.integrator = fire
+    sim.run(0)
+
+    initial_energy = fire.energy
+    steps_to_converge = 0
+    while not fire.converged:
+        sim.run(1)
+        steps_to_converge += 1
+
+    assert initial_energy >= fire.energy
```
joaander, line 124: This test ensure that FIRE runs and does something. Could you also run an offline test and post screenshots of the before/after configurations of this run on the CPU and GPU so we can visually verify that it does appear to be minimizing the energy?

joaander (changes requested):
Thanks for all your hard work, @tommy-waltmann . I have one last documentation formatting suggestion and I'd also like to see screenshots that show the method appears to be working correctly.

b-butler (approved):
Thanks @tommy-waltmann for the work. :+1:

tommy-waltmann:
Here are screenshots from a simple LJ fluid simulation:

Initial state:
image

Final state:
image

System energies:
image

tommy-waltmann:
I'd say this all looks good!

tommy-waltmann:
Here are the GPU results:

Initial State:
image

Final State:
image

Energies:
image

joaander (approved):
Thanks!

PR #1079, opened 2021-07-28
Improve snapshot api and document State

### LINKED ISSUE #819: Document and improve Snapshot API
## Description

I had a hard time figuring out how to calculate the number density of a HOOMD v3 simulation running in MPI. `sim.state.snapshot.particles.N` and `sim.state.snapshot.configuration.box` are only set on the root rank, but the snapshot has to be created on all ranks (it requires collective access). It would not be possible to use the computed density for further simulation changes on any rank but the root. Luckily there are state properties like `sim.state.N_particles` and `sim.state.box` that can be used, but this wouldn't work if the properties of interest were arrays like particle positions.

## Proposed solution

@joaander @b-butler I read through [the hoomd.State docstring](https://hoomd-blue.readthedocs.io/en/v3.0.0-beta.1/package-hoomd.html#hoomd.State) and the [snapshot description](https://hoomd-blue.readthedocs.io/en/v3.0.0-beta.1/package-hoomd.html#hoomd.State.snapshot) again after our conversation today. It describes the same things you told me today, but I didn't understand it when trying to work through it on my own. Maybe there are ways to make this clearer -- I would have benefitted most from a code snippet demonstrating the access pattern, as I suggest below.

We could document that MPI snapshot access patterns should look like this (if the information isn't needed on all ranks):
```python
snap = sim.state.snapshot  # Uses all ranks
if snap.exists:  # Only triggers on root rank
    print(snap.particles.N)
```

We should also provide a simple mpi4py example for the case where the information _is_ needed on all ranks:
```python
snap = sim.state.snapshot  # Uses all ranks
if snap.exists:  # Only triggers on root rank
    first_position = snap.particles.position[0]
# Do some mpi4py broadcasting here so all ranks have the data from first_position
```

Also for what it's worth, this code snippet deadlocks because snapshot gathering is collective (this is expected and sort-of documented).

```python
if sim.device.communicator.rank == 0:
    snap = sim.state.snapshot
```

## Additional context

@joaander suggested we might be able to provide a simple wrapper for broadcasting to all ranks, but that might be just as complicated to use as `mpi4py`, so it's not clear whether it would result in a net improvement for usability / user intuition.

joaander:
## Description

Implement the snapshot API improvements discussed in #819 . Also apply the same changes to the `State.box` property and add detailed documentation to `State` to explain all of the data it stores.

## Motivation and context

See the extensive discussion in #819.

Resolves #819

## How has this been tested?

The deprecated APIS are still present and use the new APIs.

## Change log

```
TODO: write this
```

## Checklist:


hoomd/state.py
```diff
@@ -38,33 +39,166 @@ def _create_domain_decomposition(device, box):
...
+      :math:`p_\\mathrm{tag,i} = i`. When accessing data in local snapshots,
+      particles may be in any order.
 
     Note:
-        This object should never be directly instantiated by users. There is no
-        way to set a state created outside of a `hoomd.Simulation` object to a
-        simulation. Use `hoomd.Simulation.create_state_from_gsd` and
-        `hoomd.Simulation.create_state_from_snapshot` to instantiate a
-        `State` object.
+        HOOMD stores angular momentum as a quaterion because that is the form
```
bdice, line 107:
```suggestion
        HOOMD stores angular momentum as a quaternion because that is the form
```

hoomd/state.py
```diff
@@ -38,33 +39,166 @@ def _create_domain_decomposition(device, box):
...
+            \\vec{S} = \\frac{1}{2}im(\\mathbf{q}^* \\times \\mathbf{P_S})
+
+    .. rubric:: Bonds
+
+    The state contains `N_bonds` bonds, `N_angles` angles, `N_dihedrals`
+    dihedrals, `N_impropers` impropers, and `N_special_pairs` special pairs.
+    Each of these data structures is similar, differing in the number of
+    particles in the group and what operations use them. Bonds, angles,
+    dihedrals, and impropers contain 2, 3, 4, and 4 particles per group
+    respectively. Bonds specify the toplogy used when computing energies and
```
bdice, line 132:
```suggestion
    respectively. Bonds specify the topology used when computing energies and
```

hoomd/state.py
```diff
@@ -38,33 +39,166 @@ def _create_domain_decomposition(device, box):
...
+
+    When running in serial or on 1 MPI rank, the entire simulation state is
+    stored in that process. When using more than 1 MPI rank, HOOMD-blue employs
+    a domain decomposition approach to split the simulation box an integer
+    number of times in the x, y, and z directions. Each MPI rank stores and
+    operates on the particles local to that rank, those contained within the
+    region defined by the split planes. Each MPI rank communicates with its
+    neighbors to obtain the properties of particles (ghost particles) near the
+    boundary between ranks so that it can compute interactions across the
+    boundary.
```
bdice, line 178: I moved the name "ghost particles" so it's clearer that it corresponds to the phrase "particles near the boundary between ranks."
```suggestion
    neighbors to obtain the properties of particles near the boundary between
    ranks (ghost particles) so that it can compute interactions across the
    boundary.
```

bdice (approved):
I think these access patterns make a lot of sense, and it seems like a well-reasoned improvement over the current design. I have only minor comments. (I did not verify that the Sphinx docs are rendering correctly.)

hoomd/state.py
joaander: Good catch. cSpell apparently identifies what I typed as a compound word quater-ion.

hoomd/snapshot.py
```diff
@@ -302,19 +301,30 @@ def _from_cpp_snapshot(cls, snapshot, communicator):
...
+
+        Returns:
+            ``self``
         """
         self._cpp_obj.replicate(nx, ny, nz)
+        return self
 
     def wrap(self):
-        """Wrap particles into the simulation box."""
+        """Wrap particles into the simulation box.
```
b-butler, line 321:
```suggestion
        """Wrap particles into the snapshot box.
```

hoomd/state.py
```diff
@@ -38,33 +39,166 @@ def _create_domain_decomposition(device, box):
...
+    bonding potentials used in molecular dynamics force fields. Like bonds,
+    special pairs define connections between two particles, but special pairs
+    are intended to adjust the 1-4 pairwise interactions in some molecular
+    dynamics force fields: see `md.special_pair`. Each bonded group is defined
+    by a type id, the group members, and a tag.
+
+    - :math:`t_\\mathrm{id}`: type id :math:`[\\mathrm{dimensionless}]` -
+      integer in the range ``[0,len(bond_types)``) that identifies the
+      bond's type. `bond_types` maps type ids to names with:
+      ``name = particle_types[t_id]``. Similarly, `angle_types` lists the
```
b-butler, line 144:
```suggestion
      ``name = bond_types[t_id]``. Similarly, `angle_types` lists the
```

hoomd/state.py
```diff
@@ -81,87 +215,119 @@ def __init__(self, simulation, snapshot):
...
+        Returns:
+            hoomd.Snapshot: The current simulation state
         """
         cpp_snapshot = self._cpp_sys_def.takeSnapshot_double()
         return Snapshot._from_cpp_snapshot(cpp_snapshot,
                                            self._simulation.device.communicator)
 
-    @snapshot.setter
-    def snapshot(self, snapshot):
+    def restore_snapshot(self, snapshot):
```
b-butler, line 263: I know `restore_snapshot` is the name used in v2, so there is benefit in using this for v3. However, I don't know if it is the best name. The naming seems to imply the primary use case for the method is to restore the state from a previous snapshot which if that is what we are good for is fine, but if this is meant to be a general purpose setting `set_snapshot`, `apply_snapshot`, or something similar might be a better name.

If you think that continuity here makes sense over any other reasons, then I am fine as is. The name isn't bad; I just wanted to mention this to discuss.

hoomd/state.py
```diff
@@ -38,33 +39,166 @@ def _create_domain_decomposition(device, box):
...
+      coordinates defining the position of the particle in the box
+    - :math:`\\mathbf{q}`: orientation :math:`[\\mathrm{dimensionless}]` - (s,
+      :math:`\\vec{a}`), unit quaternion defining the rotation from the
+      particle's local reference frame to the box reference frame.
+    - :math:`t_\\mathrm{id}`: type id :math:`[\\mathrm{dimensionless}]` -
+      integer in the range ``[0,len(particle_types)``) that identifies the
+      particle's type. `particle_types` maps type ids to names with:
+      ``name = particle_types[t_id]``.
+    - :math:`b_\\mathrm{id}`: body id :math:`[\\mathrm{dimensionless}]` -
+      integer that identifies the particle's body. A value of ``-1`` indicates
```
b-butler, line 76:
```suggestion
      integer that identifies the particle's rigid body. A value of ``-1`` indicates
```
I think using rigid as an adjective the first time helps users understand what is meant by body.

b-butler (changes requested):
The documentation is awesome, this will be a big improvement from the current stub documentation and help answer a lot of user questions (assuming they read the documentation). I have a few minor comments, but nothing big.

hoomd/state.py
joaander: Do you think it is OK to have a `set_snapshot` that can only set snapshots with identical particle/bond/etc.. types? I don't have a strong preference. If we choose `set_snapshot` then the corresponding operation should be `get_snapshot`.

bdice: I think the limitations on behavior are fine, regardless of the naming of `restore/take` or `set/get`. It would be helpful to add a link to the `State.create_state_from_snapshot` method in these setter/getter methods.

b-butler (approved):
LGTM :+1:

PR #1081, opened 2021-07-29
Add logger flags to write.Table

### LINKED ISSUE #1048: Table writer throws a confusing error when logging pressure
## Description

`hoomd.write.Table` doesn't set `pDataFlags[pdata_flag::pressure_tensor]` to `1` when logging `ThermodynamicQuantities.pressure` like `hoomd.write.GSD` does.

## Output

```python
  File "SIBCI.py", line 417, in <module>
    run(100)
  File "SIBCI.py", line 403, in run
    sim.run(s, write_at_start=was)
  File "/home/ssavitz/miniconda3/lib/python3.8/site-packages/hoomd/simulation.py", line 393, in run
  File "/home/ssavitz/miniconda3/lib/python3.8/site-packages/hoomd/write/table.py", line 285, in act
    self._write_row(output_dict)
  File "/home/ssavitz/miniconda3/lib/python3.8/site-packages/hoomd/write/table.py", line 264, in _write_row
    self.delimiter.join(
  File "/home/ssavitz/miniconda3/lib/python3.8/site-packages/hoomd/write/table.py", line 265, in <genexpr>
    (self._fmt(data[k], headers[k]) for k in headers)))
  File "/home/ssavitz/miniconda3/lib/python3.8/site-packages/hoomd/write/table.py", line 97, in __call__
    return self.format_num(value, column_width)
  File "/home/ssavitz/miniconda3/lib/python3.8/site-packages/hoomd/write/table.py", line 112, in format_num
    min_len_repr = int(log10(max(abs(value), 1))) + 1
ValueError: cannot convert float NaN to integer
```

If a `hoomd.write.GSD` happens to output the pressure on that frame, or `hoomd.Simulation.always_compute_pressure` is set, then it can work.

## Passible resolutions

- `hoomd.write.Table` should include logic like `hoomd.write.GSD`. (I'd add it myself, if I could easily find the segment of code where this is handled in the `GSD` module.)  This is the most reasonable solution in my opinion.
- Some `string` value like `"nan"` should be printed in the table.
- A warning should be printed when creating a `hoomd.write.Table` with a `hoomd.logging.Logger` which includes `ThermodynamicQuantities.pressure`.

## My workaround

My solution was

1. To write a simple function:
```python
def pressure():
    p = thermo.pressure

    if isnan(p):
        return "nan"
    else:
        return str(p)
```
2. Add it to the Table Logger with:
```python
tableLogger[ ("pressure") ] = (pressure, "string")
```
3. And add an additional `GSD` writer which just saves the pressure with a period that is quadruple that of table and a filename of `"/tmp/pressure.gsd"`.

Thus, in the table, I get three `nan`s per valid scalar, which I figure is good for efficiency.

### A further extension

I also added six analogous functions `Pxx`, `Pyy`, . . ., `Pyz`, which do a similar thing for `ThermodynamicQuantities.pressure_tensor`, which would otherwise be in the `hoomd.logging.Logger` category `sequence`, and hence impossible to logged in a table.  These should perhaps be combined into the `strings` category, but then `hoomd.write.Table` would have to support `strings`.

## Configuration

Versions:
- **HOOMD-blue** `v3.0.0-beta.7`
  -  (The rest is probably irrelevant)
- Python version: 3.8.5

Platform:
- Linux

Installation method:
- Compiled from source in a `Conda` `venv`

joaander:
This allows write.Table to display pressure values. See #1048
for a complete description.

## Description

Add pdata flags to `write.Table`

## Motivation and context

This allows `write.Table` to display pressure values.

Resolves #1048 

## How has this been tested?

This is tested with the added unit test.

## Change log

```
Fixed:

- ``write.Table`` can now display MD pressures.
```

## Checklist:


hoomd/md/pytest/test_table_pressure.py
```diff
@@ -0,0 +1,25 @@
...
+        hoomd.md.methods.NVT(hoomd.filter.All(), tau=1, kT=1))
+    sim.operations.integrator = integrator
+
+    logger = hoomd.logging.Logger(categories=['scalar'])
+    logger.add(thermo)
+    output = io.StringIO("")
+    table_writer = hoomd.write.Table(1, logger, output)
+    sim.operations.writers.append(table_writer)
+
+    sim.run(1)
```
b-butler, line 25: We should check for pressure in the `output`'s first line.

b-butler (changes requested):
Overall, looks good. I have one suggestion on the test.

Also, while not necessary to do now. It may be worth considering making the flags for logger backends dynamic (or providing a way to do so) to increase performance when pressure is not required, although I don't know how much return on investment that would be.

hoomd/md/pytest/test_table_pressure.py
joaander: Done.

joaander:
I have doubts on whether dynamic determination of flags is possible in the general sense. Given an arbitrary callable function that is a loggable quantity, how could you determine which flags need to be set for it?

lyrivera (approved):
This looks good to me.

b-butler:
@joaander I agree the general case is probably not easily feasible, but the specific case of our provided loggable quantities it should be fairly simple. However, we would need an option to set all flags for the logger backends, and ultimately this was just a thought, not something that needs to be done now, just if we find the logging prohibitively slow with the flags set at some point.

b-butler (approved):
LGTM

PR #1082, opened 2021-07-29
Remove Doxygen documentation

### LINKED ISSUE #211: clean up doxygen docs
**[Original report](https://bitbucket.org/glotzer/hoomd-blue/issue/211) by me.**

The doxygen docs for hoomd have not been tested in a while and need some work to get back into a usable condition.

joaander:
## Description

Remove the doxygen documentation configuration and directory.

## Motivation and context

#211 has been around for 5 years. To my knowledge, very few developers have compiled the doxygen documentation since it was posted. I don't think it is worth the effort to maintain doxygen documentation configuration files and ensure that all hoomd docs build and render properly with them.

I still recommend Javadoc style documentation in `style.rst`, not for compatibility with doxygen, but because developers most often read this documentation in their source editor and many source editors provide rich features for Javadoc style comments (e.g. displaying the documentation for an argument as you type a function call argument list).

Resolves #211

## How has this been tested?

No tests needed.

## Change log

This is an internal developer focused change, no change log entry is needed.

## Checklist:


joaander:
I considered moving `sphinx-doc` to `doc`. I decided not to because `sphinx-doc` has a git submodule in it, and git submodule removals are very disruptive to developers.

j-proc (approved):
I think I'd agree that actually building doxygen is largely unnecessary as devs would typically prefer to read it in context. I also think HOOMD is probably too large to produce any truly insightful diagrams that can be produced via doxygen. So I fully support this change. 

> I considered moving `sphinx-doc` to `doc`. I decided not to because `sphinx-doc` has a git submodule in it, and git submodule removals are very disruptive to developers.

I don't see this as an issue. I somewhat prefer the `sphinx-doc` name. More explicit is generally better for new devs so they don't need to keep checking project documentation for where things would be located. I at least use tab completion so there's really no benefit to having such short names.

joaander:
Thanks for the review!

PR #1085, opened 2021-07-29
Unify msg_file

### LINKED ISSUE #1005: Combine msg_file and shared_msg_file Device parameters
## Description

The Device class has separate options for msg_file (one file per rank), and shared_msg_file (one file shared among all ranks on the communicator - but not across partitions).

I would think that the default option users would want would be the shared message file. For example, if they gave all ranks the same `msg_file`, the output would be garbled as many ranks may write to the same file.

## Proposed solution

Nominally, we should remove `shared_msg_file` and making that the behavior of `msg_file` in MPI enable builds.

Except that we should remove the automatic extension of the file name with `.{{partition}`` - that is something that users can set in their Python script if they wish. We should document that the `msg_file` names should be different on different partitions, but leave it up to the user to choose the name (e.g. each partition's output goes into a separate signac job directory).

## Additional context

However, using the shared message file only solution prevents the current default use-case where `msg_file` is one file per rank. While not the default case, this can be useful by developers for debugging.

I still think this is the right choice. Developers needing to separate output by rank can either use the collective notice messages or otherwise redirect stdout to separate files by rank in their launch. The only other option I see is to detect when the user gives the same or different file names and open shared or non-shared files accordingly - though this is somewhat non-intuitive.

## Developer

I can make the change, though this would be a good first issue for someone wanting to get started with C++ and MPI development in HOOMD.

joaander:
## Description

Remove `shared_msg_file` and make `msg_file` always use the `shared_msg_file` behavior.

## Motivation and context

There is no need to have two options that both accomplish essentially the same effect. There is only one rare use case detailed in #1005 that this removes, but there are workarounds.

The ``nrank`` change is somewhat disconnected from this pull request, but I made it here because I realized the confusing nature of the name ``nrank`` when I wrote the example for ``msg_file`` and partitions.

Resolves #1005 

## How has this been tested?

Unit tests pass. I also ran a local simulation using partitions and ensured that the new example code works as intended.

## Change log

```
Removed:

- [breaking] ``shared_msg_file`` option to ``Device``. ``msg_file`` now has the same behavior as ``shared_msg_file``

Changed:

- [breaking] Rename ``nrank`` to ``ranks_per_partition`` in ``Communicator``.
```

## Checklist:

### DIFF
--- a/hoomd/Messenger.cc
+++ b/hoomd/Messenger.cc
@@ -83,10 +83,6 @@ Messenger::Messenger(std::shared_ptr<MPIConfiguration> mpi_config) : m_mpi_confi
     assert(m_mpi_config);
     if (m_mpi_config->getRank() != 0)
         m_notice_level = 0;
-
-#ifdef ENABLE_MPI
-    m_shared_filename = "";
-#endif
     }
 
 Messenger::Messenger(const Messenger& msg)
@@ -105,10 +101,6 @@ Messenger::Messenger(const Messenger& msg)
     m_notice_level = msg.m_notice_level;
 
     m_mpi_config = msg.m_mpi_config;
-
-#ifdef ENABLE_MPI
-    m_shared_filename = msg.m_shared_filename;
-#endif
     }
 
 Messenger& Messenger::operator=(Messenger& msg)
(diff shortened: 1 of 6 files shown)


b-butler (approved):
LGTM

PR #1087, opened 2021-08-02
Type Parameter Refactoring

### LINKED ISSUE #1068: TypeParameter multiple key setting shortcut fails after attaching
## Description

The shortcut syntax to set the parameters for multiple types throws an exception when used after attaching.

## Script

```python
import hoomd

cpu = hoomd.device.CPU()
sim = hoomd.Simulation(device=cpu, seed=1)
sim.create_state_from_gsd(filename='init.gsd')

cell = hoomd.md.nlist.Cell()
lj = hoomd.md.pair.LJ(nlist=cell)

particle_types = sim.state.particle_types
lj.params[(particle_types, particle_types)] = dict(epsilon=1, sigma=1)
lj.r_cut[(particle_types, particle_types)] = 2.5
integrator.forces.append(lj)
nvt = hoomd.md.methods.NVT(kT=1.5, filter=hoomd.filter.All(), tau=1.0)
integrator.methods.append(nvt)

sim.run(0)

lj.params[(particle_types, particle_types)] = dict(epsilon=1, sigma=2)
lj.r_cut[(particle_types, particle_types)] = 5
```
Note: `init.gsd` has particle types `['A', 'B']`

## Output

```
  File "/Users/joaander/test/charge-pppm/test-params.py", line 22, in <module>
    lj.params[(particle_types, particle_types)] = dict(epsilon=1, sigma=2)
  File "/Users/joaander/build/hoomd/hoomd/data/typeparam.py", line 35, in __setitem__
    self.param_dict[key] = value
  File "/Users/joaander/build/hoomd/hoomd/data/parameterdicts.py", line 266, in __setitem__
    for key in keys:
  File "/Users/joaander/build/hoomd/hoomd/data/parameterdicts.py", line 274, in _yield_keys
    raise KeyError("Type {} does not exist in the "
KeyError: "Type ('A', 'B') does not exist in the system."
```

## Expected output

No error.

## Configuration

Platform:
- Mac
- CPU

Installation method:
- Compiled from source

Versions:
- Python version: 3.9.6
- **HOOMD-blue** version: `master` branch

## Developer

@b-butler can you look into this?

b-butler:
## Description

This PR fixes #1068 and does some maintenance improvements to the type parameter implementation and documentation. While this does improve both documentation and implementation, this is not intended to be the final refactoring support for in-place mutation of nested data structures is still necessary, and better rearrangement of classes in `hoomd/data` should be done.

In addition, developer documentation on this infrastructure via `ARCITECTURE.md` is still missing.

## Motivation and context

This fixes #1068 and addresses some other needs for the type parameter API/internals.

## How has this been tested?

Existing tests pass.

## Change log

```
Improved documentation for ``hoomd.data.typeparam.TypeParameter``.
Fixed setting type pairs with ``TypeParameter`` instances after calling ``hoomd.Simulation.run`` (#1068).
```

## Checklist:


hoomd/data/typeparam.py
```diff
@@ -4,12 +4,87 @@
...
-    """Store parameters per type."""
-    __slots__ = ('name', 'type_kind', 'param_dict')
+class TypeParameter(MutableMapping):
+    """Implement a type based mutable mapping.
+
+    *Implements the `collections.abc.MutableMapping` interface (``__delitem__``
+    is disallowed).*
+
+    `TypeParameter` instances extend the base Python mapping interface with with
+    smart defaults, value/key validation/processing, and advanced indexing.
```
joaander, line 18: Explain that the class store parameters indexed by type name or pairs of type names.

hoomd/data/typeparam.py
```diff
@@ -4,12 +4,87 @@
...
+        # setting
+        lj.params[("A", "B")] = {"sigma": 1.0}
+        print(lj.params[("A", "B")])
+        # {"epsilon": 4.0, "sigma": 1.0}
+
+    Note:
+        Setting values for fictitious but valid types will not trigger an error
+        before calling `hoomd.Simulation.run`, but attempts to access that data
+        or set values for fictitious types afterwards will result in
+        ``KeyError`` exceptions.
```
joaander, line 76: I would reword this in a more positive way. Something covering these points.
* Before calling `run`, `TypeParameterDict` can map parameters for any number of arbitrarily named types.
* After calling run, it contains only those types that are present in the simulation `State`.
* Previously set names that are not in `State` are lost.
* It is an error to set names not present in `State` after calling run.

hoomd/data/typeparam.py
```diff
@@ -4,12 +4,87 @@
...
+
+    Warning:
+        For nested data structures, editing the internal stuctures such as a
```
joaander, line 79: Start off with something positive like "Values in this mapping are returned by copy, not reference. To modify nested data structures, such as the union shape intergrators in HPMC and the table pair potential in MD, read the current value, modify it, then write a new parameter (example)."

Also, don't users "anti-pattern". Most users won't know what that means.

hoomd/data/typeparam.py
```diff
@@ -4,12 +4,87 @@
...
 from hoomd.data.parameterdicts import AttachedTypeParameterDict
 
 
-class TypeParameter:
-    """Store parameters per type."""
-    __slots__ = ('name', 'type_kind', 'param_dict')
+class TypeParameter(MutableMapping):
+    """Implement a type based mutable mapping.
+
+    *Implements the `collections.abc.MutableMapping` interface (``__delitem__``
```
joaander, line 14: This doesn't appear to render correctly in the docs.

hoomd/data/typeparam.py
```diff
@@ -4,12 +4,87 @@
...
+    is disallowed).*
+
+    `TypeParameter` instances extend the base Python mapping interface with with
+    smart defaults, value/key validation/processing, and advanced indexing.
+
+    .. rubric:: Indexing
+
+    For getting and setting values, multiple index formats are supported. The
```
joaander, line 22: Explain basic indexing before advanced indexing.

hoomd/data/typeparam.py
```diff
@@ -34,6 +109,40 @@ def __setitem__(self, key, value):
...
+        """__delitem__ is not available for `TypeParameter` objects."""
+        raise NotImplementedError("__delitem__ is not defined for this type.")
+
+    def get(self, key, value):
+        """Get values for keys with undefined keys returning default.
+
+        Args:
+            key:
+                Valid keys specifications (depends on the expected key length).
+            default (``any``, optional):
```
joaander, line 122: `default` does not appear in the argument list for this method.

hoomd/data/typeparam.py
```diff
@@ -34,6 +109,40 @@ def __setitem__(self, key, value):
...
+                The value to default to if a key is not found in the mapping.
+                If not set, the value defaults to the mapping's default.
+
+        Returns:
+            values:
+                Returns a dict of the values for the keys asked for if multiple
```
joaander, line 128: This can handle a list of keys? 

Or is this referring to ('A', 'B') "keys"?

joaander (changes requested):
I have some suggestions to improve the documentation.

hoomd/data/typeparam.py
b-butler: All allowed methods can handle the full indexing documented above in a mostly intuitive way (still no deleting of items allowed).

joaander (commented):
Thanks, this is looking better. It needs to pass tests on CI before approval.

b-butler:
@joaander, yes they do. I added a type to the dummy state, and then checked the tests, I modified, and it ended up causing errors in other tests I didn't run :sweat_smile:. I ran all the tests and it seems to work now :crossed_fingers:

joaander (approved):
Thanks!

PR #1089, opened 2021-08-03
Support cuda 11 5

joaander:
## Description

Add support for CUDA 11.5. Patch by Peng Wang, NVIDIA.

## Motivation and context

CUDA 11.4 and 11.5 may cause `__global__ function call is not configure` errors when running HOOMD. CUDA 11.5 adds a workaround and this patch uses it. The issue is that thrust generates the same weakly linked symbols in multiple .so files and the linker is free to use whichever symbol it finds first. When the linker choose a symbol from a different .so file, the CUDA context is invalid, leading to this error.

## How has this been tested?

Peng Wang tested these changes.

## Change log

This PR updates the change log.

## Checklist:


mphoward:
@joaander I noticed, while debugging something else, that these changes are only in `v2.9.7` and not on the `trunk-*` branches. Was this on purpose?

joaander:
Yes, this is on purpose. 1) Peng Wang was not able to reproduce the same failure in the v3.0 beta releases. I haven't seen them either. 2) He suggested that the thrust developers had an alternate fix in mind for CUDA 11.6, making these changes unnecessary for CUDA 11.6+.

PR #1095, opened 2021-08-04
Make array documentation consistent in docs

### LINKED ISSUE #638: Standardize numpy array type documentation
We should standardize documentation of function parameters, class constructor parameters, and properties that accept or return numpy arrays. This stems from discussion on #618. Our current consensus is to adopt a modified version of the style currently used in [freud](https://freud.readthedocs.io/en/latest), which reads like `(N_{particles}, 3) numpy.ndarray`. The specific format we settled on is

```
(SHAPE) numpy.ndarray [of TYPE][, optional]
```
(where square brackets indicate optional components).

Ideally, we should also use this across other [glotzerlab](https://github.com/glotzerlab/) packages as well, but that's out of scope for here.

joaander:
## Description

Make sphinx documentation of array quantities consistent

* lists are documented at list[type]
* numpy arrays as discussed in #638

## Motivation and context

The documentation should be consistent.

Resolves #638

## How has this been tested?

I rendered the sphinx docs locally and checked that each modification renders correctly.

## Change log

```
Improved documentation.
```

## Checklist:

### DIFF
--- a/hoomd/box.py
+++ b/hoomd/box.py
@@ -247,7 +247,7 @@ class Box:
     # Length based properties
     @property
     def L(self):  # noqa: N802 - allow function name
-        """(3) `numpy.ndarray` of `float`: The box lengths, ``[Lx, Ly, Lz]`` \
+        """(3, ) `numpy.ndarray` of `float`: The box lengths, ``[Lx, Ly, Lz]`` \
         :math:`[\\mathrm{length}]`.
 
         Can be set with a float which sets all lengths, or a length 3 vector.
@@ -300,7 +300,7 @@ class Box:
     # Box tilt based properties
     @property
     def tilts(self):
-        """(3) `numpy.ndarray` of `float`: The box tilts, ``[xy, xz, yz]``.
+        """(3, ) `numpy.ndarray` of `float`: The box tilts, ``[xy, xz, yz]``.
 
         Can be set using one tilt for all axes or three tilts. If the box is 2D
         ``xz`` and ``yz`` will automatically be set to zero.
@@ -348,7 +348,7 @@ class Box:
     # Misc. properties
     @property
     def periodic(self):
-        """(3) `numpy.ndarray` of `bool`: The periodicity of each dimension."""
+        """(3, ) `numpy.ndarray` of `bool`: The periodicity of each dimension."""
         return _vec3_to_array(self._cpp_obj.getPeriodic(), bool)
 
     @property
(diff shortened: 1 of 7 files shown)


b-butler (commented):
This change is needed, but the only concern I have is constency when documenting the types of the array elements. I am the one that promoted `float` over `numpy.float64` since these are effectively the same type `np.ndarray.astype` treats `float` as a double precision floating point number. However, we document the size of signed/unsigned integers explicitly. This PR also switches `float32` to `float` which is less explicit.

I originally, thought that following NumPy's advice to use `float`, `int`, and `bool` as NumPy array dtypes for the documentation made sense. But if we are only doing that for `float` and not differentiating between `float64` and `float32`, we should like switch to using those. Since this is a larger change, and it is my fault for using `float` in the first place, if you agree @joaander I can make the necessary find and replace.

joaander:
> This change is needed, but the only concern I have is constency when documenting the types of the array elements. I am the one that promoted `float` over `numpy.float64` since these are effectively the same type `np.ndarray.astype` treats `float` as a double precision floating point number. However, we document the size of signed/unsigned integers explicitly. This PR also switches `float32` to `float` which is less explicit.

`float` is correct for snapshots following your recommendation to use `float` instead of `numpy.float64`:
```
>>> import hoomd
>>> sim = hoomd.Simulation(device=hoomd.device.CPU())
>>> sim.create_state_from_gsd('init.gsd')
>>> snap = sim.state.get_snapshot()
>>> snap.particles.position
array([[-0.5  ,  0.   ,  0.125],
       [ 0.5  ,  0.   ,  0.125]])
>>> snap.particles.position.dtype
dtype('float64')
```

I'm not aware of any places in HOOMD where we explicitly export `float32` arrays. If a user builds HOOMD with a single precision mode, many of these arrays (but not snapshots) would be `float32` instead of `float64`.

> I originally, thought that following NumPy's advice to use `float`, `int`, and `bool` as NumPy array dtypes for the documentation made sense. But if we are only doing that for `float` and not differentiating between `float64` and `float32`, we should like switch to using those. Since this is a larger change, and it is my fault for using `float` in the first place, if you agree @joaander I can make the necessary find and replace.

I don't think we can use `int` for the integer data types because these data types are explicitly sized and Python's `int` is not. Python also lacks a concept of an unsigned integers. At the same time, sphinx is not able to cross-link to `numpy.uint32` data structures - so I changed it to a more generic `uint32` and `int32` expecting users to know what that means. If you prefer the unlinked `numpy.uint32` and `numpy.int32` references I can change it back.

PR #1096, opened 2021-08-05
Filter updater

### LINKED ISSUE #796: Updater to force particle filter re-evaluations
## Description

Add an Updater that forces group memberships to be re-evaluated when triggered.

## Proposed solution

The updater would take a list of filters and call `updateMemberTags(true)` on their corresponding groups. This should be implemented in C++ to avoid python/C++ call overheads in cases where users choose to update groups with a small period.

## Additional context

HOOMD v3 uses filters to select particles, but caches the filter results in groups to maintain high performance.

## Developer

Seeking volunteers to implement this functionality.

b-butler:
## Description
Creates an updater that will update `ParticleGroups` associated with `ParticleFilter`s.

## Motivation and context
Resolves an open issue, and provides a needed functionality for non-static groups.

Resolves #796

## How has this been tested?
New test file.

## Change log

```
Adds: Add an updater to update the particles associated with a `hoomd.filter.ParticleFilter` instance.
```

## Checklist:

### DIFF
--- a/hoomd/pytest/test_filter_updater.py
+++ b/hoomd/pytest/test_filter_updater.py
@@ -71,9 +71,7 @@ def test_updating(simulation, filter_updater, filter_list):
     def modify_typeid(state):
         with state.cpu_local_snapshot as snapshot:
             Np = len(snapshot.particles.typeid)
-            indices = rng.choice(state.N_particles,
-                                 max(1, int(Np * 0.1)),
-                                 replace=False)
+            indices = rng.choice(Np, max(1, int(Np * 0.1)), replace=False)
             values = rng.choice([0, 1], len(indices))
             snapshot.particles.typeid[indices] = values
 
--- a/hoomd/pytest/test_filter_updater.py
+++ b/hoomd/pytest/test_filter_updater.py
@@ -1,0 +1,88 @@
+import numpy as np
+import pytest
+
+import hoomd
+import hoomd.conftest
+
+
+@pytest.fixture
+def filter_list():
+    return [
+        hoomd.filter.All(),
+        hoomd.filter.Tags([1, 2, 3]),
+        hoomd.filter.Type(["A"])
+    ]
+
+
+def test_initialization_setting(filter_list):
+    filter_updater = hoomd.update.FilterUpdater(1, [])
+    assert filter_updater.trigger == hoomd.trigger.Periodic(1)
+    assert filter_updater.filters == []
+    filter_updater.filters.extend(filter_list)
+    assert len(filter_updater.filters) == 3
+    assert filter_list == filter_updater.filters
+
+    filter_updater = hoomd.update.FilterUpdater(5, filter_list)
+    assert filter_updater.trigger == hoomd.trigger.Periodic(5)
+    assert len(filter_updater.filters) == 3
+    assert filter_list == filter_updater.filters
+    filter_updater.trigger = hoomd.trigger.After(100)
+    assert filter_updater.trigger == hoomd.trigger.After(100)
+
+
+@pytest.fixture
+def filter_updater(filter_list):
+    return hoomd.update.FilterUpdater(1, filter_list)
+
+
+@pytest.fixture(scope="function")
+def simulation(lattice_snapshot_factory, simulation_factory, filter_list):
+    sim = simulation_factory(
+        lattice_snapshot_factory(particle_types=["A", "B"]))
+    # place filters in state list manually to enable updating the particle
+    # groups.
+    for filter_ in filter_list:
+        sim.state._get_group(filter_)
+    return sim
+
+
+def test_attaching(simulation, filter_updater):
+    simulation.operations += filter_updater
+    trigger = filter_updater.trigger
+    filters = filter_updater.filters
+    simulation.run(0)
+    assert trigger == filter_updater.trigger
+    assert filters == filter_updater.filters
+    assert hasattr(filter_updater, "_cpp_obj")
+    assert filter_updater._attached
+
+
+def assert_group_match(filter_, state):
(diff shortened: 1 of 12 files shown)


hoomd/pytest/test_filter_updater.py
```diff
@@ -0,0 +1,88 @@
...
+        with state.cpu_local_snapshot as snapshot:
+            Np = len(snapshot.particles.typeid)
+            indices = rng.choice(state.N_particles,
+                                 max(1, int(Np * 0.1)),
+                                 replace=False)
+            values = rng.choice([0, 1], len(indices))
+            snapshot.particles.typeid[indices] = values
+
+    for _ in range(4):
+        modify_charge(simulation.state)
```
joaander, line 81:
```suggestion
    def modify_typeid(state):
        with state.cpu_local_snapshot as snapshot:
            Np = len(snapshot.particles.typeid)
            indices = rng.choice(state.N_particles,
                                 max(1, int(Np * 0.1)),
                                 replace=False)
            values = rng.choice([0, 1], len(indices))
            snapshot.particles.typeid[indices] = values

    for _ in range(4):
        modify_typeid(simulation.state)
```
Name the function consistent with what it does.

hoomd/update/particle_filter.py
```diff
@@ -0,0 +1,92 @@
...
+        state = copy.copy(self.__dict__)
+        state.pop("_state", None)
+        return state
+
+
+class FilterUpdater(hoomd.operation.Updater):
+    """Update sets of particles associated with a filter.
+
+    HOOMD stores the particles selected by `hoomd.filter.ParticleFilter`
+    instance that are used by it's objects to avoid recomputing them on demand.
```
joaander, line 37:
```suggestion
    HOOMD caches the particles selected by `hoomd.filter.ParticleFilter`
    instances to avoid the cost of re-running the filter on every time step.
```
Reword for clarity.

hoomd/update/particle_filter.py
```diff
@@ -0,0 +1,92 @@
...
+    This means that unless the particles selected by a filter are recomputed the
+    are static. This class provides a mechanism to update grouped particles via
+    an updater.
```
joaander, line 40:
```suggestion
    This means that unless the particles selected by a filter are recomputed the
    set particles an operation works on is static. This class provides a
    mechanism to update the cached list of particles periodically. For example,
    use it to update the particles operated on by an MD integration method so
    that the integration method applies to particles in a given region of space.
```
Add more context.

hoomd/update/particle_filter.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Note:
+        Through custom `hoomd.trigger.Trigger` subclasses, filters can be made
+        to update only when there is a known change to the particles that a
+        filter would select to save computation.
```
joaander, line 45:
```suggestion
        If needed to improve performance, use a `hoomd.trigger.Trigger`
        subclass, to update only when there is a known change to the particles
        that a filter would select.
```
Use active voice.

hoomd/update/particle_filter.py
```diff
@@ -0,0 +1,92 @@
...
+        super().__init__(trigger)
+        self._filters = hoomd.data.syncedlist.SyncedList(
+            hoomd.filter.ParticleFilter,
+            iterable=filters,
+            to_synced_list=_GroupConverter(),
+            attach_members=False)
+
+    @property
+    def filters(self):
+        """hoomd.filter.ParticleFilter: filters to update select particles."""
```
joaander, line 73:
```suggestion
        """list[hoomd.filter.ParticleFilter]: filters to update select particles."""
```

joaander (changes requested):
Excellent, thanks! I have some suggestions to improve the documentation.

hoomd/pytest/test_filter_updater.py
b-butler: Yep, this is a typo from a previous version of the test.

hoomd/pytest/test_filter_updater.py
```diff
@@ -0,0 +1,95 @@
...
+    # On MPI simulations, the group tags won't exactly match since they include
+    # particles from every rank, so two checks are necessary. One that no
+    # particles in the filters tags are not in the groups tags (below), and that
+    # all local tags in group tags are in filter tags (2nd check).
+    assert filter_tags - group_tags == set()
+    if not mpi:
+        return
+    NOT_LOCAL = 4294967295
+    with state.cpu_local_snapshot as snapshot:
+        np.all(snapshot.particles.rtag[group_tags - filter_tags] == NOT_LOCAL)
```
joaander, line 72: This check is not being performed with the default `mpi=False` argument.
You can set `mpi=True` if the check is valid even in serial simulations, or you can make it conditional on `simulation.device.communicator.num_ranks > 1`

hoomd/update/particle_filter.py
```diff
@@ -0,0 +1,95 @@
...
+        return state
+
+
+class FilterUpdater(hoomd.operation.Updater):
+    """Update sets of particles associated with a filter.
+
+    HOOMD caches the particles selected by `hoomd.filter.ParticleFilter`
+    instances to avoid the cost of re-running the filter on every time step.
+    This means that unless the particles selected by a filter are recomputed the
+    set particles an operation works on is static. This class provides a
```
joaander, line 39:
```suggestion
    set of particles an operation works on is static. This class provides a
```

joaander (changes requested):
Excellent, thanks! There are just a few minor changes needed.

SchoeniPhlippsn (commented):
I think this looks good once Josh's comments have been addressed.

joaander (approved):
Thanks!

PR #1102, opened 2021-08-18
Update HPMC move size tuner documentation for multitype systems

tcmoore3:
## Description

Add an example into the MoveSize tuner documentation about how to tune for multiple particle types.

## Motivation and context

This shows how to tune move sizes for particle types independently, which important for systems with large particle size and/or shape polydispersity.

## How has this been tested?

Built docs locally.

## Checklist:

### DIFF
--- a/hoomd/hpmc/tune/move_size.py
+++ b/hoomd/hpmc/tune/move_size.py
@@ -322,37 +322,7 @@ class MoveSize(_InternalCustomTuner):
         approach the target value, while the per-type acceptance ratios may not
         be close to the target value. This requires creating move size tuners
         for each type and leveraging the ``ignore_statistics`` flag of the shape
-        property of the HPMC integrator; see below for an example.
-
-    ::
-
-        mc_tuners = {}
-        for ptype in sim.state.particle_types:
-            mc_tuners[ptype] = hoomd.hpmc.tune.MoveSize.scale_solver(
-                moves=['a', 'd'],
-                target=0.2,
-                types=[ptype],
-                trigger=hoomd.trigger.Periodic(1),
-                max_translation_move=2*max_r,
-                max_rotation_move=1.0,
-            )
-            # get shape params, modify, and re-set
-            td = mc.shape[ptype]
-            td['ignore_statistics'] = True
-            mc.shape[ptype] = td
-
-        # now loop through and only acknowledge statistics of 1 type at a time
-        for acknowledged_type in sim.state.particle_types:
-            sim.operations.tuners.append(mc_tuners[acknowledged_type])
-            td = mc.shape[acknowledged_type]
-            td['ignore_statistics'] = False
-            mc.shape[acknowledged_type] = td
-            for _ in range(10):
-                sim.run(10)
-            td = mc.shape[acknowledged_type]
-            td['ignore_statistics'] = True
-            mc.shape[acknowledged_type] = td
-            sim.operations.tuners.remove(mc_tuners[acknowledged_type])
+        property of the HPMC integrator.
 
     """
     _internal_class = _InternalMoveSize
--- a/hoomd/hpmc/tune/move_size.py
+++ b/hoomd/hpmc/tune/move_size.py
@@ -320,9 +320,10 @@ class MoveSize(_InternalCustomTuner):
         acceptance ratio. Otherwise, the *global* acceptance ratio, a weighted
         average of the acceptance ratios for each individual particle type, will
         approach the target value, while the per-type acceptance ratios may not
-        be close to the target value. This requires creating move size tuners
-        for each type and leveraging the ``ignore_statistics`` flag of the shape
-        property of the HPMC integrator.
+        be close to the target value. This requires setting the ``types`` attribute
+        to be one type at a type while setting the ``ignore_statistics`` flag of
+        the shape property of the HPMC integrator for all other types to
+        ``True``.
 
     """
     _internal_class = _InternalMoveSize
--- a/hoomd/hpmc/tune/move_size.py
+++ b/hoomd/hpmc/tune/move_size.py
@@ -320,10 +320,10 @@ class MoveSize(_InternalCustomTuner):
         acceptance ratio. Otherwise, the *global* acceptance ratio, a weighted
         average of the acceptance ratios for each individual particle type, will
         approach the target value, while the per-type acceptance ratios may not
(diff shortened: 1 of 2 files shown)


tcmoore3:
Per a discussion with Brandon, this example may be too long for the API reference documentation. A possible fix for this would be to create a new "how-to" section of the docs, similar to what [signac does](https://docs.signac.io/en/latest/recipes.html). Then blocks of code like this that are too long for the API docs but don't quite fit into the tutorials could go in there.

hoomd/hpmc/tune/move_size.py
```diff
@@ -310,6 +312,48 @@ class MoveSize(_InternalCustomTuner):
...
+            sim.operations.tuners.append(mc_tuners[acknowledged_type])
+            td = mc.shape[acknowledged_type]
+            td['ignore_statistics'] = False
+            mc.shape[acknowledged_type] = td
+            for _ in range(10):
+                sim.run(10)
+            td = mc.shape[acknowledged_type]
+            td['ignore_statistics'] = True
+            mc.shape[acknowledged_type] = td
+            sim.operations.tuners.remove(mc_tuners[acknowledged_type])
```
b-butler, line 355: Only one tune object needs to be created,  the `obj.types` attribute only needs to be change through something like `obj.types = [acknowledged_type]`.

```suggestion
        def set_statistics(mc, type, value):
            """ Get shape params, modify, and re-set."""
            tp = mc.shape[type]
            tp["ignore_statistics"] = value
            mc.shape[type] = tp

        mc_tuner = hoomd.hpmc.tune.MoveSize.scale_solver(
            moves=['a', 'd'], target=0.2, types=[ptype],
            trigger=hoomd.trigger.Periodic(1), max_translation_move=2*max_r,
            max_rotation_move=1.0,
            )
        sim.operations += mc_tuner
        
        for ptype in sim.state.particle_types:
            set_statistics(mc, ptype, True)

        # now loop through and only acknowledge statistics of 1 type at a time
        for acknowledged_type in sim.state.particle_types:
            mc_tuner.types = [acknowledged_type]
            set_statistics(mc, acknowledged_type, False)
            sim.run(100)
            set_statistics(mc, acknowledged_type, True)
```
The suggestion is also shorter, I think which should help reduce the imposing nature of it.

hoomd/hpmc/tune/move_size.py
```diff
@@ -350,6 +394,12 @@ def scale_solver(cls,
                 considered tuned. The tolerance should not be too much lower
                 than the default of 0.01 as acceptance rates can vary
                 significantly at typical tuning rates.
+
+        Note:
+            Increasing ``gamma`` towards 1 does not necessarily speed up
+            convergence and can slow it down. In addition, large values of
+            ``gamma`` can make the solver unstable, especially when tuning
+            frequently.
```
b-butler, line 402:
```suggestion
```
I don't think this is as true as in the secant solver. This is a simple scaling, and is more stable albeit slower than the secant slower in many use cases (as far as I could tell).

sphinx-doc/module-hpmc-tune.rst
```diff
@@ -16,7 +16,7 @@ hpmc.tune
     :synopsis: Tuners for HPMC.
     :members:
 
-    .. autoclass:: MoveSize(trigger, moves, target, solver, types=None, max_move_size=None)
+    .. autoclass:: MoveSize
```
b-butler, line 19: Revert this change please. Due to the indirection used to present `MoveSize` as a _normal_ HOOMD operation rather than a custom action one, sphinx does not know the signature for the class otherwise.

There is a way to tell Python to pretend a class of method has a different signature (yay runtime patching), but we don't currently use that.

b-butler (changes requested):
Thanks @tcmoore3 for the documentation improvements. Please see my suggestions.

joaander:
> Per a discussion with Brandon, this example may be too long for the API reference documentation. A possible fix for this would be to create a new "how-to" section of the docs, similar to what [signac does](https://docs.signac.io/en/latest/recipes.html). Then blocks of code like this that are too long for the API docs but don't quite fit into the tutorials could go in there.

We can talk about this at the hoomd developers meeting.

joaander:
Should we close this PR, or remove the long form example and keep the other changes?

joaander:
@b-butler @tcmoore3 Should we close this PR, or remove the long form example and keep the other changes?

tcmoore3:
I think we should close this and save it for a how-to page if we ever decide to go that direction.

hoomd/hpmc/tune/move_size.py
```diff
@@ -310,6 +312,18 @@ class MoveSize(_InternalCustomTuner):
         rate tends towards 1. Therefore, it is recommended to pick a moderate
         maximum move size for at least the translational moves to prevent
         requiring checking periodic images.
+
+    Note:
+        In systems containing nonmonodisperse particle shapes and/or sizes, move
```
joaander, line 317:
```suggestion
        In systems containing disparate particle shapes and/or sizes, move
```

joaander:
@tcmoore3 When this is ready for another round of reviews, use the icon next to the reviewer's name to re-request a review.

joaander (commented):
@b-butler needs to review as he requested changes.

hoomd/hpmc/tune/move_size.py
```diff
@@ -310,6 +312,18 @@ class MoveSize(_InternalCustomTuner):
...
+    Note:
+        In systems containing disparate particle shapes and/or sizes, move
+        sizes for the different types should be tuned independently so that the
+        acceptances rates for the different particles are each near the target
+        acceptance ratio. Otherwise, the *global* acceptance ratio, a weighted
+        average of the acceptance ratios for each individual particle type, will
+        approach the target value, while the per-type acceptance ratios may not
+        be close to the target value. This requires creating move size tuners
+        for each type and leveraging the ``ignore_statistics`` flag of the shape
+        property of the HPMC integrator.
```
b-butler, line 325:
```suggestion
        be close to the target value. This requires setting the ``types`` attribute
        to be one type at a type while setting the ``ignore_statistics`` flag of
        the shape property of the HPMC integrator for all other types to
        ``True``.
```

b-butler (changes requested):
I have one comment/suggestion about some wording that might be misleading.

Issue #1001, opened 2021-05-21
Doc rendering issues


bdice:
Some docstrings aren't rendering correctly. I wasn't immediately sure how to fix the rendering or I would have made a PR.

https://github.com/glotzerlab/hoomd-blue/blob/c596d2c570d3c1d2d161b7f0fcccb2d4ad6e86eb/hoomd/md/force.py#L63-L64
https://hoomd-blue.readthedocs.io/en/latest/module-md-force.html#hoomd.md.force.Force.energies
![image](https://user-images.githubusercontent.com/3943761/119070037-e3113000-b9ac-11eb-855c-4b47880ccbdf.png)

vyasr:
I would guess that the problem is that Sphinx doesn't correctly handle formatting when the first line of the docstring is split over two lines. This is why so many docstrings in our other packages end up putting in a `noqa: E501` on docstrings, to avoid line breaks that result in poor formatting like this.

joaander:
This is now fixed on master. 

For future reference, there are two likely causes for formatting that looks like this.

1. A list is not indented properly on line continuation. Example (incorrect):
```
* This is a list item
    broken across lines
```

restructured text formats this as a definition list. The intention is for a single bulleted list item. To get that, line up the start of the continued line with the start of the first line (b with T in this example). Note that this means indenting by 2 chars for `*` and `-`, but 3 for `1.` in lists. Many editors auto-indent and unindent features do not work so well with this and try to force things on multiples of 4 characters.

Correct:
```
* This is a list item
  broken across lines
```

2. Google formatting in napoleon parses only the first line in a property docstring as its type and short description. When that short description is too long (example):
```
@property
def energy()
    """(float) This is a long line
    that needs to be broken."""
```

It either needs to be reworded to be shorter, or use `\` to make the line continued:
```
@property
def energy()
    """(float) This is a long line \
    that needs to be broken."""
```

There is one caveat here. When a single-line type and short description is in a `r"""` docstring, the `\` will be included in the string and will not continue the line :( The only option I found here, is to drop Google formatting in these cases and use `.. py:attribute::` directly. Example:

```
    .. py:attribute:: params

        The LJ potential parameters. The dictionary has the following keys:

        * ``epsilon`` (`float`, **required**) -
          energy parameter :math:`\varepsilon` (units: [energy])
        * ``sigma`` (`float`, **required**) -
          particle size :math:`\sigma` (units: [length])
[...]

PR #1113, opened 2021-08-25
Improve MD documentation.

bdice:
## Description
This PR fixes a number of issues in the documentation. I was reading through the MD docs to familiarize myself with features I don't use very often. As I was reading, I fixed up issues with formulas, units, and LaTeX/sphinx formatting. I edited `md.angle` through `md.external` and stopped at `md.force`, with the exception of separating function cases with semicolons, which I standardized for all instances of the LaTeX command `eqnarray` in the `md` subdirectory.

- Improve angle docs.
- Use semicolons to separate all function cases.
- Update NumPy intersphinx link.
- r is a distance, no vector needed.
- Improve docs in constrain module.
- Fix pressure tensor formula.
- Fix formulas and formatting in dihedral documentation.
- Improve documentation and fix formulas/units for external field potentials.

## Motivation and context

Improves documentation.

## Checklist:

### DIFF
--- a/hoomd/md/angle.py
+++ b/hoomd/md/angle.py
@@ -38,7 +38,7 @@ class Angle(Force):
 class Harmonic(Angle):
     r"""Harmonic angle potential.
 
-    The command angle.harmonic specifies a harmonic potential energy between
+    :py:class:`Harmonic` specifies a harmonic potential energy between
     every triplet of particles with an angle specified between them.
 
     .. math::
@@ -52,10 +52,10 @@ class Harmonic(Angle):
             The parameter of the harmonic bonds for each particle type.
             The dictionary has the following keys:
 
-            * ``k`` (`float`, **required**) - potential constant
+            * ``k`` (`float`, **required**) - potential constant :math:`k`
               :math:`[\mathrm{energy} \cdot \mathrm{radians}^{-2}]`
 
-            * ``t0`` (`float`, **required**) - rest angle
+            * ``t0`` (`float`, **required**) - rest angle :math:`\theta_0`
               :math:`[\mathrm{radians}]`
 
     Examples::
@@ -77,7 +77,7 @@ class Harmonic(Angle):
 class Cosinesq(Angle):
     r"""Cosine squared angle potential.
 
-    The command angle.cosinesq specifies a cosine squared potential energy
+    :py:class:`Cosinesq` specifies a cosine squared potential energy
     between every triplet of particles with an angle specified between them.
 
     .. math::
@@ -94,8 +94,8 @@ class Cosinesq(Angle):
             The parameter of the harmonic bonds for each particle type.
             The dictionary has the following keys:
 
-            * ``k`` (`float`, **required**) - potential constant
-              :math:`[\mathrm{energy} \cdot \mathrm{radians}^{-2}]`
+            * ``k`` (`float`, **required**) - potential constant :math:`k`
+              :math:`[\mathrm{energy}]`
 
             * ``t0`` (`float`, **required**) - rest angle :math:`\theta_0`
               :math:`[\mathrm{radians}]`
--- a/hoomd/md/bond.py
+++ b/hoomd/md/bond.py
@@ -46,8 +46,7 @@ class Harmonic(Bond):
 
         V(r) = \frac{1}{2} k \left( r - r_0 \right)^2
 
-    where :math:`\vec{r}` is the vector pointing from one particle to the other
-    in the bond.
+    where :math:`r` is the distance from one particle to the other in the bond.
 
     Attributes:
         params (TypeParameter[``bond type``, dict]):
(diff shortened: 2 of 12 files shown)


hoomd/md/angle.py
```diff
@@ -38,7 +38,7 @@ def _attach(self):
 class Harmonic(Angle):
     r"""Harmonic angle potential.
 
-    The command angle.harmonic specifies a harmonic potential energy between
+    :py:class:`Harmonic` specifies a harmonic potential energy between
```
bdice, line 41: We might prefer to remove the Sphinx role? I was copying this from other classes and used an explicit role to match the examples I saw.
```suggestion
    `Harmonic` specifies a harmonic potential energy between
```

hoomd/md/angle.py
```diff
@@ -77,7 +77,7 @@ def __init__(self):
 class Cosinesq(Angle):
     r"""Cosine squared angle potential.
 
-    The command angle.cosinesq specifies a cosine squared potential energy
+    :py:class:`Cosinesq` specifies a cosine squared potential energy
```
bdice, line 80: See above.
```suggestion
    `Cosinesq` specifies a cosine squared potential energy
```

hoomd/md/bond.py
```diff
@@ -46,8 +46,7 @@ class Harmonic(Bond):
 
         V(r) = \frac{1}{2} k \left( r - r_0 \right)^2
 
-    where :math:`\vec{r}` is the vector pointing from one particle to the other
-    in the bond.
+    where :math:`r` is the distance from one particle to the other in the bond.
```
bdice, line 49: This formula uses `r` as a distance, so we don't need to define or introduce the vector quantity / direction. Also applies below.

hoomd/md/constrain.py
```diff
@@ -209,7 +209,7 @@ class Rigid(Constraint):
         - ``diameters`` (list[float]): List of diameters of constituent
           particles
 
-        Type: `TypeParameter[``particle_type``, `dict`]
+        Type: `TypeParameter` [``particle_type``, `dict`]
```
bdice, line 212: The `TypeParameter` type definitions are inconsistent throughout the code, I did my best to improve whatever was already there based on examples elsewhere in the code.

Some use a construction like this, and I'm not sure if we have a reason to prefer one over the other:

```
    Attributes:
        params (`TypeParameter` [``dihedral type``, `dict`]):
            The parameter of the harmonic bonds for each dihedral type. The
            dictionary has the following keys:
```

hoomd/md/external/field.py
```diff
@@ -15,12 +15,14 @@
...
-    Users should not invoke `Field` directly. Documentation common to all
-    external field potentials is located here. External potentials represent
-    forces which are applied to all particles in the simulation by an external
-    agent.
+    External potentials represent forces which are applied to all particles in
+    the simulation by an external agent.
+
+    Note:
+        `Field` is the base class for all external field potentials.
+        Users should not instantiate this class directly.
```
bdice, line 25: I tried to make this match the style of other base classes' docstrings. It felt weird that the base class previously had self-referential mentions of common documentation belonging there. It sounded like a docstring declaring itself to be a docstring.

hoomd/md/external/field.py
```diff
@@ -48,7 +50,7 @@ class Periodic(Field):
 
     .. math::
 
-       V(\\vec{r}) = A * \\tanh\\left[\\frac{1}{2 \\pi p w} \\cos\\left(
+       V(\\vec{r}) = A \\tanh\\left[\\frac{1}{2 \\pi p w} \\cos\\left(
```
bdice, line 53: I removed stars in a few places where the LaTeX rendering is clear that there is an implied multiplication operation, especially because the star renders very prominently and makes me question if there is some kind of unusual product operation being used.

sphinx-doc/conf.py
```diff
@@ -28,7 +28,7 @@
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
-    'numpy': ('https://docs.scipy.org/doc/numpy', None),
+    'numpy': ('https://numpy.org/doc/stable', None),
```
bdice, line 31: NumPy moved their docs URL. This hides a warning that pops up when building the docs.

hoomd/md/angle.py
joaander: Yes, we can remove Sphinx roles.

hoomd/md/bond.py
joaander: Yes, we can make this change here.

hoomd/md/constrain.py
joaander: I have used the Sphinx `.. py:attribute::` syntax in places where Google formatting for the attribute name and type is longer than the line length limit. As discussed in #1001 I do not accept the solution that applies noqa to all lines in a docstring as that to easily allows for many style violations beyond the local line where Google style demands it.

joaander:
This looks to be ready for another round of review. @lyrivera we would greatly appreciate your review.

b-butler (approved):
:+1:

PR #1114, opened 2021-08-25
Fix problems with __eq__ in data model

b-butler:
## Description

Tackles a problem recently found in the mailing list regarding a false positive when checking equality between two `hoomd.write.Table` instances. This fixes the existing problem, as well as improves the implementation of `__eq__` for the data model generally.

The PR also removes all current warnings when running `pytest`. This wasn't necessary, but a mindless change for the most part, so I have added it as well.

## Motivation and context

Fixes the reported bug and helps prevent others in the future by making `__eq__` behave better.

## How has this been tested?

The existing tests pass which rely on the `==` operator.

## Change log

```
Fixed: Fixed errors related to equality checks between HOOMD operations.
```

## Checklist:


joaander (approved):
Thanks!

PR #1099, opened 2021-08-13
Move Active force rotational diffusion to updater

b-butler:
## Description

This PR moves the rotational diffusion part of `hoomd.md.force.Active` and subclasses to its own updater. In addition this PR addresses various bugs or limitations in other parts of the code that lead to errors or difficulties in adding the updater. For more info on those, see the git commit messages.

## Motivation and context

This addresses #898 and bugs in the existing code base.

Resolves #898

## How has this been tested?

An additional test file has been created for testing the updater.

## Change log

```
Added: A new roational diffusion updater for use with Active Forces ``hoomd.md.update.ActiveRotationalDiffusion``.
```
or
```
Changed: Active force rotaional diffusion is now managed by ``hoomd.md.update.ActiveRotationalDiffusion``.
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

PR #1053, opened 2021-07-09
Separate RATTLE integrator methods

SchoeniPhlippsn:
## Description

I put all rattle integrators in separate classes to make sure that integrator methods can be removed from the simulation.  

## Motivation and context

Resolves #998 and #1052

## How has this been tested?

Tests have been updated to the new changes.

## Change log

```

```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

PR #1118, opened 2021-08-25
Fix Integrator and Neighbor List detaching

b-butler:
## Description

Fixes `hoomd.md._DynamicIntegrator` and `hoomd.md.nlist.NList` detaching. Currently MD integrator objects do not detach any of the forces, methods, or constraints or the rigid parameter. The neighbor list is not detached by the composing potentials either. This is fixed as well through using the `_DependencyRelation` interface. This interface is cleaned up and fixed in this PR as well.

This PR cherry-picks a commit from #1099, so either a rebase and force push here or there will be needed after the other is merged.

This is also dependent on #1053 like #1099.

## Motivation and context

This fixes multiple existing bugs with detaching properly which is done when operations or other objects are removed from a simulation that is attached.

## How has this been tested?

Currently, neighbor list detaching is tested, but no tests for the md integrator currently exist. I can add the test for detaching, but I don't think all the tests should go in this PR.

## Change log

```
Fixes: Removing an integrator from a simulation after running.
```

## Checklist:

### DIFF
--- a/hoomd/md/methods/methods.py
+++ b/hoomd/md/methods/methods.py
@@ -23,15 +23,7 @@ class Method(_HOOMDBaseObject):
     Note:
         Users should use the subclasses and not instantiate `Method` directly.
     """
-
-    def _detach(self):
-        if self._attached:
-            self._unapply_typeparam_dict()
-            self._update_param_dict()
-
-            self._cpp_obj = None
-            self._notify_disconnect(self._simulation)
-        return self
+    pass
 
 
 class NVT(Method):
--- a/hoomd/md/methods/methods.py
+++ b/hoomd/md/methods/methods.py
@@ -844,7 +844,7 @@ class Langevin(Method):
 
         Langevin uses RNGs. Warn the user if they did not set the seed.
         """
-        if simulation is not None:
+        if isinstance(simulation, hoomd.Simulation):
             simulation._warn_if_seed_unset()
 
         super()._add(simulation)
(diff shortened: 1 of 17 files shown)


hoomd/md/nlist.py
```diff
@@ -82,12 +82,6 @@ class NList(_HOOMDBaseObject):
             :math:`[\mathrm{length}]`.
     """
 
-    _remove_for_pickling = _HOOMDBaseObject._remove_for_pickling + (
-        '_cpp_cell',)
-    _skip_for_equality = _HOOMDBaseObject._skip_for_equality | {
-        '_cpp_cell',
-    }
-
```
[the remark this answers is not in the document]
b-butler: We don't ever set `_cpp_cell` as far as I can see.

joaander (changes requested):
Thanks! We need to merge master into this branch and fix the failing tests before merging.

hoomd/md/methods/methods.py
```diff
@@ -23,15 +23,7 @@ class Method(_HOOMDBaseObject):
     Note:
         Users should use the subclasses and not instantiate `Method` directly.
     """
-
-    def _detach(self):
```
joaander, line 27: Is this now handled in the base class?

joaander (approved):
Thanks! I have one question for clarification.

b-butler:
@joaander, I am going to rerequest review because some major changes were necessary to get the desired behavior implemented. Some may seem unrelated, but they aren't due to requirements for avoiding copying the neighbor list unless necessary. The changes are generally cleaner regardless. I added new tests to ensure the behavior is as expected.

hoomd/md/methods/methods.py
b-butler: Yes, this keeps popping back up some how, I deleted it in another branch as well.

joaander (approved):
Thanks!

Issue #1078, opened 2021-07-28
Revise documentation to include complete theory descriptions


joaander:
## Description

Expand the documentation of the box, MD integrators, pair potentials, bond potentials, and other classes with a mathematical description of the computation performed. Where needed, include figures to help explain.

## Motivation and context

HOOMD's current documentation makes a number of assumptions that readers understand certain terms (e.g. exclusion). Not all readers have the same background in simulation codes to understand these terms, or they may be familiar with codes that use these terms differently. Documenting these with mathematical equations will make the computations that HOOMD performs more explicit and understandable by readers. After this change, there will still be built in assumptions, but the assumed knowledge will be textbook material (e.g. potential energy, Hamiltonian).

We also need to make consistent use of symbols across the user documentation. For example, we currently use `V` for both potential energy and volume.

Related to #1027

joaander:
When updating the box documentation, there is good material from hoomd v2: https://hoomd-blue.readthedocs.io/en/stable/box.html

joaander:
A few notes on things that need to be clarified in this.
* `Force.energy` and `Force.additional_energy` need to more clearly document what and what is not included.
* Same for the virial and additional virial.
* `ThermodynamicQuantities` needs to clearly document that the additional energy and virials are included.

PR #1119, opened 2021-08-27
Describe integrator reversibility.

bdice:
## Description
Adds documentation to integration methods to note whether they are reversible.

After talking further with @joaander on Slack, the initial guess of which methods are symplectic/reversible in this PR may not be accurate. This needs to be reviewed carefully for correctness. Alternatively we can just close this PR and wait for #1078.

## Motivation and context

Related to #1078.

## Change log

```

```

## Checklist:

### DIFF
--- a/hoomd/md/methods/methods.py
+++ b/hoomd/md/methods/methods.py
@@ -79,6 +79,11 @@ class NVT(Method):
 
         `NVT` integrates rotational degrees of freedom.
 
+    Note:
+        The NVT integrator is reversible. In addition to negating velocities
+        and angular momenta, the thermostat degrees of freedom must also be
+        negated.
+
     Examples::
 
         nvt=hoomd.md.methods.NVT(filter=hoomd.filter.All(), kT=1.0, tau=0.5)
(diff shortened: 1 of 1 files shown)


hoomd/md/methods/methods.py
```diff
@@ -79,6 +79,11 @@ class NVT(Method):
 
         `NVT` integrates rotational degrees of freedom.
 
+    Note:
+        The NVT integrator is reversible. In addition to negating velocities
```
bdice, line 83: This may not be true, after further discussion with @joaander in Slack. The acceleration applied to the thermostat depends on the difference between the instantaneous temperature and the set point, which may not work correctly if reversed. But if NVT isn't reversible, is the NPT integrator reversible? Does the NPT thermostat work the same way or not?

bdice:
I don't know if the information in this PR is correct or not -- I am going to close this PR and let it be handled at a later time with #1078. From what I can tell, the NVT ensemble with a Nose-Hoover thermostat is reversible, but I don't know the implementation details in HOOMD well enough to make a documentation PR that I feel confident about.

See also:
- https://docs.lammps.org/fix_nh.html
- https://manual.q-chem.com/5.3/subsec_Thermostats.html

joaander:
> I don't know if the information in this PR is correct or not -- I am going to close this PR and let it be handled at a later time with #1078. From what I can tell in the references below, the NVT ensemble with a Nose-Hoover thermostat is reversible, but I don't know the implementation details in HOOMD well enough to make a documentation PR that I feel confident about.
> 
> See also:
> 
> * https://docs.lammps.org/fix_nh.html
> * https://manual.q-chem.com/5.3/subsec_Thermostats.html

@bdice I think "time reversibility" in those contexts is with respect to negating `\delta t`. For path sampling, you were discussing a different type of reversibility, that of negating the momenta in the system.

PR #1121, opened 2021-08-28
OverdampedViscous method

### LINKED ISSUE #1100: Noiseless Brownian Dynamics
## Description

It appears the `noiseless_t` and `noiseless_r` arguments in the `Langevin` and `Brownian` methods were removed from the Python API in the conversion from hoomd v2 to v3. 

Is there a reason for this / would I be able to add them back? I would like to simulate Active Brownian Particles using Brownian Dynamics with no translational noise.

## Proposed solution
The `noiseless_t` and `noiseless_r` arguments could be added to the Python `Langevin` and `Brownian` classes and their parameter dicts, and also added to the appropriate C++ constructors.

## Additional context

The `noiseless_t` and `noiseless_r` arguments are still used in the C++ code, just automatically set to `false`.

## Developer

I have a local branch that makes these changes, I can open a pull request if there is interest in adding these arguments back.

danevans09:
## Description

This implements a new MD method in python, `OverdampedViscous`, that is an overdamped integrator with a drag force but no random force. This class uses the C++ `Brownian` integrator with `noiseless_t` and `noiseless_r` set to `True`.

## Motivation and context

In the version 2 API, `noiseless_t` and `noiseless_r` were arguments to the Python `Brownian` integrator. These arguments were removed in version 3, so this `OverdampedViscous` class re-enables simulations that are on the Brownian time scale with a drag force but have no Brownian force or torque.

Resolves #1100 

## How has this been tested?

I added the `OverdampedViscous` method to `test_methods.py`

## Change log

```
Added: Overdamped integrator with a drag force but no random force ``hoomd.md.methods.OverdampedViscous``
```

## Checklist:

### DIFF
--- a/hoomd/md/methods/rattle.py
+++ b/hoomd/md/methods/rattle.py
@@ -487,10 +487,10 @@ class OverdampedViscous(MethodRATTLE):
 
         sphere = hoomd.md.manifold.Sphere(r=10)
         odv_rattle = hoomd.md.methods.rattle.OverdampedViscous(
-        filter=hoomd.filter.All(), manifold_constraint=sphere,
-        seed=1, alpha=1.0)
-        integrator = hoomd.md.Integrator(dt=0.001, methods=[odv_rattle],
-        forces=[lj])
+            filter=hoomd.filter.All(), manifold_constraint=sphere, seed=1,
+            alpha=1.0)
+        integrator = hoomd.md.Integrator(
+            dt=0.001, methods=[odv_rattle], forces=[lj])
 
 
     Attributes:
--- a/hoomd/md/methods/rattle.py
+++ b/hoomd/md/methods/rattle.py
@@ -454,3 +454,132 @@ class Brownian(MethodRATTLE):
 
         # Attach param_dict and typeparam_dict
         super()._attach()
+
+
+class OverdampedViscous(Method):
+    r"""Overdamped viscous dynamics with RATTLE constraint.
+
+    Args:
+        filter (`hoomd.filter.ParticleFilter`): Subset of particles to
+            apply this method to.
+
+        manifold_constraint (:py:mod:`hoomd.md.manifold.Manifold`): Manifold
+            constraint.
+
+        alpha (`float`): When set, use :math:`\alpha d_i` for the
+            drag coefficient where :math:`d_i` is particle diameter.
+            Defaults to None
+            :math:`[\mathrm{mass} \cdot \mathrm{length}^{-1}
+            \cdot \mathrm{time}^{-1}]`.
+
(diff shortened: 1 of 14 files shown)


joaander (changes requested):
Thanks @danevans09 ! I have some suggestions for improvement. Also: Add OverdampedViscous to the summary and class list in the appropriate `.rst` files so they appear in the documentation.

One of the suggestions involves making minor modifications to the behavior of Brownian at the C++ level. This will implement a long requested for feature where users on the mailing list have asked for useful velocity data from their active particle integrators. You might add a `Tip:` in the method docs that this method is useful for active matter to provide a search term users might look for in the docs.

joaander (changes requested):
Thanks!

Please correct the warnings issued by pre-commit. 

Also, I forgot to mention that the RATTLE BD integrator has a separate implementation in `TwoStepRATTLEBD.h` and `TwoStepRATTLEBDGPU.cuh` those also need modifications to handle the velocity setting. I think you only need to modify the first step of the integration method as the `includeRATTLEForce` method handles the `noiseless_t` variable correctly and doesn't set velocities. The part of the code that projects the velocity into the tangent plane should be executed in both branches, the only difference is that when `noiseless_t` is set, the velocity to project is not random but is force over gamma. Ask @SchoeniPhlippsn if you have any questions about these methods.

@pepak13 it would be great to hear your thoughts on these changes.

hoomd/md/methods/rattle.py
```diff
@@ -448,6 +448,138 @@ def _attach(self):
...
+    `hoomd.md.methods.OverdampedViscous`.
+
+    Examples of using ``manifold_constraint``::
+
+        sphere = hoomd.md.manifold.Sphere(r=10)
+        odv_rattle = hoomd.md.methods.rattle.OverdampedViscous(
+        filter=hoomd.filter.All(), manifold_constraint=sphere,
+        seed=1, alpha=1.0)
+        integrator = hoomd.md.Integrator(dt=0.001, methods=[odv_rattle],
+        forces=[lj])
```
joaander, line 493:
```suggestion
        odv_rattle = hoomd.md.methods.rattle.OverdampedViscous(
            filter=hoomd.filter.All(), manifold_constraint=sphere, seed=1,
            alpha=1.0)
        integrator = hoomd.md.Integrator(
            dt=0.001, methods=[odv_rattle], forces=[lj])
```
Make the line splits more readable.

joaander (changes requested):
One last change request.

danevans09:
Ok great! Thank you for letting me implement this.

joaander (approved):
Thanks!

PR #1124, opened 2021-08-31
Fix rigid body bugs

b-butler:
## Description

Fixes multiple bugs with `hoomd.md.constrain.Rigid.create_bodies`, and fixes a
bug in setting the rigid property of `hoomd.md.integrate._DynamicIntegrator` to
None when attached.

## Motivation and context

Multiple people have found this bug(s), and it needed to be fixed.

## How has this been tested?

New tests have been added to ensure correct behavior.

## Change log

```
Fixed: ``hoomd.md.constrain.Rigid.create_bodies`` method.
Fixed: Setting rigid attribute of a MD integrator to ``None``.
```

## Checklist:


syjlee:
I just compiled this branch to test my script which didn't work before, not it works. Rigid body is created with valid rigid_body_id on the central particles. 
@b-butler Did you mention that you're going to change the name `rigid_body_id` to something else?

tommy-waltmann:
Were there formal issues that were created for the bugs this PR fixes? I don't see any linked in the PR description.

b-butler:
@syjlee yes, I am this is a separate issue where `hoomd.Snapshot` objects have a different API than `hoomd.data.LocalSnapshot` objects. The name will be changed to `body` to match the base snapshot API.

b-butler:
@tommy-waltmann There is no formal issue, just discussions in person and over slack about issues in using the rigid body constraint.

hoomd/md/pytest/test_rigid.py
```diff
@@ -136,6 +136,45 @@ def test_create_bodies(simulation_factory, two_particle_snapshot_factory,
...
+    sim.operations.integrator.rigid = None
+
+    initial_snapshot = lattice_snapshot_factory(n=10)
+    if initial_snapshot.communicator.rank == 0:
+        initial_snapshot.particles.types = ["C", "A", "B"]
+        # Grab the middle particles and a random one to ensure that particle
+        # type ordering with respect to particle tag does not matter for
+        # create_bodies.
+        initial_snapshot.particles.typeid[100:800] = 1
+        initial_snapshot.particles.typeid[55] = 1
```
tommy-waltmann, line 154: Could we choose this index (55) and range (100-800) to come from `np.random.randint`? That way each time the test was run we would get a different random index and range.

tommy-waltmann (changes requested):
Without a formal issue describing the problems this PR is fixing, I can't give a very detail oriented review, but I still had one thing to say that could improve the added testing.

hoomd/md/pytest/test_rigid.py
joaander: `conftest.py` sets the numpy random seed so that tests run with reproducible random numbers. Executing tests with different random seeds each time makes failures very challenging to debug.

tommy-waltmann: I guess this approach would only work if test output tells us which random numbers were chosen if the test fails.

hoomd/md/ForceComposite.cc
```diff
@@ -589,20 +591,19 @@ void ForceComposite::createRigidBodies()
...
                     = m_body_diameter[body_type][current_body_index];
-                snap.pos[constituent_particle_tag] = snap.pos[central_particle_tag];
+                snap.pos[constituent_particle_tag] = snap.pos[particle_tag];
 
                 // Since the central particle tags here will be [0, n_central_particles), we know
                 // that the molecule number will be the same as the central particle tag.
-                molecule_tag[constituent_particle_tag] = central_particle_tag;
+                molecule_tag[constituent_particle_tag] = particle_tag;
 
                 ++constituent_particle_tag;
```
joaander, line 605: @tommy-waltmann The issue was that `create_bodies` was setting the body tag of the particles incorrectly which led to rigid body validation errors at the start of the simulation. These changes here fix it to set `body[i] = i` for those particles that are the central particles of rigid bodies.

joaander:
This pull request is passing and ready for final review.

tommy-waltmann (approved):

PR #1127, opened 2021-09-01
Remove deprecated functionality

joaander:
## Description

Remove deprecated functionality and bundle the latest tutorials that show the new APIs. Also fix some unrelated warnings generated during the unit tests and change the github actions jobs so that PRs show a warning when validation tests don't run.

## Motivation and context

* In the beta release cycle, we remove deprecated APIs in subsequent beta releases.
* The warnings indicate that the behavior of the test may be incorrect without well defined type sizes.
* Some developers are confused by the failing "Unit test" job. Unfortunately, GitHub considers "skipped" jobs as passing required status checks. The best I can do to improve this situation is to add a helpful warning.

## How has this been tested?

I ran pytest locally.

## Change log

```
Removed:

- ``Snapshot.exists`` - use ``snapshot.communicator.rank == 0``
- ``State.snapshot`` - use ``get_snapshot`` and ``set_snapshot``
-   The ``State.box`` property setter - use ``State.set_box``
```

## Checklist:

### DIFF
--- a/hoomd/pytest/test_local_snapshot.py
+++ b/hoomd/pytest/test_local_snapshot.py
@@ -68,7 +68,7 @@ _particle_data = dict(
               new_value=np.linspace(1, 20, Np, dtype=np.uint32),
               shape=(Np,)),
     # typeid is a signed integer in C++ despite always being nonnegative
-    typeid=dict(np_type=np.signedinteger,
+    typeid=dict(np_type=np.int32,
                 value=[0, 0, 0, 1, 1],
                 new_value=[1, 1, 1, 0, 0],
                 shape=(Np,)),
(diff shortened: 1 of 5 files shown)


hoomd/pytest/test_local_snapshot.py
```diff
@@ -84,13 +84,13 @@
                   value=[5, 2, 3, 2, 5],
                   new_value=[2, 1, 0.5, 1, 2],
                   shape=(Np,)),
-    image=dict(np_type=np.signedinteger,
+    image=dict(np_type=np.int32,
```
b-butler, line 87: Is this raising warnnings on CI? I don't get any warnings locally which is strange.

b-butler (approved):
I am surprised at the warnings generated by NumPy with `np.signedinteger` and such. It doesn't do that locally, but there is nothing wrong with the change, so it all looks good to me.

hoomd/pytest/test_local_snapshot.py
joaander: I get the warnings locally on an Apple M1 MacOs system with numpy 1.20.1. I do not get them on the Linux systems I develop on. I do not know why.

The warning is:
```
lib/python3.9/site-packages/numpy/core/function_base.py:175: DeprecationWarning: Converting `np.integer` or `np.signedinteger` to a dtype is deprecated. The current result is `np.dtype(np.int_)` which is not strictly correct. Note that the result depends on the system. To ensure stable results use may want to use `np.int64` or `np.int32`.
    return y.astype(dtype, copy=False)
```

PR #1129, opened 2021-09-06
Fix hoomd.Operations.__len__

### LINKED ISSUE #1128: Maximum recursion depth when taking the length of an Operations instance
## Script

```python
import hoomd
operations = hoomd.Operations()
len(operations)
```

## Output

```
  File "/Users/joaander/build/hoomd/hoomd/operations.py", line 238, in __len__
    return len(list(self))
  File "/Users/joaander/build/hoomd/hoomd/operations.py", line 238, in __len__
    return len(list(self))
  File "/Users/joaander/build/hoomd/hoomd/operations.py", line 238, in __len__
    return len(list(self))
  [Previous line repeated 995 more times]
RecursionError: maximum recursion depth exceeded
```

## Expected output

No output.

## Configuration

Platform:
- Mac

Installation method:
- Compiled from source

Versions:
- Python version: 3.9.6
- **HOOMD-blue** version: AttributeSetError - likely present on master as well

## Developer

@b-butler can you fix this?

b-butler:
## Description

Fixes the behavior of `hoomd.Operations.__len__`.

## Motivation and context

Resolves #1128

## How has this been tested?

Tests have been added for `__len__` and `__iter__`.

## Change log

```
Fixed: Calling `hoomd.Operations.__len__` no longer raises a `RecursionError`.
```

## Checklist:

### DIFF
--- a/hoomd/operations.py
+++ b/hoomd/operations.py
@@ -235,7 +235,8 @@ class Operations(Collection):
 
     def __len__(self):
         """Return the number of operations contained in this collection."""
-        return len(list(self))
+        base_len = len(self._writers) + len(self._updaters) + len(self._tuners)
+        return base_len + (1 if self._integrator is not None else 0)
 
     @property
     def integrator(self):
--- a/hoomd/pytest/CMakeLists.txt
+++ b/hoomd/pytest/CMakeLists.txt
@@ -25,6 +25,7 @@ set(files __init__.py
           test_table.py
           test_variant.py
           test_sorter.py
+          test_operations.py
     )
 
 install(FILES ${files}
--- a/hoomd/pytest/test_operations.py
+++ b/hoomd/pytest/test_operations.py
@@ -1,0 +1,38 @@
+import hoomd
+
+
+class FakeIntegrator(hoomd.integrate.BaseIntegrator):
+    pass
+
+
+def test_len():
+    operations = hoomd.Operations()
+    # ParticleSorter automatically added
+    assert len(operations) == 1
+    operations.tuners.clear()
+    assert len(operations) == 0
+
+    operations.integrator = FakeIntegrator()
+    operations.updaters.append(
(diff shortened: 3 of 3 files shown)


hoomd/operations.py
```diff
@@ -230,12 +230,12 @@ def __contains__(self, operation):
     def __iter__(self):
         """Iterates through all contained operations."""
         integrator = (self._integrator,) if self._integrator else []
-        yield from chain(self._tuners, self._updaters, integrator,
-                         self._writers, self._computes)
```
joaander, line 234: Is this change important to fixing the issue? I liked the previous ordering as it follows the same order the operations are applied during `run` (except computes which may be triggered at an time during the step)

hoomd/operations.py
```diff
@@ -230,12 +230,12 @@ def __contains__(self, operation):
...
+        yield from chain(integrator, self._updaters, self._writers,
+                         self._tuners, self._computes)
 
     def __len__(self):
         """Return the number of operations contained in this collection."""
-        return len(list(self))
+        return len([None for _ in self])
```
joaander, line 238: Why not add the lengths of the contained lists (and add 1 if there is an integrator)?

joaander (commented):
Thanks for the fix. I have some questions.

tcmoore3 (commented):
Looks good after you address Josh's questions.

hoomd/operations.py
b-butler: I didn't do this to keep the code somewhat simpler, (utilizing the existing `__iter__`), but there would be nothing wrong with adding list lengths.

hoomd/operations.py
b-butler: No it isn't, and that is good reason to revert.

joaander (approved):
Thanks!

PR #1141, opened 2021-09-20
Revise units documentation

### LINKED ISSUE #1112: Expand documentation on unit systems
## Description

The HOOMD units documentation does an adequate job of describing the self-consistent units in HOOMD. https://hoomd-blue.readthedocs.io/en/latest/units.html This was improved in #801 / #1019, and now most (all?) quantities have units.

I think it would be useful to include additional reference information by adding some set of the following links. I would say the central purpose is to emphasize that most packages choose between a small number of unit _systems_ -- for the most part, users of MD codes don't choose their energy, length, and mass units arbitrarily. There are two systems that are the most popular, named AKMA units (angstrom, kcal/mol, amu) and MD units (nanometer, kJ/mol, amu).

## Proposed solution

Parmed's docs have a few nice tables that would benefit users: https://parmed.github.io/ParmEd/html/dimensional_analysis.html

We could adapt something like the tables in that page for HOOMD. Maybe the empty cells could contain approximations of each derived unit in SI or some other expression, which I computed for a few cases just to check that some atomistic simulation results seemed physically reasonable.

The table below shows AKMA and MD units, along with their derived units according to HOOMD's unit system. I hope this is a handy list of conversion factors that serves most users' needs, especially with common units that have tricky calculations, like time and charge.

| Unit                 | AKMA                                        | MD        |
|----------------------|---------------------------------------------|-----------|
| *energy* (base unit) | kcal/mol                                    | kJ/mol    |
| *length* (base unit) | angstrom                                    | nanometer |
| *mass*               | amu                                         | amu       |
| time                 | 48.888 fs                                  | 1 ps |
| force                |                                             |           |
| pressure             |                                             |           |
| charge               | 8.7922 x 10^-21 coulomb; 0.054877 proton charge | 1.3593 x 10^-20 coulomb; 0.084838 proton charge |
| velocity             | 0.020455 angstrom/fs                        | 1 nm/ps |
| volume               | angstrom^3                                  | nm^3      |
| area                 | angstrom^2                                  | nm^2      |

Note that other packages don't always adhere to the same conventions about the relationships between derived units. For example, HOOMD units declare that time is `sqrt(mass * length**2  / energy)`. However, some packages use an arbitrary unit like picoseconds or "AKMA time" which is ~20.455 picoseconds. Meanwhile, the AKMA time needed for compatibility with HOOMD derived units is ~48.888 fs. (Source: WolframAlpha query: [sqrt(1 amu * 1 angstrom * 1 angstrom / (1 kcal / avogadro's number))](https://www.wolframalpha.com/input/?i=sqrt%281+amu+*+1+angstrom+*+1+angstrom+%2F+%281+kcal+%2F+avogadro%27s+number%29%29&assumption=%22UnitClash%22+-%3E+%7B%22amu%22%2C+%7B%22AtomicMassUnit%22%7D%7D&assumption=%22UnitClash%22+-%3E+%7B%22kcal%22%2C+%7B%22ThermochemicalKilocalories%22%7D%7D))

Sources for other values:
- [1 ps = sqrt(1 amu * 1 nm * 1 nm / (1 kJ/avogadro's number))](https://www.wolframalpha.com/input/?i=sqrt%281+amu+*+1+nm+*+1+nm+%2F+%281+kJ%2Favogadro%27s+number%29%29)
- [8.7922 x 10^-21 coulomb = sqrt(4 * pi * permittivity of free space * 1 (kcal/avogadro's number) * 1 angstrom)](https://www.wolframalpha.com/input/?i=sqrt%284+*+pi+*+permittivity+of+free+space+*+1+%28kcal%2Favogadro%27s+number%29+*+1+angstrom%29)
- [0.054877 = sqrt(4 * pi * permittivity of free space * 1 (kcal/avogadro's number) * 1 angstrom) / proton charge](https://www.wolframalpha.com/input/?i=sqrt%284+*+pi+*+permittivity+of+free+space+*+1+%28kcal%2Favogadro%27s+number%29+*+1+angstrom%29+%2F+proton+charge)
- [1.3593 x 10^-20 coulomb = sqrt(4 * pi * permittivity of free space * 1 nm * 1kJ/avogadro's number)](https://www.wolframalpha.com/input/?i=sqrt%284+*+pi+*+permittivity+of+free+space+*+1+nm+*+1kJ%2Favogadro%27s+number%29)
- [0.084838 = sqrt(4 * pi * permittivity of free space * 1 nm * 1kJ/avogadro's number) / proton charge](https://www.wolframalpha.com/input/?i=sqrt%284+*+pi+*+permittivity+of+free+space+*+1+nm+*+1kJ%2Favogadro%27s+number%29+%2F+proton+charge)
- [1 nm/ps = sqrt(1kJ/avogadro's number / 1 amu)](https://www.wolframalpha.com/input/?i=sqrt%281kJ%2Favogadro%27s+number+%2F+1+amu%29+in+nm+%2Fps)

Some of this information could be directly copy-pasted into the docs.

As suggested previously in #801, we could also add copy-pasteable snippets that use `unyt` or another Python library for converting units. That might even be more helpful than a table with the conversion factors.

## Additional context 

OpenMM units: http://docs.openmm.org/latest/userguide/theory.html#units
LAMMPS units: https://docs.lammps.org/units.html
GROMACS units: https://manual.gromacs.org/documentation/2019/reference-manual/definitions.html#md-units

joaander:
## Description

* Make minor revisions to the units docs for consistency
* Add two commonly used systems of units to the documentation.

## Motivation and context

To teach users how to perform dimensional analysis.

Resolves #1112 

## How has this been tested?

I viewed the Sphinx documentation locally.

## Change log

```
- Added AKMA and MD units to the documentation.
```

## Checklist:

### DIFF
--- a/sphinx-doc/units.rst
+++ b/sphinx-doc/units.rst
@@ -5,7 +5,7 @@ HOOMD-blue does not adopt a particular system of units, nor does it offer a vari
 to choose from. Instead, it follows a self-consistent system of units where all derived units
 (e.g. force) are defined in terms of base units (e.g. energy / length). To adopt a system of units
 for your simulations, choose a set of base units (e.g. meters versus centimeters for length), and
-then determine what the derived unis are.
+then determine what the derived units are.
 
 .. note::
 
--- a/sphinx-doc/units.rst
+++ b/sphinx-doc/units.rst
@@ -1,20 +1,27 @@
 Units
 +++++
 
-HOOMD-blue does not adopt any particular system of units, but instead follows a self-consistent system of units. Users are free to define HOOMD-blue's base units (e.g. meters versus centimeters for length), and units of all derived quantities can be determined from the base units. For instance if the base units are chosen to be 1 meter, 1 Joule, and 1 kilogram, then velocity is in meters per second, but if one chooses nanometers for length, kilojoules per mole for energy, and 1 amu for mass then velocity is in nanometers per picosecond.
+HOOMD-blue does not adopt a particular system of units, nor does it offer a variety of systems
+to choose from. Instead, it follows a self-consistent system of units where all derived units
+(e.g. force) are defined in terms of base units (e.g. energy / length). To adopt a system of units
+for your simulations, choose a set of base units (e.g. meters versus centimeters for length), and
+then determine what the derived unis are.
 
-Note:
+.. note::
+
+    Most of the units on this page apply to MD simulations.
+
     In HPMC, the primary unit is that of length. Mass is factored out of the partition function and
     does not enter into the simulation. In addition, the scale of energy is irrelevant in athermal
     HPMC systems where overlapping energies are infinite and valid configurations have
     zero potential energy. However, energy does appear implicitly in derived units like
     :math:`[\mathrm{pressure}] = \left(\frac{\mathrm{[energy]}}{\mathrm{[length]}^3}\right)`.  In
-    HPMC, :math:`kT` is assumed to be 1 unit of energy.
+    HPMC, :math:`kT` is assumed to be 1 :math:`\mathrm{energy}`.
 
 Base Units
 ==========
 
-Base units in HOOMD-blue are:
+The base units are:
 
 - energy
 - length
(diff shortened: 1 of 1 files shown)


j-proc (commented):
Too late for approval, but looking at the live version and have to say awesome changes. The clear separation of the base and derived units is a definite benefit. I also like the conversion factors for common unit systems. 

My only other feedback is that I do wonder if we should note anywhere the typical simulation convention of keeping most values close to 1. I made that mistake early on in my work where I worked in actual nanometers and quickly got undesirable behavior due to now obviously outrageously sized boxes. I know there have been changes to neighbor-lists since then. So this may not be the case anymore.

joaander:
@j-proc sorry, I merged this too soon trying to move progress on v3 forward.

You have a good point about making particle diameters close to 1. However, I think we can fix that issue by making some other changes to HOOMD rather than asking users to choose unit systems appropriately. I've noticed some users on the mailing list and elsewhere really struggle with the concept that they can choose any value they want for their units - they really just want to use a provided system of units that they are familiar with (which is why I am now providing the two standard examples). 

99+% of the math in HOOMD is scale invariant. Even if you use SI units for a nanoscale system (or use nanometers for a macro scale simulation), the floating point math will just carry those small (or large) exponents through the calculation. There are only a few places I can think of that would cause issues, and those are the default values for some parameters that have units, like the neighbor list buffer (default 0.4) and the tolerances in the FIRE energy minimizer. There are also a few absolute tolerances in the C++ code, one checks when moments of inertia are 0 and another is used in the HPMC shape overlap.

We could solve those issues by removing the default values and forcing users to pick values. I think that is a reasonable solution for FIRE's tolerances, but I wouldn't want to burden all users by making them choose a default nlist buffer. Unfortunately, we don't have a good way to make default parameters (buffer) depend on the value of other unknown parameters (like r_cut). The best idea I have right now is to change `buffer` to `buffer_fraction` and make it always relative to the largest `r_cut` in the system. I'll open an issue to note this down.

PR #1144, opened 2021-09-21
Rename pivot_move_ratio in the Cluster updater

### LINKED ISSUE #777: Rename move_ratio in HPMC

Charlottez112:
## Motivation and context

The `pivot_move_ratio` in `hpmc.update.clusters` is actually the fraction of pivot moves. See issue #777 for more details on the motivation.

Currently this PR only fixes the names in the cluster updater in the python file. The names in the wall updater and the muvt updater are not changed since they haven't been implemented for version 3.

Resolves #777

## Checklist:

### DIFF
--- a/hoomd/hpmc/UpdaterClusters.h
+++ b/hoomd/hpmc/UpdaterClusters.h
@@ -1778,7 +1778,7 @@ template < class Shape> void export_UpdaterClusters(pybind11::module& m, const s
           .def( pybind11::init< std::shared_ptr<SystemDefinition>,
                          std::shared_ptr< IntegratorHPMCMono<Shape> > >())
         .def("getCounters", &UpdaterClusters<Shape>::getCounters)
-        .def_property("pivot_move_ratio", &UpdaterClusters<Shape>::getMoveRatio, &UpdaterClusters<Shape>::setMoveRatio)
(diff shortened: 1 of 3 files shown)


tommy-waltmann (changes requested):
Good work so far on this, there are a couple of things that I'd like to see happen before we can merge this:

1) The property `pivot_move_ratio` exported by `pybind11` in the `UpdaterClusters.h` file should have its name changed to `pivot_move_probability` to match the name change on the python side. That should fix the tests that are failing.

2) We should rename the cpp member variable `m_move_ratio` to `m_move_probability` or `m_pivot_move_probability` to match what the meaning of the variable actually is.

tommy-waltmann (approved):
Thanks for your work on this @Charlottez112 !

PR #1146, opened 2021-09-22
Fix pair.Table on the GPU

### LINKED ISSUE #1105: Issues running table potential (v3.0.0beta008) on GPU
## Description

While running a membrane simulation with tabulated potential for a cosine/squared potential. The script will run locally on CPU, but on a GPU node the process exits with an Error computing cell list (particle no longer in simulation box).

This system runs to completion with a LJ-potential in stead of a table potential on GPU (albeit not with 'chemical' stability). I run into the same GPU issues with the cosine/squared as a table potential on the molecular dynamics tutorial.

Another issue I have run into is when formulating a zeropotential for the table (V=F=np.array(n*[0])
) a cudaErrorIllegalAddress occurs on GPU. This too will run on CPU. This zero potential is quite important when using multiple types of potentials between pairs. 

## Script

Included files at [groups](https://groups.google.com/g/hoomd-users/c/AhuQh75dKXU):

    * init.gsd (initial coordinates)
    * test.py (hoomd script)
    * exectest.sh (gpu batch script)
    * nvidia-smi.txt (output nvidia-smi)
    * errorCellList.txt (error message 1)
    * cudaErrorIllegalAddress.txt (error message 2)

## Output

In included files.

## Expected output

Stable simulation run. Can provide a trajectory ran on CPU if necessary. 

## Configuration
OS: Debian 10
GPU: GTX1080ti
HOOMD version: 3.0.0dev8 
Build: conda gpu_py39hc6c7ca5_0
Python version: 3.9.6

joaander:
## Description

* Fix a race condition in PotentialPairGPU.
* Check for invalid user input in pair.Table that could lead to invalid memory accesses.
* Inform users how to define non-interacting terms in pair.Table.

## Motivation and context

The race condition caused random errors when using pair.Table - even when the tables were populated with 0's (#1105). The other changes proactively fix issues that could lead to invalid memory accesses when users provide invalid input to the table.

Resolves #1105 

## How has this been tested?

I tested the changes locally with the scripts provided in #1105. The automated unit tests for pair.Table should be affected by this bug, but for some reason they don't trigger it.

## Change log

```
Fixed:

* Race condition that lead to incorrect simulations with ``md.pair.Table``.
```

## Checklist:

### DIFF
--- a/hoomd/md/EvaluatorPairTable.h
+++ b/hoomd/md/EvaluatorPairTable.h
@@ -77,7 +77,12 @@ class EvaluatorPairTable
 
             if (V_py.size() != F_py.size())
                 {
-                throw std::runtime_error("The length of V and F arrays must be equal");
+                throw std::runtime_error("The length of V and F arrays must be equal.");
+                }
+
+            if (V_py.size() == 0)
+                {
+                throw std::runtime_error("The length of V and F must not be zero.");
                 }
 
             size_t width = V_py.size();
--- a/hoomd/md/PotentialPairGPU.cuh
+++ b/hoomd/md/PotentialPairGPU.cuh
@@ -189,6 +189,8 @@ gpu_compute_pair_forces_shared_kernel(Scalar4* d_force,
     // initialize extra shared mem
     auto s_extra = reinterpret_cast<char*>(s_ronsq + num_typ_parameters);
 
+    __syncthreads();
+
     unsigned int available_bytes = max_extra_bytes;
     for (unsigned int cur_pair = 0; cur_pair < num_typ_parameters; ++cur_pair)
         s_params[cur_pair].load_shared(s_extra, available_bytes);
--- a/hoomd/md/pair/pair.py
+++ b/hoomd/md/pair/pair.py
@@ -591,8 +591,8 @@ class Table(Pair):
         \\begin{eqnarray*}
         \\vec{F}(\\vec{r}) = & 0; & r < r_{\\mathrm{min}} \\\\
                            = & F(r)\\hat{r};
-                             & r_{\\mathrm{min}} \\le r < r_{\\mathrm{max}} \\\\
-                           = & 0; & r \\ge r_{\\mathrm{max}} \\\\
+                             & r_{\\mathrm{min}} \\le r < r_{\\mathrm{cut}} \\\\
+                           = & 0; & r \\ge r_{\\mathrm{cut}} \\\\
         \\end{eqnarray*}
 
     and the potential :math:`V(r)` is:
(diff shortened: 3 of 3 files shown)


joaander:
@YouriRan can you confirm that this branch fixes the behavior you are seeing?

joaander:
This is a small pull request that fixes a critical bug in `md.pair.Table`. Reviewers, could you please take a look and submit your review.

Charlottez112 (approved):
Thank you for fixing this! I do have a comment/question related to the `Table` potential though: I noticed that in the documentation we have `default_r_cut`, `default_r_on`, `r_min`, `r_cut` and `r_max`, and from what I understand, `r_max` is just `r_cut`, and `r_cut` , `r_min` are equivalent to `default_r_cut`, `default_r_on`? This is a little confusing to me. Not sure if it's just me or we could consider renaming some of them.

joaander:
> Thank you for fixing this! I do have a comment/question related to the `Table` potential though: I noticed that in the documentation we have `default_r_cut`, `default_r_on`, `r_min`, `r_cut` and `r_max`, and from what I understand, `r_max` is just `r_cut`, and `r_cut` , `r_min` are equivalent to `default_r_cut`, `default_r_on`? This is a little confusing to me. Not sure if it's just me or we could consider renaming some of them.

Thanks for checking the docs. I will fix this. You are correct that `r_max` should be replaced with `r_cut`. `r_min` is not related to `r_on`, `r_min` is a parameter specific to the table potential and is defined as a parameter in `params`.

b-butler (commented):
Is the reason, we couldn't we just add the check for a zero sized array in the conditional for `evalForceAndEnergy`? Or is there a reason you prefer having a size requirement.

joaander:
> Is the reason, we couldn't we just add the check for a zero sized array in the conditional for `evalForceAndEnergy`? Or is there a reason you prefer having a size requirement.

The table potential implicitly defines `V(r_cut) = 0` and `F(r_cut)=0`. Using linear interpolation for arbitrary r, the potential and force are only well defined between `r_min` and `r_cut` when the user provides at least one point at `r_min`.

b-butler (approved):
Ah, it a conceptual notation that a potential only at zero is still a potential and needs a value for the energy and force. :+1:

joaander:
> Ah, it a conceptual notation that a potential only at zero is still a potential and needs a value for the energy and force. 👍

Yes, we could change the code to implement what you suggested but they we also need to document the conceptual special case. I think its fine the way it is now.

joaander:
@YouriRan I didn't hear from you so I assume that this fixes the incorrect behavior you were getting.

YouriRan:
Excuse my late response. Just tested this and it works as expected. Many thanks!

Issue #1074, opened 2021-07-27
Update pre-commit configuration


joaander:
## Description

- [x] Fix license headers
- [x] Remove `Maintainer` lines
- [x] Any additional pre-commit fixers/checkers that may be useful

## Motivation and context

GSD and Fresnel are already using pre-commit to fix license headers.

joaander:
Additional fixers to consider:
* https://github.com/asottile/add-trailing-comma
* https://github.com/pre-commit/pygrep-hooks - https://github.com/pre-commit/pygrep-hooks
* github.com/PyCQA/isort

PR #1151, opened 2021-10-04
Custom Force for Version 3

### LINKED ISSUE #866: Port force.constant to the v3 API
## Description

Port force.constant to the v3 API.

## Motivation and context

We should probably retool the interface for force.constant in the process. Also consider removing the unused IMD updater at the same time to remove restrictions on how force.constant needs to behave.

tommy-waltmann:
## Description

This PR converts what was the constant force for v2 into a customizable force in v3. The custom force will allow users to get direct access to the force data buffers in python.

## Motivation and context

We have to port v2 APIs over to v3, but the v2 API was a little to restrictive. Making the force customizable will allow for more flexibility with minimal performance cost.

Resolves #866 

## How has this been tested?

Unit tests are written in `test_custom_force.py`

## Change log

```
Added customizable forces in python.

Changes made in local data context managers now persist after the context manager exits.

```

## Checklist:


tommy-waltmann:
We've talked about making properties gettable but not settable to ensure users use the `arrays.force[:]` syntax instead of just `arrays.force`. As I look through the code to add that, I see what looks like may have been an attempt to do this already, at line 74 in `local_access.py`. @b-butler Is this what you're intent was when you originally wrote that code?

b-butler:
@tommy-waltmann this was to enable `snap.positions = array` to work by  just setting the data in `snap.positions` to that in `array` exactly like `snap.positions[:] = array`.

tommy-waltmann:
> @tommy-waltmann this was to enable `snap.positions = array` to work by just setting the data in `snap.positions` to that in `array` exactly like `snap.positions[:] = array`.

The way the `__setattr__` is written in `_LocalAccess` does not work as you intend it to. When I attempt to set values in the properties of the local snapshots, those changes are not reflected in cpp once the code exits the context manager. I have explicitly confirmed this with tests.

The local snapshots do not work properly for the same reason we ran into before with the `snap.particles.position` syntax vs the `snap.particles.position[:]` syntax. I looked and there were no tests written for the local snapshots to make sure changes persist after leaving the context manager.

This is a significant bug, and whatever changes are made to fix it should be written so they integrate nicely with both the local snapshot and local force access APIs.

tommy-waltmann:
> > @tommy-waltmann this was to enable `snap.positions = array` to work by just setting the data in `snap.positions` to that in `array` exactly like `snap.positions[:] = array`.
> 
> The way the `__setattr__` is written in `_LocalAccess` does not work as you intend it to. When I attempt to set values in the properties of the local snapshots, those changes are not reflected in cpp once the code exits the context manager. I have explicitly confirmed this with tests.
> 
> The local snapshots do not work properly for the same reason we ran into before with the `snap.particles.position` syntax vs the `snap.particles.position[:]` syntax. I looked and there were no tests written for the local snapshots to make sure changes persist after leaving the context manager.
> 
> This is a significant bug, and whatever changes are made to fix it should be written so they integrate nicely with both the local snapshot and local force access APIs.

Update on this: @b-butler and I had a nice chat and a bug fix is on this branch. We may want to update the tests for the local snapshot API to confirm changes persist after exiting the context manager.

hoomd/ForceCompute.cc
```diff
@@ -625,7 +625,7 @@ Scalar ForceCompute::getVirial(unsigned int tag, unsigned int component)
     if (found)
         {
         ArrayHandle<Scalar> h_virial(m_virial, access_location::host, access_mode::read);
-        result = h_virial.data[m_virial_pitch * component + i];
+        result = h_virial.data[m_virial_pitch * i + component];
```
joaander, line 628: Document this and any other bugfixes in the proposed change log entry.

tommy-waltmann:
For some reason, there are appear to be tests which involve the const force compute, which we are removing in this PR, in the harmonic bond force ctest file. @joaander Can the old tests be removed in this PR, or is there some reason to keep those around?

hoomd/md/force.py
```diff
@@ -117,212 +118,144 @@ def additional_virial(self):
...
+                arrays.virial = ...
+        """
+        if self._in_context_manager:
+            raise RuntimeError("Cannot enter cpu_local_force_arrays context "
+                               "manager inside another local_force_arrays "
+                               "context manager")
+        return hoomd.data.ForceLocalAccess(self)
+
+    @property
+    def gpu_local_force_arrays(self):
```
j-proc, line 221: Just for my own clarification, although it might be beneficial to users, with the internals of HOOMD, integrators won't touch the GPU so there's no need to set this unless you're doing some other computes with a GPU implementation correct?

hoomd/md/force.py
```diff
@@ -117,212 +118,144 @@ def additional_virial(self):
...
+        return hoomd.data.ForceLocalAccessGPU(self)
+
+    @abstractmethod
+    def set_forces(self, timestep):
+        """Set the forces in the simulation loop.
 
-    # there are no coeffs to update in the constant force compute
-    def update_coeffs(self):  # noqa - will be ported to v3
+        Args:
+            timestep (int): The current timestep in the simulation.
```
j-proc, line 257: This seems like an odd parameter for something that is explicitly set or triggered. Might be helpful to clarify how it's used... which on my own looking up is:
To allow for easy access of the timestep for custom forces which are temporal in nature. Is this a significant enough usecase that we would discourage the perhaps uglier way of the users custom force class keeping a reference to the simulation object and getting it from there?

This is also making me think about efficiency if the force is constant. I'm just wanting to make sure I'm thinking about it correctly. Because the access is O(N(1)) there wouldn't be any benefit from caching these values if they were constant (the simpler old code that we're replacing). It's just a matter of if that those arrays would be kept around internally vs in a python object?

I think in general this is maybe pointing out to me that this might need some documentation about why this isn't based on custom.Action.

hoomd/md/pytest/test_custom_force.py
```diff
@@ -0,0 +1,323 @@
...
+    def __init__(self):
+        super().__init__()
+
+    def set_forces(self, timestep):
+        with self.gpu_local_force_arrays as arrays:
+            arrays.force = -5
+            arrays.potential_energy = 37
+            arrays.torque = 23
+            # cupy won't let me use two indices on arrays of length 0
+            if len(arrays.virial) > 0:
```
j-proc, line 39: Maybe an assert would make sense here? Since we apparently need cupy to run this test anyways and there's not another case I can think of that would trigger it.

hoomd/md/pytest/test_custom_force.py
```diff
@@ -0,0 +1,323 @@
...
+def _try_running_sim(sim, tsteps):
+    """Run the sim while checking that the error is raised."""
+    should_error = _using_gpu_with_cpu_device(
+        sim.operations.integrator.forces[0].__class__, sim)
+    if should_error:
+        with pytest.raises(RuntimeError):
+            sim.run(tsteps)
+    else:
+        sim.run(tsteps)
+    return should_error
```
j-proc, line 67: If I follow this logic, this is essentially allowing us to later filter out errors which just come from accessing GPU classes on the CPU. It's mainly just for the convenience to write the test once and properly ignore these errors?

hoomd/md/pytest/test_custom_force.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def __init__(self):
+        super().__init__()
+
+    def set_forces(self, timestep):
+        with self.gpu_local_force_arrays as arrays:
+            self._check_buffer_lengths(arrays)
+
+
+@pytest.mark.parametrize("force_cls",
```
j-proc, line 298: I could use a description here because I'm not as familiar with the internals for Ghost particles and MPI. I think it might help other users too when they at correctly accessing these values for MPI runs. I think the main point here is that we don't need to have any portion protected by rank==1, correct?

hoomd/data/__init__.py
```diff
@@ -4,6 +4,6 @@
 from .local_access import (AngleLocalAccessBase, BondLocalAccessBase,
                            ConstraintLocalAccessBase, DihedralLocalAccessBase,
                            ImproperLocalAccessBase, PairLocalAccessBase,
-                           ParticleLocalAccessBase)
-from .local_access_cpu import LocalSnapshot
-from .local_access_gpu import LocalSnapshotGPU
+                           ParticleLocalAccessBase, ForceLocalAccessBase)
+from .local_access_cpu import LocalSnapshot, ForceLocalAccess
+from .local_access_gpu import LocalSnapshotGPU, ForceLocalAccessGPU
```
j-proc, line 9: Do these belong in the this base level even though force and virial are nonsensical for HPMC. I feel like the answer is probably yes because there's lots of particle data that this is the case for.

hoomd/data/local_access_gpu.py
```diff
@@ -5,10 +5,11 @@
 """Implement local access classes for the GPU."""
 
 from hoomd import _hoomd
+from hoomd.md import _md
```
j-proc, line 8: What if we build with MD?

hoomd/data/__init__.py
j-proc: On second thought reading below, I think this could be an issue. I'd say this is the only thing truly blocking approval from me. The other things are clarification

j-proc (changes requested):
@tommy-waltmann, I left quite a few comments that I could use clarification on before I approve and one potential failure point with imports(the only reason I'm requesting changes). 
I did also want to mention though how awesome I think this effort is because you've managed to enable an incredible level of customization and control while also keeping things pretty simple.

hoomd/md/force.py
tommy-waltmann: I'm not quite sure what you mean by "integrators won't touch the GPU"; the integration methods have GPU implementations in MD. This property isn't intended to be set. If your simulation uses other computes who need access to the data buffers we expose here, that will still work correctly because the ArrayHandles on the force data buffers are released as soon as the context manager exits.

hoomd/md/force.py
tommy-waltmann: The reason the timestep argument is here is because the `compute` method of classes that are derived from the `Compute` class in cpp all take the timestep as an argument, so I just forwarded that argument to the python method.

I don't think there is any reason to be more restrictive than what currently exists in the cpp side of the code. If a user doesn't want to use the timestep argument, they don't have to (in fact almost all computes operate this way).

One of the reasons behind adding the ability to write custom forces in python was so that people could prototype new potentials easily, and since we don't currently have any time-dependent potentials in cpp it seems reasonable to me that someone might want to prototype a time-dependent potential and see what happens.

The reason this isn't just a custom action is because the custom action API doesn't allow a user to change the forces on a set of particles. Using custom actions, you can access the system state via local snapshots, but you can't write your own force compute. That's where custom forces can help.

hoomd/md/pytest/test_custom_force.py
tommy-waltmann: In some of the tests with MPI, there are 0 particles on one of the MPI ranks. If we put an assert here, the test will fail on that rank.

hoomd/md/pytest/test_custom_force.py
tommy-waltmann: This helper function makes sure the simulation fails to run if we are trying to use the GPU custom force class with a CPU simulation device, and otherwise runs the simulation as it is supposed to be run. It is a check that has to be done every time we run a test, so putting all the logic in one helper function is cleaner than writing the same code over and over.

Others can comment if they think there is a cleaner way to write this, I know its somewhat awkward because we have to check the return value after every time we call `_try_running_sim` so the test can exit properly.

hoomd/data/local_access_gpu.py
tommy-waltmann: Do you mean what if we build WITHOUT md? That's a good point, and perhaps these class should sit in a `hoomd.md.data` namespace to avoid errors for hpmc only builds.

hoomd/data/__init__.py
tommy-waltmann: I'll move this to an md specific namespace, see my response to your next comment.

hoomd/md/pytest/test_custom_force.py
tommy-waltmann: The reason for this test is that we have to make sure when we access that `ghost_XXX` and `XXX_with_ghost` attributes of the custom force, they have the right data in them. This test uses an LJ force to make sure there is ghost data available and then just checks to make sure the lengths of the ghost buffers have the right number of particles in them.

I could add a small comment in the docstring for this test to make sure that's clearer. Teaching users how the ghost data is organized and best accessed with custom forces is best left for a tutorial.

hoomd/md/pytest/test_custom_force.py
j-proc: I think it might benefit from a more clear name. Especially because `try` is a python keyword. Perhaps something like `run_should_fail` because that's essentially what it returns. But this function is also actually doing the run as well. Will defer to others opinion.

hoomd/data/local_access_gpu.py
j-proc: Yes that's what I meant. I couldn't remember if the flag was ENABLE_, BUILD_, etc., thanks for understanding my meaning. Edited for future clarity. 

And @b-butler @joaander are the authorities on the right way to do this. So probably let them weigh in on the correct location for this code to live before making the move.

joaander: HOOMD must import and pass tests when `BUILD_MD=off`. Do not use `try: import ... except: pass` constructions as they mask legitimate errors.

If possible, move code that needs to import `_md` to the `md` component. If not, make the non MD parts of this code work when `_md` is not available. Use the bool `hoomd.version.md_built` to see whether you can import `md` and `_md`.

hoomd/md/pytest/test_custom_force.py
tommy-waltmann: I'll wait until @joaander has a pass over this PR before deciding how to move forward with this comment.

joaander: I agree, this is confusing. It is unclear what code path any given test run is supposed to take and what it is testing.

1) If you are attempting to test that running on the GPU with a CPU device raises an error, write one specific test that does just that.

2) If this is a construction to avoid invalid parameters in the matrix resulting from invalid combinations of two separate parameters, then use a different construction. For example, instead of parameterizing the test on the force class, select the force class based on the device. If that isn't possible, use skips - but skips should be used sparingly. I prefer when they are used only when the user's hardware or software configuration prevents a test from running.

hoomd/md/CustomForceCompute.cc
```diff
@@ -0,0 +1,59 @@
...
+    m_exec_conf->msg->notice(5) << "Destroying ConstForceCompute" << endl;
+    }
+
+/*! This function calls the python set_forces method.
+    \param timestep Current timestep
+*/
+void CustomForceCompute::computeForces(uint64_t timestep)
+    {
+    // execute python callback to update the forces, if present
+    if (m_callback && !m_callback.is(py::none()))
```
joaander, line 40:
```suggestion
    if (!m_callback.is(py::none()))
```
This is redundant - make one check or the other.

hoomd/md/CustomForceCompute.cc
```diff
@@ -0,0 +1,59 @@
...
+namespace hoomd
+    {
+namespace md
+    {
+/*! \param sysdef SystemDefinition containing the ParticleData to compute forces on
+ */
+CustomForceCompute::CustomForceCompute(std::shared_ptr<SystemDefinition> sysdef)
+    : ForceCompute(sysdef)
+    {
+    m_exec_conf->msg->notice(5) << "Constructing ConstForceCompute" << endl;
```
joaander, line 26: Set `m_callback` in the constructor.

hoomd/md/CustomForceCompute.h
```diff
@@ -0,0 +1,186 @@
...
+
+    const GlobalArray<Scalar>& getVirialArray() const
+        {
+        return m_virial;
+        }
+
+    const GlobalArray<Scalar4>& getTorqueArray() const
+        {
+        return m_torque;
+        }
```
joaander, line 59: These are defined in the base class. Use those instead of duplicating code.

hoomd/md/CustomForceCompute.h
```diff
@@ -0,0 +1,186 @@
...
+
+    unsigned int getNGhosts() const
+        {
+        return m_pdata->getNGhosts();
+        }
+
+    const GlobalArray<unsigned int>& getRTags() const
+        {
+        return m_pdata->getRTags();
+        }
```
joaander, line 74: Where are these used? Why do they need to be duplicated here instead of calling `m_pdata->method()`?

hoomd/md/CustomForceCompute.h
```diff
@@ -0,0 +1,186 @@
...
+    private:
+    //! A python callback when the force is updated
+    pybind11::object m_callback;
+    };
+
+/** Make the local particle data available to python via zero-copy access
+ *
+ * */
+template<class Output>
+class PYBIND11_EXPORT LocalForceComputeData : public LocalDataAccess<Output, CustomForceCompute>
```
joaander, line 89: Local force arrays should be accessible from the base class `Force` object.

hoomd/md/data/local_access_cpu.py
```diff
@@ -0,0 +1,34 @@
...
+        potential_energy ((N_particles,) `hoomd.data.array` of ``float``):
+            Local potential energy data. :math:`[\\mathrm{energy}]`
+        torque ((N_particles, 3) `hoomd.data.array` of ``float``):
+            Local torque data. :math:`[\\mathrm{force} \\cdot \\mathrm{length}]`
+        virial ((N_particles, 6) `hoomd.data.array` of ``float``):
+            Local virial data. :math:`[\\mathrm{energy}]`
+        rtag ((N_particles_global) `hoomd.data.array` of ``int``):
+            The reverse tag of a particle. This means that the value
+            ``force_data.rtag[0]`` represents the current index accessing data
+            for the particle with tag 0.
```
joaander, line 27:
```suggestion
            The reverse tag of a particle. The value
            ``i = force_data.rtag[t]`` is the memory index accessing data
            for the particle with tag ``t``.
```
And similarly for the GPU array class.

hoomd/md/force.py
```diff
@@ -117,212 +118,145 @@ def additional_virial(self):
...
+    instantiated directly. A custom force should be derived from this class, and
+    should override a method called `set_forces` in order to set the force data
+    however they wish. Custom forces have direct, zero-copy access to
+    HOOMD-Blue's underlying force buffers via either the
+    `cpu_local_force_arrays` or `gpu_local_force_arrays` property. Choose the
+    property that corresponds to the device you wish to alter the data on. In
+    addition to zero-copy access to force buffers, custom forces have access to
+    the local snapshot API via the ``_state.cpu_local_snapshot`` or the
+    ``_state.gpu_local_snapshot`` property. See the documentation in
+    ``hoomd.state`` for more information on the local snapshot API.
```
joaander, line 134:
```suggestion
    Derive a custom force class from `Custom`, and override the `set_forces` method to  compute forces on particles. Use have direct, zero-copy access to the C++ managed buffers via either the
    `cpu_local_force_arrays` or `gpu_local_force_arrays` property. Choose the
    property that corresponds to the device you wish to alter the data on. In
    addition to zero-copy access to force buffers, custom forces have access to
    the local snapshot API via the ``_state.cpu_local_snapshot`` or the
    ``_state.gpu_local_snapshot`` property. 
    
    See Also:
      See the documentation in `hoomd.State` for more information on the local snapshot API.
```
This will need to be rewrapped and check that the `See Also:` syntax is correct to create a see also info block in the generated docs.

hoomd/md/force.py
```diff
@@ -117,212 +118,145 @@ def additional_virial(self):
...
-    Note:
-        Forces are kept constant during the simulation. If a callback should
-        re-compute particle forces every time step, it needs to overwrite the
-        old forces of **all** particles with new values.
-
-    Note:
-        Per-particle forces take precedence over a particle group, which takes
-        precedence over constant forces for all particles.
+    When accessing the local force arrays, the force, potential energy, torque,
+    and virial of any particle can be set to any value.
```
joaander, line 137: Does this need to be said?

hoomd/md/force.py
```diff
@@ -117,212 +118,145 @@ def additional_virial(self):
...
-            global const
-            const.setForce(tag=1, fvec=(1.0*timestep,2.0*timestep,3.0*timestep))
-        const = force.constant(callback=updateForces)
-    """
+            def set_forces(self, timestep):
+                with self.cpu_local_force_arrays as arrays:
+                    arrays.force = -5
+                    arrays.torque = 3
+                    arrays.potential_energy = 27
+                    arrays.virial = np.arange(6)[None, :]
```
joaander, line 150:
```suggestion
                    arrays.force[:] = -5
                    arrays.torque[:] = 3
                    arrays.potential_energy[:] = 27
                    arrays.virial[:] = np.arange(6)[None, :]
```
And similarly for all the other examples.

hoomd/md/pytest/test_custom_force.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+class MyForceGPU(md.force.Custom):
+
+    def __init__(self):
+        super().__init__()
+
+    def set_forces(self, timestep):
+        with self.gpu_local_force_arrays as arrays:
+            arrays.force = -5
```
joaander, line 35: How does this work? I thought we needed the `[:]` syntax to write to the elements of the array?

joaander (changes requested):
Thanks! See my comments inline.

hoomd/md/CustomForceCompute.h
tommy-waltmann: Needed to write a version of the getters which are declared `const` and return `const &` in order to make the templating for the `LocalForceComputeData` class to work. I'll try moving the `LocalForceComputeData` to be templated on the `ForceCompute` class before deciding how to address this.

hoomd/md/CustomForceCompute.h
tommy-waltmann: This is more awkwardness to make the templating work. The template class needs to have these methods in order for the `LocalForceComputeData` to compile

b-butler (commented):
The local access implementation looks good generally. I have some other comments to address though.

hoomd/md/CustomForceCompute.h
tommy-waltmann: Exposing the rtags could be removed, as Brandon said, but I think the others need to stay so everything can compile.

hoomd/md/CustomForceCompute.h
tommy-waltmann: Why would we want the local force arrays to be read-only? Maybe I'm misunderstanding what you mean, but by "these" you do mean the force arrays, right?

hoomd/md/CustomForceCompute.h
joaander: Can we just have one method in the base class that is `const`? It should have been that way from the beginning.

hoomd/md/CustomForceCompute.h
joaander: Read only arrays would prevent users from overwriting forces computed by `pair.LJ` for example.

hoomd/md/CustomForceCompute.h
```diff
@@ -0,0 +1,186 @@
+// Copyright (c) 2009-2021 The Regents of the University of Michigan
+// This file is part of the HOOMD-blue project, released under the BSD 3-Clause License.
+
+// Maintainer: joaander
```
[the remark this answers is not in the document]
joaander: Yes, most devs copy and paste existing code. These maintainer lines are a relic from when HOOMD was maintained differently. I have a TODO noted in #1074 to write a pre-commit hook that checks for these and remove them.

hoomd/md/pytest/test_custom_force.py
joaander: Should we update the global snapshots to use the same shorthand?

hoomd/md/CustomForceCompute.h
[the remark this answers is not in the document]
b-butler: Okay

hoomd/md/pytest/test_custom_force.py
b-butler: I am fine with either switch. Where setting directly is an error or setting directly is allowed.

hoomd/md/CustomForceCompute.h
tommy-waltmann: Can do

hoomd/md/pytest/test_custom_force.py
tommy-waltmann: I am a fan of allowing bracketless notation to work, but not advertising it. I can update examples and unit tests to use the bracket notation if need be.

If we enforce the bracket notation, we need to raise errors when users use the bracketless notation so there aren't silent issues with user code.

tommy-waltmann:
@b-butler "You can use test classes for this as I do for the local particle data tests. One class defines the tests and runs on the CPU. The other inherits from the first and runs all test on CPU and GPU."

I only see one class in the file `test_local_snapshots.py` which is used for testing, so your comment confuses me.

hoomd/md/pytest/test_custom_force.py
```diff
@@ -0,0 +1,326 @@
...
+    energies2 = integrator2.forces[0].energies
+    if sim.device.communicator.rank == 0:
+        npt.assert_allclose(energies1, energies2)
+
+    torques1 = integrator.forces[0].torques
+    torques2 = integrator2.forces[0].torques
+    if sim.device.communicator.rank == 0:
+        npt.assert_allclose(torques1, torques2)
+
+    assert integrator.forces[0].virials == integrator2.forces[0].virials
```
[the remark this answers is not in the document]
tommy-waltmann: When virials aren't computed because the proper flags aren't set, the `force.virials` property returns `None`. `npt.assert_allclose` can't compare two None values.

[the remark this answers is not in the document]
b-butler: Why does this not fail then? This should return an array of Booleans.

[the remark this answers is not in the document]
tommy-waltmann: In this particular test, accessing the virials property just returns `None` because I don't set the flags to compute the virials. I don't know whether that is the desired behavior or not.

tommy-waltmann:
A TODO list for completing this PR, anyone can feel free to add things if they want:

1. Rewrite tests so the test matrix doesn't require so many skips.
2. Write tests for non-custom force classes (like LJ) which ensures getting direct access to their local buffers is read-only.
3. Fix pre-commit issue with importing cupy.
4. Clarify the logic in the `_try_running_sim` test helper function.

j-proc (approved):
Everything I brought up last time has been addressed or discussed. And it's received quite a bit of attention beyond just getting the functionality in, so aside from making sure the test fails aren't this (it doesn't seem to be) then I think this is a go. 👍 great work @tommy-waltmann

joaander (approved):
This is great! Thanks!

PR #1153, opened 2021-10-06
Tests for checking loggables match expected namespace

### LINKED ISSUE #922: Test for logging names

Charlottez112:
## Description

This PR adds tests that check the names of the log gables match the expected namespace.

## Motivation and context

Resolves #922 

## Checklist:

### DIFF
--- a/hoomd/hpmc/integrate.py
+++ b/hoomd/hpmc/integrate.py
@@ -199,14 +199,6 @@ class HPMCIntegrator(BaseIntegrator):
 
     # TODO need to validate somewhere that quaternions are normalized
 
-    @property
-    def type_shapes(self):
-        """list[dict]: Description of shapes in ``type_shapes`` format."""
-        raise NotImplementedError(
-            "You are using a shape type that is not implemented! "
-            "If you want it, please modify the "
-            "hoomd.hpmc.integrate.HPMCIntegrator.get_type_shapes function.")
-
     def _return_type_shapes(self):
         type_shapes = self._cpp_obj.getTypeShapesPy()
         ret = [json.loads(json_string) for json_string in type_shapes]
--- a/hoomd/hpmc/pytest/test_boxmc.py
+++ b/hoomd/hpmc/pytest/test_boxmc.py
@@ -5,7 +5,8 @@
 """Test hoomd.hpmc.update.BoxMC."""
 
 import hoomd
-from hoomd.conftest import operation_pickling_check
+from hoomd.logging import LoggerCategories
+from hoomd.conftest import operation_pickling_check, logging_check
 import pytest
 import numpy as np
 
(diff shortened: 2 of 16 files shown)


Charlottez112:
I still need to add tests for several other classes that have loggables, but just wanted to push the current changes and have someone take a look if this is what we want. Also I'm not sure what's the most efficient way to test loggables related to `md.force.Force`, since it seems like there's no test specifically for it.

joaander:
@b-butler should review these changes.

Charlottez112:
There's not a pytest for `hpmc.integrate`, so I put the test that checks its loggable names in `test_compute_sdf.py`. And like I mentioned above, the loggables associated with `md.force.Force` are checked in 'test_potential.py'. Otherwise I think I've checked all loggables.

joaander:
@Charlottez112 @b-butler I would like to see this completed and merged. Do you have any questions?

Charlottez112:
> `type_shapes` is not implemented in the base class `HPMCIntegrator`. It is implemented only in those subclass integrators that support the protocol. 

Should we also add the `loggable` label under the `shape` parameters in the doc then? Because all other loggable quantities have a line like `(Loggable: category=”sequence”)`. Also I'm not sure where I should put the test for checking `type_shapes`.

joaander:
> > `type_shapes` is not implemented in the base class `HPMCIntegrator`. It is implemented only in those subclass integrators that support the protocol.
> 
> Should we also add the `loggable` label under the `shape` parameters in the doc then? Because all other loggable quantities have a line like `(Loggable: category=”sequence”)`. Also I'm not sure where I should put the test for checking `type_shapes`.

Do not manually add the `Loggable:` documentation. The implemented `type_shapes` are defined as loggable with `@log(category='object', requires_run=True)`. These should get the automatically generated `Loggable:` documentation? @b-butler why aren't they?

You can add a new `test_` file and check the specific HPMC integrators that support `type_shapes` there.

b-butler:
@joaander my guess is that since the property `type_shapes` exist in `HPMCIntegrator` that it is preventing the same property to show in the derived classes. This also may have something to do with the `show-inheritance` option.

joaander:
> @joaander my guess is that since the property `type_shapes` exist in `HPMCIntegrator` that it is preventing the same property to show in the derived classes. This also may have something to do with the `show-inheritance` option.
> 
> This would mean the docstrings are correct but just not shown which is consistent with the current documentation.

Can we fix by removing the method from the base class?

Charlottez112:
> You can add a new `test_` file and check the specific HPMC integrators that support `type_shapes` there.

I realized that I can still test this by just calling a specific subclass integrator, so I just put it in `test_shape.py` to let the tests past first. I'm not sure what else I need to do. It seems like we haven't decided what to do?

joaander (commented):
@b-butler is the one with the change request, he needs to re-review.

b-butler (approved):
:+1:

PR #1162, opened 2021-10-26
Mesh data structure

SchoeniPhlippsn:
## Description
This PR introduces the backbone data structure for meshes which can be used as flexible walls/membranes. The meshes are particle-based and are constituted of particle triplets that make up a surface triangulation.

The PR also contains a first class of bond forces that are applied to the mesh data structure. This shows how I plan to use meshes in general. Meshes are attached to forces in a similar fashion as neighbour lists are attached to pair potentials. 

## Motivation and context

These changes make flexible walls and membranes possible. In addition, it can be expanded to introduce elastic bending forces, area conservation forces, dynamic bonding, etc...

## How has this been tested?

Added pytests and tested sphinx documentation

## Change log

```
Introduction of mesh data structure
```

## Checklist:

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 3.0.1
+current_version = 3.1.0
 commit = False
 tag = False
 parse = ^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))(?:\.(?P<number>0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)))?$
--- a/.github/ISSUE_TEMPLATE/support.md
+++ b/.github/ISSUE_TEMPLATE/support.md
@@ -7,7 +7,7 @@ assignees: ''
 
 ---
 
-## Please request support of the mailing list
+## Please request support from the mailing list
 
 Please ask user support questions on the hoomd-users mailing list:
 https://groups.google.com/d/forum/hoomd-users
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,7 +36,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3.0.0
+        uses: actions/checkout@v3.0.2
         with:
           fetch-depth: 0
           submodules: true
--- a/.github/workflows/templates/release.yml
+++ b/.github/workflows/templates/release.yml
@@ -23,7 +23,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3.0.0
+        uses: actions/checkout@v3.0.2
         with:
           fetch-depth: 0
           submodules: true
--- a/.github/workflows/templates/test.yml
+++ b/.github/workflows/templates/test.yml
@@ -49,7 +49,7 @@ env:
 <% endmacro %>
 <% set build_steps %>
     - name: Checkout
-      uses: actions/checkout@v3.0.0
+      uses: actions/checkout@v3.0.2
       with:
         path: code
         submodules: true
@@ -250,10 +250,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3.0.0
+      - uses: actions/checkout@v3.0.2
         with:
           path: code
-      - uses: actions/cache@v3.0.0
+      - uses: actions/cache@v3.0.2
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-openstacksdk==0.61.0
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,7 +68,7 @@ jobs:
       run: rm -rf ./*
 
     - name: Checkout
-      uses: actions/checkout@v3.0.0
+      uses: actions/checkout@v3.0.2
       with:
         path: code
         submodules: true
@@ -274,7 +274,7 @@ jobs:
       run: rm -rf ./*
 
     - name: Checkout
-      uses: actions/checkout@v3.0.0
+      uses: actions/checkout@v3.0.2
       with:
         path: code
         submodules: true
@@ -464,10 +464,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3.0.0
+      - uses: actions/checkout@v3.0.2
         with:
           path: code
-      - uses: actions/cache@v3.0.0
+      - uses: actions/cache@v3.0.2
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-openstacksdk==0.61.0
--- a/.mailmap
+++ b/.mailmap
@@ -31,6 +31,7 @@ Paul Dodd <pdodd@umich.edu> Paul M Dodd <Paul@MacAttack.local>
 Paul Dodd <pdodd@umich.edu> Paul M Dodd <Paul@141-213-172-105.vpn.umnet.umich.edu>
 Paul Dodd <pdodd@umich.edu> Paul M Dodd <Paul@host-112.subnet-108.med.umich.edu>
 Paul Dodd <pdodd@umich.edu> Paul M Dodd <Paul@host-226.subnet-108.med.umich.edu>
+Paul Dodd <pdodd@umich.edu> the_real_pdodd <pdodd@umich.edu>
 Vyas Ramasubramani <vyasr@nvidia.com> Vyas <vramasub@umich.edu>
 Vyas Ramasubramani <vyasr@nvidia.com> vramasub <vramasub@umich.edu>
 Vyas Ramasubramani <vyasr@nvidia.com> Vyas Ramasubramani <vyas.ramasubramani@gmail.com>
@@ -47,6 +48,9 @@ Eric Harper <harperic@umich.edu> harperic <harperic@umich.edu>
 Bryan VanSaders <bvansade@umich.edu> Bryan Vansaders <bvansade@umich.edu>
 Luis Y. Rivera-Rivera <lyrivera@umich.edu> lyrivera <lyrivera@santo.engin.umich.edu>
 Luis Y. Rivera-Rivera <lyrivera@umich.edu> Luis Yamil Rivera-Rivera <lyrivera@umich.edu>
+Luis Y. Rivera-Rivera <lyrivera@umich.edu> lyrivera <lyrivera@umich.edu>
+Luis Y. Rivera-Rivera <lyrivera@umich.edu> lyrivera <47360365+lyrivera@users.noreply.github.com>
+Luis Y. Rivera-Rivera <lyrivera@umich.edu> Luis Yamil Rivera-Rivera <lyrivera@flux-login1.arc-ts.umich.edu>
 Isaac Bruss <ibruss@umich.edu> Isaac Bruss <ibruss@fidrych.engin.umich.edu>
 Isaac Bruss <ibruss@umich.edu> Isaac Bruss <isaacbruss@Isaacs-MacBook-Pro.local>
 Lin Yang <amesyang@gmail.com>
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,7 @@
 ci:
   autoupdate_schedule: quarterly
   autoupdate_branch: 'trunk-patch'
+  autofix_prs: false
 
 exclude: (?x)(
         ^hoomd/extern/|
--- a/BUILDING.rst
+++ b/BUILDING.rst
@@ -138,7 +138,7 @@ Clone using Git_::
 
    $ git clone --recursive https://github.com/glotzerlab/hoomd-blue
 
-Release tarballs are also available as `GitHub release`_ assets: `Download hoomd-v3.0.1.tar.gz`_.
+Release tarballs are also available as `GitHub release`_ assets: `Download hoomd-v3.1.0.tar.gz`_.
 
 .. seealso::
 
@@ -151,7 +151,7 @@ Release tarballs are also available as `GitHub release`_ assets: `Download hoomd
     Execute ``git submodule update --init`` to fetch the submodules each time you switch branches
     and the submodules show as modified.
 
-.. _Download hoomd-v3.0.1.tar.gz: https://github.com/glotzerlab/hoomd-blue/releases/download/v3.0.1/hoomd-v3.0.1.tar.gz
+.. _Download hoomd-v3.1.0.tar.gz: https://github.com/glotzerlab/hoomd-blue/releases/download/v3.1.0/hoomd-v3.1.0.tar.gz
 .. _GitHub release: https://github.com/glotzerlab/hoomd-blue/releases
 .. _git book: https://git-scm.com/book
 .. _Git: https://git-scm.com/
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,20 +7,24 @@ Change Log
 v3.x
 ----
 
-v3.1.0 (not yet released)
-^^^^^^^^^^^^^^^^^^^^^^^^^
+v3.1.0 (2022-04-27)
+^^^^^^^^^^^^^^^^^^^
 
 *Added*
 
 * Support LLVM 13 when ``ENABLE_LLVM=on``.
-* ``hoomd.md.pair.LJGauss`` Lennard-Jones-Gaussian pair potential.
-* ``hoomd.md.alchemy.methods.NVT`` Alchemical molecular dynamics integration method.
-* ``hoomd.md.alchemy.pair.LJGauss`` Lennard-Jones-Gaussian pair potential with alchemical degrees of
-  freedom.
+* ``hoomd.md.pair.LJGauss`` - Lennard-Jones-Gaussian pair potential.
+* ``hoomd.md.alchemy.methods.NVT`` - Alchemical molecular dynamics integration method.
+* ``hoomd.md.alchemy.pair.LJGauss`` - Lennard-Jones-Gaussian pair potential with alchemical degrees
+  of freedom.
+* ``hoomd.hpmc.update.Shape`` - Alchemical hard particle Monte Carlo through shape change moves.
+* ``hoomd.hpmc.shape_move.Elastic`` - Shape move with elastic potential energy penalty.
+* ``hoomd.hpmc.shape_move.ShapeSpace`` - Moves in a user defined shape space.
+* ``hoomd.hpmc.shape_move.Vertex`` - Translate shape vertices.
 
 *Changed*
 
-* Change HPMC fugacity to a per-type quantity.
+* HPMC fugacity is now a per-type quantity.
 * Improved documentation.
 * [developers] Reduced the time needed for incremental builds.
 * [developers] Reduced memory needed to compile HOOMD.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ add_subdirectory (CMake)
 
 ################################
 ## Version information
-set(HOOMD_VERSION_RAW "3.0.1")
+set(HOOMD_VERSION_RAW "3.1.0")
 string(REGEX MATCH "(.*)\\.(.*)\\.(.*)$" _hoomd_version_match ${HOOMD_VERSION_RAW})
 set(HOOMD_VERSION_MAJOR ${CMAKE_MATCH_1})
 set(HOOMD_VERSION_MINOR ${CMAKE_MATCH_2})
--- a/hoomd/RNGIdentifiers.h
+++ b/hoomd/RNGIdentifiers.h
@@ -65,6 +65,9 @@ struct RNGIdentifier
     static const uint8_t HPMCMonoPatch = 39;
     static const uint8_t UpdaterClusters2 = 40;
     static const uint8_t HPMCMonoChainMove = 41;
+    static const uint8_t UpdaterShapeUpdate = 42;
+    static const uint8_t UpdaterShapeConstruct = 43;
+    static const uint8_t HPMCShapeMoveUpdateOrder = 44;
     };
 
     } // namespace hoomd
--- a/hoomd/hpmc/CMakeLists.txt
+++ b/hoomd/hpmc/CMakeLists.txt
@@ -37,6 +37,7 @@ set(_hpmc_sources   module.cc
                     module_convex_polyhedron.cc
                     module_convex_spheropolyhedron.cc
                     ExternalFieldWall.cc
+                    ShapeUtils.cc
                     UpdaterBoxMC.cc
                     UpdaterQuickCompress.cc
                     IntegratorHPMC.cc
@@ -78,6 +79,7 @@ set(_hpmc_headers
     ShapeConvexPolyhedron.h
     ShapeEllipsoid.h
     ShapeFacetedEllipsoid.h
+    ShapeMoves.h
     ShapePolyhedron.h
     ShapeSimplePolygon.h
     ShapeSphere.h
@@ -91,6 +93,7 @@ set(_hpmc_headers
     UpdaterClustersGPUDepletants.cuh
     UpdaterMuVT.h
     UpdaterQuickCompress.h
+    UpdaterShape.h
     XenoCollide2D.h
     XenoCollide3D.h
     )
@@ -170,6 +173,7 @@ set(files   compute.py
             __init__.py
             integrate.py
             update.py
+            shape_move.py
     )
 
 install(FILES ${files}
--- a/hoomd/hpmc/ShapeMoves.h
+++ b/hoomd/hpmc/ShapeMoves.h
@@ -1,0 +1,700 @@
+// Copyright (c) 2009-2022 The Regents of the University of Michigan.
+// Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+#ifndef _SHAPE_MOVES_H
+#define _SHAPE_MOVES_H
+
+#include "GSDHPMCSchema.h"
+#include "Moves.h"
+#include "ShapeUtils.h"
+#include "hoomd/extern/quickhull/QuickHull.hpp"
+#include <Eigen/Dense>
+#include <hoomd/Variant.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+namespace hoomd
+    {
+
+namespace hpmc
+    {
+
+template<typename Shape> class ShapeMoveBase
+    {
+    public:
+    ShapeMoveBase(std::shared_ptr<SystemDefinition> sysdef,
+                  std::shared_ptr<IntegratorHPMCMono<Shape>> mc)
+        : m_mc(mc), m_sysdef(sysdef)
+        {
+        m_ntypes = this->m_sysdef->getParticleData()->getNTypes();
+        m_volume.resize(m_ntypes, 0);
+        m_step_size.resize(m_ntypes, 0);
+        }
+
+    virtual ~ShapeMoveBase() {};
+
+    //! prepare is called at the beginning of every update()
+    virtual void prepare(uint64_t timestep) { }
+
+    //! construct is called for each particle type that will be changed in update()
+    virtual void update_shape(uint64_t,
+                              const unsigned int&,
+                              typename Shape::param_type&,
+                              hoomd::RandomGenerator&)
+        {
+        }
+
+    //! retreat whenever the proposed move is rejected.
+    virtual void retreat(uint64_t timestep, unsigned int type) { }
+
+    Scalar getStepSize(std::string typ)
+        {
+        unsigned int typid = getValidateType(typ);
+        return this->m_step_size[typid];
+        }
+
+    void setStepSize(std::string typ, Scalar volume)
+        {
+        unsigned int typid = getValidateType(typ);
+        this->m_step_size[typid] = volume;
+        }
+
+    Scalar getVolume(std::string typ)
+        {
+        unsigned int typid = getValidateType(typ);
+        return this->m_volume[typid];
+        }
+
+    void setVolume(std::string typ, Scalar volume)
+        {
+        unsigned int typid = getValidateType(typ);
+        this->m_volume[typid] = volume;
+        }
+
+    unsigned int getValidateType(std::string typ)
+        {
+        unsigned int typid = this->m_sysdef->getParticleData()->getTypeByName(typ);
+        if (typid >= this->m_ntypes)
+            {
+            throw std::runtime_error("Invalid particle type.");
+            }
+        return typid;
+        }
+
+    Scalar getMoveProbability()
+        {
+        return this->m_move_probability;
+        }
+
+    void setMoveProbability(Scalar move_probability)
+        {
+        this->m_move_probability = fmin(move_probability, 1.0);
+        }
+
+    virtual Scalar computeLogBoltmann(uint64_t timestep,
+                                      const unsigned int& N,
+                                      const unsigned int type_id,
+                                      const typename Shape::param_type& shape_new,
+                                      const Scalar& inew,
+                                      const typename Shape::param_type& shape_old,
+                                      const Scalar& iold)
+        {
+        return (Scalar(N) / Scalar(2.0)) * log(inew / iold);
+        }
+
+    virtual Scalar computeEnergy(uint64_t timestep,
+                                 const unsigned int& N,
+                                 const unsigned int type_id,
+                                 const typename Shape::param_type& shape,
+                                 const Scalar& inertia)
+        {
+        return 0.0;
+        }
+
+    protected:
+    std::shared_ptr<IntegratorHPMCMono<Shape>> m_mc;
+    std::shared_ptr<SystemDefinition> m_sysdef;
+    unsigned m_ntypes;
+    std::vector<Scalar> m_volume;
+    std::vector<Scalar> m_step_size;
+    Scalar m_move_probability;
+    std::vector<vec3<Scalar>> m_centroids;
+
+    }; // end class ShapeMoveBase
+
+template<typename Shape> class PythonShapeMove : public ShapeMoveBase<Shape>
+    {
+    public:
+    PythonShapeMove(std::shared_ptr<SystemDefinition> sysdef,
+                    std::shared_ptr<IntegratorHPMCMono<Shape>> mc)
+        : ShapeMoveBase<Shape>(sysdef, mc)
+        {
+        m_params.resize(this->m_ntypes);
+        m_params_backup.resize(this->m_ntypes);
+        }
+
+    void prepare(uint64_t timestep)
+        {
+        m_params_backup = m_params;
+        }
+
+    void update_shape(uint64_t timestep,
+                      const unsigned int& type_id,
+                      typename Shape::param_type& shape,
+                      hoomd::RandomGenerator& rng)
+        {
+        for (unsigned int i = 0; i < m_params[type_id].size(); i++)
+            {
+            Scalar stepsize = this->m_step_size[type_id];
+            Scalar a = fmax(-stepsize, -(m_params[type_id][i]));
+            Scalar b = fmin(stepsize, (1.0 - m_params[type_id][i]));
+            hoomd::UniformDistribution<Scalar> uniform(a, b);
+            Scalar r = hoomd::detail::generate_canonical<double>(rng);
+            Scalar x = (r < this->m_move_probability) ? uniform(rng) : 0.0;
+            m_params[type_id][i] += x;
+            }
+        pybind11::object d = m_python_callback(type_id, m_params[type_id]);
+        pybind11::dict shape_dict = pybind11::cast<pybind11::dict>(d);
+        shape = typename Shape::param_type(shape_dict);
+        }
+
+    void retreat(uint64_t timestep, unsigned int type)
+        {
+        // move has been rejected.
+        m_params[type] = m_params_backup[type];
+        }
+
+    pybind11::list getParams(std::string typ)
+        {
+        unsigned int type_id = this->getValidateType(typ);
+        pybind11::list ret;
+        for (unsigned int i = 0; i < m_params[type_id].size(); i++)
+            {
+            ret.append(m_params[type_id][i]);
+            }
+        return ret;
+        }
+
+    void setParams(std::string typ, pybind11::list params)
+        {
+        unsigned int type_id = this->getValidateType(typ);
+        auto N = pybind11::len(params);
+        m_params[type_id].resize(N);
+        m_params_backup[type_id].resize(N);
+        for (unsigned int i = 0; i < N; i++)
+            {
+            m_params[type_id][i] = params[i].cast<Scalar>();
+            m_params_backup[type_id][i] = params[i].cast<Scalar>();
+            }
+        }
+
+    pybind11::object getCallback()
+        {
+        return m_python_callback;
+        }
+
+    void setCallback(pybind11::object python_callback)
+        {
+        m_python_callback = python_callback;
+        }
+
+    private:
+    std::vector<std::vector<Scalar>>
+        m_params_backup;                       // tunable shape parameters to perform trial moves on
+    std::vector<std::vector<Scalar>> m_params; // tunable shape parameters to perform trial moves on
+    // callback that takes m_params as an argiment and returns a Python dictionary with the shape
+    // params.
+    pybind11::object m_python_callback;
+    };
+
+class ConvexPolyhedronVertexShapeMove : public ShapeMoveBase<ShapeConvexPolyhedron>
+    {
+    public:
+    typedef typename ShapeConvexPolyhedron::param_type param_type;
+
+    ConvexPolyhedronVertexShapeMove(std::shared_ptr<SystemDefinition> sysdef,
+                                    std::shared_ptr<IntegratorHPMCMono<ShapeConvexPolyhedron>> mc)
+        : ShapeMoveBase<ShapeConvexPolyhedron>(sysdef, mc)
+        {
+        this->m_centroids.resize(this->m_ntypes, vec3<Scalar>(0, 0, 0));
+        initializeMassProperties();
+        }
+
+    void initializeMassProperties()
+        {
+        auto& mc_params = this->m_mc->getParams();
+        for (unsigned int i = 0; i < this->m_ntypes; i++)
+            {
+            detail::MassProperties<ShapeConvexPolyhedron> mp(mc_params[i]);
+            this->m_centroids[i] = mp.getCenterOfMass();
+            }
+        }
+
+    void setVolume(std::string typ, Scalar volume)
+        {
+        unsigned int typid = getValidateType(typ);
+        this->m_volume[typid] = volume;
+        auto& mc_params = this->m_mc->getParams();
+        param_type shape = mc_params[typid];
+        detail::MassProperties<ShapeConvexPolyhedron> mp(shape);
+        Scalar current_volume = mp.getVolume();
+        vec3<Scalar> dr = this->m_centroids[typid] - mp.getCenterOfMass();
+        OverlapReal scale = (OverlapReal)fast::pow(volume / current_volume, 1.0 / 3.0);
+        this->scaleParticleVolume(shape, dr, scale);
+        this->m_mc->setParam(typid, shape);
+        }
+
+    void scaleParticleVolume(param_type& shape, vec3<Scalar> dr, OverlapReal scale)
+        {
+        Scalar rsq = 0.0;
+        for (unsigned int i = 0; i < shape.N; i++)
+            {
+            shape.x[i] += static_cast<OverlapReal>(dr.x);
+            shape.x[i] *= scale;
+            shape.y[i] += static_cast<OverlapReal>(dr.y);
+            shape.y[i] *= scale;
+            shape.z[i] += static_cast<OverlapReal>(dr.z);
+            shape.z[i] *= scale;
+            vec3<Scalar> vert(shape.x[i], shape.y[i], shape.z[i]);
+            rsq = fmax(rsq, dot(vert, vert));
+            }
+        shape.diameter = OverlapReal(2.0 * fast::sqrt(rsq));
+        }
+
+    void prepare(uint64_t timestep)
+        {
+        m_step_size_backup = this->m_step_size;
+        }
+
+    void update_shape(uint64_t timestep,
+                      const unsigned int& type_id,
+                      param_type& shape,
+                      hoomd::RandomGenerator& rng)
+        {
+        // perturb the shape.
+        for (unsigned int i = 0; i < shape.N; i++)
+            {
+            if (hoomd::detail::generate_canonical<double>(rng) < this->m_move_probability)
+                {
+                vec3<Scalar> vert(shape.x[i], shape.y[i], shape.z[i]);
+                move_translate(vert, rng, this->m_step_size[type_id], 3);
+                shape.x[i] = static_cast<OverlapReal>(vert.x);
+                shape.y[i] = static_cast<OverlapReal>(vert.y);
+                shape.z[i] = static_cast<OverlapReal>(vert.z);
+                }
+            }
+        detail::MassProperties<ShapeConvexPolyhedron> mp(shape);
+        Scalar volume = mp.getVolume();
+        vec3<Scalar> dr = this->m_centroids[type_id] - mp.getCenterOfMass();
+        OverlapReal scale
+            = static_cast<OverlapReal>(fast::pow(this->m_volume[type_id] / volume, 1.0 / 3.0));
+        scaleParticleVolume(shape, dr, scale);
+        this->m_step_size[type_id] *= scale;
+        }
+
+    void retreat(uint64_t timestep, unsigned int type)
+        {
+        this->m_step_size[type] = m_step_size_backup[type];
+        }
+
+    private:
+    std::vector<Scalar> m_step_size_backup;
+    }; // end class ConvexPolyhedronVertexShapeMove
+
+template<class Shape> class ElasticShapeMoveBase : public ShapeMoveBase<Shape>
+    {
+    public:
+    ElasticShapeMoveBase(std::shared_ptr<SystemDefinition> sysdef,
+                         std::shared_ptr<IntegratorHPMCMono<Shape>> mc)
+        : ShapeMoveBase<Shape>(sysdef, mc)
+        {
+        }
+
+    typedef typename Shape::param_type param_type;
+
+    void setStiffness(std::shared_ptr<Variant> stiff)
+        {
+        m_k = stiff;
+        }
+
+    std::shared_ptr<Variant> getStiffness() const
+        {
+        return m_k;
+        }
+
+    pybind11::dict getReferenceShape(std::string typ)
+        {
+        unsigned int typid = this->getValidateType(typ);
+        return m_reference_shapes[typid].asDict();
+        }
+
+    virtual void setReferenceShape(std::string typ, pybind11::dict v) = 0;
+
+    protected:
+    std::shared_ptr<Variant> m_k; // shape move stiffness
+    // shape to reference shape move against
+    std::vector<param_type, hoomd::detail::managed_allocator<param_type>> m_reference_shapes;
+    };
+
+template<class Shape> class ElasticShapeMove : public ElasticShapeMoveBase<Shape>
+    {
+    };
+
+template<>
+class ElasticShapeMove<ShapeConvexPolyhedron> : public ElasticShapeMoveBase<ShapeConvexPolyhedron>
+    {
+    public:
+    typedef typename Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic> MatrixSD;
+    typedef typename Eigen::Matrix<Scalar, 3, 3> Matrix3S;
+
+    ElasticShapeMove(std::shared_ptr<SystemDefinition> sysdef,
+                     std::shared_ptr<IntegratorHPMCMono<ShapeConvexPolyhedron>> mc)
+        : ElasticShapeMoveBase<ShapeConvexPolyhedron>(sysdef, mc)
+        {
+        m_reference_shapes.resize(this->m_ntypes);
+        m_F.resize(this->m_ntypes, Matrix3S::Identity());
+        m_F_last.resize(this->m_ntypes, Matrix3S::Identity());
+        }
+
+    void prepare(uint64_t timestep)
+        {
+        m_F_last = m_F;
+        }
+
+    //! construct is called at the beginning of every update()
+    void update_shape(uint64_t timestep,
+                      const unsigned int& type_id,
+                      param_type& param,
+                      hoomd::RandomGenerator& rng)
+        {
+        Matrix3S F_curr;
+        // perform a scaling move
+        if (hoomd::detail::generate_canonical<double>(rng) < this->m_move_probability)
+            {
+            generateExtentional(F_curr, rng, this->m_step_size[type_id] + 1.0);
+            }
+        else // perform a rotation-scale-rotation move
+            {
+            quat<Scalar> q(1.0, vec3<Scalar>(0.0, 0.0, 0.0));
+            move_rotate<3>(q, rng, 0.5);
+            Matrix3S rot, rot_inv, scale;
+            Eigen::Quaternion<double> eq(q.s, q.v.x, q.v.y, q.v.z);
+            rot = eq.toRotationMatrix();
+            rot_inv = rot.transpose();
+            generateExtentional(scale, rng, this->m_step_size[type_id] + 1.0);
+            F_curr = rot * scale * rot_inv;
+            }
+        m_F[type_id] = F_curr * m_F[type_id];
+        auto transform = F_curr.cast<OverlapReal>();
+        Scalar dsq = 0.0;
+        for (unsigned int i = 0; i < param.N; i++)
+            {
+            vec3<OverlapReal> vert(param.x[i], param.y[i], param.z[i]);
+            param.x[i]
+                = transform(0, 0) * vert.x + transform(0, 1) * vert.y + transform(0, 2) * vert.z;
+            param.y[i]
+                = transform(1, 0) * vert.x + transform(1, 1) * vert.y + transform(1, 2) * vert.z;
+            param.z[i]
+                = transform(2, 0) * vert.x + transform(2, 1) * vert.y + transform(2, 2) * vert.z;
+            vert = vec3<Scalar>(param.x[i], param.y[i], param.z[i]);
+            dsq = fmax(dsq, dot(vert, vert));
+            }
+        param.diameter = OverlapReal(2.0 * fast::sqrt(dsq));
+        }
+
+    Matrix3S getEps(unsigned int type_id)
+        {
+        return 0.5 * (m_F[type_id].transpose() + m_F[type_id]) - Matrix3S::Identity();
+        }
+
+    Matrix3S getEpsLast(unsigned int type_id)
+        {
+        return 0.5 * (m_F_last[type_id].transpose() + m_F_last[type_id]) - Matrix3S::Identity();
+        }
+
+    //! retreat whenever the proposed move is rejected.
+    void retreat(uint64_t timestep, unsigned int type)
+        {
+        // we can swap because m_F_last will be reset on the next prepare
+        m_F[type] = m_F_last[type];
+        }
+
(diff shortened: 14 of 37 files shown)


hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,803 @@
...
+
+                for (unsigned int i = 0; i < group_size_half; ++i)
+                    {
+                    unsigned int tag1 = g.tag[i];
+                    unsigned int idx1 = h_rtag.data[tag1];
+                    unsigned int num = h_n_groups.data[idx1]++;
+
+                    members_t h;
+
+                    h.idx[group_size - 1] = ((typeval_t)this->m_group_typeval[cur_group]).type;
```
b-butler, line 478: Static_cast

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,803 @@
+// Copyright (c) 2009-2021 The Regents of the University of Michigan
```
b-butler, line 1: Are all the methods reimplemented here necessary? Also, is there any way to use the base class method?

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,803 @@
...
+// This file is part of the HOOMD-blue project, released under the BSD 3-Clause License.
+
+// Maintainer: jglaser
```
b-butler, line 4: Search for this in your files and remove.

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,803 @@
...
+    if (export_struct)
+        Group::export_to_python(m);
+
+    pybind11::class_<T, std::shared_ptr<T>>(m, name.c_str())
+        .def(pybind11::init<std::shared_ptr<ParticleData>, unsigned int>())
+        .def(pybind11::init<std::shared_ptr<ParticleData>, const typename TriangleData::Snapshot&>())
+        .def("initializeFromSnapshot", &T::initializeFromSnapshot)
+        .def("takeSnapshot", &T::takeSnapshot)
+        .def("addBondedGroup", &T::addBondedGroup);
+    }
```
b-butler, line 783: This seems a subset of `export_BondedGroupData`. Could you just use that?

hoomd/MeshGroupData.cu
```diff
@@ -0,0 +1,251 @@
...
+        // scatter groups to destinations
+        block_size = 256;
+        n_blocks = (group_size * n_groups) / block_size + 1;
+
+        hipLaunchKernelGGL(gpu_mesh_scatter_kernel<group_size>,
+                           dim3(n_blocks),
+                           dim3(block_size),
+                           0,
+                           0,
+                           n_groups * group_size_half,
```
b-butler, line 200: I there a way to not use this `group_size_half` it seem to be the reason we can't reuse this kernel.

hoomd/MeshGroupData.cuh
```diff
@@ -0,0 +1,73 @@
...
+//#else
+//! Forward declaration of group_storage
+//template<unsigned int group_size> union group_storage;
+
+//! Forward declaration of packed_storage
+//template<unsigned int group_size> struct packed_storage;
+
+//! Forward declaration of typeval union
+//union typeval_union;
+//#endif
```
b-butler, line 53: If this is not necessary it is not necessary.

hoomd/MeshDefinition.h
```diff
@@ -0,0 +1,106 @@
...
+
+    BondData::Snapshot getBondData();
+
+    void updateTriangleData();
+
+    void updateMeshData();
+
+
+    TriangleData::Snapshot triangle_data;             //!< The triangle data accessible in python
+    Scalar m_mesh_energy;         //!< check if dynamic bonding has changed the mesh data
```
b-butler, line 87: This comment does not seem to go with the variable

b-butler, line 87: These can be private and accessable through a getter.

hoomd/MeshDefinition.h
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+    BondData::Snapshot getBondData();
+
+    void updateTriangleData();
+
+    void updateMeshData();
+
+
+    TriangleData::Snapshot triangle_data;             //!< The triangle data accessible in python
```
b-butler, line 86: Why are we storing the snapshot and the data rather than generating the snapshot and returning it when asked?

hoomd/MeshDefinition.cc
```diff
@@ -0,0 +1,83 @@
...
+
+    m_sysdef = sysdef;
+    m_data_changed = false;
+    m_meshtriangle_data
+        = std::shared_ptr<MeshTriangleData>(new MeshTriangleData(m_sysdef->getParticleData(), 1));
+    m_meshbond_data
+        = std::shared_ptr<MeshBondData>(new MeshBondData(m_sysdef->getParticleData(), 1));
+
+    m_mesh_energy = 0;
+    m_mesh_energy_old = 0;
```
b-butler, line 37: Can be done in the initializer list.

hoomd/MeshDefinition.cc
```diff
@@ -0,0 +1,83 @@
...
+    {
+    pybind11::class_<MeshDefinition, std::shared_ptr<MeshDefinition>>(m, "MeshDefinition")
+        .def(pybind11::init<>())
+        .def(pybind11::init<std::shared_ptr<SystemDefinition> >())
+        .def("getMeshTriangleData", &MeshDefinition::getMeshTriangleData)
+        .def("getMeshBondData", &MeshDefinition::getMeshBondData)
+        .def("getBondData", &MeshDefinition::getBondData)
+        .def("updateTriangleData", &MeshDefinition::updateTriangleData)
+        .def("updateMeshData", &MeshDefinition::updateMeshData)
+        .def_readonly("triangles", &MeshDefinition::triangle_data)
```
b-butler, line 77: Given the complexity of these attributes they are still mutable in Python, but I don't think that is reflected in C++.

hoomd/data/typeparam.py
```diff
@@ -197,6 +197,11 @@ def _attach(self, cpp_obj, sim):
             self.param_dict)
         return self
 
+    def _attach_with_mesh(self, cpp_obj, mesh):
+        self.param_dict = AttachedTypeParameterDict(
+            cpp_obj, self.name, getattr(mesh, self.type_kind), self.param_dict)
+        return self
+
```
b-butler, line 204: I would really like to find a way to make this unnecessary.

b-butler: One way could be to abstract out the way to query for types to the constructor. There should be a way to do this by either extending the abstraction or fitting this into the existing one.

hoomd/mesh.py
```diff
@@ -0,0 +1,151 @@
...
+                self._remove()
+
+    @property
+    def size(self):
+        """(int): Number of triangles in the mesh."""
+        if self._attached:
+            self._update_triangles()
+            return self._cpp_obj.triangles.N
+        else:
+            return self._size
```
b-butler, line 66:
```suggestion
        return self._size
```

hoomd/mesh.py
```diff
@@ -0,0 +1,151 @@
...
+            self._size = newN
+
+    @property
+    def typeid(self):
+        """((*N*,) `numpy.ndarray` of ``uint32``): Triangle type id."""
+        if self._attached:
+            self._update_triangles()
+            return self._cpp_obj.triangles.typeid
+        else:
+            return self._typeid
```
b-butler, line 82:
```suggestion
        return self._typeid
```

b-butler: Do this elsewhere as well.

hoomd/mesh.py
```diff
@@ -0,0 +1,151 @@
...
+            return self._cpp_obj.triangles.types
+        else:
+            return self._types
+
+    @types.setter
+    def types(self, newtypes):
+        if self._attached:
+            self._cpp_obj.triangles.types = newtypes
+        else:
+            self._types = newtypes
```
b-butler, line 107: Are we letting the number of types change after construction? We removed dynamic number of types for the state.

hoomd/pytest/test_mesh.py
```diff
@@ -0,0 +1,127 @@
+import copy as cp
+import hoomd
+from hoomd.mesh import Mesh
+import numpy
+import pytest
+from hoomd.error import DataAccessError
+
+
+@pytest.fixture(scope='session')
+def dihedral_snapshot_factory(device):
```
b-butler, line 10: Do you need the dihedrals here?

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,251 @@
...
+        elif (not self._added and mesh._added
+              and mesh._simulation != simulation):
+            raise RuntimeError(
+                f"Mesh associated with {self} is associated with "
+                f"another simulation.")
+        super()._add(simulation)
+        # this ideopotent given the above check.
+        self._mesh._add(simulation)
+        # This is ideopotent, but we need to ensure that if we change
+        # neighbor list when not attached we handle correctly.
```
b-butler, line 57:
```suggestion
        # mesh when not attached we handle correctly.
```

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,251 @@
...
+        self._add_dependency(self._mesh)
+
+    def _attach(self):
+        """Create the c++ mirror class."""
+        if not self._mesh._added:
+            self._mesh._add(self._simulation)
+        else:
```
b-butler, line 64: This should be unnecessary now.

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,251 @@
...
+        self._apply_typeparam_dict_with_mesh(self._cpp_obj, self.mesh)
+
+    def _apply_typeparam_dict_with_mesh(self, cpp_obj, mesh):
+        for typeparam in self._typeparam_dict.values():
+            try:
+                typeparam._attach_with_mesh(cpp_obj, mesh)
+            except ValueError as err:
+                raise err.__class__(
+                    f"For {type(self)} in TypeParameter {typeparam.name} "
+                    f"{str(err)}")
```
b-butler, line 90: If we are going to overwrite this, than we should do so with the same name to indicate what we are doing. Then we can call `super()._attach()`

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,251 @@
...
+        if self._added:
+            if mesh._added and self._simulation != mesh._simulation:
+                raise RuntimeError("Mesh and forces must belong to the same "
+                                   "simulation or SyncedList.")
+            self._mesh._add(self._simulation)
+        self._mesh = mesh
+
+    @property
+    def _children(self):
+        return [self.mesh]
```
b-butler, line 111: Unnecessary
```suggestion
```

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,251 @@
...
+    Examples::
+
+        harmonic = mesh.bond.Harmonic(mesh)
+        harmonic.params['wall'] = dict(k=3.0, r0=2.38)
+        harmonic.params['vesicle'] = dict(k=10.0, r0=1.0)
+    """
+    _cpp_class_name = "PotentialMeshBondHarmonic"
+
+    def __init__(self, mesh):
+        params = TypeParameter("params", "types",
```
b-butler, line 144: One way we could unify type specification would be to accept a type string or callable and the callable would return the valid types.

hoomd/md/pytest/test_meshbond.py
```diff
@@ -0,0 +1,195 @@
...
+    Tether_forces3 = [[0, 0, 0], [0.036666, 0.021169, 0],
+                      [5.358389, 3.093667, 0]]
+    Tether_energies = [0, 0.000463152, 0.1472802]
+
+    harmonic_args_and_vals = []
+    FENE_args_and_vals = []
+    Tether_args_and_vals = []
+    for i in range(3):
+        harmonic_args_and_vals.append(
+            (_harmonic_arg_list[i][0], _harmonic_arg_list[i][1],
```
b-butler, line 60:
```suggestion
            (*_harmonic_arg_list[i],
```
And so on for the other two.

hoomd/md/pytest/test_meshbond.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@pytest.mark.parametrize("mesh_bond_cls, potential_kwargs",
+                         get_mesh_bond_and_args())
+def test_before_attaching(mesh_bond_cls, potential_kwargs):
+    mesh = hoomd.mesh.Mesh()
+    mesh_bond_potential = mesh_bond_cls(mesh)
+    mesh_bond_potential.params['meshbond'] = potential_kwargs
+
+    assert mesh == mesh_bond_potential.mesh
```
b-butler, line 112:
```suggestion
    assert mesh is mesh_bond_potential.mesh
```

hoomd/md/pytest/test_meshbond.py
```diff
@@ -0,0 +1,195 @@
...
+    for key in potential_kwargs:
+        np.testing.assert_allclose(mesh_bond_potential.params['meshbond'][key],
+                                   potential_kwargs[key],
+                                   rtol=1e-6)
+
+    mesh1 = hoomd.mesh.Mesh()
+    mesh_bond_potential.mesh = mesh1
+    assert mesh1 == mesh_bond_potential.mesh
```
b-butler, line 120:
```suggestion
    assert mesh1 is mesh_bond_potential.mesh
```

b-butler (changes requested):
I didn't review everything thoroughly, but I did look at everything besides the C++ implementation of the mesh potentials.

For some of the classes it would be nice if we could rely more on inheritance as the reimplemented methods look very similar.

hoomd/md/CMakeLists.txt
```diff
@@ -143,6 +143,9 @@ set(_md_headers ActiveForceComputeGPU.h
                 PotentialBondGPU.h
                 PotentialBondGPU.cuh
                 PotentialBond.h
+                PotentialMeshBondGPU.h
+                PotentialMeshBondGPU.cuh
+                PotentialMeshBond.h
```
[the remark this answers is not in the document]
SchoeniPhlippsn: I used the name `PotentialMeshBond.ext` as I plan to add more potentials that are applied to the mesh triangles rather than the mesh bonds (ie. bending forces or area conservation forces).

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,251 @@
...
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeconverter import OnlyTypes
+import hoomd
+import warnings
+import copy
+
+validate_mesh = OnlyTypes(Mesh)
+
+
+class MeshBond(Force):
```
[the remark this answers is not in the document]
SchoeniPhlippsn: The same idea applies here.

hoomd/md/mesh/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2009-2021 The Regents of the University of Michigan
+# This file is part of the HOOMD-blue project, released under the BSD 3-Clause
+# License.
+
+"""Mesh potentials for molecular dynamics."""
+
+from . import bond
```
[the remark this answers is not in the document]
SchoeniPhlippsn: The same idea applies here.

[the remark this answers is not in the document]
SchoeniPhlippsn: Again, this is done in anticipation of other types of potentials to come. I will later add `bending`, `conserve` etc.

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,803 @@
...
+        }
+    else
+#endif
+        {
+        this->m_type_mapping = snapshot.type_mapping;
+
+	if(bond)
+	    {
+            typeval_t t;
+            t.type = snapshot.type_id[0];
```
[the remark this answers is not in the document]
SchoeniPhlippsn: This is something I wanted to ask in the next hoomd meeting. As users will only pass the triangulation with the corresponding bonds being calculated in the back, I don't exactly know how to treat the bond types (usually we have more bonds than triangles). Maybe we should allow only for uniform meshes and not even set individual types for each triangle.

hoomd/MeshGroupData.cc
SchoeniPhlippsn: I think it is necessary, but I will go over them again. Maybe I can come up with a better utilization of the base class method.

hoomd/MeshGroupData.cc
SchoeniPhlippsn: It's not a subset. The initialization with `TriangleData::Snapshot` is not used in `export_BondedGroupData`.

hoomd/MeshDefinition.h
SchoeniPhlippsn: It depends if we want to change the mesh triangulation after initialization. For me, it is not extremely necessary. However, I came up with a case (cut a slit in the mesh) where manipulating the mesh might be useful.

hoomd/mesh.py
SchoeniPhlippsn: I am open to prevent the manipulation of the mesh after initialization.

hoomd/MeshDefinition.h
SchoeniPhlippsn: energy can be private. The triangle_data has to be public to allow for access with python.

hoomd/MeshDefinition.h
```diff
@@ -0,0 +1,105 @@
...
+    <b>Initializing</b>
+
+    A default constructed MeshDefinition is full of NULL shared pointers. Such is intended to be
+   assigned to by one created by a SystemInitializer.
+
+    Several other default constructors are provided, mainly to provide backward compatibility to
+   unit tests that relied on the simple initialization constructors provided by ParticleData.
+
+    \ingroup data_structs
+*/
```
joaander, line 53: This documentation is not relevant as it was copied and pasted from `SystemDefinition`. Please rewrite and provide a full description of the various mesh data structures and describe how they interact. We especially need to document clearly where the data structures store indices vs tags and which arrays these indicies or tags refer to.

hoomd/MeshGroupData.cc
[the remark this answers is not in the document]
joaander: It depends on whether you have use-cases where you want to apply different mesh potential parameters to different types of mesh sub units. My initial impression is that types would not be needed.

hoomd/BondedGroupData.h
```diff
@@ -252,11 +263,14 @@ class BondedGroupData
 
     virtual ~BondedGroupData();
 
+    //! Initialize internal memory
+    void initialize();
+
     //! Initialize from a snapshot
-    virtual void initializeFromSnapshot(const Snapshot& snapshot);
```
joaander, line 256: These should remain virtual. Future subclasses may need to override these methods.

hoomd/md/PotentialMeshBond.h
```diff
@@ -0,0 +1,360 @@
...
+
+namespace hoomd
+    {
+namespace md
+    {
+/*! Mesh bond potential with evaluator support
+
+    \ingroup computes
+*/
+template<class evaluator> class PotentialMeshBond : public ForceCompute
```
joaander, line 29: Aside from the fact that this class operations on "mesh bonds" instead of "bonds", are there any other differences with `PotentialBond`?

joaander (changes requested):
* I would like to see more overview documentation on how the mesh data structures are organized and how to use them.
* If we plan to keep mesh types, then they should be an independent type list and not tied to particle types.
* `mesh` is going to be a large module with significant amounts of Python and C++ code that interoperate tightly. We should move this new code to its own top level module `mesh` (like dem and mpcd) in the `hoomd/mesh` directory.
* It appears that the mesh bond potential class is a complete duplication of the bond potential. We should reuse the existing code instead of duplicating it.

mphoward:
> * `mesh` is going to be a large module with significant amounts of Python and C++ code that interoperate tightly. We should move this new code to its own top level module `mesh` (like dem and mpcd) in the `hoomd/mesh` directory.

Just wanted to mention that I am quite interested to use mesh structures in `hoomd.mpcd` in future, so I was very excited to see this feature being added. `hoomd.mpcd` requires `hoomd.md` to build, and I have had no issues with that dependency if you want to move the code.

hoomd/MeshGroupData.cu
SchoeniPhlippsn: I haven't removed it but I think I found the problem that fixes the issue.

SchoeniPhlippsn:
I think this is ready for another review.

joaander:
My apologies for the delay in reviewing. I will review this pull request after Jan 3, 2022.

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,808 @@
...
+            unsigned int group_idx = flag - this->m_next_flag - 1;
+            members_t g = this->m_groups[group_idx];
+
+            std::ostringstream oss;
+            oss << name << ".*: " << name << " ";
+            for (unsigned int k = 0; k < group_size_half; ++k)
+                oss << g.tag[k] << ((k != group_size_half - 1) ? ", " : " ");
+            oss << "incomplete!" << std::endl;
+            this->m_exec_conf->msg->error() << oss.str();
+            throw std::runtime_error("Error building GPU group table.");
```
b-butler, line 581: Just throw the exception with the error streams error message.

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,808 @@
...
+        }
+
+    snapshot.type_mapping = this->m_type_mapping;
+
+    return index;
+    }
+
+namespace detail
+    {
+template<class T, typename Group>
```
b-butler, line 758: use class or typename for both

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,808 @@
...
+            unsigned int group_tag = *active_tag_it;
+            rtag_it = rtag_map.find(group_tag);
+            if (rtag_it == rtag_map.end())
+                {
+                this->m_exec_conf->msg->error() << endl
+                                                << "Could not find " << name << " " << group_tag
+                                                << ". Possible internal error?" << endl
+                                                << endl;
+                throw std::runtime_error("Error gathering " + std::string(name) + "s");
+                }
```
b-butler, line 735: Same drill here.

b-butler (changes requested):
Thanks for the changes. I know it was a lot of work. I have some more comments to address in the mean time, and I am sure Josh will after break.

hoomd/BondedGroupData.cc
```diff
@@ -795,9 +795,7 @@ void BondedGroupData<group_size, Group, name, has_type_mapping>::rebuildGPUTable
                 }
 
             // find the maximum number of groups
-            for (unsigned int i = 0; i < N; i++)
-                if (h_n_groups.data[i] > num_groups_max)
-                    num_groups_max = h_n_groups.data[i];
+            num_groups_max = *std::max_element(h_n_groups.data, h_n_groups.data + N);
```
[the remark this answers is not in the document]
SchoeniPhlippsn: The length of this array is always greater than zero.

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,808 @@
...
+                        this->m_exec_conf->msg->error() << oss.str();
+                        throw std::runtime_error("Error building GPU group table.");
+                        }
+
+                    h_n_groups.data[idx]++;
+                    }
+                }
+
+            // find the maximum number of groups
+            num_groups_max = *std::max_element(h_n_groups.data, h_n_groups.data + N);
```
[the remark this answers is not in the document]
SchoeniPhlippsn: Like in `BondedGroupData.cc` the length of this array is also always greater than zero.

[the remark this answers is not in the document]
b-butler: Awesome, I didn't take the time to sure for sure as I figured you knew. I am glad I suggested the appropriate change last time :smile:.

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,248 @@
...
+        mesh = validate_mesh(value)
+        if self._added:
+            if mesh._added and self._simulation != mesh._simulation:
+                raise RuntimeError("Mesh and forces must belong to the same "
+                                   "simulation or SyncedList.")
+            self._mesh._add(self._simulation)
+        self._mesh = mesh
+
+
+class Harmonic(MeshBond):
```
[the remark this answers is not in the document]
SchoeniPhlippsn: I want to keep it as a `TypeParameter` just in case we want to allow multiple types per mesh in the future. So I changed it to your second suggestion.

hoomd/MeshDefinition.h
```diff
@@ -0,0 +1,111 @@
...
+    //! Constructs a NULL MeshDefinition
+    MeshDefinition();
+    //! Constructs a MeshDefinition with a simply initialized ParticleData
+    MeshDefinition(std::shared_ptr<SystemDefinition> sysdef);
+
+    //! Access the mesh triangle data defined for the simulation
+    std::shared_ptr<MeshTriangleData> getMeshTriangleData()
+        {
+        return m_meshtriangle_data;
+        }
```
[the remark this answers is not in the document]
SchoeniPhlippsn: This will be used mainly by the potentials which can be applied to the meshes later. The bond potential only needs mesh bonds. Other potentials like Helfrich bending potential or area conservation also have to get the mesh triangle data . I am working on those potentials in tandem with this PR.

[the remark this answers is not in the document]
b-butler: Makes sense. :+1:

hoomd/MeshDefinition.h
```diff
@@ -0,0 +1,121 @@
...
+
+#ifndef __MESH_DEFINITION_H__
+#define __MESH_DEFINITION_H__
+
+namespace hoomd
+    {
+#ifdef ENABLE_MPI
+//! Forward declaration of Communicator
+class Communicator;
+#endif
```
joaander, line 25:
```suggestion
```
Is this needed? I don't see `Communicator` used in this file.

hoomd/MeshDefinition.h
```diff
@@ -0,0 +1,121 @@
...
+    Scalar getEnergy()
+        {
+        return m_mesh_energy;
+        }
+
+    void updateTriangleData();
+
+    void updateMeshData();
+
+    TriangleData::Snapshot triangle_data; //!< The triangle data accessible in python
```
joaander, line 101: There should be one source of truth and data should not be stored more than once to limit memory usage. Global snapshots should be constructed when needed and passed back to the user.

hoomd/data/typeparam.py
```diff
@@ -191,10 +191,9 @@ def default(self):
     def default(self, value):
         self.param_dict.default = value
 
-    def _attach(self, cpp_obj, sim):
```
joaander, line 194: @b-butler Is this change to the core HOOMD typeparameter API appropriate?

hoomd/md/AllBondPotentials.h
```diff
@@ -29,21 +31,42 @@ namespace hoomd
...
-typedef PotentialBond<EvaluatorBondFENE> PotentialBondFENE;
+typedef PotentialBond<EvaluatorBondFENE, BondData> PotentialBondFENE;
 //! Bond potential force compute for Tethering forces
-typedef PotentialBond<EvaluatorBondTether> PotentialBondTether;
+typedef PotentialBond<EvaluatorBondTether, BondData> PotentialBondTether;
+
+//! Mesh Bond potential force compute for harmonic forces
+typedef PotentialBond<EvaluatorBondHarmonic, MeshBondData> PotentialMeshBondHarmonic;
+//! Mesh Bond potential force compute for FENE forces
+typedef PotentialBond<EvaluatorBondFENE, MeshBondData> PotentialMeshBondFENE;
```
joaander, line 43: FENE is typically a potential used for polymer simulations to avoid bond crossings. Do you see a strong need for users of meshes to utilize FENE potentials?

hoomd/md/PotentialBondGPU.h
```diff
@@ -64,11 +69,14 @@ class PotentialBondGPU : public PotentialBond<evaluator>
...
 template<class evaluator,
-         hipError_t gpu_cgbf(const kernel::bond_args_t& bond_args,
+         class Bonds,
+         int group_size,
+         hipError_t gpu_cgbf(const kernel::bonds_args_t<group_size>& bond_args,
                              const typename evaluator::param_type* d_params,
                              unsigned int* d_flags)>
-PotentialBondGPU<evaluator, gpu_cgbf>::PotentialBondGPU(std::shared_ptr<SystemDefinition> sysdef)
-    : PotentialBond<evaluator>(sysdef)
+PotentialBondGPU<evaluator, Bonds, group_size, gpu_cgbf>::PotentialBondGPU(
```
joaander, line 77: Would it be possible to derive `group_size` at compile time from the `Bonds` class? That would reduce the number of template arguments and limit the chance for error.

joaander (changes requested):
Thanks @SchoeniPhlippsn this is much cleaner now! I made a few small suggestions to improve and have a few questions. 

I didn't review the documentation this time, will get that next time. I'll leave @b-butler to review the Python API.

joaander:
I pushed a commit to help debut the test failure in MPI. The error message is: "RuntimeError: Error: bond 2 3 is incomplete.".

This means that the code is attempting to compute the bond potential between two particles, but one or more of those particles is not available on the local rank. Do you connect the meshbond data structure to the MPI communication routines? If not, then the bonded particles won't be transmitted to neighboring ranks as ghost particles and could trigger this message.

hoomd/data/typeparam.py
b-butler: @joaander to keep the current flexibility is it somewhat necessary since we do not know from the simulation the mesh types. I mentioned using the `ParameterDict` (`_param_dict`) for mesh potentials but @SchoeniPhlippsn wanted to keep the flexibility for multiple mesh types in the future.

SchoeniPhlippsn:
The MPI code works now so I think this is ready for another round of reviews

hoomd/MeshDefinition.cc
```diff
@@ -0,0 +1,111 @@
...
+    triangle_data.bcast(0, m_sysdef->getParticleData()->getExecConf()->getMPICommunicator());
+#endif
+    return triangle_data;
+    }
+
+//! Triangle array setter
+void MeshDefinition::setTriangleData(pybind11::array_t<int> triangles)
+    {
+    TriangleData::Snapshot triangle_data = getTriangleData();
+    // TriangleData::Snapshot triangle_data;
```
joaander:
```suggestion
```

hoomd/MeshDefinition.h
```diff
@@ -0,0 +1,142 @@
...
+/*! \file MeshDefinition.h
+    \brief Defines the MeshDefinition class
+ */
+
+#ifdef __HIPCC__
+#error This header cannot be compiled by nvcc
+#endif
+
+#include "BondedGroupData.h"
+//#include "IntegratorData.h"
```
joaander:
```suggestion
```

hoomd/MeshGroupData.h
```diff
@@ -0,0 +1,118 @@
...
+    //! Constructor for empty MeshGroupData
+    MeshGroupData(std::shared_ptr<ParticleData> pdata, unsigned int n_group_types);
+
+    //! Constructor to initialize from a snapshot
+    MeshGroupData(std::shared_ptr<ParticleData> pdata, const TriangleData::Snapshot& snapshot);
+
+    virtual ~MeshGroupData();
+
+//! Initialize from a snapshot
+// using MeshGroupData<group_size,Group,name,snap,bond>::initializeFromSnapshot;
```
joaander:
```suggestion
// mask the base class initializeFromSnapshot(const Snapshot& snapshot)
```

hoomd/MeshGroupData.h
```diff
@@ -0,0 +1,118 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Woverloaded-virtual"
+    void initializeFromSnapshot(const TriangleData::Snapshot& snapshot);
+#pragma clang diagnostic pop
+
+    //! Take a snapshot
+    // using MeshGroupData<group_size,Group,name,snap,bond>>::takeSnapshot;
```
joaander:
```suggestion
```

hoomd/MeshGroupData.h
```diff
@@ -0,0 +1,118 @@
...
+    std::map<unsigned int, unsigned int> takeSnapshot(snap& snapshot) const;
```
joaander, line 71: Why is this `snap&` and `initializeFromSnapshot` takes an explicit `TriangleData::Snapshot&`?

hoomd/md/AllBondPotentials.h
```diff
@@ -8,10 +8,12 @@
 #include "EvaluatorBondHarmonic.h"
 #include "EvaluatorBondTether.h"
 #include "PotentialBond.h"
+//#include "PotentialMeshBond.h"
```
joaander:
```suggestion
```

hoomd/md/AllBondPotentials.h
```diff
@@ -8,10 +8,12 @@
...
 
 #ifdef ENABLE_HIP
 #include "AllDriverPotentialBondGPU.cuh"
 #include "PotentialBondGPU.h"
+//#include "PotentialMeshBondGPU.h"
```
joaander:
```suggestion
```

hoomd/md/AllBondPotentials.h
```diff
@@ -27,21 +29,42 @@ namespace hoomd
...
+typedef PotentialBond<EvaluatorBondHarmonic, MeshBondData> PotentialMeshBondHarmonic;
+//! Mesh Bond potential force compute for FENE forces
+typedef PotentialBond<EvaluatorBondFENE, MeshBondData> PotentialMeshBondFENE;
+//! Mesh Bond potential force compute for Tethering forces
+typedef PotentialBond<EvaluatorBondTether, MeshBondData> PotentialMeshBondTether;
 
 #ifdef ENABLE_HIP
 //! Bond potential force compute for harmonic forces on the GPU
-typedef PotentialBondGPU<EvaluatorBondHarmonic, kernel::gpu_compute_harmonic_forces>
+typedef PotentialBondGPU<EvaluatorBondHarmonic, BondData, 2, kernel::gpu_compute_harmonic_forces>
```
joaander: #1305 (not merged yet) removes `AllBondPotentials.h`. These changes will need to be propagated to the template `.inc` files after that is merged.

hoomd/md/NeighborList.cc
```diff
@@ -737,6 +737,8 @@ void NeighborList::setSingleExclusion(std::string exclusion)
     if (exclusion == "bond")
         {
         addExclusionsFromBonds();
+        if (m_meshbond_data)
```
joaander: Make a new `meshbond` exclusion option so that users can separately control `bond` and `meshbond` exclusions.

hoomd/md/mesh/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2009-2022 The Regents of the University of Michigan.
+# Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+"""Mesh potentials for molecular dynamics."""
```
joaander, line 4: Add a brief description of the potentials and what they can be used for. Include a "See Also:" block that links to the original paper(s).

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,149 @@
+# Copyright (c) 2009-2022 The Regents of the University of Michigan.
+# Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+"""Mesh Bond potentials."""
```
joaander: Revise the docstrings in this module following those in the current `bond.py`. Add a note that emphasizes how mesh bonds compute over a meshbond data structure that is separate from the bond data structure.
Include a `See Also:` block pointing to `hoomd.mesh` for details on initializing the mesh bonds.

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def __init__(self, mesh):
+        params = TypeParameter("params", "types",
+                               TypeParameterDict(k=float, r0=float, len_keys=1))
+        self._add_typeparam(params)
+
+        super().__init__(mesh)
+
+
+class FENE(MeshPotential):
```
joaander:
```suggestion
class FENEWCA(MeshPotential):
```
This is now `FENEWCA` in `bonds.py`. Rename for consistency.

hoomd/md/mesh/potential.py
```diff
@@ -0,0 +1,101 @@
...
+import copy
+
+validate_mesh = OnlyTypes(Mesh)
+
+
+class MeshPotential(Force):
+    """Constructs the bond potential applied to a mesh.
+
+    Note:
+        :py:class:`MeshBond` is the base class for all bond potentials applied
```
joaander:
```suggestion
        :py:class:`MeshPotential` is the base class for all mesh potentials applied
```

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
+# Copyright (c) 2009-2022 The Regents of the University of Michigan.
+# Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+"""Implement Mesh."""
```
joaander: Add overview documentation, including a `See Also:` block that links to the objects that use meshes.

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
...
+        else:
+            self.size = len(triag)
+        self._triangles = triag
+
+    @log(category='sequence', requires_run=True)
+    def bonds(self):
+        """((*N*, 2) `numpy.ndarray` of ``uint32``): Mesh bonds.
+
+        A list of tuples of particle ids which encodes the
+        bonds within the mesh structure.
```
joaander, line 129: Are the bonds automatically identified from the triangles input?

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
...
+                self._remove()
+                return
+            if self._added:
+                self._remove()
+
+    @log(category='sequence')
+    def triangles(self):
+        """((*N*, 3) `numpy.ndarray` of ``uint32``): Mesh triangulation.
+
+        A list of triplets of particle ids which encodes the
```
joaander:
```suggestion
        A list of triplets of particle tags which encodes the
```

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
...
+from collections.abc import Sequence
+from hoomd.logging import log
+import numpy as np
+
+
+class Mesh(_HOOMDBaseObject):
+    """Data structure combining multiple particles into a mesh.
+
+    The mesh is defined by an array of triangles tht make up a
+    triangulated surface.
```
joaander:
```suggestion
    The mesh is defined by an array of triangles that make up a
    triangulated surface of particles. Each triangle consists of 
    three particle tags.
```

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
...
+        A list of tuples of particle ids which encodes the
+        bonds within the mesh structure.
+        """
+        if self._attached:
+            return self._cpp_obj.getBondData().group
+
+    @log(requires_run=True)
+    def energy(self):
+        """(float): Surface energy of the mesh."""
+        return self._cpp_obj.getEnergy()
```
joaander: How is this energy computed? Include a LaTeX formula that describes the computation unambiguously.

Also, is it correct that `Mesh` computes its own energy? Shouldn't the energy be the property of the meshbond and other computes?

sphinx-doc/module-hoomd-mesh.rst
```diff
@@ -0,0 +1,20 @@
...
+
+.. autosummary::
+    :nosignatures:
+
+    Mesh
+
+.. rubric:: Details
+
+.. automodule:: hoomd.mesh
+    :synopsis: HOOMD mesh data structure
```
joaander:
```suggestion
    :synopsis: Triangular mesh made of particles.
```

sphinx-doc/module-hoomd-mesh.rst
```diff
@@ -0,0 +1,20 @@
+.. Copyright (c) 2009-2022 The Regents of the University of Michigan.
+.. Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+hoomd.mesh
+------------------
```
joaander:
```suggestion
-----------
```

joaander (changes requested):
Thanks! I have some small tweaks to the code and suggestions to improve the documentation.

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,149 @@
...
+from hoomd.data.typeparam import TypeParameter
+from hoomd.data.parameterdicts import TypeParameterDict
+
+
+class Harmonic(MeshPotential):
+    r"""Harmonic bond potential.
+
+    :py:class:`Harmonic` specifies a harmonic potential energy between all
+    particles that share an edge within the mesh. For more details see
+    :py:class:`hoomd.md.bond.Harmonic`.
```
b-butler:
```suggestion
    `Harmonic` specifies a harmonic potential energy between all
    particles that share an edge within the mesh. For more details see
    `hoomd.md.bond.Harmonic`.
```

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Args:
+        mesh (:py:mod:`hoomd.mesh.Mesh`): Mesh data structure constraint.
```
b-butler:
```suggestion
        mesh (`hoomd.mesh.Mesh`): Mesh data structure constraint.
```
This is a class anyways.

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,149 @@
...
+
+            * ``k`` (`float`, **required**) - potential constant
+              :math:`[\mathrm{energy} \cdot \mathrm{length}^{-2}]`
+
+            * ``r0`` (`float`, **required**) - rest length
+              :math:`[\mathrm{length}]`
+
+    Examples::
+
+        harmonic = mesh.bond.Harmonic(mesh)
```
b-butler:
```suggestion
        harmonic = hoomd.md.mesh.bond.Harmonic(mesh)
```

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,149 @@
...
+
+    :py:class:`Harmonic` specifies a harmonic potential energy between all
+    particles that share an edge within the mesh. For more details see
+    :py:class:`hoomd.md.bond.Harmonic`.
+
+    Args:
+        mesh (:py:mod:`hoomd.mesh.Mesh`): Mesh data structure constraint.
+
+    Attributes:
+        parameter (TypeParameter[dict]):
```
b-butler:
```suggestion
        params (TypeParameter[dict]):
```

hoomd/md/mesh/bond.py
```diff
@@ -0,0 +1,149 @@
...
+            The parameter of the harmonic bonds for the defined mesh.
+            As the mesh can only have one type a type name does not have
+            to be stated. The dictionary has the following keys:
```
b-butler: Not true any more. This is still on all bond potentials.

hoomd/md/nlist.py
```diff
@@ -95,6 +101,14 @@ def __init__(self, buffer, exclusions, rebuild_check_delay, check_dist):
         params["exclusions"] = exclusions
         self._param_dict.update(params)
 
+        self._mesh = validate_mesh(mesh)
+
+    def _attach(self):
+        if self._mesh:
```
b-butler:
```suggestion
        if self._mesh is not None:
```
While your code is considered Pythonic, I find this explicit check to be clearer than relying on the truthiness of the variable.

hoomd/md/pytest/test_meshbond.py
```diff
@@ -0,0 +1,246 @@
+# Copyright (c) 2009-2022 The Regents of the University of Michigan.
```
b-butler, line 1: Are any of the tested objects supposed to be picklable or loggable? If so we should add the checks for that.

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
...
+        type name. Only one type per mesh can be defined!
+
+    Examples::
+
+        mesh = mesh.Mesh()
+        mesh.triangles = [[0,1,2],[0,2,3],[0,1,3],[1,2,3]]
+
+    """
+
+    def __init__(self, name=["mesh"]):
```
b-butler: In Python mutable default arguments are ill-advised as they can actually mutate and cause unexpected behavior.
```suggestion
    def __init__(self, name="mesh"):
```

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class Mesh(_HOOMDBaseObject):
+    """Data structure combining multiple particles into a mesh.
+
+    The mesh is defined by an array of triangles tht make up a
+    triangulated surface.
+
+    Args:
+        name ([`str`]): name of the mesh that also acts as a
```
b-butler:
```suggestion
        name (str): name of the mesh that also acts as a
```

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
...
+        type name. Only one type per mesh can be defined!
```
b-butler:
```suggestion
        type name.
```

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
...
+
+    """
+
+    def __init__(self, name=["mesh"]):
+
+        param_dict = ParameterDict(size=int,
+                                   types=OnlyIf(
+                                       to_type_converter([
+                                           str,
+                                       ] * 1),
```
b-butler:
```suggestion
                                   types=OnlyTypes([str], postprocessing=lambda x: [x], strict=True),
```
You can't use this directly, but this is more what you want, I think.

b-butler: You may have to keep preprocessing; I am not sure if querying from C++ with this will error or not.

You could also just accept the single string and configure the C++ side that way.

Or if this is to enable more types later, than we should leave as is and close this suggestion. Use a tuple in the `name` argument then.

hoomd/mesh.py
```diff
@@ -0,0 +1,117 @@
...
+        self._triangles = triag
+
+    @log(category='sequence', requires_run=True)
+    def bonds(self):
+        """((*N*, 2) `numpy.ndarray` of ``uint32``): Mesh bonds.
+
+        A list of tuples of particle ids which encodes the
+        bonds within the mesh structure.
+        """
+        if self._attached:
```
b-butler:
```suggestion
```
`requires_run` deals with this.

hoomd/pytest/test_mesh.py
```diff
@@ -0,0 +1,99 @@
...
+def test_empty_mesh(simulation_factory, two_particle_snapshot_factory):
+
+    sim = simulation_factory(two_particle_snapshot_factory(d=2.0))
+    mesh = Mesh()
+
+    assert mesh.size == 0
+    assert mesh.types == ["mesh"]
+    assert len(mesh.triangles) == 0
+    with pytest.raises(DataAccessError):
+        mesh.bonds == 0
```
b-butler:
```suggestion
        mesh.bonds
```

hoomd/pytest/test_mesh.py
```diff
@@ -0,0 +1,99 @@
...
+    with pytest.raises(TypeConversionError):
+        mesh.types = ["me", "sh"]
+    mesh.types = ["mesh"]
+    assert mesh.types == ["mesh"]
+
+    mesh.triangles = numpy.array([[0, 1, 2], [1, 2, 3]])
+
+    assert mesh.size == 2
+    assert numpy.array_equal(mesh.triangles, numpy.array([[0, 1, 2], [1, 2,
+                                                                      3]]))
```
b-butler:
```suggestion
    mesh_triangles = numpy.array([[0, 1, 2], [1, 2, 3]])
    mesh.triangles = mesh_triangles

    assert mesh.size == 2
    assert numpy.array_equal(mesh.triangles, mesh_triangles)
```

hoomd/pytest/test_mesh.py
```diff
@@ -0,0 +1,99 @@
...
+    with pytest.raises(MutabilityError):
+        mesh.types = ["vesicle"]
+    with pytest.raises(MutabilityError):
+        mesh.size = 3
+
+    mesh.triangles = numpy.array([[0, 1, 2], [1, 2, 3]])
+
+    assert mesh.size == 2
+    assert numpy.array_equal(mesh.triangles, numpy.array([[0, 1, 2], [1, 2,
+                                                                      3]]))
```
b-butler:
```suggestion
    mesh_triangles = numpy.array([[0, 1, 2], [1, 2, 3]])
    mesh.triangles = mesh_triangles

    assert mesh.size == 2
    assert numpy.array_equal(mesh.triangles, mesh_triangles)
```

sphinx-doc/module-md-mesh-bond.rst
```diff
@@ -0,0 +1,26 @@
+.. Copyright (c) 2009-2022 The Regents of the University of Michigan.
+.. Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+md.mesh.bond
+--------------
```
b-butler:
```suggestion
md.mesh.bond
------------
```

sphinx-doc/module-md-mesh.rst
```diff
@@ -0,0 +1,27 @@
+.. Copyright (c) 2009-2022 The Regents of the University of Michigan.
+.. Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+md.mesh
+--------------
```
b-butler:
```suggestion
md.mesh
-------
```

hoomd/BondedGroupData.cc
```diff
@@ -551,6 +566,18 @@ BondedGroupData<group_size, Group, name, has_type_mapping>::getMembersByIndex(
     return m_groups[group_idx];
     }
 
+/*! \param idx Tag of bonded group
+ * \return Type of bonded group
```
b-butler: The return type is void. This documentation doesn't actually describe what this function does.

hoomd/BondedGroupData.h
```diff
@@ -108,6 +108,17 @@ void serialize(Archive& ar, hoomd::group_storage<4>& s, const unsigned int versi
...
+//! Serialization of group_storage<6> (meshtriangles)
+template<class Archive>
+void serialize(Archive& ar, hoomd::group_storage<6>& s, const unsigned int version)
+    {
+    ar& s.tag[0];
+    ar& s.tag[1];
+    ar& s.tag[2];
+    ar& s.tag[3];
+    ar& s.tag[4];
+    ar& s.tag[5];
```
b-butler:
```suggestion
    ar(s.tag[0], s.tag[1], s.tag[2], s.tag[3], s.tag[4], s.tag[5]);
```
Preferred way to do this in cereal (https://uscilab.github.io/cereal/assets/doxygen/classcereal_1_1InputArchive.html#a88c7141a24a934978b1ef9894215e67f)

hoomd/BondedGroupData.h
```diff
@@ -652,22 +672,24 @@ class BondedGroupData
...
 
+    Nano::Signal<void()> m_group_num_change_signal; //!< Signal that is triggered when groups are
+                                                    //!< added or deleted (globally)
+                                                    //
+#ifdef ENABLE_HIP
+    GPUArray<unsigned int> m_condition; //!< Condition variable for rebuilding GPU table on the GPU
+    unsigned int m_next_flag;           //!< Next flag value for GPU table rebuild
+#endif
     private:
     bool m_groups_dirty; //!< Is it necessary to rebuild the lookup-by-index table?
```
b-butler: Not a change made, but this documentation has a (?)?

hoomd/MeshDefinition.h
```diff
@@ -0,0 +1,142 @@
...
+
+    The data structure stored in MeshDefinition is all the edge bonds and triangles within
+    a particle mesh. It is used to combine a set of particles into a bonded mesh. In doing so
+    we can asign different potentials (as bond or surface potentials) to the mesh data structure.
+    The class has to access system information such as the particle positions to define the mesh
+    vertices.
+
+    As any data structure class in MeshDefinition can potentially reference any
+   other, other classes can simply use the mesh data by giving the shared pointer to the referenced
+   class to the constructor of the onem that needs to refer to it. Note that using this setup, there
```
b-butler:
```suggestion
   class to the constructor of the one that needs to refer to it. Note that using this setup, there
```

hoomd/md/PotentialBond.h
```diff
@@ -345,6 +365,18 @@ template<class T> void export_PotentialBond(pybind11::module& m, const std::stri
...
+    \tparam T class type to export. \b Must be an instantiated PotentialBOnd class template.
+*/
+template<class T> void export_PotentialMeshBond(pybind11::module& m, const std::string& name)
+    {
+    pybind11::class_<T, ForceCompute, std::shared_ptr<T>>(m, name.c_str())
+        .def(pybind11::init<std::shared_ptr<SystemDefinition>, std::shared_ptr<MeshDefinition>>())
+        .def("setParams", &T::setParamsPython)
+        .def("getParams", &T::getParams);
+    }
+
```
b-butler: This will need to be moved to a source file. Also, @joaander given the template would we need to create an `.inc` file and use the CMake template feature?

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,1009 @@
...
+void MeshGroupData<group_size, Group, name, snap, bond>::initializeFromSnapshot(
+    const TriangleData::Snapshot& snapshot)
+    {
+    // check that all fields in the snapshot have correct length
+    if (this->m_exec_conf->getRank() == 0 && !snapshot.validate())
+        {
+        this->m_exec_conf->msg->error()
+            << "init.*: invalid " << name << " data snapshot." << std::endl
+            << std::endl;
+        throw std::runtime_error(std::string("Error initializing ") + name + std::string(" data."));
```
b-butler: With errors, we usually populate the error with the message now.

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,1009 @@
...
+            bonds[1].tag[0] = triag.tag[1];
+            bonds[1].tag[1] = triag.tag[2];
+            bonds[1].tag[2] = group_idx;
+            bonds[1].tag[3] = group_idx;
+            bonds[2].tag[0] = triag.tag[2];
+            bonds[2].tag[1] = triag.tag[0];
+            bonds[2].tag[2] = group_idx;
+            bonds[2].tag[3] = group_idx;
+
+            for (unsigned int i = 0; i < all_groups.size(); ++i)
```
b-butler, line 152:
```suggestion
            // Remove any duplicate bonds.
            for (unsigned int i = 0; i < all_groups.size(); ++i)
```

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,1009 @@
...
+                }
+            unsigned int j = 0;
+            for (j = 0; j < bond_id.size(); ++j)
+                {
+                triag.tag[3 + j] = bond_id[j];
+                }
+
+            for (unsigned int i = 0; i < bonds.size(); ++i)
+                {
+                triag.tag[3 + j] = (unsigned int)all_helper.size();
```
b-butler:
```suggestion
                triag.tag[3 + j] = static_cast<unsigned int>(all_helper.size());
```

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,1009 @@
...
+            }
+        }
+    else
+#endif
+        {
+        this->m_type_mapping = snapshot.type_mapping;
+
+        typeval_t t;
+        t.type = 0;
+        // t.type = snapshot.type_id[0];
```
b-butler:
```suggestion
```

hoomd/MeshGroupData.cc
```diff
@@ -0,0 +1,1009 @@
...
+                {
+                unsigned int group_tag = *active_tag_it;
+                rank_rtag_it = rank_rtag_map.find(group_tag);
+                if (rank_rtag_it == rank_rtag_map.end())
+                    {
+                    this->m_exec_conf->msg->error() << endl
+                                                    << "Could not find " << name << " " << group_tag
+                                                    << " on any processor. " << endl
+                                                    << endl;
+                    throw std::runtime_error("Error gathering " + std::string(name) + "s");
```
b-butler: Use `std::ostringstream` as elsewhere

hoomd/md/PotentialBond.h
joaander: Yes, I can make this change after merging #1305.

joaander: Done.

hoomd/MeshGroupData.h
SchoeniPhlippsn: Both mesh bonds and mesh triangles are constructed from `TriangleData::Snapshot&`, as the user only has to provide the triangulated data. The bond data which is based on this triangulation is extracted automatically in `MeshGroupData.cc`. To write mesh bond and mesh triangle data, however, they need to return `BondData::Snapshot` and `TriangleData::Snapshot`, respectively.

hoomd/mesh.py
SchoeniPhlippsn: Yes!

hoomd/md/pytest/test_meshbond.py
SchoeniPhlippsn: I don't think they have to be pickable or loggable as the regular bond potentials tests aren't testing for it either.

b-butler: @SchoeniPhlippsn, we test both in `hoomd/md/pytest/test_potential.py` so that seems to more of an oversight than the expectation. Could you add them for these test, and I can make an issue for the bond tests.

hoomd/mesh.py
```diff
@@ -103,13 +140,7 @@ def bonds(self):
...
-            return self._cpp_obj.getBondData().group
-
-    @log(requires_run=True)
-    def energy(self):
-        """(float): Surface energy of the mesh."""
-        return self._cpp_obj.getEnergy()
+        return self._cpp_obj.getBondData().group
 
     def _preprocess_type(self, typename):
         if isinstance(typename, Sequence):
```
b-butler:
```suggestion
        if not isinstance(typename, str) and isinstance(typename, Sequence):
```

hoomd/mesh.py
```diff
@@ -103,13 +140,7 @@ def bonds(self):
...
-        if self._attached:
-            return self._cpp_obj.getBondData().group
-
-    @log(requires_run=True)
-    def energy(self):
-        """(float): Surface energy of the mesh."""
-        return self._cpp_obj.getEnergy()
+        return self._cpp_obj.getBondData().group
 
     def _preprocess_type(self, typename):
```
b-butler, line 133: The last line (return statement) needs be unindented once.

hoomd/md/mesh/bond.py
```diff
@@ -101,16 +138,15 @@ class Tether(MeshPotential):
 
     :py:class:`Tether` specifies a Tethering potential energy between all
     particles that share an edge within the mesh. For more details see
-    :py:class:`hoomd.md.bond.Tethering`.
+    :py:class:`hoomd.md.bond.Tether`.
```
b-butler:
```suggestion
    `hoomd.md.bond.Tether`.
```

hoomd/md/mesh/bond.py
```diff
@@ -101,16 +138,15 @@ class Tether(MeshPotential):
 
     :py:class:`Tether` specifies a Tethering potential energy between all
```
b-butler:
```suggestion
    `Tether` specifies a Tethering potential energy between all
```

hoomd/mesh.py
```diff
@@ -75,6 +103,15 @@ def _remove_dependent(self, obj):
...
 
+    @property
+    def name(self):
+        """(`str`): Mesh name."""
+        return self.types[0]
+
+    @name.setter
+    def name(self, name):
+        self.types = [name]
+
```
b-butler: Do you expect to extend to multiple types in version 3? If so this will introduce problems.

b-butler (changes requested):
I just have some small comments this time around.

hoomd/mesh.py
SchoeniPhlippsn: For now, no. I might go for multiple triangle types later, though.

joaander: Then we should remove the name argument, and assign a default name in `types`. That way, when a future change adds support for multiple types, there is a clearly defined default behavior that the user can override.

joaander (approved):
Thanks!

We can merge this once @b-butler's `name` argument change request is adressed.

hoomd/mesh.py
SchoeniPhlippsn: done

b-butler (approved):

PR #1155, opened 2021-10-13
Use pre-commit official clang-format hook

b-butler:
## Description

Moves to using the pre-commit mirror of clang-format.

## Motivation and context

This allows clang-format to run automatically with all other formatters and linters with no special handling by users.

Resolves #???

## How has this been tested?

My local development environment and pre-commit install handled this fine.

## Change log

```
Added: clang-format to default run pre-commit hooks.
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

PR #1165, opened 2021-11-15
Fix pre commit clang format

joaander:
## Description

Apply clang-format to CUDA files (This change was applied and then reverted for unexplained reasons in #1155 ). 
Also, remove the now unneeded style check GitHub action.

## Motivation and context

CUDA files should be formatted along with all other C++ code.
`pre-commit.ci` now runs clang-format so `style_check.yml` now performs duplicate checks.

## How has this been tested?

The CI run on this pull request will test the changes.

## Change log

Internal change, no change log needed.

## Checklist:


hoomd/BondedGroupData.cuh
```diff
@@ -19,14 +19,14 @@ namespace hoomd
 //! Sentinel value
 const unsigned int GROUP_NOT_LOCAL = 0xffffffff;
 
-//! Storage for group members (GPU declaration)
-template<unsigned int group_size> union group_storage {
+    //! Storage for group members (GPU declaration)
+    template<unsigned int group_size> union group_storage {
     unsigned int tag[group_size]; // access 'tags'
     unsigned int idx[group_size]; // access 'indices'
     };
```
tommy-waltmann, line 26: This style of indentation doesn't seem consistent with what we enforce for all the other parts of hoomd. Did it get set this way on purpose, or do we need to fix some `clang-format` setting?

hoomd/BondedGroupData.cuh
```diff
@@ -19,14 +19,14 @@ namespace hoomd
...
 
-//! A union to allow storing a Scalar constraint value or a type integer (GPU declaration)
-union typeval_union {
+    //! A union to allow storing a Scalar constraint value or a type integer (GPU declaration)
+    union typeval_union {
     unsigned int type;
     Scalar val;
     };
```
tommy-waltmann, line 32: Same comment as above here.

hoomd/BondedGroupData.cuh
joaander: @tommy-waltmann #1155 introduced many of these changes as well. The `conda-forge` build of `clang-format 12.0.0` formats the code one way and the `pypi` build of `clang-format 12.0.1` and `13.0.0` format it this new way.

I cannot find any `clang-format` setting that controls this. `clang-format` regularly introduces styling changes with every new release. It seems strange that such a change would result from `12.0.0` to `12.0.1`, but it also seems strange that the `pypi` build would format code differently from the `conda-forge` build.

If you have the time to invest, feel free to find and change the appropriate setting or find the corresponding change in `clang-format` and propose a fix upstream.

tommy-waltmann (approved):

PR #1174, opened 2021-12-03
Include full error message details in exception messages.

### LINKED ISSUE #811: Use specific exception classes where appropriate

joaander:
## Description

Report full error message in the exception messages. Remove them from the `m_exec_conf->error()` output.

## Motivation and context

1) In some cases, the exception would tear down stderr before the full error message would be printed.
2) Even if the error message was printed, this change gives the user the full details of the error in one place rather than spread across many lines of output.

Resolves #811 

## How has this been tested?

The changes compile.

## Change log

```
Changed:

* Report full error message in the exception messages.
```

## Checklist:

### DIFF
--- a/hoomd/BondedGroupData.cc
+++ b/hoomd/BondedGroupData.cc
@@ -941,7 +941,7 @@ void BondedGroupData<group_size, Group, name, has_type_mapping>::rebuildGPUTable
             for (unsigned int k = 0; k < group_size; ++k)
                 oss << g.tag[k] << ((k != group_size - 1) ? ", " : " ");
             oss << "incomplete!";
-            throw std::runtime_error(s.str);
+            throw std::runtime_error(oss.str());
             }
 
         if (flag == m_next_flag)
--- a/hoomd/BondedGroupData.cc
+++ b/hoomd/BondedGroupData.cc
@@ -1046,8 +1046,8 @@ BondedGroupData<group_size, Group, name, has_type_mapping>::takeSnapshot(Snapsho
                 if (rank_rtag_it == rank_rtag_map.end())
                     {
                     std::ostringstream s;
-                    s << "Could not find " << name << " " << group_tag
-                      << " on any processor." throw std::runtime_error(s.str());
+                    s << "Could not find " << name << " " << group_tag << " on any processor.";
+                    throw std::runtime_error(s.str());
                     }
 
                 // store tag in index
--- a/hoomd/BondedGroupData.cc
+++ b/hoomd/BondedGroupData.cc
@@ -493,7 +493,7 @@ BondedGroupData<group_size, Group, name, has_type_mapping>::getGroupByTag(unsign
             std::ostringstream s;
             s << "Trying to get type or constraint value of " << name << " " << tag
               << " which does not exist!";
-            throw runtime_error(std::string(s.str()));
+            throw runtime_error(s.str());
             }
 
         typeval = m_group_typeval[group_idx];
--- a/hoomd/BondedGroupData.cc
+++ b/hoomd/BondedGroupData.cc
@@ -435,7 +435,7 @@ unsigned int BondedGroupData<group_size, Group, name, has_type_mapping>::getNthT
         {
         std::ostringstream s;
         s << name << " index " << n << " out of bounds!"
-                                  << "The number of " << name << "s is " << getNGlobal();
+          << "The number of " << name << "s is " << getNGlobal();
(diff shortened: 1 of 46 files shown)


hoomd/BondedGroupData.cc
```diff
@@ -490,9 +490,10 @@ BondedGroupData<group_size, Group, name, has_type_mapping>::getGroupByTag(unsign
         {
         if (group_idx == GROUP_NOT_LOCAL)
             {
-            m_exec_conf->msg->error() << "Trying to get type or constraint value of " << name << " "
-                                      << tag << " which does not exist!" << endl;
-            throw runtime_error(std::string("Error getting ") + name);
+            std::ostringstream s;
+            s << "Trying to get type or constraint value of " << name << " " << tag
+              << " which does not exist!";
+            throw runtime_error(std::string(s.str()));
```
b-butler, line 496:
```suggestion
            throw runtime_error(s.str());
```

hoomd/md/NeighborList.cc
```diff
@@ -1103,20 +1101,16 @@ void NeighborList::addOneThreeExclusionsFromTopology()
 
         if (nBondsA >= MAXNBONDS)
             {
-            m_exec_conf->msg->error()
-                << "nlist: Too many bonds to process exclusions for particle with tag: " << tagA
-                << endl;
-            m_exec_conf->msg->error() << "Maximum allowed is currently: " << MAXNBONDS - 1 << endl;
```
b-butler, line 1109: Do we not want to inform users of the maximum bonds they can have?

hoomd/md/PPPMForceCompute.cc
```diff
@@ -91,38 +89,31 @@ void PPPMForceCompute::setParams(unsigned int nx,
 
         if (!is_pow2(m_mesh_points.x) || !is_pow2(m_mesh_points.y) || !is_pow2(m_mesh_points.z))
             {
-            m_exec_conf->msg->error()
-                << "The number of mesh points along the every direction must be a power of two!"
-                << std::endl;
-            throw std::runtime_error("Error initializing charge.pppm");
+            std::ostringstream s;
+            s << "The number of mesh points along the every direction must be a power of two!";
+            throw std::runtime_error(s.str());
```
b-butler, line 94:
```suggestion
            throw std::runtime_error(
                "The number of mesh points along the every direction must be a power of two!");
```

b-butler (changes requested):
I only have a few small comments.

hoomd/md/NeighborList.cc
joaander: Done.

b-butler (commented):
:+1:

b-butler (approved):

mariaward10 (commented):
It looks good to go!

Issue #878, opened 2020-12-04
Refactor HPMC external fields and port to v3 API


joaander:
## Description

Port needed HPMC external fields to v3. We should also refactor the internals of how these fields are implemented as the code is very complex and hard to follow.

## Motivation and context

This issue is closely related with #551 which deals specifically with the walls and crosses over to MD as well.

As for the non-walls, we need to consider:
* hpmc.field.callback - What problem does this solve? Would it be better to provide a different interface for custom external potentials? Certainly in v3 it should be possible to leverage local snapshots.
* hpmc.field.external_field_composite - We should remove this and consider alternatives (i.e. HPMCIntegrator could store a list of external fields).
* hpmc.field.frenkel_ladd_energy - We should remove this. It is just a shortcut for lattice_field + remove_drift. Users should configure and manage those separately if they wish to.
* hpmc.field.lattice_field - Needed for frenkel-ladd compuations.

joaander:
Note: #937 enables support for C++ run time compiled external potentials. It validates that the external class is specifically of that type. We will need to generalize that when we enable additional external potentials when working on this issue.

tcmoore3:
@b-butler Doing Frenkel–Ladd calculations, which will require the simultaneous use of `lattice_field` and `remove_drift` after this issue is addressed, would be a good item in our how-to section of the docs.

tcmoore3:
> * hpmc.field.callback - What problem does this solve? Would it be better to provide a different interface for custom external potentials? Certainly in v3 it should be possible to leverage local snapshots.

Unless I'm missing something, this problem is solved with the jit-compiled `CPPExternalPotential`s, and we can safely remove `hpmc.field.callback`.

PR #1177, opened 2021-12-08
Port LatticeField to v3

tcmoore3:
## Description

The primary purpose of this PR is to port the lattice field to the v3 API. The current implementation only allows a single external field to be applied to the systems. This limitation will be addressed in a future PR.

## Motivation and context

This PR covers a portion of #878. 

## How has this been tested?

I have tested locally that particles remain in their reference positions and orientations with strong springs, and that their fluctuations grow in size with decreasing spring constants. I will add unit tests.

## Change log

```
Changed hoomd.hpmc.field.lattice_field to hoomd.hpmc.field.LatticeField
```

## Checklist:


hoomd/hpmc/ExternalFieldLattice.h
```diff
@@ -22,256 +22,278 @@ namespace hoomd
...
+
+        // connect updateMemberTags() method to maximum particle number change signal
+        //m_pdata->getGlobalParticleNumberChangeSignal()
+        //    .connect<ExternalFieldLattice, &ExternalFieldLattice::slotGlobalParticleNumChange>(this);
+        }  // end constructor
+
+    //! Destructor
+    ~ExternalFieldLattice()
+        {
+        /*
```
tcmoore3, line 54: I couldn't get this to compile

hoomd/hpmc/ExternalFieldLattice.h
```diff
@@ -22,256 +22,278 @@ namespace hoomd
...
+        //const BoxDim* const box_new = &m_pdata->getGlobalBox();
         const Scalar4 *position_old = position_old_arg, *orientation_old = orientation_old_arg;
-        const BoxDim* box_old = box_old_arg;
+        //const BoxDim* box_old = box_old_arg;
         if (!position_old)
             position_old = position_new;
         if (!orientation_old)
             orientation_old = orientation_new;
-        if (!box_old)
+        /*if (!box_old)
```
tcmoore3, line 276: I left the box stuff commented out here. We either need to remove this stuff or allow this field with variable box sizes (which will take a little more thought about the right way to do).

hoomd/hpmc/ExternalFieldLattice.h
```diff
@@ -348,237 +360,74 @@ template<class Shape> class ExternalFieldLattice : public ExternalFieldMono<Shap
...
+     */
     Scalar calcE(const unsigned int& index,
                  const vec3<Scalar>& position,
                  const quat<Scalar>& orientation,
                  const Scalar& scale = 1.0)
         {
         Scalar energy = 0.0;
-        if (m_latticePositions.isValid())
-            {
+        //if (m_lattice_positions.isValid())
```
tcmoore3, line 420: This commented out code will leave once I figure out the the particle change number signal bit.

hoomd/hpmc/field.py
```diff
@@ -32,11 +37,10 @@ def __init__(self):
         # nothing else to do.
 
 
-class lattice_field(_external):
+class LatticeField(_HOOMDBaseObject):
```
tcmoore3, line 40: This class will need to inherit from a different (probably new) class that all of the external fields inherit from.

hoomd/hpmc/field.py
```diff
@@ -79,162 +83,102 @@ class lattice_field(_external):
 
     """
 
+    def __init__(self, reference_positions, reference_orientations,
+                 k_translational, k_orientational, sym_ors):
+        pass
+        self.compute_name = "lattice_field"
```
tcmoore3, line 89: Do we need this?

hoomd/hpmc/ExternalFieldLattice.h
tcmoore3: This works now.

hoomd/hpmc/ExternalFieldLattice.h
tcmoore3: I figured that other bit out. This is gone now.

tcmoore3:
Some to-dos:

- [ ] Update (write?) unit tests for `LatticeField` class
- [ ] Decide what to do when the box can change and a `LatticeField` is active

hoomd/hpmc/field.py
tcmoore3: Done.

hoomd/hpmc/field.py
tcmoore3: No.

hoomd/hpmc/field.py
```diff
@@ -32,208 +37,136 @@ def __init__(self):
...
+    :py:class:`LatticeField` specifies that a harmonic spring is added to every
+    particle:
 
-    .. warning::
-        The lattice energies and standard deviations logged by this class are multiplied by the spring constant.
+    .. math::
 
-    Example::
+        V_{i}(r_i)  = k_{trans} \cdot (r_i-r_{0,i})^2 \\
+        V_{i}(q_i)  = k_{rot} \cdot (q_i-q_{0,i})^2
```
joaander, line 72: Add a more complete description of how this is computed with respect to the given symmetries.

sphinx-doc/module-hpmc-field.rst
```diff
@@ -0,0 +1,17 @@
...
+------------
+
+.. rubric:: Overview
+
+.. py:currentmodule:: hoomd.hpmc.field
+
+.. autosummary::
+    :nosignatures:
+
+    LatticeField
```
joaander, line 11: I find `hpmc.field.LatticeField` both redundant and not informative. How about `hpmc.field.Harmonic` or `hpmc.field.Einstein`? I'm open to other suggestions.

hoomd/hpmc/ExternalFieldLattice.h
joaander: I say remove. Future developers can add scaling back in the correct way if they find a strong use case for it.

hoomd/hpmc/ExternalFieldLattice.h
```diff
@@ -346,180 +360,9 @@ template<class Shape> class ExternalFieldLattice : public ExternalFieldMono<Shap
...
-        Scalar second_moment = m_EnergySqSum / double(m_num_samples);
-        return sqrt(second_moment - (first_moment * first_moment));
-        }
-
     protected:
-    // These could be a little redundant. think about this more later.
+    //! Calculate the energy associated with the deviation of a single particle from its reference
+    //! position
     Scalar
     calcE_trans(const unsigned int& index, const vec3<Scalar>& position, const Scalar& scale = 1.0)
```
joaander, line 367: Remove `scale` here too and in the other methods.

joaander (changes requested):
Thanks! This is looking good. I've requested some changes.

Also, be careful with `git commit -a`, it appears you committed a different version of the tutorials. Revert with:
```
cd sphinx-doc/tutorial
git checkout 92038ecc811ca450acd3c2912eb818c4ebece06b
cd ..
git add tutorial
git commit
```

hoomd/hpmc/field.py
```diff
@@ -32,208 +37,136 @@ def __init__(self):
...
+    """
 
-    * :math:`k_r` - translational spring constant.
-    * :math:`r_{o}` - lattice positions (in distance units).
-    * :math:`k_q` - rotational spring constant.
-    * :math:`q_{o}` - lattice orientations (quaternion)
 
-    Once initialized, the compute provides the following log quantities that can be logged via analyze.log:
+class LatticeField(ExternalField):
+    R"""Restrain particles on a lattice
```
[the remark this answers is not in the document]
tcmoore3: Just to be clear, this docstring needs the `r"""` because of the backslashes in the equations, right?

hoomd/hpmc/field.py
```diff
@@ -32,208 +37,136 @@ def __init__(self):
...
+
+    :py:class:`LatticeField` specifies that a harmonic spring is added to every
+    particle:
 
-    .. warning::
-        The lattice energies and standard deviations logged by this class are multiplied by the spring constant.
+    .. math::
 
-    Example::
+        V_{i}(r_i)  = k_{trans} \cdot (r_i-r_{0,i})^2 \\
```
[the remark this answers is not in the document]
tcmoore3: Yes, definitely.

hoomd/hpmc/field.py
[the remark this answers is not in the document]
joaander: Perhaps, but not necessarily. In may places in the docs, I need to use `"""` and escape every backslash so that the single backslash character can be used for line continuation to correctly format sphinx docs while meeting the flake8 checks.

hoomd/hpmc/field.py
```diff
@@ -32,208 +37,136 @@ def __init__(self):
...
-        hpmc.field.lattice_field(mc=mc, position=fcc_lattice, k=1000.0);
-        log = analyze.log(quantities=['lattice_energy'], period=100, filename='log.dat', overwrite=True);
+    where :math:`k_{trans}` and :math:`k_{rot}` correspond to the arguments
+    ``k_translational`` and ``k_rotational``, respectively, :math:`r_i` and
+    :math:`q_i` are the position and orientation of particle :math:`i`, and the
+    :math:`0` subscripts lattice quantities.
+
+    Note:
+        The factor of 1/2 is not included in the formulas; specify your spring
+        constants accordingly.
```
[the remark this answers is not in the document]
tcmoore3: I made that change.

hoomd/hpmc/field.py
tcmoore3: Done.

sphinx-doc/module-hpmc-field.rst
tcmoore3: I changed this to `hpmc.field.Harmonic`, and also updated the C++ class to match.

hoomd/hpmc/field.py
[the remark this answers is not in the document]
tcmoore3: I changed it to include the factor of 1/2.

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -0,0 +1,468 @@
...
+                          pybind11::array_t<double> r0,
+                          Scalar k,
+                          pybind11::array_t<double> q0,
+                          Scalar q,
+                          pybind11::array_t<double> symRotations)
+        : ExternalFieldMono<Shape>(sysdef), m_k_translational(k), m_k_rotational(q)
+        {
+        setReferencePositions(r0);
+        setReferenceOrientations(q0);
+        setSymmetricallyEquivalentOrientations(symRotations); // TODO: check for identity?
```
joaander, line 44: Reading the code, it appears that the user is required to include the identity in the list. This should be documented.

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -0,0 +1,468 @@
...
+        std::vector<size_t> dims(2);
+        dims[0] = this->m_reference_positions.size();
+        dims[1] = 3;
+        // the cast from vec3<Scalar>* to Scalar* is safe since vec3 is tightly packed without any
+        // padding. This also makes a copy so, modifications of this array do not effect the
+        // original reference positions.
+        const auto reference_array = pybind11::array_t<Scalar>(
+            dims,
+            reinterpret_cast<const Scalar*>(this->m_reference_positions.data()));
+        // This is necessary to expose the array in a read only fashion through C++
```
joaander, line 181: If this returns a copy, why does it make it not writeable?
@b-butler does this need any special handling to interoperate with #1178?

hoomd/hpmc/field.py
```diff
@@ -32,208 +37,142 @@ def __init__(self):
...
+    where :math:`k_{trans}` and :math:`k_{rot}` correspond to the arguments
+    ``k_translational`` and ``k_rotational``, respectively, :math:`r_i` and
+    :math:`q_i` are the position and orientation of particle :math:`i`, and the
+    :math:`0` subscripts denote the given reference quantities. When multiple
+    orientations are passed to ``symmetries``, the rotational energy for each
+    particle is computed for not only the orientation of the particle at its
+    current orietation, but also at the orientation resulting from the rotation
+    represented by each quaternion in ``symmetries`` yielding a list of
+    energies :math:`\mathbb{V}`; the contribution to the energy is the minimum
+    value in this list, :math:`\mathrm{min}(\mathbb{V})`.
```
joaander, line 87: Encode this description in the math equation succinctly. Something like:
```\frac{1}{2} k_{rot} \cdot \min_j \left[ (q_i-q_{0,i} \cdot q_{\mathrm{symmetry},j})^2 \right]```
Then the text description becomes: "where q_{\mathrm{symmetry}} is the given set of symmetric orientations."

File: hoomd/hpmc/field.py [hunk 15]
```diff
@@ -32,208 +37,142 @@ def __init__(self):
...
```
joaander, line 88: Need an `Attributes:` section that documents the `param_dict` entries. It can mostly be copied and pasted from `Args:`.

joaander (changes requested):
Thanks! Some additional changes are needed.

hoomd/hpmc/ExternalFieldHarmonic.h
tcmoore3: Added.

hoomd/hpmc/ExternalFieldHarmonic.h
tcmoore3: Deferring to @b-butler

hoomd/hpmc/field.py
tcmoore3: Yes, beautiful. I couldn't figure out how to write this succinctly because I was unaware of the `\min` LaTeX command. Changed.

hoomd/hpmc/ExternalFieldHarmonic.h
b-butler: The reason this is not writable (I believe this is a copy paste from another updater) is because it would be deceiving as to what is happening. It is the copy-modify-write problem again. Given the fact that NumPy is not part of the Python standard library I did not create an array class that continually updates. It is possible though.

joaander: That makes sense. This is attempting to give some notion of `const` or immutability with the read/modify/write mechanism.

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -0,0 +1,468 @@
...
+            quat<Scalar> dq = q0 - equiv_orientation;
+            dqmin = (i == 0) ? norm2(dq) : fmin(dqmin, norm2(dq));
+            }
+        return Scalar(0.5) * m_k_rotational * dqmin;
+        }
+
+    Scalar calcE_rot(const unsigned int& index, const Shape& shape)
+        {
+        if (!shape.hasOrientation())
+            return Scalar(0.0);
```
joaander, line 398: Add a note in the documentation that the rotational energy is zero for non-orientable particles (such as `hpmc.integrate.Sphere`).

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -0,0 +1,468 @@
...
+#endif
+
+        return dE;
+        }
+
+    /** This function gets called during the update step in the MC integrator
+     *
+     * What purpose does this function serve?
+     */
+    Scalar getEnergy(uint64_t timestep)
```
joaander, line 313: Would users be interested in accessing energy components separately in `energy_translational` and `energy_rotational`? If so, we could add these properties. Keep the `energy` property that computes the total.

joaander (changes requested):
Thanks for the fast turnaround. There are just a few more updates needed. 
In addition to the inline comments, `Harmonic` should live somewhere under `hpmc.external`. Let's discuss on Slack.

The suggestion to add separate properties for translational and rotational energy is optional, but would be nice.

hoomd/hpmc/ExternalFieldHarmonic.h
tcmoore3: Yes, I think users would be interested in something like that. I haven't worked out the details but I imagine this would be useful for getting rotational and translational contributions to the free energy somehow. And it's easy enough to add, so I'll put it in.

hoomd/hpmc/external/field.py
```diff
@@ -1,243 +1,223 @@
...
-        enlist = hoomd.hpmc.data._param.ensure_list
-        self.cpp_compute = cls(hoomd.context.current.system_definition,
-                               enlist(position), float(k), enlist(orientation),
-                               float(q), enlist(symmetry))
-        hoomd.context.current.system.addCompute(self.cpp_compute,
-                                                self.compute_name)
-        if not composite:
-            mc.set_external(self)
+class Harmonic(ExternalField):
+    r"""Restrain particle positions and orientations with harmonic springs.  # noqa: E501,D415,W505
```
joaander, line 51:
```suggestion
    r"""Restrain particle positions and orientations with harmonic springs.
```
As written, the `noqa` is part of the docstring. We need to format the docstring appropriately so that it obeys the linter rules.

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -306,11 +306,12 @@ template<class Shape> class ExternalFieldHarmonic : public ExternalFieldMono<Sha
         return dE;
         }
 
-    /** This function gets called during the update step in the MC integrator
+    /** This function gets exported to python to act as a property
```
joaander, line 309: Except that this isn't exported as a property with pybind. Typically the first sentence in a documentation comment describes what the method does: "Compute the total external energy on the system."

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -407,17 +408,25 @@ template<class Shape> class ExternalFieldHarmonic : public ExternalFieldMono<Sha
      * acceptance criteria, since it's the energy difference that matters for the latter.
      */
     Scalar
-    calcE(const unsigned int& index, const vec3<Scalar>& position, const quat<Scalar>& orientation)
+    calcE(const unsigned int& index, const vec3<Scalar>& position, const quat<Scalar>& orientation,
+            bool include_translational=true, bool include_rotational=true)
```
joaander: I prefer not to introduce boolean parameters that get must get passed down the chain. Instead implement `getEnergyTranslational` and `getEnergyRotational` as separate methods.

hoomd/hpmc/ExternalFieldHarmonic.h
joaander: On second review, I noticed some issues here. The method only the `calcE_rot(const unsigned int& index, const Shape& shape)` knows whether the shape is orientable. I haven't search the whole codebase, but it appears this method is not even called. `getEnergy`, `energydiff`, and `calculateDeltaE` all call the other signature.

1) Remove `calcE_rot` and see if it compiles, that would verify it is not used.
2) Update the docs to remove the mention of a zero rotational energy. Users are responsible for setting `k_rotational=0` for non-orientable particles.

hoomd/hpmc/external/field.py
```diff
@@ -1,243 +1,223 @@
...
-
-        Example::
-            mc = hpmc.integrate.sphere(seed=415236);
-            lattice = hpmc.field.lattice_field(mc=mc, position=fcc_lattice, k=exp(15));
-            run(20000)
-            avg_eng = lattice.get_average_energy() //  should be about 1.5kT
+    @log(requires_run=True)
+    def energy_translational(self):
+        """float: The energy associated with positional fluctuations \
+            :math:`[\\mathrm{energy}]`.  # noqa
```
joaander, line 201:
```suggestion
            :math:`[\\mathrm{energy}]`. 
```
Don't include noqa in docstrings.

hoomd/hpmc/external/field.py
```diff
@@ -1,243 +1,223 @@
...
-        timestep = hoomd.context.current.system.getCurrentTimeStep()
-        return self.cpp_compute.getSigma(timestep)
+        timestep = self._simulation.timestep
+        return self._cpp_obj.getEnergy(timestep, False, True)
 
 
-class external_field_composite(_external):
-    R""" Manage multiple external fields.
+class external_field_composite(_External):  # noqa
+    R"""Manage multiple external fields.  # noqa: this will change in v3
```
joaander, line 220:
```suggestion
    r"""Manage multiple external fields.
```
Don't include noqa in docstrings.

I notice there are many more docstrings that include noqa, please fix all of them.

hoomd/hpmc/external/field.py
```diff
@@ -329,8 +309,8 @@ def add_field(self, fields):
             self.cpp_compute.addExternal(field.cpp_compute)
 
 
-class wall(_external):
-    R""" Manage walls (an external field type).
+class wall(_External):  # noqa: name will change in v3
```
joaander, line 312: `wall` doesn't belong in the `field.py` file. You can create a `wall.py` to keep the old v2 code for now and add it to the exceptions for pre-commit and flake8 instead of specifying noqa on every line.

joaander (changes requested):
Thanks! There are more updates needed before merging.

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -0,0 +1,482 @@
...
+                }
+
+            const size_t N_particles = ref_pos.shape(0);
+            const size_t dim = ref_pos.shape(1);
+            if (N_particles != m_pdata->getNGlobal() || dim != 3)
+                {
+                throw std::runtime_error("The array must be of shape (N_particles, 3).");
+                }
+            const double* rawdata = static_cast<const double*>(ref_pos.data());
+            for (size_t i = 0; i < std::min<unsigned long>(N_particles, m_pdata->getNGlobal()); i++)
```
b-butler, line 81: You error if `N_particles != m_pdata->getNGlobal()` so no need for the min call.

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -0,0 +1,482 @@
...
+                {
+                const size_t array_index = i * 3;
+                this->m_reference_positions[i] = vec3<Scalar>(rawdata[array_index],
+                                                              rawdata[array_index + 1],
+                                                              rawdata[array_index + 2]);
+                }
+            }
```
b-butler, line 88:
```suggestion
            for (size_t i = 0; i < N_particles; i+=3)
                {
                this->m_reference_positions[i] = vec3<Scalar>(rawdata[i]],
                                                              rawdata[i + 1],
                                                              rawdata[i + 2]);
                }
            }
```

b-butler: This can be applied to the other loops for this purpose.

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -0,0 +1,482 @@
...
+        }
+
+    /** Calculate the change in energy for trial moves
+     *
+     * This function currently ignores any information associated with box changes.
+     * However, this function only gets called in the box updater, so it should not ignore box
+     * changes. But why would you have a lattice field and a box updater active in the same
+     * simulation?
+     */
+    double calculateDeltaE(const Scalar4* const position_old_arg, // why is this a Scalar4?
```
b-butler, line 265:
```suggestion
    double calculateDeltaE(const Scalar4* const position_old_arg,
```

hoomd/hpmc/ExternalFieldHarmonic.h
```diff
@@ -0,0 +1,482 @@
...
+
+    protected:
+    //! Calculate the energy associated with the deviation of a single particle from its reference
+    //! position
+    Scalar calcE_trans(const unsigned int& index, const vec3<Scalar>& position)
+        {
+        ArrayHandle<unsigned int> h_tags(m_pdata->getTags(),
+                                         access_location::host,
+                                         access_mode::read);
+        int3 dummy = make_int3(0, 0, 0);
```
b-butler, line 366: This is only used in box wap. Go ahead and create it there.

hoomd/hpmc/external/field.py
```diff
@@ -1,243 +1,223 @@
...
+            `float`): the reference orientations, to which particles are
+            restrained :math:`[\mathrm{dimensionless}]`.
+        k_translational (`float`): translational spring constant
+            :math:`[\mathrm{energy} \cdot \mathrm{length}^{-2}]`.
+        k_rotational (`float`): rotational spring constant
+            :math:`[\mathrm{energy}]`.
+        symmetries ((*N_sym*, 4) `numpy.ndarray` of
+            `float`): the orientations that are equivalent through symmetry,
+            i.e., the rotation quaternions that leave the particles unchanged
+            :math:`[\mathrm{dimensionless}]`.
```
b-butler, line 67: Add note here that this must include the identity quaternion rather than a note below.

hoomd/hpmc/external/field.py
```diff
@@ -1,243 +1,223 @@
...
-            for k in ks:
-              lattice.set_params(k=k, q=0.0);
-              lattice.reset();
-              run(1000)
+    def __init__(self, reference_positions, reference_orientations,
+                 k_translational, k_rotational, symmetries):
+        param_dict = ParameterDict(
+            reference_positions=NDArrayValidator(dtype=np.float32,
+                                                 shape=(None, 3)),
+            reference_orientations=NDArrayValidator(dtype=np.float32,
```
b-butler, line 117: You require double's in the signatures in C++ but use single precision here. That involves a needless copy. Change one or the other.

hoomd/hpmc/external/field.py
b-butler: Most of these are left overs from v2. For those I would recommend removal. Most of these should cause errors. For those that may be useful for v3, we can do as you suggest for walls and for the rest using noqa seems appropriate to not create work that will likely have to be redone.

For the classes and functions ported though, they should not have noqa.

hoomd/hpmc/pytest/test_external.py
```diff
@@ -0,0 +1,188 @@
...
+import pytest
+import numpy as np
+
+valid_constructor_args = [
+    dict(reference_positions=[[0, 0, 0]],
+         reference_orientations=[[1, 0, 0, 0]],
+         k_translational=1.0,
+         k_rotational=1.0,
+         symmetries=[[1, 0, 0, 0]]),
+]
```
b-butler, line 17: If we are only going to test one combination of parameters no need to make the test parametrized. If we wanted to test multiple combinations then that would make sense.

hoomd/hpmc/pytest/test_external.py
```diff
@@ -0,0 +1,188 @@
...
+    else:
+        reference_positions = [[0, 0, 0], [0, 0, 0]]
+        reference_orientations = [[1, 0, 0, 0], [1, 0, 0, 0]]
+    lattice = hoomd.hpmc.external.field.Harmonic(
+        reference_positions=reference_positions,
+        reference_orientations=reference_orientations,
+        k_translational=k_trans,
+        k_rotational=1.0,
+        symmetries=[[1, 0, 0, 0]])
+    mc.external_potential = lattice
```
b-butler, line 124: This code appears in most tests. It would be a great canidate for a fixture that returns a simulation populated with the integrator and external potential.

hoomd/hpmc/pytest/test_external.py
```diff
@@ -0,0 +1,188 @@
...
+    # make some moves and make sure the different energies are not zero
+    sim.run(10)
+    assert lattice.energy_rotational != 0.0
+    assert lattice.energy_translational != 0.0
+    assert np.allclose(lattice.energy,
+                       lattice.energy_translational + lattice.energy_rotational)
+
+    # set k_rotational to zero and ensure the rotational energy is zero
+    lattice.k_rotational = 0
+    sim.run(0)
```
b-butler, line 144: This should be unnecessary.

hoomd/hpmc/pytest/test_external.py
```diff
@@ -0,0 +1,188 @@
...
+    assert lattice.energy_rotational == 0.0
+
+    # set k_translational to zero and ensure the translational energy is zero
+    lattice.k_translational = 0
+    sim.run(0)
```
b-butler, line 149: This should be unnecessary.

b-butler (changes requested):
The PR overall looks good. I have some comments to address mostly for code cleanness or clarity.

hoomd/hpmc/ExternalFieldHarmonic.h
tcmoore3: We resolved this in real-time.

hoomd/hpmc/external/field.py
tcmoore3: A lot of the `noqa`s are indeed in docs for classes that have not been ported over to v3 yet. I removed several of those classes and there are no more `noqa`s in the classes relevant to this PR.

hoomd/hpmc/ExternalFieldHarmonic.h
tcmoore3: Good catch! Changed.

hoomd/hpmc/ExternalFieldHarmonic.h
tcmoore3: Do you just mean move it to the line before the `box.wrap()` line?

hoomd/hpmc/external/field.py
tcmoore3: Changed.

hoomd/hpmc/pytest/test_external.py
tcmoore3: Added more sets of parameters to test.

hoomd/hpmc/pytest/test_external.py
tcmoore3: Feel free to clean this up if desired.

hoomd/hpmc/pytest/test_external.py
tcmoore3: Even if the energy property has `requires_run=True`?

hoomd/hpmc/pytest/test_external.py
tcmoore3: Even if the energy property has `requires_run=True`?

hoomd/hpmc/ExternalFieldHarmonic.h
tcmoore3: > This can be applied to the other loops for this purpose.

...with appropriate modifications of course (:

tcmoore3:
@joaander @b-butler This is ready for another review. My preference is to go ahead and merge this in if there are real issues and only cosmetic issues. Feel free to go ahead and make any style changes you prefer.

joaander (approved):
Thanks!

I pushed my suggested documentation style changes directly to the branch.

hoomd/hpmc/pytest/test_external.py
b-butler: Yes because run is called above.

joaander:
Tests are now failing: https://github.com/glotzerlab/hoomd-blue/runs/4870480987?check_suite_focus=true#step:6:6008

I fixed the pre-commit failure on master, you can merge the fix in.

b-butler (approved):

Issue #705, opened 2020-07-31
BoxResize box assignment updates initial reference


joaander:
## Description

BoxResize holds onto the initial BoxDim it is given by reference. When assigning a new Python Box object, the original reference is updated by the Box assignment operator.

## Script

```python
# Include a minimal script that reproduces the problem
import hoomd

device = hoomd.device.CPU()
sim = hoomd.Simulation(device)
sim.create_state_from_gsd('test.gsd')

box1_initial = hoomd.Box.cube(10)
box2_initial = hoomd.Box.cube(20)
resize = hoomd.update.BoxResize(box1_initial, box2_initial, hoomd.variant.Ramp(0, 0, 1, 100000), 100)

sim.operations.updaters.append(resize)
sim.run(1)
print("box1_initial:", box1_initial.Lx, " / resize.box1: ", resize.box1.Lx)
print(resize.box1 is box1_initial)

box1_initial.Lx = 25
print("box1_initial:", box1_initial.Lx, " / resize.box1: ", resize.box1.Lx)
print(resize.box1 is box1_initial)

box1_update = hoomd.Box.cube(30)
resize.box1 = box1_update
print("box1_initial:", box1_initial.Lx, " / box1_update: ", box1_update.Lx, " / resize.box1: ", resize.box1.Lx)
print(resize.box1 is box1_update)

```

## Output

```
box1_initial: 10.0  / resize.box1:  10.0
True
box1_initial: 25.0  / resize.box1:  25.0
True
box1_initial: 30.0  / box1_update:  30.0  / resize.box1:  30.0
True
```

## Expected output

```
box1_initial: 10.0  / resize.box1:  10.0
True
box1_initial: 25.0  / resize.box1:  25.0
True
box1_initial: 25.0  / box1_update:  30.0  / resize.box1:  30.0
True
```

## Other notes

We could quickly fix this bug by using pointers instead of references in `BoxResize`, however the existence of the bug exposes an anti-pattern - two member variables that need to be kept in sync. Perhaps we could solve this like we do with Trigger and ParticleFilter and make python features of `Box` available as a subclass of `BoxDim`?

joaander:
While this bug needs to be fixed, it is not critical for the 3.0.0 release.

b-butler:
I recommend `Box` inheriting from `BoxDim`. I don't know why I didn't do that to start.

PR #1178, opened 2021-12-09
Supported Nested Modification of Data Structures

b-butler:
## Description

Adds support for a read, modify, write model of modification automatically with HOOMD's attributes even with syncing to C++. This involves various other changes. These are relevant in that they arise from bug fixes with increased testing and changes to interal APIs.

## Motivation and context

This is to prevent users from modifying copies where changes do not persist.

## How has this been tested?

Many tests have been added for the changes.

## Change log

```
Added: HOOMD attributes support nested modification.
```

## Checklist:


b-butler (commented):
- [x] Add documentation to `ARCHITECTURE.md`
- [x] Add isolation tests to `TypeParameterDict`
- [ ] Add isolation tests to `ParameterDict`

hoomd/update/box_resize.py
```diff
@@ -55,17 +63,17 @@ class BoxResize(Updater):
...
+                               box2=Box,
                                variant=Variant,
                                filter=ParticleFilter)
         params['box1'] = box1
         params['box2'] = box2
         params['variant'] = variant
         params['trigger'] = trigger
         params['filter'] = filter
+        for attr in ("box1", "box2"):
+            params._set_special_getset(attr, _box_getter, _box_setter)
```
joaander, line 76: Does this fix #705?

hoomd/pytest/test_simulation.py
```diff
@@ -333,8 +333,9 @@ def test_run_limit(simulation_factory, lattice_snapshot_factory):
         sim.run(-1)
 
 
-def test_seed(simulation_factory, lattice_snapshot_factory):
-    sim = simulation_factory()
```
joaander, line 337: Why this change? `simulation_factory` takes care of some important things like reducing the sorter grid size to run unit tests faster. While the sorter is never activated in this test, this pattern may get copied and pasted elsewhere.

hoomd/conftest.py
```diff
@@ -71,6 +71,7 @@ def make_simulation(snapshot=None, domain_decomposition=None):
                 sim.create_state_from_snapshot(snapshot)
             else:
                 sim.create_state_from_snapshot(snapshot, domain_decomposition)
+        sim.seed = numpy.random.randint(100_000)
```
joaander, line 74: I see, this is the change is why the see test no longer users `simulation_factory`. Use a fixed seed so that failures are as repeatable as possible.

joaander (changes requested):
Thanks! This is a very complete effort to provide an abstract API that operates the way users expect. There are a few changes I would like to see to the test rig, but otherwise this is good to merge.

hoomd/update/box_resize.py
b-butler: No idea but I could look into it.

hoomd/pytest/test_simulation.py
b-butler: Yes this is to prevent developers from not setting a seed when they should and avoid testing error messages.

hoomd/update/box_resize.py
joaander: No problem, we can solve in another PR if appropriate. Though it is related in spirit to the changes you mare making in this PR.

b-butler: I have verified that currently this is still a bug.

joaander (approved):
Thanks!

PR #1189, opened 2022-01-05
Remove diameter usage from FENE and DLVO

joaander:
## Description

Make `delta` in `FENE` and `a1,a2` in DLVO explicit type parameters. Also, rename `FENE` to `FENEWCA` to more clearly label what the potential computes.

## Motivation and context

This avoids confusion regarding the delta value, cutoff radii and the hard-coded `1.0` parameter as discussed in #1050.

Resolves #1182

## How has this been tested?

I adjusted the existing unit tests to provide potential parameters consistent with the particle diameters.

## Change log

```
Changed:

* [breaking] Renamed ``md.bond.FENE`` to ``md.bond.FENEWCA``
* ``md.bond.FENEWCA`` takes a user provided ``delta`` parameter and ignores the particle diameters.
* [breaking] ``md.pair.DLVO`` takes user provided ``a1`` and ``a2`` parameters and ignores the particle diameters.
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

Issue #1181, opened 2021-12-14
Remove SLJ


joaander:
## Description

Remove the SLJ pair potential.

## Motivation and context

The new ExpandedLJ potential in #1152 replaces SLJ. Remove SLJ in the next beta release after adding ExpandedLJ.

joaander:
@cbkerr would you like to make this change?

cbkerr:
Yes!

PR #1190, opened 2022-01-05
Remove unused r_on and default_r_on parameters.

### LINKED ISSUE #1047: Remove r_on for potentials that do not support "xplor" mode
## Description
Many potentials do not support the "xplor" energy/force shifting, but allow setting `default_r_on` and `obj.r_on[(type, type)]`. This does not make much sense as they can never be meaningfully set. I think removing the typeparameter and constructor argument would be best.

## Motivation and context
I believe this would increase the API coherence.

joaander:
## Description

Removed unused `r_on` and `default_r_on` parameters in MD pair forces.

## Motivation and context

Some potentials do not support xplor smoothing, so the `r_on` parameters are unused.

Resolves #1047 

## How has this been tested?

Existing unit tests pass.

I manually checked that `r_on` is indeed not present:
```
>>> nlist = hoomd.md.nlist.Cell(buffer=0.4)
>>> f = hoomd.md.pair.Table(nlist=nlist)
>>> f.r_on
  File "/Users/joaander/build/hoomd/hoomd/operation.py", line 57, in __getattr__
    raise AttributeError("Object {} has no attribute {}".format(
AttributeError: Object <class 'hoomd.md.pair.pair.Table'> has no attribute r_on
```

## Change log

```
Changed:

* Removed the `r_on` attribute and `default_r_on` constructor argument from pair potentials that do not use it.
```

## Checklist:

## Note

Merge after #1189


tommy-waltmann (commented):
There was some discussion in #1047 about adding documentation explaining which smoothing modes are/aren't supported by which classes, and I think that would be a good idea to add here.

joaander:
> There was some discussion in #1047 about adding documentation explaining which smoothing modes are/aren't supported by which classes, and I think that would be a good idea to add here.

Done.

tommy-waltmann:
I see this done in every class except for `SLJ`, is that an error or it that because it will soon be removed?

joaander:
> I see this done in every class except for `SLJ`, is that an error or it that because it will soon be removed?

SLJ will be removed in the February release: #1181

tommy-waltmann (approved):

PR #1192, opened 2022-01-10
Add linker options for GNU only on non-Apple systems

### LINKED ISSUE #1191: Invalid linker options when combining gcc with non-gnu linker
## Description

Compiling of several tests fails to complete when using gcc on systems where the standard linker does not recognize the options `--allow-shlib-undefined` or `--no-as-needed`. The reason is found in lines 57-60 of `hoomd/md/test/CMakeLists.txt`,
where these linker options are added on the condition that CMAKE_CXX_COMPILER_ID equals "GNU".

This happens e.g. in MacOS when using a gcc installed through macports, which uses the system linker through xcode. I tested this in two settings (see below) and in both cases compiling failed, but commenting out lines 57-60 in the file mentioned above fixed the issue.

## Script

```bash
mkdir build
cd build
cmake -DCMAKE_C_COMPILER=$(which gcc-mp-11) -DCMAKE_CXX_COMPILER=$(which g++-mp-11) ..
make
```

## Output
CMake output for reference: [cmake_output.txt](https://github.com/glotzerlab/hoomd-blue/files/7837688/cmake_output.txt)
Only showing where it goes wrong:
```
[ 73%] Linking CXX executable test_berendsen_integrator
ld: unknown option: --allow-shlib-undefined
collect2: error: ld returned 1 exit status
make[2]: *** [hoomd/md/test/test_berendsen_integrator] Error 1
make[1]: *** [hoomd/md/test/CMakeFiles/test_berendsen_integrator.dir/all] Error 2
```
If compiling with -j8, several of these errors show up for various test exectuables.

## Expected output

Succesful compilation.

## Configuration

Platform (both giving the same result):
- Mac (Intel Mac with MacOS 10.14 and gcc 8)
- Mac (M1 Mac with MacOS 12.1 with gcc 11) (this is the system I used to generate the output above)

Installation method:
- Compiled from source

Versions:
- Python version: 3.9.9
- **HOOMD-blue** version: 3.0.0-beta12
- gcc 11.2.0
- cmake 3.22.1

## Developer
The lines the issue is about were added by @joaander but I'm also happy to contribute. I do not know what kinds of systems there are on which gnu compilers tend to be paired with non-gnu linkers. But if my understanding is correct that  one simply does not use GNU linkers in MacOS, this should be expected to happen on all MacOS systems regardless of how exactly the gnu compilers were obtained, and I would suggest to only add the linker options on non-Darwin systems by changing line 57 of the offending file to 
```
    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND NOT "${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Darwin")
```
But I'd rather see someone with more CMake experience make the call on what the best solution is.

I should add that the other obvious solution (telling Mac users to use Clang instead of GCC) gave me trouble as well, but I will address that in a separate issue.

wouterel:
Following instruction that during beta release cycle all work should be based on master.

## Description
Fix issue #1191 

## Motivation and context
Using gcc on Apple systems gave linker errors because the MacOS linker does not recognize some options that are necessary when using gcc on systems with the GNU linker. See issue #1191 for details.

Resolves #1191 

## How has this been tested?
Identified problem on two different Mac systems and tested that compiling works when the offending linker options are removed from CMakeLists.txt.

Tested the fix with CMake 3.22 on an M1 Mac with MacOS 12.1 and gcc 11: The offending options are no longer generated and compilation now succeeds. The resulting code passes both ctest and pytest.

## Change log

```
Removed invalid linker options when using gcc on Apple systems.
```

## Checklist:

### DIFF
--- a/hoomd/md/test/CMakeLists.txt
+++ b/hoomd/md/test/CMakeLists.txt
@@ -54,7 +54,7 @@ foreach (CUR_TEST ${TEST_LIST} ${MPI_TEST_LIST})
 
     add_dependencies(test_all ${CUR_TEST})
 
-    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" AND NOT APPLE)
         # these options are needed to avoid linker errors with GCC
         set(additional_link_options "-Wl,--allow-shlib-undefined -Wl,--no-as-needed")
     endif()


joaander:
Thanks! I know this is a small pull request, but even small things make a difference. Could you add your name and affiliation to the file `credits.rst`?

wouterel:
Thanks. Even though my github account has not been involved in the project before, my name was already in the credits, because I worked with Simone Ciarella on the RevCross feature.

joaander:
> Thanks. Even though my github account has not been involved in the project before, my name was already in the credits, because I worked with Simone Ciarella on the RevCross feature.

Understood, thanks.

joaander (approved):
Thanks!

PR #1193, opened 2022-01-11
Add Communicator.walltime

joaander:
## Description

Add the `walltime` property to `Communicator`.

## Motivation and context

One feature from v2 missing in v3 is the ability to stop the simulation run when a time limit has passed. I don't plan to add this explicitly, but the v3 API is designed to make this possible for the user to implement Python code. 

However, the user's code could deadlock when using `time.time()` in an MPI simulation and the node clocks were not *exactly* synchronized. To prevent this, I'm adding `Communicator.walltime` which returns the elapsed time since the communicator's creation, broadcasted to all ranks. With this, a user can implement a robust wall time limit with code like this (to appear in a tutorial on workflows soon).
```
while sim.timestep < end_step:
    sim.run(min(100_000, end_step - sim.timestep))
    
    if sim.device.communicator.walltime + sim.walltime >= WALLTIME_LIMIT:
        break
```

## How has this been tested?

I have tested the method locally.

## Change log

```
Added:

* ``Communicator.walltime`` - the wall clock time since creating the ``Communicator``.
```

## Checklist:

### DIFF
--- a/hoomd/ExecutionConfiguration.h
+++ b/hoomd/ExecutionConfiguration.h
@@ -4,7 +4,7 @@
 #ifndef __EXECUTION_CONFIGURATION__
 #define __EXECUTION_CONFIGURATION__
 
-// ensure that HOOMDMath.h is the first thing included
+// ensure that HOOMDMath.h is the first header included to work around broken mpi headers
 #include "HOOMDMath.h"
 
 #ifdef ENABLE_MPI
--- a/hoomd/MPIConfiguration.h
+++ b/hoomd/MPIConfiguration.h
@@ -4,8 +4,8 @@
 #pragma once
 
 // ensure that HOOMDMath.h is the first thing included
-#include "HOOMDMath.h"
 #include "ClockSource.h"
+#include "HOOMDMath.h"
 
 #ifdef ENABLE_MPI
 #include <mpi.h>
--- a/hoomd/MPIConfiguration.h
+++ b/hoomd/MPIConfiguration.h
@@ -3,13 +3,14 @@
 
 #pragma once
 
-// ensure that HOOMDMath.h is the first thing included
-#include "ClockSource.h"
+// ensure that HOOMDMath.h is the first header included to work around broken mpi headers
 #include "HOOMDMath.h"
 
 #ifdef ENABLE_MPI
 #include <mpi.h>
 #endif
+
+#include "ClockSource.h"
 
 /*! \file MPIConfiguration.h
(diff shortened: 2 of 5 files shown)


b-butler (approved):
LGTM!

hoomd/MPIConfiguration.h
```diff
@@ -4,6 +4,7 @@
 #pragma once
 
 // ensure that HOOMDMath.h is the first thing included
```
j-proc, line 6: This comment doesn't make sense with the style guides. Is there a purpose/reason to break that convention?

hoomd/MPIConfiguration.h
```diff
@@ -4,6 +4,7 @@
...
+#include "ClockSource.h"
```
j-proc, line 7: This works on Linux, but I'm not sure if it's provided by all systems. Verify please. Is there a way we can use a standard library or reason not to?

j-proc (changes requested):
I'd at least like to ask if there's a way to add a test for this efficiently. Although I don't think this looks to be something that would be easy to break.

hoomd/MPIConfiguration.h
joaander: Some MPI stacks `#define` things that break headers included by `HOOMDMath.h`. If `<mpi.h>` is included first, HOOMD fails to compile on these systems.

hoomd/MPIConfiguration.h
j-proc: I wasn't aware it was already used. Clearly it's no issue to include it another place.

hoomd/MPIConfiguration.h
j-proc: Might be good to reword the comment to reflect that then.

joaander:
> 
I added a crude test.

hoomd/ExecutionConfiguration.h
```diff
@@ -4,7 +4,7 @@
 #ifndef __EXECUTION_CONFIGURATION__
 #define __EXECUTION_CONFIGURATION__
 
-// ensure that HOOMDMath.h is the first thing included
+// ensure that HOOMDMath.h is the first header included to work around broken mpi headers
```
j-proc, line 7:
```suggestion
// ensure this block is included first to work around broken mpi headers
```

j-proc (approved):
Left a suggestion if it's the whole block and not just HOOMDMath.h but the rest is fine either way.

PR #1195, opened 2022-01-12
Fix/custom force aniso

tommy-waltmann:
## Description

Added a property that should have been part of #1151 . Adds a settable flag so users can alert the simulation to integrate over the anisotropic degrees of freedom if their python force has a torque.

## Motivation and context

Ideally, this would have been a part of #1151 , but it was merged before we realized it needed to be there.

## How has this been tested?

Tests have been added in the `test_custom_force.py` file;

## Change log

```
Added flag for integrating over rotational degrees of freedom with custom forces.
```

## Checklist:

### DIFF
--- a/hoomd/md/force.py
+++ b/hoomd/md/force.py
@@ -196,8 +196,7 @@ class Custom(Force):
 
         class MyCustomForce(hoomd.force.Custom):
             def __init__(self):
-                super().__init__()
-                self.aniso = True
+                super().__init__(aniso=True)
 
             def set_forces(self, timestep):
                 with self.cpu_local_force_arrays as arrays:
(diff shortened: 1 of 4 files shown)


hoomd/md/force.py
```diff
@@ -210,6 +211,12 @@ def set_forces(self, timestep):
     property name with either the prefix ``ghost_`` of the suffix
     ``_with_ghost``.
 
+    Note:
+        If updating the torque array in the set_forces method, make sure to set
+        the `aniso` property of the custom force object to `True` so the
+        simulation knows to update the rotational degrees of freedom in the
+        system.
```
joaander, line 218: This makes it sound like users should set `aniso` on a step by step basis. However, the implementation assumes that `isAnisotropic` is a static property determined at class instantiation. 

`aniso` should not be a modifiable parameter dict entry, it should be set in `__init__` and remain static.

I would then reword this section of the docs to read:
"Set `aniso=True` if your custom force produces non-zero torques on particles."

joaander (changes requested):

joaander (approved):
Great, thanks!

PR #1196, opened 2022-01-13
Automatically update group dof

### LINKED ISSUE #687: Automatically update the number of degrees of freedom in Group

joaander:
## Description

Automatically update the number of degrees of freedom in each group when the user performs actions that are likely to change it.

## Motivation and context

Users expect MD integration methods to thermostat the system correctly and `ThermodynamicQuantities.kinetic_temperature` to return the correct value - even when they make changes to their system such as:
* Adding particles.
* Changing the `integrate_rotational_dof` setting.
* Changing integrators or integration methods.
* Changing the particles an integration method acts on with `FilterUpdater`.

When changing moments of inertia with local snapshots, users still need to manually call `update_group_dof`. I do not want to incur the performance hit for all users of local snapshots where performance is critical.

Resolves #687

## How has this been tested?
I added unit tests.

## Change log

```
Changed:

* Call ``update_group_dof`` implicitly in ``set_snapshot``, when changing integrators or integration methods, and on steps where ``FilterUpdater`` acts on the system.
* [breaking] ``update_group_dof`` defers counting the degrees of freedom until the next timestep or the next call to ``Simulation.run``.
```

## Checklist:

### DIFF
--- a/hoomd/md/pytest/test_update_group_dof.py
+++ b/hoomd/md/pytest/test_update_group_dof.py
@@ -26,7 +26,7 @@ def snapshot_factory(device):
                 [1, 1, 1],
             ]
             snap.particles.angmom[:] = [[0, 2, 4, 6]] * 3
-            snap.particles.types = ['A']
+            snap.particles.types = ['A', 'B']
 
         return snap
 
@@ -82,6 +82,7 @@ def test_local_snapshot(simulation_factory, snapshot_factory):
     sim.run(0)
 
     # reduce the rotational dof by modifying the moment of inertia of particle 0
+    # TODO: Fix
     with sim.state.cpu_local_snapshot as snapshot:
         N = len(snapshot.particles.position)
         idx = snapshot.particles.rtag[0]
@@ -136,7 +137,7 @@ def test_set_method(simulation_factory, snapshot_factory):
     sim = simulation_factory(snapshot)
     filter_all = hoomd.filter.All()
     method1 = hoomd.md.methods.Langevin(filter=filter_all, kT=1.0)
-    integrator1 = hoomd.md.Integrator(
+    integrator = hoomd.md.Integrator(
         0.005,
         methods=[method1],
         integrate_rotational_dof=True,
@@ -144,7 +145,7 @@ def test_set_method(simulation_factory, snapshot_factory):
 
     method2 = hoomd.md.methods.NVT(filter=filter_all, kT=1.0, tau=1.0)
 
-    sim.operations.integrator = integrator1
+    sim.operations.integrator = integrator
     thermo = hoomd.md.compute.ThermodynamicQuantities(filter=filter_all)
     sim.operations.add(thermo)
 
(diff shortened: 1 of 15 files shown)


joaander:
I am investigating the failing test.

hoomd/md/pytest/test_update_group_dof.py
```diff
@@ -0,0 +1,242 @@
...
+    # check initial degrees of freedom in snapshot
+    sim.run(0)
+    assert thermo.translational_degrees_of_freedom == 9
+    assert thermo.rotational_degrees_of_freedom == 7
+
+    # add a particle and check after set_snapshot
+    if snapshot.communicator.rank == 0:
+        snapshot.particles.N = 4
+        if snapshot.communicator.rank == 0:
+            snapshot.particles.moment_inertia[3] = [0, 0, 1]
```
tommy-waltmann, line 59:
```suggestion
        snapshot.particles.moment_inertia[3] = [0, 0, 1]
```
Since you already know you're on rank 0 at this point, the extra check shouldn't be necessary

hoomd/md/pytest/test_update_group_dof.py
```diff
@@ -0,0 +1,242 @@
...
+            snap.particles.N = 3
+            snap.particles.position[:] = [[0, 1, 0], [-1, 1, 0], [1, 1, 0]]
+            snap.particles.velocity[:] = [[0, 0, 0], [0, -1, 0], [0, 1, 0]]
+            snap.particles.moment_inertia[:] = [
+                [2.0, 0, 0],
+                [1, 1, 1],
+                [1, 1, 1],
+            ]
+            snap.particles.angmom[:] = [[0, 2, 4, 6]] * 3
+            snap.particles.types = ['A', 'B']
```
tommy-waltmann, line 28: Why set multiple particle types if no particles are of the `B` type?

hoomd/md/pytest/test_update_group_dof.py
```diff
@@ -0,0 +1,242 @@
...
+            snapshot.particles.moment_inertia[3] = [0, 0, 1]
+    sim.state.set_snapshot(snapshot)
+
+    sim.run(0)
+    assert thermo.translational_degrees_of_freedom == 12
+    assert thermo.rotational_degrees_of_freedom == 8
+
+
+def test_local_snapshot(simulation_factory, snapshot_factory):
+    """Test dof update after local snapshot release."""
```
tommy-waltmann, line 68:
```suggestion
    """Test dof doesn't update after local snapshot release."""
```

tommy-waltmann (changes requested):
Just a couple small things to fix.

hoomd/md/pytest/test_update_group_dof.py
joaander: `test_filter_updater` uses the B type and a type filter to change the number of degrees of freedom with a `FilterUpdater`. The other tests don't use it, but it takes a few characters less code to provide it in the snapshot fixture regardless.

joaander:
@tommy-waltmann I just pushed a change to the way the group cache is stored in C++. The logic is the same, but now there are two caches, one in Python (a dict of dicts) and one in C++ (a std::vector) that are kept in sync. We only ever add items to the cache, so the synchronization process is trivial.

tommy-waltmann (approved):
Once the checks/tests pass, this can be merged.

b-butler (approved):
Looks good

PR #1207, opened 2022-01-21
Miscellaneous box fixes/changes

### LINKED ISSUE #705: BoxResize box assignment updates initial reference

### LINKED ISSUE #1070: Box.lattice_vectors and Box.matrix are similar in functionality but differ in implementation
## Description

It appears that `Box.lattice_vectors` and `Box.matrix` are two different implementations of the same thing, but transposed from one another. Lattice vectors calls into C++ but the matrix is implemented in pure Python. The matrix property can be set, while lattice vectors cannot. The docs are not clear about the relationship between the two. We should consolidate the implementation even if we keep both properties (and document the transpose between them). I’m happy to contribute a fix for this if maintainers can help decide on what to do.

https://github.com/glotzerlab/hoomd-blue/blob/d46de2f693cbc597dff4a64406476cb161b0f742/hoomd/box.py#L354-L362

https://github.com/glotzerlab/hoomd-blue/blob/d46de2f693cbc597dff4a64406476cb161b0f742/hoomd/box.py#L380-L397

https://github.com/glotzerlab/hoomd-blue/blob/d46de2f693cbc597dff4a64406476cb161b0f742/hoomd/box.py#L399-L410

## Proposed solution

Questions:
1. Should both `lattice_vectors` and `matrix` exist? (If not, I would remove `lattice_vectors` and keep `matrix`.)
2. Should the implementation(s) use C++ or pure Python? (I would prefer to keep everything in C++ if possible, which may be best to do by adding a method or two that handles this functionality in C++.)
3. If `lattice_vectors` is kept, should it be settable? (I think that it should match the settable nature of `matrix`, but as a lower triangular matrix since it is transposed.)

## Additional context

For another example, the freud library provides a `to_matrix()` method and provides lattice vectors individually as properties `v1`, `v2`, `v3`.

## Developer

I am willing to help with this if maintainers can offer guidance on the questions above.

b-butler:
- Remove Getar
- Switch BoxDim to use std::shared_ptr for Python
- Refactor matrix property in box

## Description

Changes the exposing of the `BoxDim` class in Python through a shared pointer and removing the `matrix` and `lattice_vectors` properties from the Python box class.

## Motivation and context

Resolves #1070 
Resolves #705

## How has this been tested?
Tests have been appropriately modified

## Change log

```
Changed [breaking] - `hoomd.Box` no longer has a `matrix` property use the new `to_matrix` and existent `from_matrix` classmethod instead.
Removed [breaking] - ``hoomd.Box.lattice_vectors`` property no longer exists.
```

## Checklist:


hoomd/ParticleData.cc
```diff
@@ -199,9 +199,41 @@ ParticleData::~ParticleData()
...
+    if (!box)
+        {
+        throw std::runtime_error("Expected non-null shared pointer to a box.");
+        }
+    m_global_box = std::make_shared<const BoxDim>(*box);
+
+#ifdef ENABLE_MPI
+    if (m_decomposition)
+        {
+        bcast(m_global_box, 0, m_exec_conf->getMPICommunicator());
```
joaander, line 226: This calls HOOMD's `bcast` in `HOOMDMPI.h`. Does it broadcast the contents of the box? Or the shared pointer value? In either case, I'd be more confident in this code it it broadcast a raw BoxDim object instead.

hoomd/ParticleData.cc
```diff
@@ -214,29 +246,29 @@ void ParticleData::setGlobalBox(const BoxDim& box)
     assert(box.getPeriodic().x);
     assert(box.getPeriodic().y);
     assert(box.getPeriodic().z);
-    m_global_box = box;
+    m_global_box = std::make_shared<const BoxDim>(box);
 
 #ifdef ENABLE_MPI
     if (m_decomposition)
         {
         bcast(m_global_box, 0, m_exec_conf->getMPICommunicator());
```
joaander, line 254: Same comment as above.

hoomd/SnapshotSystemData.h
```diff
@@ -54,10 +54,7 @@ template<class Real> struct SnapshotSystemData
     PairData::Snapshot pair_data;             //!< The pair data
 
     //! Constructor
-    SnapshotSystemData()
-        {
-        dimensions = 3;
-        }
+    SnapshotSystemData() : dimensions {3}, global_box {new BoxDim()} { }
```
joaander, line 57:
```suggestion
    SnapshotSystemData() : dimensions(3), global_box(new BoxDim()) { }
```
Use an initialization style consistent with the rest of the code base.

hoomd/box.py
```diff
@@ -377,14 +382,10 @@ def volume(self):
     def volume(self, volume):
         self.scale((volume / self.volume)**(1 / self.dimensions))
 
-    @property
```
joaander, line 380: This comment applies to line 355 (above).

Remove `lattice_vectors` to be consistent with the proposed change log entry.

joaander (changes requested):
Thanks!

There are merge conflicts to resolve and a few changes to make before this can be merged.

b-butler:
Additional changes have been made that directly result from changes made to aid the handling of `Box` and `BoxDim` objects.

hoomd/hpmc/UpdaterQuickCompress.cc
```diff
@@ -160,7 +157,7 @@ BoxDim UpdaterQuickCompress::getNewBox(uint64_t timestep)
     double scale = uniform(rng);
 
     // TODO: This slow. We will implement a general reusable fix later in #705
-    BoxDim target_box = m_target_box.attr("_cpp_obj").cast<BoxDim>();
+    const BoxDim target_box = *m_target_box;
```
b-butler, line 160:
```suggestion
    const auto& target_box = *m_target_box;
```

b-butler:
@joaander to get ctests to pass, I had to change the tolerance on one test for mpcd. I have no idea why this is the case. We likely shouldn't merge with the change, but I wanted to ensure the code I added works at least in all other respects. Any insight you have would be appreciated.

hoomd/mpcd/SystemData.h
```diff
@@ -57,7 +57,7 @@ class PYBIND11_EXPORT SystemData
     //! Get the current global simulation box
     const BoxDim& getGlobalBox() const
```
joaander, line 58:
```suggestion
    const BoxDim getGlobalBox() const
```
Let's not return references to values managed by shared pointers.

I don't think this would cause the problem with the unit test.

mphoward: Agreed, it seems unlikely this is the issue

mphoward:
@b-butler do you have the output of that test (`streaming_method_test`) when it was failing? Does it fail CPU and GPU, or GPU only? That is a simple test of particle wrapping through the PBCs during ballistic motion, so something bad is going on if you needed to relax the tolerances (i.e., it is probably not wrapping). Somebody deleted all of the Python level tests for MPCD during the port to v3 (since they hadn't been converted to pytest), but those would probably also fail some basic checks related to this.

Your changes look like simple grep replace, so I'm not sure why that would cause this to fail... Is there anything funny about passing `std::shared_ptr` by const reference rather than by value? I always pass by value and let the compiler sort it out. Otherwise, maybe there is something more malicious about returning a reference to a managed object that @joaander pointed out. Definitely weird.

joaander:
> @b-butler do you have the output of that test (`streaming_method_test`) when it was failing? Does it fail CPU and GPU, or GPU only? That is a simple test of particle wrapping through the PBCs during ballistic motion, so something bad is going on if you needed to relax the tolerances (i.e., it is probably not wrapping). 

I was unable to reproduce the failing test on the CPU. Reverting the tolerance change so we can see the output in GitHub actions.

>Somebody deleted all of the Python level tests for MPCD during the port to v3 (since they hadn't been converted to pytest), but those would probably also fail some basic checks related to this.

The MPCD API has not been updated to the support v3, so it was not possible to update the tests. The previous MPCD python tests can be accessed in the git history.

b-butler:
Thank you @joaander! I apologize for the headache. I realized halfway into my change that I shouldn't return references to objects stored in `std::shared_ptr` but must have missed some of my old code, and good catch on the spotting the reference to a temporary.

hoomd/mpcd/ConfinedStreamingMethodGPU.cuh
```diff
@@ -41,7 +41,7 @@ struct stream_args_t
     Scalar4* d_vel;                   //!< Particle velocities
     const Scalar mass;                //!< Particle mass
     const mpcd::ExternalField* field; //!< Applied external field on particles
-    const BoxDim& box;                //!< Simulation box
```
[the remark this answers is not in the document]
joaander: It was likely copied and pasted from the potential pair `pair_args_t` struct, or a similar place where the reference was used. These structs were an attempt at limiting the changes users need to make to plugins when the argument list to the GPU kernel driver changes.

hoomd/ParticleData.cc
```diff
@@ -73,7 +73,7 @@ std::string getDefaultTypeName(unsigned int id)
     Type mappings assign particle types "A", "B", "C", ....
 */
 ParticleData::ParticleData(unsigned int N,
-                           const BoxDim& global_box,
+                           const std::shared_ptr<const BoxDim> global_box,
```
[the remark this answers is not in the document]
b-butler: It doesn't need it, but all the const modify should do in this case is prevent a change in what the smart pointer points to, I believe which is what we want since we are copying the box anyways.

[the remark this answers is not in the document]
mphoward: Yah agreed, it just reads weirdly for a `shared_ptr`, like if you used `const unsigned int N` rather than `unsigned int N` above, for example. It doesn't hurt anything, but it's not a pattern I'm used to seeing.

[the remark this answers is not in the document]
b-butler: I have inconsistently been striving to make as many things `const` as possible when editing the C++ code, but if you feel strongly, I can change it.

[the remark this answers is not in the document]
mphoward: I don't feel strongly, so whatever style you and @joaander want to enforce is fine by me! The reason I thought it looked unusual is that the raw pointer equivalent is `const BoxDim* const` (const pointer to const), and this is not really found elsewhere in HOOMD. The const qualifier could keep someone from accidentally trying to modify the pointer itself when passed by value, but this would be a big whoops (like trying to assign into `N`, for example).

Issue #355, opened 2019-03-01
Mixed precision MD


joaander:
## Description

Implement mixed precision for MD simulations.

## Motivation

Compared to a full double precision build, mixed precision may offer performance benefits - especially on GeForce cards. Compared to a full single precision build, mixed precision simulations will conserve energy and momentum significantly better.

## Implementation details

The most reasonable mixed precision model for HOOMD is to maintain particle coordinates in double, compute forces in single, and accumulate forces in double. HOOMD is too general for fixed precision force calculations, and double-single implementations do not offer enough benefits given the implementation complexity.

Care must be taken in testing a new mixed precision implementation evaluating the energy and momentum conservation of a number of test systems that exercise the relevant code paths (pair potentials, bond potentials, ....)

The HPMC mixed precision and a new MD mixed precision type should be merged and controlled by the same compile time option.

## Question for debate

Should we continue to support single, double, and mixed precision builds? Certainly this will be helpful in developer testing, but it may require significant maintenance overhead in the future. If we decide to keep support more than one combination, which do we test and validate for users? Only mixed?

mphoward:
> The most reasonable mixed precision model for HOOMD is to maintain particle coordinates in double, compute forces in single, and accumulate forces in double. HOOMD is too general for fixed precision force calculations, and double-single implementations do not offer enough benefits given the implementation complexity.

Agreed, we can't reasonably assume a certain number of decimal places are needed, and double-single is really only beneficial on something like GeForce where fp64 is crippled. In contrast, even Tesla cards might still benefit for using a mixture of native fp32 and fp64 ops where appropriate.

> Should we continue to support single, double, and mixed precision builds? Certainly this will be helpful in developer testing, but it may require significant maintenance overhead in the future. If we decide to keep support more than one combination, which do we test and validate for users? Only mixed?

I had thought about this a little bit as a potential project for someone. One flexible solution might be to replace the current `Scalar` by two floating point types: `Double` and `Float` (need better names so someone doesn't fat-finger the shift key and get the native type by mistake). `Double` would refer to all values that we believe should be fp64 in a mixed precision build (particle coordinates, forces, accumulation, etc.), while `Float` would be for anything we believe could be fp32 in a mixed precision build without sacrificing accuracy (result of `dr` for pair forces, internal force compute, etc.). It would still be valid to hard code (native) `double` where it is absolutely unacceptable to use fp32, and there should more or less never be a spot that uses (native) `float`.

[...]

PR #1211, opened 2022-01-26
Add hoomd.version.floating_point_precision and summarize features

joaander:
## Description

Add hoomd.version.floating_point_precision.

## Motivation and context

Allow user scripts and plugins to determine what floating point precision used in the build.

## How has this been tested?

Tested locally:
```bash
$ python3 -c "import hoomd; print(hoomd.version.floating_point_precision)"
(64, 32)
```

## Change log

```
Added:

* ``hoomd.version.floating_point_precision`` - Floating point width in bits for the particle properties and local calculations.
```

## Checklist:

### DIFF
--- a/hoomd/HOOMDVersion.cc
+++ b/hoomd/HOOMDVersion.cc
@@ -173,4 +173,15 @@ std::string BuildInfo::getInstallDir()
     return std::string(HOOMD_INSTALL_PREFIX) + "/" + std::string(PYTHON_SITE_INSTALL_DIR);
     }
 
+std::pair<unsigned int, unsigned int> BuildInfo::getFloatingPointPrecision()
+    {
+    #if defined(SINGLE_PRECISION)
+    return std::make_pair(32, 32);
+    #elif !defined(SINGLE_PRECISION) && defined(ENABLE_HPMC_MIXED_PRECISION)
+    return std::make_pair(64, 32);
+    #elif !defined(SINGLE_PRECISION) && !defined(ENABLE_HPMC_MIXED_PRECISION)
+    return std::make_pair(64, 64);
+    #endif
+    }
+
     } // namespace hoomd
--- a/hoomd/HOOMDVersion.cc
+++ b/hoomd/HOOMDVersion.cc
@@ -175,13 +175,13 @@ std::string BuildInfo::getInstallDir()
 
 std::pair<unsigned int, unsigned int> BuildInfo::getFloatingPointPrecision()
     {
-    #if defined(SINGLE_PRECISION)
+#if defined(SINGLE_PRECISION)
     return std::make_pair(32, 32);
-    #elif !defined(SINGLE_PRECISION) && defined(ENABLE_HPMC_MIXED_PRECISION)
+#elif !defined(SINGLE_PRECISION) && defined(ENABLE_HPMC_MIXED_PRECISION)
     return std::make_pair(64, 32);
-    #elif !defined(SINGLE_PRECISION) && !defined(ENABLE_HPMC_MIXED_PRECISION)
+#elif !defined(SINGLE_PRECISION) && !defined(ENABLE_HPMC_MIXED_PRECISION)
     return std::make_pair(64, 64);
-    #endif
+#endif
     }
 
     } // namespace hoomd
--- a/sphinx-doc/features.rst
+++ b/sphinx-doc/features.rst
@@ -43,8 +43,6 @@ method for electrostatics. HOOMD-blue enables active matter simulations with `md
 `md.update.ActiveRotationalDiffusion`. At runtime, `hoomd.version.md_built` indicates whether the
 build supports MD simulations.
 
-md.update
-
 .. seealso::
 
     Tutorial: :doc:`tutorial/01-Introducing-Molecular-Dynamics/00-index`
@@ -77,7 +75,7 @@ preliminary support for AMD GPUs. CPU support is always enabled. GPU support mus
 compile time with the ``ENABLE_GPU`` CMake option (see :doc:`building`). Select the device to use at
 run time with the `device <hoomd.device>` module. Unless otherwise stated in the documentation,
 **all** operations and methods support GPU execution. At runtime, `hoomd.version.gpu_enabled` indicates
-whether the build supports run time compilation.
+whether the build supports GPU devices.
 
 MPI
 ---
@@ -100,7 +98,7 @@ the `device.Device.num_cpu_threads` property. In this release, threading support
 very limited and only applies to implicit depletants in `hpmc.integrate.HPMCIntegrator`, and
 `hpmc.pair.user.CPPPotentialUnion`. Threading must must be enabled at compile time with the
 ``ENABLE_TBB`` CMake option (see :doc:`building`). At runtime, `hoomd.version.tbb_enabled` indicates
-whether the build supports MPI.
+whether the build supports threaded execution.
 
 Run time compilation
 --------------------
@@ -117,15 +115,15 @@ Mixed precision
 ---------------
 
 HOOMD-blue performs computations with mixed floating point precision. There is a **high precision**
-type and a **reduced precision** type. All particle properties are stored in the **high precision**
-type, and most operations also perform all computations with **high precision**. Operations that do
-not mention "Mixed precision" in their documentation perform all calculations in **high percision**.
-Some operations use **reduced precision** when possible to improve performance, as detailed in the
+type and a **reduced precision** type. All particle properties are stored in the high precision
+type, and most operations also perform all computations with high precision. Operations that do not
+mention "Mixed precision" in their documentation perform all calculations in high percision. Some
+operations use reduced precision when possible to improve performance, as detailed in the
 documentation for each operation. In this release, only `hpmc` implements mixed precision.
 
 The precision is set at compile time with the ``SINGLE_PRECISION`` and
-``ENABLE_HPMC_MIXED_PRECISION`` CMake options (see :doc:`building`). By default, the **high
-precision** width is 64 bits and the **reduced precision** width is 32 bits. At runtime,
+``ENABLE_HPMC_MIXED_PRECISION`` CMake options (see :doc:`building`). By default, the high precision
+width is 64 bits and the reduced precision width is 32 bits. At runtime,
 `hoomd.version.floating_point_precision` indicates the width of the floating point types.
 
 Plugins
--- a/sphinx-doc/features.rst
+++ b/sphinx-doc/features.rst
@@ -8,7 +8,7 @@ Hard particle Monte Carlo
 -------------------------
 
 HOOMD-blue can perform simulations of hard particles using the Monte Carlo method (`hpmc`). Hard
-particles are defined by their shape and the `HPMC integrator <hpmc.integrate>` supports
+particles are defined by their shape, and the `HPMC integrator <hpmc.integrate>` supports
 polygons, spheropolygons, polyhedra, spheropolyhedra, ellipsoids, faceted ellipsoids, spheres,
 indented spheres, and unions of shapes. HPMC can make both constant volume and constant pressure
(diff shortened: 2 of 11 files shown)


joaander:
@RainierBarrett or @whitead, please review and comment on whether this API fits your needs.

b-butler (changes requested):
Looks good to me. I would document that the second value only currently effects some values in HPMC and the first is for MD and particular data in HPMC.

hoomd/HOOMDVersion.cc
```diff
@@ -173,4 +173,15 @@ std::string BuildInfo::getInstallDir()
...
     }
 
+std::pair<unsigned int, unsigned int> BuildInfo::getFloatingPointPrecision()
+    {
+#if defined(SINGLE_PRECISION)
+    return std::make_pair(32, 32);
+#elif !defined(SINGLE_PRECISION) && defined(ENABLE_HPMC_MIXED_PRECISION)
+    return std::make_pair(64, 32);
+#elif !defined(SINGLE_PRECISION) && !defined(ENABLE_HPMC_MIXED_PRECISION)
+    return std::make_pair(64, 64);
```
cbkerr, line 183: Some of the conditions look redundant. Is that for being explicit?

```suggestion
#elif defined(ENABLE_HPMC_MIXED_PRECISION)
    return std::make_pair(64, 32);
#elif
    return std::make_pair(64, 64);
```

cbkerr (commented):
I can't comment on the use cases for this but see inline comment.

hoomd/HOOMDVersion.cc
joaander: I'm planning ahead. Per request by users in #355, all 4 options will possible when we enable MD mixed precision.

joaander:
> Looks good to me. I would document that the second value only currently effects some values in HPMC and the first is for MD and particular data in HPMC.

I'd prefer to keep the API docs for `floating_point_precision` clean and simple - else we will forget to update them in the future. I don't see an obvious place in the documentation to provide an overview of the mixed precision capabilities offered by the code. Do you think we should add a "Features" section to discuss this?

whitead:
Exactly what we need. Thanks!

cbkerr (approved):

b-butler:
@joaander  is the documentation saying that the first number represents storage and the second local calculations? If so, I didn't find it clean what each individual value in `floating_point_precision` represented, and being more explicit on what each index represents would be more helpful. We don't need to add HPMC specific documentation if we are trying to keep this the same when mixed precision MD is added.

b-butler:
Also, having a features section would be good that outlines the default behavior and available options.

joaander:
> Also, having a features section would be good that outlines the default behavior and available options.

In a future PR we could add this.

joaander:
> @joaander is the documentation saying that the first number represents storage and the second local calculations? If so, I didn't find it clean what each individual value in `floating_point_precision` represented, and being more explicit on what each index represents would be more helpful. We don't need to add HPMC specific documentation if we are trying to keep this the same when mixed precision MD is added.

I expanded the documentation. At this level, the docs can't possibly list every location where the two precisions are used. Perhaps in the mixed precision PR I could update the API docs for every operation and document when they implement mixed precision.

joaander:
I didn't like the long description in `version.py` so I added `features.rst` which summarizes all HOOMD-blue features including those affected by compile time options. I also updated the docs of those operations for selectively implemented features.

b-butler (approved):
Thanks, sorry for the extra work, but I think this will help user much better understand the values found in `hoomd.version.floating_point_precision`.

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+.. Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+Features
+========
+
+Hard particle Monte Carlo
+-------------------------
+
+HOOMD-blue can perform simulations of hard particles using the Monte Carlo method (`hpmc`). Hard
+particles are defined by their shape and the `HPMC integrator <hpmc.integrate>` supports
```
cbkerr, line 11:
```suggestion
particles are defined by their shape, and the `HPMC integrator <hpmc.integrate>` supports
```

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+commonly used in atomistic and coarse-grained force fields (`md.angle`, `md.bond`,
+`md.constrain.Distance`, `md.dihedral`, `md.improper`, `md.special_pair`) and can model rigid bodies
+(`md.constrain.Rigid`). External fields `md.external.field` apply potentials based only on the
+particle's position and orientation, including walls (`md.external.wall`) to confine particles in a
+specific region of space. `md.long_range` provides long ranged interactions, including the PPPM
+method for electrostatics. HOOMD-blue enables active matter simulations with `md.force.Active` and
+`md.update.ActiveRotationalDiffusion`. At runtime, `hoomd.version.md_built` indicates whether the
+build supports MD simulations.
+
+md.update
```
cbkerr, line 46: typo?

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+-------------------
+
+HOOMD-blue can execute simulations on CPUs or GPUs. Typical simulations run more efficiently on
+GPUs for system sizes larger than a few thousand particles, although this strongly depends on the
+details of the simulation. The provided binaries support NVIDIA GPUs. Build from source to enable
+preliminary support for AMD GPUs. CPU support is always enabled. GPU support must be enabled at
+compile time with the ``ENABLE_GPU`` CMake option (see :doc:`building`). Select the device to use at
+run time with the `device <hoomd.device>` module. Unless otherwise stated in the documentation,
+**all** operations and methods support GPU execution. At runtime, `hoomd.version.gpu_enabled` indicates
+whether the build supports run time compilation.
```
cbkerr, line 80: Does run time compilation mean JIT code?

Or does it mean "able to run on GPU"?

```suggestion
whether the build supports running on the GPU.
```

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+
+.. seealso::
+
+    Tutorial: :doc:`tutorial/03-Parallel-Simulations-With-MPI/00-index`
+
+Threading
+---------
+
+Some operations in HOOMD-blue can use multiple CPU threads in a single process. Control this with
+the `device.Device.num_cpu_threads` property. In this release, threading support in HOOMD-blue is
```
cbkerr, line 99: "in this release"

Is there a way to set a reminder to update this when that changes? Maybe just a code comment

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+very limited and only applies to implicit depletants in `hpmc.integrate.HPMCIntegrator`, and
+`hpmc.pair.user.CPPPotentialUnion`. Threading must must be enabled at compile time with the
+``ENABLE_TBB`` CMake option (see :doc:`building`). At runtime, `hoomd.version.tbb_enabled` indicates
+whether the build supports MPI.
```
cbkerr, line 103: Reading this I'm confused about the discrepancy between the flag saying TBB and the description saying MPI again.

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+
+Run time compilation
+--------------------
+
+Some operations allow the user to provide arbitrary C++ code which HOOMD-blue compiles at run time
+and executes during the simulation. `hpmc.pair.user` and `hpmc.external.user` enable users to apply
+arbiratray pair and external potentials to particles in HPMC simulations. `hpmc.pair.user` is
```
cbkerr, line 110:
```suggestion
arbitrary pair and external potentials to particles in HPMC simulations. `hpmc.pair.user`
```

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+the `device.Device.num_cpu_threads` property. In this release, threading support in HOOMD-blue is
+very limited and only applies to implicit depletants in `hpmc.integrate.HPMCIntegrator`, and
+`hpmc.pair.user.CPPPotentialUnion`. Threading must must be enabled at compile time with the
+``ENABLE_TBB`` CMake option (see :doc:`building`). At runtime, `hoomd.version.tbb_enabled` indicates
+whether the build supports MPI.
+
+Run time compilation
+--------------------
+
+Some operations allow the user to provide arbitrary C++ code which HOOMD-blue compiles at run time
```
cbkerr, line 108:
```suggestion
Some operations allow the user to provide arbitrary C++ code that HOOMD-blue compiles at run time
```

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+and executes during the simulation. `hpmc.pair.user` and `hpmc.external.user` enable users to apply
+arbiratray pair and external potentials to particles in HPMC simulations. `hpmc.pair.user` is
+supports both CPUs and NVIDIA GPUs while `hpmc.external.user` is only supports CPUs. Run time
```
cbkerr, line 111:
```suggestion
supports both CPUs and NVIDIA GPUs while `hpmc.external.user` only supports CPUs. Run time
```

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+compilation must be enabled at compile time with the ``ENABLE_LLVM`` CMake option (see
+:doc:`building`). At runtime, `hoomd.version.llvm_enabled` indicates whether the build supports run
+time compilation.
+
+Mixed precision
+---------------
+
+HOOMD-blue performs computations with mixed floating point precision. There is a **high precision**
+type and a **reduced precision** type. All particle properties are stored in the **high precision**
+type and most operations also perform all computations with **high precision**. Operations that to
```
cbkerr, line 121:
```suggestion
type, and most operations also perform all computations with **high precision**. Operations that do
```

sphinx-doc/features.rst
```diff
@@ -0,0 +1,139 @@
...
+supports both CPUs and NVIDIA GPUs while `hpmc.external.user` is only supports CPUs. Run time
+compilation must be enabled at compile time with the ``ENABLE_LLVM`` CMake option (see
+:doc:`building`). At runtime, `hoomd.version.llvm_enabled` indicates whether the build supports run
+time compilation.
+
+Mixed precision
+---------------
+
+HOOMD-blue performs computations with mixed floating point precision. There is a **high precision**
+type and a **reduced precision** type. All particle properties are stored in the **high precision**
```
cbkerr, line 120: I would not bold the high/reduced precision terms after they are first mentioned. It provides too much emphasis. Perhaps use a code syntax style?

cbkerr (commented):
Small comments on the documentation. I will commit a batch of quick grammar fixes but some others remain.

sphinx-doc/features.rst
joaander: Yes.

sphinx-doc/features.rst
joaander: Fixed.

sphinx-doc/features.rst
joaander: Right, copy and paste bug. Fixed.

sphinx-doc/features.rst
joaander: I just removed the bold.

sphinx-doc/features.rst
joaander: We just need to be diligent. There is no direct connection between what is written in the code and the docs. This may never change. At best I would hope that we remember to update the local docs for each operation when we change them. But the overview docs are the most likely to get out of date.

cbkerr (approved):
Thanks for those tweaks, looks good now!

sphinx-doc/features.rst
cbkerr: maybe just tag with like "as of 3.0" so readers know when it was update?
```suggestion
the `device.Device.num_cpu_threads` property. As of version 3.0, threading support in HOOMD-blue is
```

PR #1213, opened 2022-01-26
Add How To guide for molecular systems.

joaander:
## Description

Add a how to guide for molecular systems in molecular dynamics simulations.

## Motivation and context

This is a common question on the mailing list.

## How has this been tested?

I executed the example script locally.

## Change log

```
Added:

* Guide on how to model molecular systems.
```

## Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,6 +27,7 @@ ignore =
 # D214 ignore overindented sections in Trigger - this is Google napoleon formatting
 per-file-ignores =
     */pytest/*.py:D100,D101,D102,D103,D104,D105,D106
+    sphinx-doc/howto/*.py:D100,D101,D102,D103,D104,D105,D106
     */__init__.py: F401
     hoomd/version.py: F401
     hoomd/trigger.py: D214
--- a/sphinx-doc/howto/howto.rst
+++ b/sphinx-doc/howto/howto.rst
@@ -1,8 +1,8 @@
 .. Copyright (c) 2009-2022 The Regents of the University of Michigan.
 .. Part of HOOMD-blue, released under the BSD 3-Clause License.
 
-Molecular dynamics
-===================
+How to guides
+=============
 
 How to model molecular systems
 ------------------------------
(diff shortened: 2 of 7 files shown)


tommy-waltmann (commented):
One thing to consider here is that having the "how to model molecular systems" howto nested inside a molecular dynamics section might make it harder for people to find what they're looking for.

joaander:
> One thing to consider here is that having the "how to model molecular systems" howto nested inside a molecular dynamics section might make it harder for people to find what they're looking for.

Would you prefer a single page for all MD, MC, and other how tos?

joaander:
> > One thing to consider here is that having the "how to model molecular systems" howto nested inside a molecular dynamics section might make it harder for people to find what they're looking for.

I moved the new documentation to `howto.rst` per your request. I still think that once we add a few dozen howtos that it will be difficult to find anything in such a long document. However, we can deal with that later.

tommy-waltmann:
To more fully explain my suggestion from before, I was trying to suggest having a "HOW TO GUIDES" section heading in blue on the left sidebar (similar to "PYTHON API", "TUTORIALS", and "GETTING STARTED") and then under that section heading, have an entry for each how to guide (similar to how under PYTHON API, one can see hoomd hoomd.hpmc, hoomd.md).

That way people could more immediately see what they're looking for, instead of having to to know they need to click another "molecular dynamics" heading to reveal the howto guide they are looking for.

I don't know exactly how many how to guides we plan on putting in the documentation, but I see your point about the left panel looking a little too messy if we have a lot of them. I'd say if we plan on having <=10 how to guides, we don't need the extra level of organization.

We can organize them so its a bit easier to use now, or we can organize them so we don't have to re-organize in the future, It just depends on how many howto's we think we'll end up having after a while.

joaander:
> > That way people could more immediately see what they're looking for, instead of having to to know they need to click another "molecular dynamics" heading to reveal the howto guide they are looking for.
> 
> OK, we can try this for now. As you say we can put only a small number on the left panel without making navigation difficult. Beyond that and people will need to use the main page table of contents, the index, or the search feature to find things.

FYI, the left panel contents already requires scrolling on 1080p displays and mobile devices. There is a small amount of space left on 1440p.

tommy-waltmann (approved):

PR #1221, opened 2022-02-04
Add the bond, angle, and dihedral table potentials to the v3 API.

### LINKED ISSUE #360: Port Table potentials to v3 API

joaander:
## Description

Add the bond, angle, and dihedral table potentials to the v3 API.

## Motivation and context

Researchers that utilize these features should be able to upgrade to v3.

Resolves #360

## How has this been tested?

I added Table tests to the relevant unit tests file. In the process, I refactored the unit tests so it is easier for developers to manage test cases.

## Change log

```
Added:

* ``md.bond.Table`` - Tabulated bond potential.
* ``md.angle.Table`` - Tabulated angle potential.
* ``md.dihedral.Table`` - Tabulated dihedral potential.

Changed:

* [breaking] - renamed ``md.angle.Cosinesq`` to ``md.angle.CosineSquared``.
```

## Checklist:

### DIFF
--- a/hoomd/md/BondTablePotential.h
+++ b/hoomd/md/BondTablePotential.h
@@ -74,6 +74,12 @@ class PYBIND11_EXPORT BondTablePotential : public ForceCompute
     /// Get the parameters for a particular type.
     pybind11::dict getParams(std::string type);
 
+    /// Get the width
+    unsigned int getWidth()
+        {
+        return m_table_width;
+        }
+
 #ifdef ENABLE_MPI
     //! Get ghost particle fields requested by this pair potential
     /*! \param timestep Current time step
--- a/hoomd/md/BondTablePotential.h
+++ b/hoomd/md/BondTablePotential.h
@@ -68,6 +68,12 @@ class PYBIND11_EXPORT BondTablePotential : public ForceCompute
                           Scalar rmin,
                           Scalar rmax);
 
+    /// Set parameters from Python.
+    void setParamsPython(std::string type, pybind11::dict params);
+
+    /// Get the parameters for a particular type.
+    pybind11::dict getParams(std::string type);
+
 #ifdef ENABLE_MPI
     //! Get ghost particle fields requested by this pair potential
     /*! \param timestep Current time step
--- a/hoomd/md/TableAngleForceCompute.cc
+++ b/hoomd/md/TableAngleForceCompute.cc
@@ -9,9 +9,6 @@
     \brief Defines the TableAngleForceCompute class
 */
 
-#undef VT0
-#undef VT1
-
 using namespace std;
 
 // SMALL a relatively small number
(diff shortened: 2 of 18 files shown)


hoomd/md/TableAngleForceCompute.cc
```diff
@@ -9,9 +9,6 @@
     \brief Defines the TableAngleForceCompute class
 */
 
-#undef VT0
-#undef VT1
-
```
tommy-waltmann, line 14: Why were these removed?

hoomd/md/TableDihedralForceCompute.cc
```diff
@@ -12,9 +12,6 @@
 
 using namespace std;
 
-#undef VT0
-#undef VT1
-
```
tommy-waltmann, line 17: Same as above

tommy-waltmann (commented):
Overall looks good, just a small question in addition to the ones below:

What happened to the tabulated angle potential that was in v2? For the tabulated bond and dihedral potentials, an old class was removed to make a new class for v3, but I don't see such a class removed to make way for the v3 tabulated angle potential.

hoomd/md/TableAngleForceCompute.cc
joaander: They are unnecessary. Nowhere does the code `#define VT0`  or `#define VT1`.

joaander:
> Overall looks good, just a small question in addition to the ones below:
> 
> What happened to the tabulated angle potential that was in v2? For the tabulated bond and dihedral potentials, an old class was removed to make a new class for v3, but I don't see such a class removed to make way for the v3 tabulated angle potential.

The `angle.table` class was not there to be removed. I did not check the git history to see when it was removed or who removed it.

tommy-waltmann (approved):

hoomd/md/BondTablePotential.h
```diff
@@ -68,6 +68,18 @@ class PYBIND11_EXPORT BondTablePotential : public ForceCompute
...
+    void setParamsPython(std::string type, pybind11::dict params);
+
+    /// Get the parameters for a particular type.
+    pybind11::dict getParams(std::string type);
+
+    /// Get the width
+    unsigned int getWidth()
+        {
+        return m_table_width;
+        }
```
ispivack, line 81: I was under the impression that it is bad form to include implementation in the header in order to not accidentally violate the "one definition rule". (I am new to C++, so this could be incorrect)

joaander: C++ developers commonly include the implementation for short methods in the header file. This allows the compiler to inline those functions when they are called for performance. When the implementation is in the .cc file, there is no opportunity for inlining and the method must be called with all the overhead of a function call.

We add preprocessor macros near the top of the file to create an inclusion guard. This ensures that the method is only defined once.

PR #1226, opened 2022-02-08
Remove unused QuaternionMath.h header.

joaander:
All code now uses the quaternion implementation in VectorMath.h.

## Description

Remove the unused QuaternionMath.h header.

## Motivation and context

There is no need to maintain unused code. `QuaternionMath.h` was part of the rigid body integrators in HOOMD v1. As far as I know, it has been unused since the new rigid body code written in v2.0. The files that "#include" it before this pull request did so due to copy and pasted file headers, not because they actually used functions defined in the file.

## How has this been tested?

i tested that HOOMD compiles and unit test pass.

## Change log

Internal change, no log entry necessary.

## Checklist:


b-butler (approved):
Almost made -500 LoC.

PR #1235, opened 2022-02-14
ALJ Ellipsoids Bugfix

### LINKED ISSUE #1233: ALJ Ellipsoids Segfault
## Description

Any ALJ simulation involving ellipsoids (and therefore spheres) on the master branch will segfault.

The bug occurs because the input shape parameters (vertices/faces) for ellipsoids are empty lists. The empty lists are synced down to cpp resulting in empty lists of vertices/faces, where the GJK code has assumptions that the list of vertices is non-empty

In v2 this was not a problem because the list of vertices in cpp for ellipsoids had a single entry [0, 0, 0], the faces and face offsets lists also had a single entry which was 0

## Script

```python
# Just initialize in a simple cubic lattice.
sim = simulation_factory(
    lattice_snapshot_factory(particle_types=["A"], a=1, n=3,
                               dimensions=3))

sim.seed = 123

alj = md.pair.aniso.ALJ(default_r_cut=1.1, nlist=md.nlist.Cell(0.1))

alj.shape["A"] = {
    "vertices": [],
    "faces": [],
    "rounding_radii": 0.2
}

alpha = 0  # Make it WCA-only (no attraction)
eps_att = 1.0
alj.params[("A", "A")] = {
    "epsilon": 1.0,
    "sigma_i": 0.4,
    "sigma_j": 0.4,
    "alpha": alpha
}
integrator = md.Integrator(dt=1e-4, integrate_rotational_dof=True,
                           forces=[alj])
nve = md.methods.NVE(filter=hoomd.filter.All())
integrator.methods.append(nve)

sim.run(1)
```

## Output

Segmentation fault (core dumped)

## Expected output

No Output

## Configuration

Arch Linux vislab workstation

Platform:
- Mac
- Linux
- CPU
- GPU

Installation method:
- Compiled from source

Versions:
- Python version: 3.10
- **HOOMD-blue** version: commit hash 7f797ab

## Developer

Either @b-butler or I can take care of this.

tommy-waltmann:
## Description

Fixes #1233 

## Motivation and context

This was a bug involved in how data is stored on the cpp side for ellipsoids. It would be good to get this onto master before the next release if possible

Resolves #1233 

## How has this been tested?

I have added a test which runs a simulation for 5 time steps with an ellipsoid. I have also used this potential for my simluations, and it appears to work correctly for spherical particles.

## Change log

```
Fixed: Seg Fault in computing ALJ interactions for ellipsoids
```

## Checklist:

### DIFF
--- a/hoomd/md/EvaluatorPairALJ.h
+++ b/hoomd/md/EvaluatorPairALJ.h
@@ -274,11 +274,9 @@ template<unsigned int ndim> class EvaluatorPairALJ
             // and we don't need to iterate over polytope based attributes.
             const auto vertices = shape["vertices"].cast<pybind11::list>();
             const auto N_vertices = static_cast<unsigned int>(len(vertices));
-            verts = ManagedArray<vec3<Scalar>>(N_vertices, managed);
 
             const auto faces_ = shape["faces"].cast<pybind11::list>();
             const auto N_faces = static_cast<unsigned int>(len(faces_));
-            face_offsets = ManagedArray<unsigned int>(N_faces, managed);
 
             // Simulating an ellipsoid just return
             if (N_vertices == 0 && N_faces == 0)
@@ -300,6 +298,9 @@ template<unsigned int ndim> class EvaluatorPairALJ
                 {
                 throw std::runtime_error("Cannot have empty vertices list.");
                 }
+
+            verts = ManagedArray<vec3<Scalar>>(N_vertices, managed);
+            face_offsets = ManagedArray<unsigned int>(N_faces, managed);
 
             // Unpack the list[list] of vertices into a ManagedArray.
             for (unsigned int i = 0; i < N_vertices; ++i)
--- a/hoomd/md/EvaluatorPairALJ.h
+++ b/hoomd/md/EvaluatorPairALJ.h
@@ -283,7 +283,14 @@ template<unsigned int ndim> class EvaluatorPairALJ
             // Simulating an ellipsoid just return
             if (N_vertices == 0 && N_faces == 0)
                 {
-                faces = ManagedArray<unsigned int>(0, managed);
+                // for ellipsoids, there needs to be a 0 vertex
+                // because that is the 0 of the support function
+                verts = ManagedArray<vec3<Scalar>>(1, managed);
+                verts[0] = vec3<Scalar>(0, 0, 0);
+                faces = ManagedArray<unsigned int>(1, managed);
+                faces[0] = 0;
+                face_offsets = ManagedArray<unsigned int>(1, managed);
+                face_offsets[0] = 0;
                 return;
                 }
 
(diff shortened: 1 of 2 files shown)


joaander:
I will release beta.14 this week. @b-butler and @inhyukcho, please review promtply.

hoomd/md/EvaluatorPairALJ.h
```diff
@@ -283,7 +283,14 @@ template<unsigned int ndim> class EvaluatorPairALJ
             // Simulating an ellipsoid just return
             if (N_vertices == 0 && N_faces == 0)
```
vyasr, line 284: Minor nit: If you moved this block up a bit you could avoid assigning verts/faces/face_offsets twice. It doesn't really matter since the initial ManagedArrays will be empty anyway, just a note.

hoomd/md/EvaluatorPairALJ.h
```diff
@@ -283,7 +283,14 @@ template<unsigned int ndim> class EvaluatorPairALJ
...
                 {
-                faces = ManagedArray<unsigned int>(0, managed);
+                // for ellipsoids, there needs to be a 0 vertex
+                // because that is the 0 of the support function
```
vyasr, line 287:
```suggestion
                // The GJK implementation assumes that all shapes have at least one
                // vertex. For ellipsoids, we must use the origin as that vertex since it
                // guarantees that the polyhedral component of the support function will
                // always be 0 and not contribute to the total support function.
```

hoomd/md/EvaluatorPairALJ.h
```diff
@@ -555,7 +574,7 @@ template<unsigned int ndim> class EvaluatorPairALJ
                           shape_j->rounding_radii,
                           shape_i->has_rounding,
                           shape_j->has_rounding);
-                assert(success && !overlap);
+                // assert(success && !overlap);
```
vyasr, line 577: I left this assertion here intentionally since it's a handy way to detect when something is wrong. If a simulation crashes and it appears to be a bug in the overlap calculation (like the one that @jekw and @thiv89 fixed somewhat recently), recompiling in debug mode and running would immediately tell you that if this assertion failed. I think it's better to just add a comment as to why this is useful.
```suggestion
                // Unphysical ALJ simulation results may be the result of
                // invalid collision detection from GJK, which will normally
                // occur silently. This assertion helps debug such errors by
                // failing fast in debug mode if GJK failed to converge.
                assert(success && !overlap);
```

hoomd/md/pytest/test_aniso_pair.py
```diff
@@ -282,8 +284,8 @@ def to_type_parameter_dicts(types, argument_dict):
                                    (0.01, 0.1, 2.0)],
                 "faces": [
                     md.pair.aniso.ALJ.get_ordered_vertices(vertices)[1]
-                    for vertices in shape_vertices
-                ]
+                    for vertices in shape_vertices[:-1]
+                ] + [[]]
```
vyasr, line 288: Why this change?

vyasr (changes requested):

hoomd/md/pytest/test_aniso_pair.py
tommy-waltmann: This adds the vertices input for the ellipsoids test case, which needs to be an empty list.

hoomd/md/EvaluatorPairALJ.h
```diff
@@ -1598,7 +1624,7 @@ template<> std::string EvaluatorPairALJ<2>::getShapeSpec() const
     std::ostringstream shapedef;
     const ManagedArray<vec3<Scalar>>& verts(shape_i->verts); //! Shape vertices.
     const unsigned int N = verts.size();
-    if (N == 0)
+    if (N < 2)
```
b-butler, line 1627:
```suggestion
    if (N == 1)
```

hoomd/md/EvaluatorPairALJ.h
```diff
@@ -1635,7 +1661,7 @@ template<> std::string EvaluatorPairALJ<3>::getShapeSpec() const
     std::ostringstream shapedef;
     const ManagedArray<vec3<Scalar>>& verts(shape_i->verts);
     const unsigned int N = verts.size();
-    if (N == 0)
+    if (N < 2)
```
b-butler, line 1664:
```suggestion
    if (N == 2)
```

tommy-waltmann: Should this suggestion say, ``if N==1``?

hoomd/md/pytest/test_aniso_pair.py
tommy-waltmann: My bad on the previous comment, I meant to say that it gives the faces input. Passing an empty list into `get_ordered_vertices` causes an error, so I have to explicitly add the empty list here.

hoomd/md/EvaluatorPairALJ.h
b-butler: Yes it should.

b-butler (approved):
:+1:

hoomd/md/pytest/test_aniso_pair.py
vyasr: I don't think the error can come from `get_ordered_vertices`, otherwise the code would fail without every constructing the list in the comprehension. Do you mean that `make_aniso_spec` fails, or maybe the simulation itself?

I realize this is a bit out of scope of what is currently changed in this PR, but in light of what we're testing I'm going to suggest a small pair of modifications to this last block modifying the `valid_params_list`.
1. Rather than wrapping all of the construction of `alj_arg_dict` in a try-except, can we just wrap the call to `get_ordered_vertices`? That makes it much clearer what triggers an error.
2. Rather than indiscriminately adding the extra empty set of faces, can we do that in an `if shape_vertices` block, again so that it's clear why this is needed?

tommy-waltmann: I'd rather address this in another small PR, this bugfix is important and needs to get merged so we can have it in the next beta release. Is everything in this PR other than this okay for merge?

vyasr: I can approve this PR so you can move forward, but I'd strongly recommend addressing this either in this PR or ASAP after it's merged. As the test is currently written it wasn't clear to reviewers that these changes to tests are linked to the bug fixed in this PR, which means any developer reworking these tests in the future could easily omit an equivalent test accidentally and allow bugs like to recur.

vyasr (approved):

hoomd/md/pytest/test_aniso_pair.py
joaander: Merging now as the release deadline is tomorrow. @tommy-waltmann will make a follow up PR to address this comment.

PR #1244, opened 2022-02-17
Update ALJ Tests

tommy-waltmann:
## Description

This PR fixes some tests for the ALJ potential. It both addresses @vyasr 's questions on #1235 which were not addressed before the PR was merged, and it properly adds a test case which was improperly added in #1235 

## Motivation and context

A bug was found in the ALJ potential on #1233 , a fix was made on #1235 , but there were other comments to address.

## How has this been tested?

tests have been updated according to @vyasr 's comments on the previous PR #1235 . Test cases which we improperly added in #1235 have also been fixed so they properly set up the test cases.

## Change log

```
Correct ALJ tests
```

## Checklist:

### DIFF
--- a/hoomd/md/pytest/test_aniso_pair.py
+++ b/hoomd/md/pytest/test_aniso_pair.py
@@ -300,19 +300,15 @@ def _valid_params(particle_types=['A', 'B']):
             'average_simplices': [True, False, True]
         }, 2),
         'shape': ({
-            "vertices":
-                shape_vertices,
+            "vertices": shape_vertices,
             "rounding_radii": [(0.1, 0.01, 0.15), (0.0, 0.0, 0.0)],
             "faces": shape_faces
         }, 1)
     }
 
     valid_params_list.append(
-        make_aniso_spec(
-            md.pair.aniso.ALJ,
-            to_type_parameter_dicts(particle_types, alj_arg_dict1)
-        )
-    )
+        make_aniso_spec(md.pair.aniso.ALJ,
+                        to_type_parameter_dicts(particle_types, alj_arg_dict1)))
 
     return valid_params_list
 
--- a/hoomd/md/pytest/test_aniso_pair.py
+++ b/hoomd/md/pytest/test_aniso_pair.py
@@ -258,8 +258,6 @@ def _valid_params(particle_types=['A', 'B']):
         [(0.5, -0.5, -0.5), (0.5, 0.5, -0.5), (0.5, 0.5, 0.5), (-0.5, 0.5, 0.5),
          (-0.5, 0.5, -0.5), (-0.5, -0.5, 0.5), (0.5, -0.5, 0.5),
          (-0.5, -0.5, -0.5)],
-        # ellipsoid
-        []
     ]
 
     # ALJ.get_ordered_vertices only works if coxeter can be imported, so we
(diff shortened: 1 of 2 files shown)


hoomd/md/pytest/test_aniso_pair.py
```diff
@@ -258,16 +279,14 @@ def to_type_parameter_dicts(types, argument_dict):
...
          (-0.5, -0.5, -0.5)],
-        # ellipsoid
-        []
     ]
 
     # ALJ.get_ordered_vertices only works if coxeter can be imported, so we
-    # check the ModuleNotFoundError and don't add it to the list for the
-    # available tests.
+    # check the RuntimeError and don't add it to the list for the available
+    # tests if coxeter is not available.
```
vyasr, line 286: I think this comment is incorrect. If the problem was not being able to import coxeter we _should_ be catching a `ModuleNotFoundError`, and maybe we need to add a second except clause here so that developers can run HOOMD tests without coxeter installed. The current except clause is present because `get_ordered_vertices` can fail if it tries to construct a `coxeter.ConvexPolyhedron` with 0 vertices, right? It looks like that is not an issue anymore anyway because of the way that you have split the params into the `alj_arg_dict0` and `alj_arg_dict1` parsing. Am I misunderstanding that?

vyasr (changes requested):
Thanks for fixing this Tommy, it'll be good to get these tests set up to prevent future regressions.

hoomd/md/pytest/test_aniso_pair.py
tommy-waltmann: yeah, good catch. It should be catching a `ModuleNotFoundError` and the comment should reflect that. We should still be able to run the test without coxeter installed once I change the `except RuntimeError` to `Except ModuleNotFoundError`.

joaander: Please test locally in an env without coxeter. The CI test environments all have coxeter installed.

joaander: The code that attempts to use coxeter uses this code:
```
        try:
            import coxeter
        except ImportError as error:
            raise RuntimeError(
                "Method requires coxeter as a dependency.") from error
```
There is no guarantee that the reason Python cannot import coxeter is because of a `ModuleNotFoundError`. From HOOMD's perspective, the failure to import coxeter is a `FuntimeError` and the `from error` correctly passes the source exception back to the user so they can debug.

This test is working fine without coxeter installed.

If you are looking to improve this code, you could simplify the code by removing the need for coxeter in the tests. Hard-code the faces, areas, and other parameters that are currently pulled from coxeter in `test_alj.py` and `test_aniso_pair.py`. Then all these tests can run whether the user has coxeter installed or not.

The only test that should require coxeter is a unit test for `get_ordered_vertices` that ensures it returns the expected values.

joaander (commented):
Why does `get_ordered_verticies` exist? It is a very thin wrapper for `coxeter` and could be replaced by a howto or short example script.

tommy-waltmann:
`get_ordered_vertices` has existed for as long as I have been working on this PR, you'd have to ask the person who added it initially. I'll hard code the info taken from calls to `get_ordered_vertices`. Removing `get_ordered_vertices` would also allow us to remove an optional dependency, but I think that should be a part of another PR.

vyasr (approved):
I'm approving this PR as is. I would be fine with removing `get_ordered_vertices` in a follow-up and updating documentation to point users to coxeter.

joaander (approved):
Thanks!

PR #1138, opened 2021-09-17
Implement long range correction for LJ potential

jennyfothergill:
## Description

Adds analytical tail correction to the LJ pair potential according to eq 5 from https://doi.org/10.1021/jp980939v

## Motivation and context

This change is needed for the correct implementation of certain forcefields, e.g., the TrappeUA forcefield.

Resolves #1139 

## How has this been tested?

(WIP) Comparing potential energy and instantaneous pressure difference of identical system with and without tail correction to that obtained in GROMACS, LAMMPS, GOMC, Cassandra, and MCCCS-MN.

Engines | Energy w/ tail correction (kJ/mol) | Energy w/o tail correction (kJ/mol) | Energy difference (kJ/mol) | Pressure w/ tail correction (kJ/mol/nm^3) | Pressure w/o tail correction (kJ/mol/nm^3) | Pressure difference (kJ/mol/nm^3)
-- | -- | -- | -- | -- | -- | --
HOOMD | 459209.3835 | 459337.6130 | -128.2295 | 30389.01054 | 30393.02325 | -4.01271145
GROMACS 2021.1 | 459060.5625 | 459317.1563 | -256.5938 |   |   | 0
GROMACS 2020.6 | 459188.9063 | 459317.1563 | -128.2500 |   |   |  
LAMMPS | 459215.4639 | 459343.6617 | -128.1978 | 30388.44712 | 30392.45966 | -4.01253809
GOMC | 459216.0800 | 459340.8000 | -124.7200 |   |   | 0
Cassandra | 459210.0760 | 459338.3060 | -128.2300 |   |   | 0
MCCCS-MN | 459206.6527 | 459334.8815 | -128.2287 | 30489.28295 | 30491.28953 | -2.006581786

## Change log

```
- Add  energy and pressure tail correction to LJ Pair Potential according to eq 5 from https://doi.org/10.1021/jp980939v.
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.
- [x] Make sure mode is "none" when using tail correction.
- [x] Validate against all engines.
- [ ] Improve performance.

PR #1246, opened 2022-02-18
Check if rcut is zero before computing tail correction

jennyfothergill:
## Description

I don't know how this didn't get included in #1138 because at some point I'm sure I had this locally. (Sorry!) Anyway, this checks whether rcut is 0 before computing the tail correction term. 

## Motivation and context

Without this fix if any rcut value is 0, `lj.additional_energy` will be nan

## How has this been tested?

I am testing this in the [reproducibility_study](https://github.com/mosdef-hub/reproducibility_study).
What would be the best way to add a test that would catch this failure? My idea would be something like a MWE of lennard-jonesium with tail correction enabled, run 0 steps, and assert that lj.additional_energy is a float value.

## Change log

```
Fix bug where hoomd.md.pair.LJ.additional_energy is nan when tail_correction is enabled with pairs for which r_cut is 0
```

## Checklist:

### DIFF
--- a/hoomd/md/EvaluatorPairLJ.h
+++ b/hoomd/md/EvaluatorPairLJ.h
@@ -233,6 +233,10 @@ class EvaluatorPairLJ
 
     DEVICE Scalar evalPressureLRCIntegral()
         {
+        if (rcutsq == 0)
+            {
+            return Scalar(0.0);
+            }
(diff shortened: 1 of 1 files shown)


b-butler (approved):
Thanks @jennyfothergill!

PR #1258, opened 2022-02-23
Remove IntegratorData class and usage

### LINKED ISSUE #1228: Remove IntegratorData
## Description

Remove the unused IntegratorData API.

## Motivation and context

Integration methods should store their additional degrees of freedom in local variables. The v3 Python API allows users to access these degrees of freedom as needed.

b-butler:
## Description

Integrator degrees of freedom are now stored in the integration methods
themselves.

## Motivation and context

Resolves #1228

## Change log

```
- Removed: (dev) ``IntegratorData`` class and its usages. It is replaced by ``std::vector`` instances that are integrator class members.
```

## Checklist:


hoomd/md/TwoStepNPTMTK.h
```diff
@@ -255,6 +245,9 @@ class PYBIND11_EXPORT TwoStepNPTMTK : public IntegrationMethodTwoStep
 
     Scalar m_gamma; //!< Optional damping factor for box degrees of freedom
 
+    std::vector<Scalar> m_thermostat; //!< thermostat degrees of freedom
+    std::vector<Scalar> m_barostat;   //!< barostat degrees of freedom
```
tommy-waltmann, line 249: Since the sizes of these arrays are static, why not declare them as `std::array` instead of `std::vector`? It would make their intended usage clearer to someone trying to understand the code.

tommy-waltmann (commented):
This is almost there, just one small comment.

hoomd/md/TwoStepNPTMTK.h
b-butler: No reason, that is a good idea.

joaander: I suggested offline to use `m_xi`, `m_eta`, etc... in place of the thermostat array to avoid the potential for error when extracting values from the array by index.

tommy-waltmann: That's a good idea, could also just document which index corresponds to which variable in a comment above where the member variable is declared.

b-butler: Oh, @joaander I misunderstood your suggestion. I can switch to that model then. Although I may create simple structs to serve that purpose then.

hoomd/md/TwoStepNPTMTK.h
```diff
@@ -43,6 +43,26 @@ namespace md
...
+    /// Barostat variables
+    struct Barostat
+        {
+        Scalar nu_xx;
+        Scalar nu_xy;
+        Scalar nu_xz;
+        Scalar nu_yy;
+        Scalar nu_yz;
+        Scalar nu_zz;
+        };
```
tommy-waltmann, line 64: I these structs should be declared in the private portion of the class.

hoomd/md/TwoStepNVTMTK.h
```diff
@@ -41,6 +41,14 @@ namespace md
...
     {
     public:
+    struct Thermostat
+        {
+        Scalar xi = 0;
+        Scalar eta = 0;
+        Scalar xi_rot = 0;
+        Scalar eta_rot = 0;
+        };
+
```
tommy-waltmann, line 51: Same as above comment

tommy-waltmann (changes requested):
One small change, then this is ready to go

hoomd/md/TwoStepNPTMTK.h
b-butler: I moved these to the `protected` section since that is where the struct instance is stored.

tommy-waltmann (approved):

joaander (approved):
Thanks!

PR #1260, opened 2022-02-24
Remove get_ordered_vertices

tommy-waltmann:
## Description

This PR removes the convenience method `get_ordered_vertices` of the ALJ potential in favor of having users just use `coxeter` in their scripts. This PR removes coxeter as an optional dependency for hoomd.

## Motivation and context

On #1244 we fixed the way ALJ tests were written and the decision was also made that `get_ordered_vertices` was no longer necessary as a convenience method for ALJ.

## How has this been tested?

No updates to testing are needed for this, because all tests were removed of their coxeter dependency in #1244

## Change log

```
Remove `get_ordered_vertices` and optional coxeter dependency from `hoomd.md.pair.aniso.ALJ`.
```

## Checklist:

### DIFF
--- a/BUILDING.rst
+++ b/BUILDING.rst
@@ -120,10 +120,6 @@ Install prerequisites
 
 - LLVM >= 10.0, < 13
 
-**To use** `md.pair.aniso.ALJ.get_ordered_vertices`:
-
-- coxeter
-
 **To build the documentation:**
 
 - sphinx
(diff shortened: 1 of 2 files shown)


vyasr (approved):
LGTM

j-proc (approved):
Makes sense. My only comment is that "The vertices must be ordered." is a vague. Ordered how? If coxeter has it documented could always just link there.

vyasr:
> Makes sense. My only comment is that "The vertices must be ordered." is a vague. Ordered how? If coxeter has it documented could always just link there.

Good point. The requirement is that the vertices must be ordered counterclockwise with respect to the face normal vector pointing outward from the origin. `coxeter` automatically orders the faces this way because it is also necessary for other geometric calculations like the signed area to work correctly, but HOOMD should document that as well. @tommy-waltmann could you add that note?

joaander (approved):
Thanks! Sorry I missed this in the initial branch.

PR #1274, opened 2022-03-15
Refactor actions workflow to separate build and test runners.

joaander:
## Description

* Refactor the GitHub Actions workflows to reduce build times.
* Use jetstream2 to compile hoomd and reduce the latency to run a full batch of unit tests in CI.
* Remove CUDA compute arches that aren't in the test pool.
* Add builds with `BUILD_MD=off`, `BUILD_HPMC=off` and both `BUILD_MD` and `BUILD_HPMC` off.
* Fix failing build with `BUILD_MD=off`.

## Motivation and context

Reducing the build time will improve developer productivity. Adding tests with components disabled ensures that tests will pass with these configurations.

The jetstream2 instances make fewer total CPU cores available than hosted runners, but individual runners have 8 cores - allowing CPU builds of HOOMD in 2-3 minutes (compare to 12 on minutes on hosted runners) and GPU builds in 13 minutes (compare to 80+ minutes on hosted runners). This reduces the time to complete a single build, but *many* queued builds may take longer. If this becomes a problem, we can always explore the possibility expanding the jetstream2 pool.

At this time, jetstream2 GPUs cannot be used to run the GPU tests as A100 vGPUs do not support unified memory. It is not clear whether this will ever be supported.

## How has this been tested?

The CI runs on this pull request are the test.

## Checklist:

### DIFF
--- a/.github/workflows/make_workflows.py
+++ b/.github/workflows/make_workflows.py
@@ -30,19 +30,23 @@ if __name__ == '__main__':
     for name, configuration in configurations.items():
         for entry in configuration:
             if entry['config'].startswith('[cuda'):
-                entry['runner'] = "[self-hosted,GPU]"
+                entry['build_runner'] = "[self-hosted,jetstream2,CPU]"
+
+                entry['test_runner'] = "[self-hosted,GPU]"
                 # device options needed to access the GPU devices on the runners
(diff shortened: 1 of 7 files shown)


joaander:
Tests are passing and the run time is much improved. Merging without review. We can always adjust this as needed if we run into difficulties.

PR #1264, opened 2022-03-04
Add trigger member to Updater and Analyzer

b-butler:
## Description

Adds a `m_trigger` member to `Updater` and `Analyzer` in C++. Also removes the `_TriggeredOperation` class in Python.

## Motivation and context

Simplifies the code.

Resolves #???

## How has this been tested?

Existing tests pass.

## Change log

```
Changed: [developer] ``Updater`` and ``Analyzer`` in C++ have a `m_trigger` member now.
Changed: [developer] `_TriggeredOperation` has been moved to `TriggeredOperation` and custom trigger setting and getting logic removed.
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/master/CONTRIBUTING.md).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/master/ContributorAgreement.md).
- [x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

PR #1290, opened 2022-03-31
Fix TPS updating

b-butler:
## Description

Fixes a bug that only updates simulation TPS at the end of every run call.

## Motivation and context

This makes it difficult to accurately measure performance and unintuitive logger output.

Resolves #???

## How has this been tested?

A check for TPS updating has been added to the tests.

## Change log

```
Fixed: ``Simulation.tps`` now updates every step of the run.
```

## Checklist:

### DIFF
--- a/hoomd/System.cc
+++ b/hoomd/System.cc
@@ -167,9 +167,9 @@ void System::run(uint64_t nsteps, bool write_at_start)
             if ((*analyzer->getTrigger())(m_cur_tstep))
                 analyzer->analyze(m_cur_tstep);
             }
-        }
 
-    updateTPS();
+        updateTPS();
+        }
 
     // propagate Python exceptions related to signals
     if (PyErr_CheckSignals() != 0)
--- a/hoomd/pytest/test_simulation.py
+++ b/hoomd/pytest/test_simulation.py
@@ -123,7 +123,7 @@ def test_tps(simulation_factory, two_particle_snapshot_factory):
     assert sim.tps == 0
 
     list_writer = ListWriter(sim, "tps")
-    sim.operations += list_writer
+    sim.operations.writers.append(hoomd.write.CustomWriter(action=list_writer, trigger=hoomd.trigger.Periodic(1)))
     sim.run(10)
     tps = list_writer.data
     assert len(np.unique(tps)) > 1
@@ -137,7 +137,7 @@ def test_walltime(simulation_factory, two_particle_snapshot_factory):
     assert sim.walltime == 0
 
     list_writer = ListWriter(sim, "walltime")
-    sim.operations += list_writer
+    sim.operations.writers.append(hoomd.write.CustomWriter(action=list_writer, trigger=hoomd.trigger.Periodic(1)))
     sim.run(10)
     walltime = list_writer.data
     assert all(a > b for a, b in zip(walltime[1:], walltime[:-1]))
--- a/hoomd/pytest/test_simulation.py
+++ b/hoomd/pytest/test_simulation.py
@@ -123,7 +123,7 @@ def test_tps(simulation_factory, two_particle_snapshot_factory):
     assert sim.tps == 0
 
     list_writer = ListWriter(sim, "tps")
-    sim += list_writer
+    sim.operations += list_writer
     sim.run(10)
     tps = list_writer.data
     assert len(np.unique(tps)) > 1
@@ -137,7 +137,7 @@ def test_walltime(simulation_factory, two_particle_snapshot_factory):
     assert sim.walltime == 0
 
     list_writer = ListWriter(sim, "walltime")
-    sim += list_writer
+    sim.operations += list_writer
     sim.run(10)
     walltime = list_writer.data
     assert all(a > b for a, b in zip(walltime[1:], walltime[:-1]))
--- a/hoomd/pytest/test_simulation.py
+++ b/hoomd/pytest/test_simulation.py
@@ -123,7 +123,9 @@ def test_tps(simulation_factory, two_particle_snapshot_factory):
     assert sim.tps == 0
 
     list_writer = ListWriter(sim, "tps")
-    sim.operations.writers.append(hoomd.write.CustomWriter(action=list_writer, trigger=hoomd.trigger.Periodic(1)))
+    sim.operations.writers.append(
+        hoomd.write.CustomWriter(action=list_writer,
+                                 trigger=hoomd.trigger.Periodic(1)))
     sim.run(10)
     tps = list_writer.data
     assert len(np.unique(tps)) > 1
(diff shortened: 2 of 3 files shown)


hoomd/pytest/test_simulation.py
```diff
@@ -115,14 +115,21 @@ def test_run(simulation_factory, lattice_snapshot_factory):
...
+
+    sim.run(10)
+    tps = sim.tps
+    assert tps > 0
+    assert tps == sim.tps
 
-    assert sim.tps > 0
+    # Test that tps updates
+    sim.run(10)
+    assert sim.tps != tps
```
cbkerr: Doesn't this have a tiny chance of failing?
Can we do something like increase the r_cut on the 2nd run to make sure it runs slower?

hoomd/System.cc
```diff
@@ -167,9 +167,9 @@ void System::run(uint64_t nsteps, bool write_at_start)
             if ((*analyzer->getTrigger())(m_cur_tstep))
                 analyzer->analyze(m_cur_tstep);
             }
-        }
 
-    updateTPS();
+        updateTPS();
```
joaander, line 171: This was working at one point: https://hoomd-blue.readthedocs.io/en/v3.0.0/tutorial/02-Logging/04-Writing-Formatted-Output.html

It was broken in #1264 : https://github.com/glotzerlab/hoomd-blue/pull/1264/files#diff-a241487071270e42a76be6ecb1d44bfeb8a0b5512de965d893dfd8517db0e347

hoomd/pytest/test_simulation.py
```diff
@@ -115,14 +115,21 @@ def test_run(simulation_factory, lattice_snapshot_factory):
     assert sim.operations._scheduled
 
 
-def test_tps(simulation_factory, lattice_snapshot_factory):
+def test_tps(simulation_factory, two_particle_snapshot_factory):
```
joaander, line 118: The fix in this pull request also fixes the `walltime` property. Add a similar test (or include in this one renamed to `test_tps_and_walltime`).

hoomd/pytest/test_simulation.py
joaander: Yes, there is a small chance this could fail. There are no particle interactions and no r_cut to change in this test. To ensure different timings, add a custom action that sleep different amounts of time in the two tests.

hoomd/pytest/test_simulation.py
```diff
@@ -115,14 +115,21 @@ def test_run(simulation_factory, lattice_snapshot_factory):
...
     assert sim.tps is None
 
-    sim = simulation_factory(lattice_snapshot_factory())
-    sim.run(100)
+    sim = simulation_factory(two_particle_snapshot_factory())
+    assert sim.tps == 0
+
+    sim.run(10)
+    tps = sim.tps
+    assert tps > 0
```
joaander: This doesn't test that `tps` and `walltime` update during the run. Use `conftest.ListWriter` instances to log these quantities and check that the bug is indeed fixed.

joaander (changes requested):

hoomd/System.cc
b-butler: Yep, I noticed that I messed it up.

hoomd/pytest/test_simulation.py
b-butler: I now run 10 steps and only require that at least 2 TPS values exist. I feel this is so unlikely not to occur with the given clock to the degree of floating point accuracy that this should be robust. If you disagree though, I will add a custom action (or if tests prove me wrong :sweat_smile: )

joaander: I don't disagree and am not patient enough to run this continually until it fails. The rate of failure is likely now much smaller than the rate of random false failures on other tests in the codebase.

joaander (approved):
Thanks!
@cbkerr do you approve as well?

cbkerr (approved):
Looks good, thanks!

joaander:
Tests are failing: `TypeError: unsupported operand type(s) for +=: 'Simulation' and 'ListWriter'`

tcmoore3 (approved):
LGTM once the failing test is fixed. 

Good lesson in testing *everything*.

hoomd/pytest/test_simulation.py
```diff
@@ -115,14 +115,32 @@ def test_run(simulation_factory, lattice_snapshot_factory):
...
     sim = simulation_factory()
     assert sim.tps is None
 
-    sim = simulation_factory(lattice_snapshot_factory())
-    sim.run(100)
+    sim = simulation_factory(two_particle_snapshot_factory())
+    assert sim.tps == 0
+
+    list_writer = ListWriter(sim, "tps")
+    sim.operations += list_writer
```
joaander:
```suggestion
    sim.operations.writers.append(hoomd.write.CustomWriter(action=list_writer, trigger=hoomd.trigger.Periodic(1)))
```

hoomd/pytest/test_simulation.py
```diff
@@ -115,14 +115,32 @@ def test_run(simulation_factory, lattice_snapshot_factory):
...
+def test_walltime(simulation_factory, two_particle_snapshot_factory):
+    sim = simulation_factory()
+    assert sim.walltime == 0
+
+    sim = simulation_factory(two_particle_snapshot_factory())
+    assert sim.walltime == 0
 
-    assert sim.tps > 0
+    list_writer = ListWriter(sim, "walltime")
+    sim.operations += list_writer
```
joaander:
```suggestion
    sim.operations.writers.append(hoomd.write.CustomWriter(action=list_writer, trigger=hoomd.trigger.Periodic(1)))
```

joaander (changes requested):
This should fix the failing tests.

joaander (approved):

joaander:
Well, it wasn't long before one of these tests failed:
https://github.com/glotzerlab/hoomd-blue/runs/5885953832?check_suite_focus=true#step:6:6476
```
>       assert all(a > b for a, b in zip(walltime[1:], walltime[:-1]))
E       assert False
```

I'm adding a `print(walltime)` so we can see how this is failing when it triggers next time.

joaander:
Actually, we don't need to wait. I printed it and examined the output when passing:
```
[0.0, 6e-06, 6e-06, 7e-06, 7e-06, 8e-06, 9e-06, 9e-06, 1e-05, 1e-05]
```
With no operations, timesteps can occur at a rate faster than the timer resolution. We need to make the check `>=`.

joaander:
With this sort of timer resolution, it may not be long before the TPS test fails as there are only a small discrete number of TPS values possible at this speed :)
```
print(tps): [0.0, 142857.14285714287, 250000.0, 375000.0, 444444.44444444444, 555555.5555555555, 600000.0, 636363.6363636364, 727272.7272727273, 750000.0]
```

b-butler:
Hmm, well there goes my intuition. I will have the fix posthaste.

joaander:
I put the quick `>=` fix for walltime on #1296 as it was failing checks there.

PR #1300, opened 2022-04-11
Zero forces in CustomForceCompute

### LINKED ISSUE #1298: hoomd.md.force.Custom should zero forces
## Description

When users set a subset of the force array, forces from previous timesteps remain.

`CustomForceCompute::computeForces` should write 0's to all forces, potential energies, torqes, and virials before calling `set_forces`.

## Script

```python
class customforce(hoomd.md.force.Custom):
        def __init____(self):
            super().__init__(aniso=True)
        def set_forces(self,timestep):
                with sim.state.cpu_local_snapshot as data:
                    idx = data.particles.rtag[0]
                    N = len(data.particles.position)
                with self.cpu_local_force_arrays as arrays:
                    if idx < N:
                        arrays.force[idx] = [10,0,0]
```

## Configuration

Platform:
- Mac
- Linux
- GPU
- CPU

b-butler:
## Description

Zero forces in `hoomd.md.force.CustomForce`.

## Motivation and context

Resolves #1298

## How has this been tested?

Test has been added.

## Change log

```
Fixed: Forces are zeroed in ``hoomd.md.CustomForce`` between time steps.
```

## Checklist:

### DIFF
--- a/hoomd/md/pytest/test_custom_force.py
+++ b/hoomd/md/pytest/test_custom_force.py
@@ -392,7 +392,6 @@ def test_force_zeroing(force_simulation_factory, two_particle_snapshot_factory):
                 force.force[:] += timestep
                 force.torque[:] += timestep
                 force.virial[:] += timestep
-                print(f"{force.virial.shape=}")
 
     snap = two_particle_snapshot_factory()
     test_force = TestForceZeroing()
--- a/hoomd/md/pytest/test_custom_force.py
+++ b/hoomd/md/pytest/test_custom_force.py
@@ -378,3 +378,40 @@ def test_torques_update(local_force_names, two_particle_snapshot_factory,
         if sim.device.communicator.rank == 0:
             assert np.count_nonzero(snap.particles.orientation
                                     - initial_orientations)
+
+
+def test_force_zeroing(force_simulation_factory, two_particle_snapshot_factory):
+
+    class TestForceZeroing(hoomd.md.force.Custom):
+
+        def __init__(self):
+            super().__init__(aniso=True)
+
+        def set_forces(self, timestep):
+            with self.cpu_local_force_arrays as force:
+                force.force[:] += timestep
+                force.torque[:] += timestep
+                force.virial[:] += timestep
+                print(f"{force.virial.shape=}")
+
+    snap = two_particle_snapshot_factory()
+    test_force = TestForceZeroing()
+    sim = force_simulation_factory(test_force, snap)
+    if sim.device.communicator.rank == 0:
+        snap.particles.moment_inertia[:] = [[1, 1, 1], [1, 1, 1]]
+    sim.state.set_snapshot(snap)
+
+    _skip_if_gpu_device_and_no_cupy(sim)
+    sim.operations.integrator.integrate_rotational_dof = True
+    sim.always_compute_pressure = True
+
+    for _ in range(3):
+        sim.run(1)
+        forces = test_force.forces
+        torques = test_force.torques
+        virials = test_force.virials
+        timestep = sim.timestep
+        if sim.device.communicator.rank == 0:
+            assert np.allclose(forces, timestep)
+            assert np.allclose(torques, timestep)
+            assert np.allclose(virials, timestep)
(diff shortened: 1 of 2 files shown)


hoomd/md/pytest/test_custom_force.py
```diff
@@ -378,3 +378,37 @@ def test_torques_update(local_force_names, two_particle_snapshot_factory,
...
+    sim.state.set_snapshot(snap)
+
+    _skip_if_gpu_device_and_no_cupy(sim)
+    sim.operations.integrator.integrate_rotational_dof = True
+    sim.always_compute_pressure = True
+
+    for _ in range(3):
+        sim.run(1)
+        snap = sim.state.get_snapshot()
+        step = sim.sim.timestep
```
tommy-waltmann:
```suggestion
        step = sim.timestep
```

hoomd/md/pytest/test_custom_force.py
```diff
@@ -378,3 +378,37 @@ def test_torques_update(local_force_names, two_particle_snapshot_factory,
         if sim.device.communicator.rank == 0:
             assert np.count_nonzero(snap.particles.orientation
                                     - initial_orientations)
+
+
+def test_force_zeroing(force_simulation_factory, two_particle_snapshot_factory):
```
tommy-waltmann, line 383: How does this test make sure the forces are zeroed every timestep before calling `set_forces`? It doesn't seem to me that its doing that.

tommy-waltmann (changes requested):

hoomd/md/pytest/test_custom_force.py
b-butler: Thanks!

hoomd/md/pytest/test_custom_force.py
b-butler: If they weren't zeroed then the force arrays would be the sum of all timesteps up to that point not simply the current timestep.

joaander: I'm not seeing it either. You assign the whole array to the value `timestep`. Maybe you mean to use `+= timestep`?

hoomd/md/pytest/test_custom_force.py
```diff
@@ -378,3 +378,40 @@ def test_torques_update(local_force_names, two_particle_snapshot_factory,
...
+
+        def __init__(self):
+            super().__init__(aniso=True)
+
+        def set_forces(self, timestep):
+            with self.cpu_local_force_arrays as force:
+                force.force[:] = timestep
+                force.torque[:] = timestep
+                force.virial[:] = timestep
+                print(f"{force.virial.shape=}")
```
tommy-waltmann:
```suggestion

```

tommy-waltmann (changes requested):
My comment from the last review still applies: setting the force arrays to a value each time step and then checking they are the value they were set to doesn't ensure the forces are being zeroed every time step.

hoomd/md/pytest/test_custom_force.py
b-butler: That is what I was thinking in my head @joaander; but apparently not in my code. Sorry @tommy-waltmann for the confusion on my part.

tommy-waltmann (approved):
Looks ready to go to me.

hoomd/md/pytest/test_custom_force.py
```diff
@@ -378,3 +378,39 @@ def test_torques_update(local_force_names, two_particle_snapshot_factory,
...
+
+    _skip_if_gpu_device_and_no_cupy(sim)
+    sim.operations.integrator.integrate_rotational_dof = True
+    sim.always_compute_pressure = True
+
+    for _ in range(3):
+        sim.run(1)
+        forces = test_force.forces
+        torques = test_force.torques
+        virials = test_force.virials
```
syjlee, line 411: Just from the blob of the class TestForceZeroing, should there be `s` on test_force.forces/torques/virials?

joaander: Yes, the Python and C++ APIs use slightly different names. In Python, it is customary to use the plural form to denote lists or arrays as a Python variable may be of any type. This is not necessary in C++ where every variable has a well defined type and that type makes it clear whether the variable represents an array or a single item.

Thanks for taking the time to review and to note a potential problem.

PR #1305, opened 2022-04-20
Reduce time and memory needed to compile HOOMD.

### LINKED ISSUE #614: Improve compilation times and memory requirements
## Description

Improve compilation times and memory requirements. Currently, `module-md.cc` and the various HPMC integrator builds take a long time to build and use several GB of memory for each file.

## Proposed solution

Instantiate pair potential, HPMC integrator, and other classes with one class per file. 

## Additional context

This will create a small additional maintenance burden with more files to track in `CMakeLists.txt`. However, the benefits to decreases in the compilation time and the reduced memory requirements will be well worth it.

## Developer

I will implement this after HOOMD v3 enters the beta phase.

joaander:
## Description

Reduce the time and memory needed to build the `md` component by:
* Breaking large class template instantiations across multiple files, using CMake's `configure_file` to reduce maintenence.
* Skip building `shift_mode=1` PotentialPairAniso kernels where the potential doesn't use `shift_mode`.

## Motivation and context

The time and memory needed to build HOOMD was larger than needed. Especially `module-md.cc` which required more than 3GB of RAM to compile, took over 1 minute, *and* had so many dependencies that almost any change to the md code would cause it to rebuild:

Here is the build trace of `trunk-minor` building *just* the `md` component on `cheme-hodges`
original-md
Profiled with [ninjatracing](https://github.com/nico/ninjatracing) and visualized with [perfetto](https://ui.perfetto.dev/):

After this pull request, `md` can be built in about half the total time, there is no long `module-md.cc` build and the largest memory usage by a single file compilation is about 1 GB.
updated-md

On, `cheme-hodges`, all of HOOMD (including all optional dependencies, such as LLVM and MPI) now completes in about 4 minutes and the largest memory usage by a single file compilation is 1.5 GB.

This pull request focuses on `md`. Possibly more work could be done on `hpmc` following this pattern, but I will leave that for a future pull request. As demonstrated by the time and memory needed to build all of HOOMD, there aren't any major issues left to solve. The best we could do is prevent the current 18 second build of `hpmc/module.cc` from triggering on unrelated changes.

Resolves #614

## How has this been tested?

Existing unit tests pass.

TODO:

- [x] Add deprecation to `deprecated.rst`

## Change log

```
Changed

* Reduced time and memory needed to compile HOOMD-blue from source.

Deprecated:

* ``mode`` parameter in ``hoomd.md.pair.aniso.ALJ`` (``mode`` is documented, but has no effect since v3.0.0).
* ``mode`` parameter in ``hoomd.md.pair.aniso.Dipole`` (``mode`` is documented, but has no effect since v3.0.0).
```

## Checklist:


hoomd/md/export_PotentialPair.cc.inc
```diff
@@ -0,0 +1,16 @@
...
+
+#include "hoomd/md/PotentialPair.h"
+#include "hoomd/md/EvaluatorPair@_evaluator_cpp@.h"
+
+namespace hoomd { namespace md { namespace detail {
+
+void export_PotentialPair@_evaluator@(pybind11::module &m)
+    {
+    export_PotentialPair<EvaluatorPair@_evaluator_cpp@>(m, "PotentialPair@_evaluator@");
+    }
```
tommy-waltmann: Is this way of having CMake generate new files helpful for the build time, or does it just make code neater in `module.cc`?

hoomd/md/TwoStepNVE.h
```diff
@@ -62,7 +62,6 @@ class PYBIND11_EXPORT TwoStepNVE : public IntegrationMethodTwoStep
 namespace detail
     {
 //! Exports the TwoStepNVE class to python
-void export_TwoStepNVE(pybind11::module& m);
     } // end namespace detail
```
tommy-waltmann: Remove the detail namespace here as well.

hoomd/md/test/test_nve_integrator.cc
```diff
@@ -420,8 +421,9 @@ void nve_updater_aniso_test(std::shared_ptr<ExecutionConfiguration> exec_conf,
     std::shared_ptr<NeighborList> nlist_1(new NeighborListBinned(sysdef_1, r_cut, r_buff));
 
     nlist_1->setStorageMode(NeighborList::full);
-    std::shared_ptr<AnisoPotentialPairGB> fc_1
-        = std::shared_ptr<AnisoPotentialPairGB>(new AnisoPotentialPairGB(sysdef_1, nlist_1));
+    std::shared_ptr<AnisoPotentialPair<EvaluatorPairGB>> fc_1
+        = std::shared_ptr<AnisoPotentialPairGB>(
+            new AnisoPotentialPair<EvaluatorPairGB>(sysdef_1, nlist_1));
```
tommy-waltmann:
```suggestion
    auto fc_1 = std::make_shared<AnisoPotentialPair<EvaluatorPairGB>>(sysdef_1, nlist_1));
```
While we're fixing this line, let's just make it a lot less verbose.

hoomd/md/test/test_nvt_mtk_integrator.cc
```diff
@@ -193,8 +194,9 @@ void test_nvt_mtk_integrator_aniso(std::shared_ptr<ExecutionConfiguration> exec_
     std::shared_ptr<NeighborList> nlist_1(new NeighborListBinned(sysdef_1, r_cut, r_buff));
 
     nlist_1->setStorageMode(NeighborList::full);
-    std::shared_ptr<AnisoPotentialPairGB> fc_1
-        = std::shared_ptr<AnisoPotentialPairGB>(new AnisoPotentialPairGB(sysdef_1, nlist_1));
+    std::shared_ptr<AnisoPotentialPair<EvaluatorPairGB>> fc_1
+        = std::shared_ptr<AnisoPotentialPair<EvaluatorPairGB>>(
+            new AnisoPotentialPair<EvaluatorPairGB>(sysdef_1, nlist_1));
```
tommy-waltmann:
```suggestion
    auto fc_1 = std::make_shared<AnisoPotentialPair<EvaluatorPairGB>>(sysdef_1, nlist_1));
```
Same here.

hoomd/test/test_gpu_array.cc
```diff
@@ -16,6 +16,9 @@
 using namespace hoomd::test;
 #endif
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wpessimizing-move"
```
tommy-waltmann, line 20: Why were the pragmas added to this file?

sphinx-doc/deprecated.rst
```diff
@@ -18,3 +18,9 @@ v3.x
    * - Particle diameters
      - Potentials such as `md.pair.ExpandedLJ`.
      - v3.0.0
+   * - ``hoomd.md.pair.aniso.ALJ.mode`` parameter
+     - ``mode`` has no effect.
+     - v3.1.0
+   * - ``hoomd.md.pair.aniso.Dipole.mode`` parameter
+     - ``mode`` has no effect.
+     - v3.1.0
```
tommy-waltmann: To be clear, even before the mode parameter for these classes was officially declared deprecated, the mode parameter did nothing, right?

hoomd/md/PotentialBondGPU.cuh
```diff
@@ -242,9 +242,10 @@ __global__ void gpu_compute_bond_forces_kernel(Scalar4* d_force,
     This is just a driver function for gpu_compute_bond_forces_kernel(), see it for details.
 */
 template<class evaluator>
-hipError_t gpu_compute_bond_forces(const kernel::bond_args_t& bond_args,
-                                   const typename evaluator::param_type* d_params,
-                                   unsigned int* d_flags)
+__attribute__((visibility("default"))) hipError_t
```
tommy-waltmann, line 245: Why does the visibility attribute need to be added here?

tommy-waltmann (changes requested):
Here is a first round of comments, I'll add some more throughout the week as I continue to dig through this.

hoomd/md/export_PotentialPair.cc.inc
joaander: This is not valid C++ syntax. The CMake command `configure_file` uses this as a template to generate valid C++ files at configure time: https://github.com/glotzerlab/hoomd-blue/blob/c87d24f91d8b70a40437dedb4d9710226880dcee/hoomd/md/CMakeLists.txt#L477-L523

The alternative is to create the 100+ files like this by hand rather than using loops in CMake with template files. They must be separate files so they can be compiled in parallel to build quickly and with low memory requirements.

hoomd/test/test_gpu_array.cc
joaander: The 3D code path for GJK started generating warnings when compiled for 2D shapes:
https://github.com/glotzerlab/hoomd-blue/blob/c87d24f91d8b70a40437dedb4d9710226880dcee/hoomd/md/GJK_SV.h#L461-L468

I don't know why the warnings started with this pull request, but on CI I require that we build warning free.

sphinx-doc/deprecated.rst
joaander: Yes.

hoomd/md/PotentialBondGPU.cuh
joaander: I added it to be consistent with the others (pair, external, etc...). It would be needed for any C++ unit tests that use PotentialBond*GPU classes. With pybind11, the visibility is hidden so the methods are not available to executables linked to the `_md` library.

hoomd/md/TwoStepNVE.h
joaander: Good catch.

jekw90:
It is really nice that the building time for Hoomd completes in about 4 minutes! This is really awesome.
I will go on a long vacation soon, and I have been busy finishing up stuffs before going on the vacation.
So, I think that it would be nice if other peeps review this pull-request.
Thank you.

sphinx-doc/deprecated.rst
joaander: I changed the documentation to clarify this point.

hoomd/md/export_PotentialPair.cc.inc
joaander: Using CMake `configure_file` has the added benefit that new pair/bond/external/... potentials require less effort to add:
1. Add evaluator.h
2. Add name to the appropriate `_evaluators` list
3. Add python wrapper.

hoomd/test/test_gpu_array.cc
b-butler: We do use C++17 features now, but as this must compile on the GPU we cannot use `if constexpr`. Or at least I assume this is the case.

joaander: You could try it and see if it works.

vyasr: You can use `if constexpr` in device code.

joaander:
> It is really nice that the building time for Hoomd completes in about 4 minutes! This is really awesome. I will go on a long vacation soon, and I have been busy finishing up stuffs before going on the vacation. So, I think that it would be nice if other peeps review this pull-request. Thank you.

Thanks for letting us know, @jekw90.

hoomd/test/test_gpu_array.cc
joaander: @vyasr nvcc gives an error as we pass in the c++14 standard (we still support CUDA 10).

In any case, `num_used` is not a constexpr because it depends on the `W_used` input. I added a `else if (ndim==3)` check in that final clause which seems to work as `ndim` is a template parameter known at compile time. Is this change correct?

hoomd/md/AnisoPotentialPairALJ2.h
```diff
@@ -0,0 +1,11 @@
...
+// Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+#pragma once
+
+#include "AnisoPotentialPair.h"
+#include "EvaluatorPairALJ.h"
+
+#ifdef ENABLE_HIP
+#include "AnisoPotentialPairGPU.h"
+#endif
```
tommy-waltmann: I grepped the codebase and no file includes this header, why is it here?

hoomd/md/AnisoPotentialPairGB.cc
```diff
@@ -0,0 +1,24 @@
...
+namespace hoomd
+    {
+namespace md
+    {
+template class AnisoPotentialPair<EvaluatorPairGB>;
+
+namespace detail
+    {
+template void export_AnisoPotentialPair<EvaluatorPairGB>(pybind11::module& m,
+                                                         const std::string& name);
```
tommy-waltmann, line 14: Is this necessary here when the template instantiation is called immediately below it?

hoomd/md/PotentialExternal.h
```diff
@@ -4,6 +4,7 @@
 #include "hoomd/ForceCompute.h"
 #include "hoomd/GPUArray.h"
 #include "hoomd/GlobalArray.h"
+#include "hoomd/managed_allocator.h"
```
tommy-waltmann, line 7: Why was this line added? The managed allocator class doesn't appear in the diff for this file.

hoomd/md/PotentialExternalGPUKernel.cu.inc
```diff
@@ -0,0 +1,16 @@
...
+// Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+#include "hoomd/md/PotentialExternalGPU.cuh"
+#include "hoomd/md/EvaluatorExternal@_evaluator@.h"
+
+namespace hoomd { namespace md { namespace kernel {
+template
+__attribute__((visibility("default")))
+hipError_t
+gpu_cpef<EvaluatorExternal@_evaluator@>(const external_potential_args_t& external_potential_args,
```
tommy-waltmann: In other places in this PR, we have changed the abbreviated names to more verbose versions. In this case, the changed name would be `gpu_compute_potential_external_forces`.

hoomd/test/test_gpu_array.cc
vyasr: Ah I thought @b-butler was implying that `if constexpr` was not supported by `nvcc` even with C++17. I did not realize that you had upgraded the requirement for only the host compiler. 

Yes, the change that you included is technically equivalent. Unlike with an `if constexpr` you are not guaranteed that the compiler will optimize that branch out, so it's still possible that you'll get errors, but I guess even with -O1 the dead code optimizer is happy to remove branches that are clearly defined at compile time. I assume that you'll still see the warning in debug builds without optimization, but that probably doesn't matter so much.

If you use the `if (ndim == 3)` condition, it would be helpful to add an extra comment there since it may not be obvious why that condition looks different from everything else in the cascade. Something like this might be appropriate (I'm being very verbose and descriptive, feel free to shorten to whatever level of detail that you think is appropriate):
> The maximum number of points used to define a simplex in N dimensions is N+1. Therefore, `num_used` can only be 4 in three dimensions, so this branch is unreachable in 2D code paths. While a simple `else` clause is sufficient for correctness since one of the preceding three branches will always be taken in 2D, this approach is problematic because even compiling `s3d` invocations into 2D code paths leads to compiler warnings. To suppress these warnings, we instead condition this branch on the dimensionality of the system (which is a compile-time constant) so that the dead code optimizer will compile it out and avoid the warnings.

vyasr: Also just FYI my Github handle is @vyasr. You've tagged somebody else in your last message :)

tommy-waltmann (commented):
Next round of comments.

hoomd/test/test_gpu_array.cc
joaander: > Also just FYI my Github handle is @vyasr. You've tagged somebody else in your last message :)

Yes, I know what your handle is, but GitHub's mandatory auto-complete does not.

> If you use the `if (ndim == 3)` condition, it would be helpful to add an extra comment there since it may not be obvious why that condition looks different from everything else in the cascade. Something like this might be appropriate (I'm being very verbose and descriptive, feel free to shorten to whatever level of detail that you think is appropriate):

Even if we could use `constexpr`, it would still need to be `ndim==3` because `num_used` is not known at compile time. I will add the comment.

hoomd/md/AnisoPotentialPairALJ2.h
joaander: Because I forgot to remove it.

hoomd/md/PotentialExternal.h
joaander: Functions from `managed_allocator` are used in the file (`make_managed_shared`), but `managed_allocator.h` was previously included only in files before `PotentialExternal.h` in other files. In the current code, remove this include and the compilation will fail with an error.

hoomd/md/PotentialExternalGPUKernel.cu.inc
joaander: In other places, I didn't change any names - I removed an abbreviated template name and started using the full name. The original code base used `gpu_cpef` as the original full name for this potential for unknown reasons.

hoomd/md/AnisoPotentialPairGB.cc
joaander: No.

hoomd/md/PotentialTersoffGPU.cuh
```diff
@@ -118,7 +118,7 @@ __device__ double myAtomicAdd(double* address, double val)
 
 // workaround for HIP bug
 #ifdef __HIP_PLATFORM_HCC__
-inline __device__ float myAtomicAdd(float* address, float val)
+inline __device__ inline float myAtomicAdd(float* address, float val)
```
tommy-waltmann:
```suggestion
__device__ inline float myAtomicAdd(float* address, float val)
```

hoomd/md/EvaluatorPairALJ.h
```diff
@@ -1619,7 +1625,7 @@ EvaluatorPairALJ<3>::computeSimplexInteractions(const vec3<Scalar>& a,
 // Note: This method assumes that shape_i == shape_j. This should be valid for
 // all cases, and this logic will be moved up to the AnisoPotentialPair in
 // HOOMD 3.0.
-template<> std::string EvaluatorPairALJ<2>::getShapeSpec() const
+template<> inline std::string EvaluatorPairALJ<2>::getShapeSpec() const
```
tommy-waltmann, line 1628: Does explicitly stating inline in a function implemented in a header actually do anything?

hoomd/md/CMakeLists.txt
```diff
@@ -290,26 +270,267 @@ set(_md_cu_sources ActiveForceComputeGPU.cu
...
+
+# generate pybind11 export cc files
+set(_manifolds Cylinder
+               Diamond
+               Ellipsoid
+               Gyroid
+               Plane
+               Primitive
+               Sphere)
+
```
tommy-waltmann:
```suggestion
```
This is a redefinition of the list above.

tommy-waltmann (changes requested):
Just a couple more things. At this point, I have looked through all the files in this PR.

hoomd/md/PotentialTersoffGPU.cuh
joaander: Without the inline, you get multiple defined symbol errors when the header is included in multiple compilation units.

hoomd/md/EvaluatorPairALJ.h
joaander: Without the inline, you get multiple defined symbol errors when the header is included in multiple compilation units.

hoomd/md/PotentialTersoffGPU.cuh
joaander: Sure, we can change the order. The compiler will accept these in either order.

tommy-waltmann: Inline was being declared twice here. My suggestion was to get rid of one of them.

joaander: Sorry, I misread the diff. One inline is enough. Not sure why I added a 2nd except that I was getting a compile error with this at one point while developing this branch.

hoomd/md/EvaluatorPairALJ.h
tommy-waltmann: I didn't know that was possible when the header had an include guard.

joaander: The include guard prevents the function from being defined twice in the same compilation unit. Multiple defined symbol errors occur at the linking step when linking multiple compilation units together.

tommy-waltmann: ok, then disregard this comment.

tommy-waltmann (approved):

hoomd/test/test_gpu_array.cc
vyasr: > Even if we could use constexpr, it would still need to be ndim==3 because num_used is not known at compile time.

I followed the reasoning, but requested the comment to help future readers of the code who may not.

PR #1316, opened 2022-05-05
Clean up workspace and HOME at end of test run.

joaander:
## Description

Delete the workspace and HOME directories at the end of GitHub Actions test runs.

## Motivation and context

The self-update process on the self-hosted GitHub Actions runners fails to restart the runner. The only error I see in the logs is: `Access to the path '/home/azure/actions-runner/_work/_temp/_github_home/.nv' is denied.`. This path is owned by root because GitHub Actions bind mounts `/home/azure/actions-runner/_work/_temp/_github_home/` as the home directory, then runs jobs in the container as root. This process is hard coded in the runner and is not configurable.

This fix cleans up all files that could be made as the docker root user before exiting so that error should not longer occur.

## How has this been tested?

I will check the logs after the CI tests on this pull request run.

## Change log

Internal change, no entry needed.

## Checklist:

### DIFF
--- a/.github/workflows/templates/test.yml
+++ b/.github/workflows/templates/test.yml
@@ -162,12 +162,15 @@ env:
     steps:
     - name: Clean workspace
       run: ( shopt -s dotglob nullglob; rm -rf ./* )
+      shell: bash
 <% endset %>
 <% set post_cleanup_steps %>
     - name: Clean workspace
       run: ( shopt -s dotglob nullglob; rm -rf ./* )
+      shell: bash
     - name: Clean HOME
       run: ( shopt -s dotglob nullglob; rm -rf $HOME/* )
+      shell: bash
 <% endset %>
 <% block jobs %>
 # Use multiple jobs to reduce the amount of time spent on GPU runners. Use CPU runners for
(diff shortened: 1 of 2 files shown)


joaander:
Confirmed that the actions runner is now able to clean up `_github_home` properly. Time will tell if this fixes the update process. I can't trigger that manually.

b-butler (approved):

joaander:
This is a crude solution, but unfortunately I can't identify any alternatives.
If would be nice GitHub Actions would run with the same uid inside the container as outside: https://stackoverflow.com/questions/27925006/share-files-between-host-system-and-docker-container-using-specific-uid. I tried, but we cannot accomplish this with the container `options:` key.

It would be nice if we could use the started/completed hooks to cleanup, but these run in the host system and would not be able to delete files created by the root user inside the docker container: https://docs.github.com/en/actions/hosting-your-own-runners/running-scripts-before-or-after-a-job

For future reference, some github actions, such as `cache` run post conditions that examine the state of the workspace at the end of the job. Deleting all the files in the workspace would break that. We don't currently use `cache` in any of the jobs that need to be cleaned, but will need to keep this in mind if we ever do.

PR #1317, opened 2022-05-06
Add -O3 to jit compiler options for cpu

tcmoore3:
## Description

Adds an optimization flag to the cpu compile options for the user-defined HPMC pair potentials.

## Motivation and context
No optimizations are currently being passed to the JIT compiler for pair potentials, causing poor performance when using the `hpmc.pair.user` classes.

## How has this been tested?
For 1672 hexagons with(out) square-well patches at 80% packing fraction, with mpi and 16 ranks, with docker image  `bridges2` unless otherwise noted, and not including compilation time:
| hoomd version | patches? | mean tps|
| -- | -- | -- | 
| 2.9.6 | no | 2927|
|2.9.6 | yes | 1233|
|3.1.0|no|2300|
|3.1.0|yes|143 (191 on `cheme-hodges`)|
|3.1.0 with "-O3" for patch code compilation | yes | 1169 (on `cheme-hodges`)|
|3.1.0 with "-O3 -funroll-loops" for patch code compilation | yes | 1166 (on `cheme-hodges`)|

## Change log

```
- Add optimization flag to compiler options for user-defined pair potentials in HPMC
```

## Checklist:

### DIFF
--- a/hoomd/_compile.py
+++ b/hoomd/_compile.py
@@ -23,7 +23,7 @@ def _get_hoomd_include_path():
 
 def get_cpu_include_options():
     """Get the source code include path for HOOMD's include files."""
-    return ['-I', str(_get_hoomd_include_path())]
+    return ['-I', str(_get_hoomd_include_path()), '-O3']
 
 
 def get_gpu_compilation_settings(gpu):
--- a/hoomd/_compile.py
+++ b/hoomd/_compile.py
@@ -21,8 +21,11 @@ def _get_hoomd_include_path():
     return hoomd_include_path.resolve()
 
 
-def get_cpu_include_options():
-    """Get the source code include path for HOOMD's include files."""
+def get_cpu_compiler_arguments():
+    """Get the arguments to pass to the compiler.
+
+    These arguments must include the include patch for HOOMD's include files.
+    """
     return ['-I', str(_get_hoomd_include_path()), '-O3']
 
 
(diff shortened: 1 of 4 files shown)


tommy-waltmann (approved):
Thanks for this! It's really crazy how just adding a flag can just make the performance 10x faster

hoomd/_compile.py
```diff
@@ -23,7 +23,7 @@ def _get_hoomd_include_path():
 
 def get_cpu_include_options():
     """Get the source code include path for HOOMD's include files."""
-    return ['-I', str(_get_hoomd_include_path())]
```
joaander, line 26: `-O3` is an optimization option, not an include option. Rename the method to `get_cpu_compiler_arguments` and update the docstring accordingly.

joaander (changes requested):
My apologies. All the information I read indicated that this would not work, so I didn't try it. The documentation and answers on stackoverflow say that optimization is done at the LLVM level, not clang and optimization passes need to be added on the module:

https://releases.llvm.org/13.0.1/docs/tutorial/MyFirstLanguageFrontend/LangImpl04.html#llvm-optimization-passes
https://stackoverflow.com/questions/24088403/compile-using-clang-optimisation-passes-flags

I thought I had left a TODO in the code to check the performance and add optimization passes later, but that seems to have been lost.

Does performance improve at all with `-funroll-loops`? Your code loops over a fixed number of patches, so that might help.

tcmoore3:
> Does performance improve at all with `-funroll-loops`? Your code loops over a fixed number of patches, so that might help.

It does not, at least for this system. I updated the table in the description to show this.

joaander (approved):
Thanks!

PR #1318, opened 2022-05-09
Clarify where constraint_groups are to be set in md.constrain.Distance documentation

joaander:
## Description

In `md.constrain.Distance`, clarify that `constraint_group` members should be set in the system `State`.

## Motivation and context

Some users expressed confusion.

## How has this been tested?

I built the docs locally and inspected the new Note box.

## Change log

```
* Improved documentation.
```

## Checklist:

### DIFF
--- a/hoomd/md/constrain.py
+++ b/hoomd/md/constrain.py
@@ -72,12 +72,8 @@ class Distance(Constraint):
             - d_{ij}^2 = 0
 
     Where :math:`i` and :math:`j` are the the particle tags in the
-    ``constraint_group`` and :math:`d_{ij}` is the constraint distance.
-    Define any number of constraint groups in the system state.
-
-    Note:
-        Define ``constraint_group`` members and :math:`d_{ij}` in the
-        constraints property of `hoomd.State`.
+    ``constraint_group`` and :math:`d_{ij}` is the constraint distance as given
+    by the `system state <hoomd.State>`_.
 
     The method sets the second derivative of the Lagrange multipliers with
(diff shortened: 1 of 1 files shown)


hoomd/md/constrain.py
```diff
@@ -75,6 +75,10 @@ class Distance(Constraint):
     ``constraint_group`` and :math:`d_{ij}` is the constraint distance.
     Define any number of constraint groups in the system state.
 
+    Note:
+        Define ``constraint_group`` members and :math:`d_{ij}` in the
+        constraints property of `hoomd.State`.
```
b-butler: Users cannot directly edit a constraints property. They have to go through the snapshot or local snapshot (presumably the snapshot since you cannot set `N_constraints` in a local snapshot). We should mention that in the note here, as there is no direct constraints property of the `State`, and property is a special word idiomatically in Python.

b-butler (changes requested):
This is a useful change, but I just have one comment related to potential confusion for Python users.

hoomd/md/constrain.py
joaander: I do not want to mention snapthot or gsd files here. That is a basic concept that users should be familiar with. If we mention it here, then for consistency, we would also need to mention it in every single place that positions, orientations, velocities, masses, charges, bonds, dihedrals, and all other properties of the state are used.

I will reword this so it doesn't trigger an idomatic Python interpretation.

b-butler (approved):

joaander:
This is a very small change, merging with one review.

PR #1321, opened 2022-05-09
Find clang correctly in relocated conda-forge builds

joaander:
## Description

Store the LLVM prefix path by itself in the data segment and append `/bin/clang` at runtime.

## Motivation and context

We would like to build the conda-forge binaries with `ENABLE_LLVM=on`.

Even though HOOMD uses `libclang` to perform the compilation, a `clang` executable is still needed to discover the correct header search path. To ensure that we use the correct `clang` executable, HOOMD embeds the path to LLVM found at compile time. In conda binary installations, the user install path is different from the install path at build time, so conda performs a search and replace to update these hard coded paths: https://docs.conda.io/projects/conda-build/en/latest/resources/make-relocatable.html

However, we found that conda replaced `BUILD_PREFIX/bin/clang` with `INSTALL_PREFIX`, truncating the `/bin/clang` (likely because it inserts a NULL character after the end of `INSTALL_PREFIX`). This leads to build errors when using HPMC patch potentials in conda environments with hoomd installed.

## How has this been tested?

This `hoomd-feedstock` pull request: https://github.com/conda-forge/hoomd-feedstock/pull/65

## Change log

```
*Added*

- Set `ENABLE_LLVM=on` in conda binary builds.
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/ClangCompiler.cc
+++ b/hoomd/hpmc/ClangCompiler.cc
@@ -126,7 +126,11 @@ std::unique_ptr<llvm::Module> ClangCompiler::compileCode(const std::string& code
     std::cout << "##  ###  ###     ###     ##     ##        ##  ##" << std::endl;
     std::cout << "################################################" << std::endl;
     std::cout << "resource_path = " << resource_path << std::endl;
-    std::cout << "clang_args = " << clang_arg_c_strings << std::endl;
+    std::cout << "clang_args = " << std::endl;
+    for (int i = 0; i < clang_args.size(); i++)
+        {
+        std::cout << clang_args[i] << std::endl;
+        }
     clang::driver::Driver driver(resource_path + "/bin/clang",
                                  llvm::sys::getDefaultTargetTriple(),
                                  diagnostics_engine);
--- a/hoomd/hpmc/ClangCompiler.cc
+++ b/hoomd/hpmc/ClangCompiler.cc
@@ -126,7 +126,7 @@ std::unique_ptr<llvm::Module> ClangCompiler::compileCode(const std::string& code
     std::cout << "##  ###  ###     ###     ##     ##        ##  ##" << std::endl;
     std::cout << "################################################" << std::endl;
     std::cout << "resource_path = " << resource_path << std::endl;
-    std::cout << "clang_args = " << clang_args_c_strings << std::endl;
+    std::cout << "clang_args = " << clang_arg_c_strings << std::endl;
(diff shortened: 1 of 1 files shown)


hoomd/hpmc/ClangCompiler.cc
```diff
@@ -91,8 +92,18 @@ std::unique_ptr<llvm::Module> ClangCompiler::compileCode(const std::string& code
...
+    // at runtime instead of compile time. When it is added at compile time, conda rewrites an
+    // embedded string "$BUILD_PREFIX/bin/clang" with "$INSTALL_PREFIX", truncating the string.
+    volatile const char* llvm_install_prefix = HOOMD_LLVM_INSTALL_PREFIX;
+
+    // standard string methods do not accept volatile char arrays, extract the string manually
+    std::string clang_exec_with_path;
+    for (volatile const char* c = llvm_install_prefix; *c != 0; ++c)
+        {
+        clang_exec_with_path.push_back(*c);
+        }
```
tommy-waltmann, line 105: Is the volatile declaration to prevent any compile-time removal of this code? Do we know the volatile declaration is needed here (and in the above variable) to prevent errors?

tommy-waltmann (commented):
I just want to bring up one point for discussion.

hoomd/hpmc/ClangCompiler.cc
joaander: `volatile` is to provide absolute certainty that the compiler can't remove the variable at compile time. It was smart enough to convert `std::string(LLVM_INSTALL_PREFIX) + "/bin/clang"` into its own constant at compile time. It may or may not be smart enough to convert this for loop - but the `volatile` forces it to consider the array as run time variable: https://en.cppreference.com/w/cpp/language/cv

> Every access (read or write operation, member function call, etc.) made through a glvalue expression of volatile-qualified type is treated as a visible side-effect for the [purposes of optimization](https://en.cppreference.com/w/cpp/language/as_if) (that is, within a single thread of execution, **volatile accesses cannot be optimized out** ...

tommy-waltmann (approved):

joaander:
Any comments, @SchoeniPhlippsn?

SchoeniPhlippsn (approved):

tcmoore3:
Needed for https://github.com/conda-forge/hoomd-feedstock/pull/65

tcmoore3:
Needed for conda-forge/hoomd-feedstock#65

PR #1339, opened 2022-05-26
ALJ Documentation Fixes

tommy-waltmann:
## Description

Changes the ALJ documentation so `sigma_i` and `sigma_j` use the word diameter instead of radius. Also clarify what the i and j refer to.

## Motivation and context

@cbkerr Noted that radius is incorrect, and diameter is the correct word.

## How has this been tested?

I have built the documentation locally and the CI will do the same.

## Change log

```
Corrections to the ALJ documentation.
```

## Checklist:

- [ ] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/CONTRIBUTING.md).

### DIFF
--- a/hoomd/md/pair/aniso.py
+++ b/hoomd/md/pair/aniso.py
@@ -256,26 +256,45 @@ class ALJ(AnisotropicPair):
     .. _Ramasubramani, V.  et. al. 2020: https://doi.org/10.1063/5.0019735
 
     .. math::
-        U(r, r_c) &= U_0(r) &+ U_c(r_c)
 
-            &= 4 \varepsilon \left[ \left( \frac{\sigma}{r} \right)^{12} -
-            \left( \frac{\sigma}{r} \right)^{6} \right] &+
-            4 \varepsilon_c \left[ \left( \frac{\sigma_c}{r_c} \right)^{12} -
-            \left( \frac{\sigma_c}{r_c} \right)^{6} \right]
+        U(r, r_c) = U_0(r) + U_c(r_c)
 
-    where :math:`\varepsilon` is BLANK, :math:`\varepsilon_s` BLANK,
-    :math:`sigma` is BLANK, :math:`sigma_c` is BLANK.
-    :math:`r` is the center-center distance, :math:`r_c` is the contact distance.
+    The first term is the central interaction :math:`U_0`, the standard
+    center-center interaction between two Lennard-Jones particles. The second
+    term is the contact interaction :math:`U_c`, computed from the smallest
+    distance between the surfaces of the two shapes, :math:`r_c`. The central
+    and contact interactions are defind as follows:
 
-    The first term is the "central" interaction :math:`U_0`, the standard center-center interaction between two
-    Lennard-Jones particles. The second term is the contact interaction :math:`U_c` computed
-    from the smallest distance between the surfaces of the two shapes,
-    :math:`r_c`.
+    .. math::
 
-    The total potential energy is therefore the sum of
-    two interactions, a central Lennard-Jones potential and a radially-shifted
-    Lennard-Jones potential where the shift is anisotropic and depends on the
-    extent of the shape in each direction.
+        &U_0(r) = 4 \varepsilon \left[ \left( \frac{\sigma}{r} \right)^{12} -
+        \left( \frac{\sigma}{r} \right)^{6} \right]
+
+        &U_c(r_c) = 4 \varepsilon_c \left[ \left( \frac{\sigma_c}{r_c}
+        \right)^{12} - \left( \frac{\sigma_c}{r_c} \right)^{6} \right]
+
+    where :math:`r` is the center-center distance, :math:`r_c` is the contact
+    distance, :math:`\varepsilon` (``epsilon``) is the strength of both the
+    central and contact interactions, :math:`\varepsilon_c` is an energy
+    coefficient proportional to :math:`\varepsilon`, :math:`\sigma` is the
+    interaction distance of the central term computed as the average of
+    :math:`\sigma_i` (``sigma_i``) and :math:`\sigma_j` (``sigma_j``). Lastly,
+    we use the contact ratios :math:`\beta_i` (``contact_ratio_i``) and
+    :math:`\beta_j` (``contact_ratio_j``) to compute :math:`\sigma_c` as
+    follows:
+
+    .. math::
+
+        \sigma_c &= \frac{1}{2} \left[\sigma_{ci} + \sigma_{cj} \right]
+
+        \sigma_{ci} &= \beta_i \cdot \sigma_i
+
+        \sigma_{cj} &= \beta_j \cdot \sigma_j
+
+    The total potential energy is therefore the sum of two interactions, a
+    central Lennard-Jones potential and a radially-shifted Lennard-Jones
+    potential where the shift is anisotropic and depends on the extent of the
+    shape in each direction.
 
     Like a standard LJ potential, each term has an independent cutoff beyond
     which it decays to zero. The behavior of these cutoffs is dependent on
--- a/hoomd/md/pair/aniso.py
+++ b/hoomd/md/pair/aniso.py
@@ -270,17 +270,18 @@ class ALJ(AnisotropicPair):
         &U_0(r) = 4 \varepsilon \left[ \left( \frac{\sigma}{r} \right)^{12} -
         \left( \frac{\sigma}{r} \right)^{6} \right]
 
-        &U_c(r_c) = 4 \varepsilon_c(\varepsilon) \left[ \left( \frac{\sigma_c}{r_c}
-        \right)^{12} - \left( \frac{\sigma_c}{r_c} \right)^{6} \right]
+        &U_c(r_c) = 4 \varepsilon_c(\varepsilon) \left[ \left(
+        \frac{\sigma_c}{r_c} \right)^{12} - \left( \frac{\sigma_c}{r_c}
+        \right)^{6} \right]
 
-    where :math:`\varepsilon` (`epsilon <params>`) affects strength of both the central
-    and contact interactions, :math:`\varepsilon_c` is an energy coefficient set
-    proportional to :math:`\varepsilon` to preserve the shape, :math:`\sigma` is
-    the interaction distance of the central term computed as the average of
-    :math:`\sigma_i` (`sigma_i <params>`) and :math:`\sigma_j` (`sigma_j <params>`). Lastly,
-    `ALJ` uses the contact ratios :math:`\beta_i` (`contact_ratio_i <params>`) and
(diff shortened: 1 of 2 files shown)


hoomd/md/pair/aniso.py
```diff
@@ -400,6 +400,10 @@ class ALJ(AnisotropicPair):
         Type: `hoomd.data.typeparam.TypeParameter` [`tuple` [``particle_types``,
         ``particle_types``], `dict`]
 
+    Note:
+        Parameters which end with i and j refer to the first and second particle
+        type in the parameter tuple, respectively.
```
joaander: I don't understand this. HOOMD sets the pair parameters symmetrically. From the user perspective:
```
alj.params[('A', 'B')] = dict(epsilon=1.1, sigma_i=2, sigma_j=3, alpha=0)
assert alj.params[('B', 'A')] == dict(epsilon=1.1, sigma_i=2, sigma_j=3, alpha=0)
```
This documentation would indicate that `sigma_i` applies to 'A' and `sigma_j` applies to 'B' AND contradicts itself by stating that `sigma_i` applies to 'B' and `sigma_j` applies to 'A' in the 2nd line.

joaander (changes requested):

hoomd/md/pair/aniso.py
joaander: Code that demonstrates the symmetric nature of the parameters:
https://github.com/glotzerlab/hoomd-blue/blob/5f9041a58a62457e5b68e87bce9a8f51896824b3/hoomd/md/AnisoPotentialPair.h#L347-L354

tommy-waltmann: This is true and my original addition to the documentation could have been better worded. The implementation of the ALJ potential is symmetric under the exchange of the labels `i` and `j`. In fact, its also symmetric under the exchange of the contact ratio labels. In other words
``` python
alj.params[('A', 'B')] = dict(epsilon=1.0, sigma_i=1.0, sigma_j=2.0, contact_ratio_i=0.2, contact_ratio_j=0.1, alpha=0)
```
Evaluates the same potential as 
``` python
alj.params[('A', 'B')] = dict(epsilon=1.0, sigma_i=1.0, sigma_j=2.0, contact_ratio_i=0.1, contact_ratio_j=0.2, alpha=0)
```
This is because internally the average of the two parameters is always used instead of the individual values. Internally it computes something like
```
sigma_ij = Scalar(0.5) * (_params.sigma_i + _params.sigma_j)
contact_sigma_ij = Scalar(0.5) * (_params.contact_sigma_i + _params.contact_sigma_j)
```
and then just works with those values instead of the original individual parameters.

When a user sets the parameters, however, they will naturally want to associate the `i` parameters with one particle type and the `j` parameters with another particle type. In fact, they need to do this in order to get accurate simulation results. As an example,
``` python
# one of these sigma/contact_sigma values physically represents the A type
alj.params[('A', 'B')] = dict(epsilon=1.0, sigma_i=1.0, sigma_j=2.0, contact_ratio_i=0.1, contact_ratio_j=0.2, alpha=0)

# the same sigma/contact_sigma values that represent the A type above must also show up in this dictionary
alj.params[('A', 'C')] = dict(...)
```

The best way to document the behavior is to make both points clear: the actual potential is symmetric w.r.t. labels `i` and `j`, but when setting parameters for all typepairs, the same `sigma`/`contact_sigma` values must be shared across different typepairs that share one particle type in order to get proper simulation results.

joaander: Yes, it would be great to document exactly how the values in the given equation are computed from the user input parameters.

tommy-waltmann:
I've pushed a re-wording of the documentation note, @cbkerr does the new documentation answer your questions about how to set pair parameters?

cbkerr:
I'll have other improvements to make to the documentation once I understand the source code.
- Need to add a note about the tag of particles `i` and `j`and how that affects how we get results from the GJK algorithm.
- I'd like to see a suggested heuristic for how to decide what `dt` to use.
- In using ALJ, I started by reading the docs. I didn't understand it enough, so I read the paper, but different notation and symbols are used, so it's still pretty confusing.

Things I don't understand after reading the documentation and the ALJ paper include:
- How does `contact_ratio` relate to the terms in the paper like $\sigma_r$?
- Why does `rounding_radii` affect the `type_shape` while `contact_ratio` does not? However, both affect step size needed and the amount of rounding.
  - How is the `type_shape` determined?
- Why can't the user set both $\epsilon$ and $\epsilon_s$ values defined in the paper? See p.4, top of left column. It's not clear which $\epsilon$ the `epsilon` parameter refers to.

tommy-waltmann:
I'll respond to each of these points individually.

tommy-waltmann:
>     * Need to add a note about the tag of particles `i` and `j`and how that affects how we get results from the GJK algorithm.

Can you explain this further? The particle tags are used to get the particle information, which is plugged into the equations defining the potential. This is similar to every other pair potential.

tommy-waltmann:
>     * I'd like to see a suggested heuristic for how to decide what `dt` to use.

We typically do not do this for pair potentials. Use the same method you use to get the `dt` for any other simulation.

tommy-waltmann:
>     * In using ALJ, I started by reading the docs. I didn't understand it enough, so I read the paper, but different notation and symbols are used, so it's still pretty confusing.

The equations we use in the documentation should be written in more depth, there are many variables in our equations which are either not defined or not explained well enough right now. However, we cannot change the names of the potential parameters at this point.

tommy-waltmann:
>     * How does `contact_ratio` relate to the terms in the paper like σr?

Adding more explicit equations will help with understanding here.

tommy-waltmann:
>     * Why does `rounding_radii` affect the `type_shape` while `contact_ratio` does not? However, both affect step size needed and the amount of rounding.

Both `rounding_radii` and `contact_ratio` should affect the `type_shapes`. Please show me an example of this behavior offline.

tommy-waltmann:
>     * Why can't the user set both ϵ and ϵs values defined in the paper? See p.4, top of left column. It's not clear which ϵ the `epsilon` parameter refers to.

I see that the paper's epsilon and our epsilon are different values. The epsilon pair parameter is the one the is documented as `epsilon` in the hoomd documentation. The other epsilon (epsilon_c in the hoomd documentation) is computed from the `epsilon` pair parameter. I'll add an equation for this to the documentation, but I should note that this equation is already in the paper.

cbkerr:
> I'll add an equation for this to the documentation, but I should note that this equation is already in the paper.

Ah! I wasn't sure if those formulas eq (6) in paper were implemented or if we had to be careful to implement them.

joaander (changes requested):
There are unfinished changes that need to be completed. For example, I assume we will replace `BLANK` with something meaningful. I have no other comments at this time.

hoomd/md/pair/aniso.py
```diff
@@ -256,21 +256,48 @@ class ALJ(AnisotropicPair):
...
+        &U_c(r_c) = 4 \varepsilon_c \left[ \left( \frac{\sigma_c}{r_c}
+        \right)^{12} - \left( \frac{\sigma_c}{r_c} \right)^{6} \right]
+
+    where :math:`r` is the center-center distance, :math:`r_c` is the contact
+    distance, :math:`\varepsilon` (``epsilon``) is the strength of both the
+    central and contact interactions, :math:`\varepsilon_c` is an energy
+    coefficient proportional to :math:`\varepsilon`, :math:`\sigma` is the
+    interaction distance of the central term computed as the average of
+    :math:`\sigma_i` (``sigma_i``) and :math:`\sigma_j` (``sigma_j``). Lastly,
+    we use the contact ratios :math:`\beta_i` (``contact_ratio_i``) and
```
joaander:
```suggestion
    `ALJ` uses the contact ratios :math:`\beta_i` (``contact_ratio_i``) and
```

hoomd/md/pair/aniso.py
```diff
@@ -256,21 +256,48 @@ class ALJ(AnisotropicPair):
...
-            \left( \frac{\sigma_c}{r_c} \right)^{6} \right]
-
-    The first term is the standard center-center interaction between two
-    Lennard-Jones spheres. The second term is a contact interaction computed
-    based on the smallest distance between the surfaces of the two shapes,
-    :math:`r_c`. The total potential energy can thus be viewed as the sum of
-    two interactions, a central Lennard-Jones potential and a shifted
-    Lennard-Jones potential where the shift is anisotropic and depends on the
-    extent of the shape in each direction.
+
```
joaander, line 258: This comment applies a few lines above:

```
    `ALJ` computes the Lennard-Jones force between anisotropic particles as
    described in `Ramasubramani, V.  et. al. 2020`_, using the formula:
```
Remove the "we" as the documentation is not written from the point of view of the authors.

joaander (changes requested):
Thanks, this is a great improvement. I have a a few small change requests.

What is the resolution to Corwin's issues with type_shapes and the rounding?

joaander (approved):
Thanks!

hoomd/md/pair/aniso.py
```diff
@@ -250,27 +250,53 @@ class ALJ(AnisotropicPair):
           .. deprecated:: v3.1.0
 
     `ALJ` computes the Lennard-Jones force between anisotropic particles as
-    described in `Ramasubramani, V.  et. al. 2020`_. Specifically we implement
-    the formula:
+    described in `Ramasubramani, V.  et. al. 2020`_, using the formula:
```
lyrivera: I believe it should be `et al.` (only one period). Same for other appearances.

lyrivera (approved):
Left one small comment. Otherwise, LGTM.

hoomd/md/pair/aniso.py
```diff
@@ -250,27 +250,53 @@ class ALJ(AnisotropicPair):
...
-    Lennard-Jones potential where the shift is anisotropic and depends on the
-    extent of the shape in each direction.
+
+        U(r, r_c) = U_0(r) + U_c(r_c)
+
+    The first term is the central interaction :math:`U_0`, the standard
+    center-center interaction between two Lennard-Jones particles. The second
+    term is the contact interaction :math:`U_c`, computed from the smallest
+    distance between the surfaces of the two shapes, :math:`r_c`. The central
+    and contact interactions are defind as follows:
```
cbkerr:
```suggestion
    and contact interactions are defined as follows:
```

cbkerr (approved):
Huge improvement, thanks for helping address my confusions!


>What is the resolution to Corwin's issues with type_shapes and the rounding?

I did some tests and I think it's working correctly. We may open another PR to address how shapes are output later but that's a bigger change out of the scope of this one

vyasr:
It sounds like this PR is handled, and I haven't been following, but let me know if any part of these docs need clarification from me.

PR #1340, opened 2022-06-01
Create *_like typing objects

### LINKED ISSUE #1319: Inconsistant use of `hoomd.filter` and `hoomd.filter.ParticleFilter`
## Description
We document the type as `hoomd.filter` for some constructors that take a filter and `hoomd.filter.ParticleFilter` for others. Since `hoomd.filter.Custom` is not a `ParticleFilter` we document the type as `hoomd.filter` in some places. Regardless of preference we should be consistent in this.

## Fix
Choose a style and stick to it.

b-butler:
## Description

Creates the 
- `variant_like`
- `trigger_like`
- `filter_like`
objects, and documents them. In addition, I modified existing documentation to use these. I attempted to use `grep` to get all instances of the previous documentation.

## Motivation and context

Resolves #1319

## How has this been tested?

Documentation builds.

## Change log

```
Added: New ``variant_like``, ``trigger_like``, and ``filter_like`` typing objects for documentation.
```

## Checklist:

### DIFF
--- a/hoomd/box.py
+++ b/hoomd/box.py
@@ -3,6 +3,9 @@
 
 """Implement Box."""
 
+import abc
+import typing
+
 import numpy as np
 from functools import partial
 import hoomd._hoomd as _hoomd
@@ -259,25 +262,16 @@ class Box:
         r"""Initialize a Box instance from a box-like object.
 
         Args:
-            box:
-                A box-like object
+            box (box_like): A box-like object.
 
         Note:
-           Objects that can be converted to HOOMD-blue boxes include lists like
-           ``[Lx, Ly, Lz, xy, xz, yz]``, dictionaries with keys ``'Lx',
-           'Ly', 'Lz', 'xy', 'xz', 'yz'``, objects with attributes ``Lx, Ly,
-           Lz, xy, xz, yz``, 3x3 matrices (see `from_matrix`), or existing
-           `hoomd.Box` objects.
-
-           If any of ``Lz, xy, xz, yz`` are not provided, they will be set to 0.
-
            If all values are provided, a triclinic box will be constructed.
            If only ``Lx, Ly, Lz`` are provided, an orthorhombic box will
            be constructed. If only ``Lx, Ly`` are provided, a rectangular
            (2D) box will be constructed.
 
         Returns:
-            :class:`hoomd.Box`: The resulting box object.
+            hoomd.Box: The resulting box object.
         """
         if np.asarray(box).shape == (3, 3):
             # Handles 3x3 matrices
@@ -511,3 +505,67 @@ class Box:
     def __reduce__(self):
         """Reduce values to picklable format."""
         return (type(self), (*self.L, *self.tilts))
+
+
+class BoxInterface(abc.ABC):
+    """The class interface which HOOMD considers to be a box-like object.
+
+    Note:
+        This class is exclusively used for help with typing and documentation in
+        HOOMD, and is not meant to be used.
+    """
+
+    @property
+    @abc.abstractmethod
+    def Lx(self) -> float:  # noqa: N802: Allow function name
+        """Length in x direction."""
+        pass
+
+    @property
+    @abc.abstractmethod
+    def Ly(self) -> float:  # noqa: N802: Allow function name
+        """Length in y direction."""
+        pass
+
+    @property
+    @abc.abstractmethod
+    def Lz(self) -> float:  # noqa: N802: Allow function name
+        """Length in z direction."""
+        pass
+
+    @property
+    @abc.abstractmethod
+    def xy(self) -> float:
+        """Tilt factor in the xy plane."""
+        pass
+
+    @property
+    @abc.abstractmethod
+    def xz(self) -> float:
+        """Tilt factor in the xy plane."""
+        pass
+
+    @property
+    @abc.abstractmethod
+    def yz(self) -> float:
+        """Tilt factor in the xy plane."""
+        pass
+
+
+box_like = typing.Union[Box, BoxInterface, typing.Sequence[float],
+                        typing.Mapping[str, float], np.ndarray]
+"""Objects that are or can be converted to `Box`.
+
+This includes
+
+* `hoomd.Box` objects
+* objects with attributes ``Lx, Ly, Lz xy, xz, yz``
+* lists like ``[Lx, Ly, Lz, xy, xz, yz]``
+* dictionaries with keys ``'Lx', 'Ly', 'Lz', 'xy', 'xz', 'yz'``
+* 3x3 NumPy arrays or object convertible to a 3x3 array (see
+  `hoomd.Box.from_matrix`)
+
+Note:
+    If any of ``Lz, xy, xz, yz`` for these different types are not provided,
+    they are considered 0.
+"""
--- a/hoomd/box.py
+++ b/hoomd/box.py
@@ -558,12 +558,12 @@ box_like = typing.Union[Box, BoxInterface, typing.Sequence[float],
 
 This includes
 
-* `hoomd.Box` objects
-* objects with attributes ``Lx, Ly, Lz xy, xz, yz``
-* lists like ``[Lx, Ly, Lz, xy, xz, yz]``
-* dictionaries with keys ``'Lx', 'Ly', 'Lz', 'xy', 'xz', 'yz'``
-* 3x3 NumPy arrays or object convertible to a 3x3 array (see
-  `hoomd.Box.from_matrix`)
+* `hoomd.Box` objects.
+* Objects with attributes ``Lx, Ly, Lz, xy, xz, yz``.
+* Lists like ``[Lx, Ly, Lz, xy, xz, yz]``.
+* Dictionaries with keys ``'Lx', 'Ly', 'Lz', 'xy', 'xz', 'yz'``.
+* 3x3 NumPy arrays or objects convertible to a 3x3 array (see
+  `hoomd.Box.from_matrix`).
 
 Note:
     If any of ``Lz, xy, xz, yz`` for these different types are not provided,
--- a/hoomd/data/typeconverter.py
+++ b/hoomd/data/typeconverter.py
@@ -45,8 +45,7 @@ def variant_preprocessing(variant):
         try:
             return Constant(float(variant))
         except Exception:
-            raise ValueError(
-                "Expected a hoomd.variant.Variant or float object.")
+            raise ValueError("Expected a hoomd.variant.variant_like object.")
 
 
 def box_preprocessing(box):
--- a/hoomd/data/typeconverter.py
+++ b/hoomd/data/typeconverter.py
@@ -31,7 +31,7 @@ def trigger_preprocessing(trigger):
         try:
             return Periodic(period=int(trigger), phase=0)
         except Exception:
-            raise ValueError("Expected a hoomd.trigger.Trigger or int object.")
+            raise ValueError("Expected a hoomd.trigger.trigger_like object.")
 
 
 def variant_preprocessing(variant):
--- a/hoomd/md/compute.py
+++ b/hoomd/md/compute.py
@@ -292,8 +292,8 @@ class HarmonicAveragedThermodynamicQuantities(Compute):
     """Compute harmonic averaged thermodynamic properties of particles.
 
     Args:
-        filter (`hoomd.filter`): Particle filter to compute thermodynamic
-            properties for.
+        filter (hoomd.filter.filter_like): Particle filter to compute
+            thermodynamic properties for.
         kT (float): Temperature of the system :math:`[\\mathrm{energy}]`.
         harmonic_pressure (float): Harmonic contribution to the pressure
             :math:`[\\mathrm{pressure}]`. If omitted, the HMA pressure can
@@ -323,10 +323,10 @@ class HarmonicAveragedThermodynamicQuantities(Compute):
 
 
     Attributes:
-        filter (hoomd.filter.ParticleFilter): Subset of particles compute
+        filter (hoomd.filter.filter_like): Subset of particles compute
             thermodynamic properties for.
 
-        kT (hoomd.variant.Variant): Temperature of the system
+        kT (float): Temperature of the system
             :math:`[\\mathrm{energy}]`.
 
         harmonic_pressure (float): Harmonic contribution to the pressure
--- a/hoomd/md/force.py
+++ b/hoomd/md/force.py
@@ -427,7 +427,7 @@ class Active(Force):
         Args:
             trigger (hoomd.trigger.Trigger): Select the timesteps to update
                 rotational diffusion.
-            rotational_diffusion (hoomd.variant.Variant or float): The
+            rotational_diffusion (hoomd.variant.variant_like): The
                 rotational diffusion as a function of time or a constant.
 
         Returns:
--- a/hoomd/md/force.py
+++ b/hoomd/md/force.py
@@ -425,7 +425,7 @@ class Active(Force):
         """Create a rotational diffusion updater for this active force.
 
         Args:
-            trigger (hoomd.trigger.Trigger): Select the timesteps to update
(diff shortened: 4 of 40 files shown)


j-proc:
Love the work and changes. It seems the readthedocs build failed. I'll try to build the docs locally just to check over them, but even if you missed a few places, this is a wonderful change! I particularly loved all those "float or Variant"s just being variant_like. Will definitely make documentation and integration with many modern tools even easier. Python's getting typing-happy and I'm here for it.

j-proc (approved):
Built the docs to check since there was an error in the automated one. I only got the standard warnings about links to the tutorials. So it all looks good to me.

I think you did a great job finding most if not all of the places these changes apply. I think it's better to get the change into the trunk so that new documentation will use it. Any missed places will be easy to address when found.

joaander:
The doc build failure comes from sphinx 5.0: `WARNING: Invalid configuration value found: 'language = None'`. We have fixed this in `trunk-patch`, but I had not merged it up into `trunk-minor` before this PR was made. Will update this branch now.

hoomd/data/typeconverter.py
```diff
@@ -45,8 +45,7 @@ def variant_preprocessing(variant):
         try:
             return Constant(float(variant))
         except Exception:
-            raise ValueError(
-                "Expected a hoomd.variant.Variant or float object.")
+            raise ValueError("Expected a hoomd.variant.variant_like object.")
```
tommy-waltmann, line 48: About ten lines above this one, there is another value error whose error message should be changed to be like this one.

hoomd/md/compute.py
```diff
@@ -329,10 +329,10 @@ class HarmonicAveragedThermodynamicQuantities(Compute):
 
 
     Attributes:
-        filter (hoomd.filter.ParticleFilter): Subset of particles compute
+        filter (hoomd.filter.filter_like): Subset of particles compute
             thermodynamic properties for.
 
-        kT (hoomd.variant.Variant): Temperature of the system
+        kT (float): Temperature of the system
```
tommy-waltmann, line 335: Will this accept a variant or not? The change in documentation seems to indicate that it will no longer accept a temperature variant, which does not seem right to me.

hoomd/md/force.py
```diff
@@ -427,7 +427,7 @@ def create_diffusion_updater(self, trigger, rotational_diffusion):
         Args:
             trigger (hoomd.trigger.Trigger): Select the timesteps to update
                 rotational diffusion.
```
tommy-waltmann: Should this be converted to `trigger-like` also?

tommy-waltmann (changes requested):
After having looked some of the changes over, I have some comments, but I also have three general points for discussion:

1. Was the ability to give `*_like` objects always present in hoomd and just not documented?

2. There seem to be a lot of inconsistencies between arguments which take a `*_like` object, and what type is returned by the associated attribute. For example, setting `kT` in the NVT integrator takes a `variant_like` object as the input argument. When I access the `kT` attribute of `NVT`, it is documentation to return a `Variant`, not a `variant_like` object. Also in `NVT`, the trigger argument is documented to take `trigger_like` object, but the associated attribute is documented to return a `trigger_like` object. Which is the correct behavior? Should attributes return the `*_like` object, or should they return the `Variant`/`Trigger` object?

3. I think it would be good to document `*_like` objects like 
```
`variant_like`
```
instead of as
```
variant_like
```
, so sphinx will create a link directly to our explanation of what the `*_like` objects mean.

hoomd/md/compute.py
b-butler: It never did.

hoomd/md/force.py
b-butler: Yes it should!

b-butler:
@tommy-waltmann 
> 1. Was the ability to give `*_like` objects always present in hoomd and just not documented?

Yes it was, and I think it was documented to some extent, just not consistently.

> 2. There seem to be a lot of inconsistencies between arguments which take a `*_like` object, and what type is returned by the associated attribute. For example, setting `kT` in the NVT integrator takes a `variant_like` object as the input argument. When I access the `kT` attribute of `NVT`, it is documentation to return a `Variant`, not a `variant_like` object. Also in `NVT`, the trigger argument is documented to take `trigger_like` object, but the associated attribute is documented to return a `trigger_like` object. Which is the correct behavior? Should attributes return the `*_like` object, or should they return the `Variant`/`Trigger` object?

In general the attribute is a `Trigger` or `Variant` so we should use those not the `*_like` typing object. For filters given we don't convert them, it should be `filter_like` in both places.

> 3. I think it would be good to document *_like objects like 
>`variant_like`
> instead of as
> `variant_like`

This should already happen without need for `

tommy-waltmann:
I have grepped the codebase, and I think there are a few more places we should change to `*_like` objects

for `trigger_like`:
- `hpmc.nec.tune`, there are many places in this file.
- custom operations

for `variant_like`:
- `methods.NPT`'s `_preprocess_stress` method value error message. Same for `methods.NPH`.

tommy-waltmann:
Last thing to talk about is tests. I wouldn't expect any new tests to be written, but modify some of the current ones so the int/floats are passed to triggers/variants, etc. instead of always using the Trigger or Variant objects.

joaander:
> > 2. There seem to be a lot of inconsistencies between arguments which take a `*_like` object, and what type is returned by the associated attribute. For example, setting `kT` in the NVT integrator takes a `variant_like` object as the input argument. When I access the `kT` attribute of `NVT`, it is documentation to return a `Variant`, not a `variant_like` object. Also in `NVT`, the trigger argument is documented to take `trigger_like` object, but the associated attribute is documented to return a `trigger_like` object. Which is the correct behavior? Should attributes return the `*_like` object, or should they return the `Variant`/`Trigger` object?
> 
> In general the attribute is a `Trigger` or `Variant` so we should use those not the `*_like` typing object. For filters given we don't convert them, it should be `filter_like` in both places.

Yes, this is key. The `*_like` is used only on initialization or setting. For example, if the user gives an `int` it is converted to a trigger so that the attribute returns the trigger:
```
>>> object.trigger = 1
>>> print(object.trigger)
hoomd.trigger.Periodic(1)
```
We document the type that the attribute returns from the getter, not what it accepts in the setter. Instead, we document it in the class constructor where a parameter sets the initial value of the attribute.

vyasr:
Honestly my first reaction at seeing this was a little confused because this is essentially just codifying ducktyping behavior that everyone expects from Python. On second thought, I sort of like having concrete representations of these "types". However, I worry that it implies something stronger than how users might expect ducktyping to behave in Python. Imposing strong typing in this way may be necessary in some cases because we do end up working in C++, not Python, but perhaps not in all of them. A more Pythonic approach (although perhaps not applicable here) would be to instead define all the `*_like` aliases as `abc.ABC` types with a specific API, rather than using a `typing.Union` to proscribe a specific set of actual types.

Consider `variant-like`, for instance. Right now, it's documented as a `Union[Variant, float]`, but wouldn't an `int` work too because it would be transparently cast to a float by the [VariantConstant constructor](https://github.com/glotzerlab/hoomd-blue/blob/26947c6ed9bd29e27c31e824b8e63ca7912c53c7/hoomd/Variant.h#L62)? Moreover, while I don't know exactly what the limitations of the current API are (Pybind exported classes will require stricter types than you'd normally expect in Python and I doubt that you've wrapped all the cases), wouldn't you really expect any object that implements `__float__` to work since it's convertible to a float (although I don't think that will actually work right now)? I don't think HOOMD has to support all the possible cases, and the above is just an example. I do think that Python ducktyping is built around _not_ specifying types so explicitly, and at least in some of the user-facing classes using the types in the way that they're defined now may be more restrictive than you really intend.

(Just some food for thought, I don't intend to block this PR for any reason on these considerations.)

b-butler:
@vyasr these are good points. I don't think we can just define an interface and use that (e.g. `abc.ABC`) as this would be overly restrictive. For instance no `Variant` is convertible to a float, `__float__`, so by that definition we couldn't accept a `Variant` according to the ABC class. On the other hand, if we codify the `Variant` interface, then we cannot use `float` object's anymore since they don't use that interface. We could do something like, `typing.Union[hoomd.variant.Variant, numbers.Real]` though (not sure this is what you are suggesting) which is more broad if perhaps a bit confusing to new users who are not familiar with Python's ABCs.

Also, when you see a base type in a typing object definition, do you expect it to be strict? I typically think that implicit conversions are allowed. This seems to be what mypy (https://mypy.readthedocs.io/en/latest/duck_type_compatibility.html) does.

Edit: I should also mention that `filter_like` is strict.

vyasr:
I'm more suggesting that `variant_like` should be a common term used to define "anything that can be converted to a variant" rather than a strict list of types. This is in line with e.g. how numpy docs use "array-like". freud's [defines a system](https://freud.readthedocs.io/en/latest/modules/locality.html#freud.locality.AABBQuery.from_system) functionally in one place, then uses "system" everywhere else. Variant (for example) describes the concept, and anything that can be coerced to it is valid.

That's a good point about ducktyping with the `typing` module, yes I agree with that definition. I think that their definition of [structural subtyping](https://mypy.readthedocs.io/en/latest/protocols.html#protocol-types) is probably closer to what I'm suggesting, although as you say that is also not quite right since e.g. `int` doesn't provide the Variant API. What I'm really saying is that "anything that can be passed to `Variant(...)` or one of its subclasses is a valid `variant` in an argument".

joaander:
> What I'm really saying is that "anything that can be passed to `Variant(...)` or one of its subclasses is a valid `variant` in an argument".

This isn't true though. We only accept `Variant` objects and python objects that will successfully convert to float with `float(a)`: https://github.com/glotzerlab/hoomd-blue/blob/a2f63502adc467f3ff555616d0e27bb25d5ca9fa/hoomd/data/typeconverter.py#L37-L49

The `Variant` must be a subclass of `hoomd.variant.Variant` and cannot be duck typed. It must inherit the base class defined in C++ and exposed with pybind11.

For `Constant`'s parameter, `pybind11` applies C++ style type widening. It allows `int` which it upcasts to `Scalar`. `pybind11` does NOT apply full Python duck typing e.g. `'5'`  is invalid:
```
>>> v = hoomd.variant.Constant(5)
>>> v(0)
5.0
>>> v = hoomd.variant.Constant("5")
  File "/scratch/joaander/build/hoomd/hoomd/variant.py", line 96, in __init__
    _hoomd.VariantConstant.__init__(self, value)
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. hoomd._hoomd.VariantConstant(value: float)

Invoked with: '5'
```

However, the explicit use of `float(variant)` in `variant_preprocessing` would allow `'5'` or any other object that converts successfully via Python's `float()` method.

`variant_like` nicely gives a single anchor for the sphinx documentation element. This is much improved over the previous code where each individual method had to replicate the documentation and did so inaccurately and inconsistently. 

If there is a better item to use to anchor the `variant_like` documentation, let us know. But PEP 484 states that these type hints are to be used by IDEs or linters and it is therefore not binding:
>  Python will remain a dynamically typed language, and the authors have no desire to ever make type hints mandatory, even by convention.

Is it so bad if a user's linter notifies them that the integer they pass to `kT` will be converted to a float? Do Python type linters really enforce 100% type equality or do they allow type widening? If the user were to pass a string instead, wouldn't it be to their advantage to notify them that the string will be converted to a float?

We could update the language in `variant_like` to correctly document the behavior:
```Accepts:
* Any subclass of `Variant`
* Values that convert to float. These are internally converted to variants of type `Constant` via: ``Constant(float(variant))``.
```
and similarly for `trigger_like`.

b-butler:
As a middle ground, I think @joaander documentation suggestion makes sense.

We could document the float convertibility through structural typing as @vyasr mentions, we would need to define a float conversion interface
```python
def FloatConvertible(ABC):
    @abstractmethdo
    def __float__(self):
        pass
```
however, I am not sure how much this gains us.

vyasr:
@joaander I completely agree that adding these anchors is a big improvement for documentation, and I also agree that type hints are hints only and would never expect them to be binding. My only concern was that, by virtue of essentially being an "allowlist" of types, the current typing approach might result in too narrow of a specification. In essence, the question to me is whether you intend (at some point) to relax the current constraints by implementing something more like my wording of

> What I'm really saying is that "anything that can be passed to Variant(...) or one of its subclasses is a valid variant in an argument".

If that were to become the implementation, then type hints seem like too narrow a choice because there are different types of variants and you would need a nontrival input preprocessing function to determine what variant to create. The logic of that function would then need to be translated into appropriate documentation for the parameter.

If that is _not_ the intention then I think that the current approach is fine. I wouldn't try to be super pedantic about documenting the int->float conversion, I only brought that up as a small example of the larger conversion question. If conversions are only happening according to C++ type widening then I don't think there will be much room for confusion. Python programmers probably aren't thinking about the difference between ints and floats anyway. @b-butler I agree that your suggestion, while possible, probably doesn't buy us much. In fact, it wouldn't even be accurate since as @joaander pointed out something like `float('5')` would match your type interface but wouldn't actually be valid with the current behavior. I say let's leave it as is, then.

joaander:
> In fact, it wouldn't even be accurate since as @joaander pointed out something like `float('5')` would match your type interface but wouldn't actually be valid with the current behavior. I say let's leave it as is, then.

But `'5'` would work. The implementation explicitly calls `float()` at the Python level to convert and does not rely on `pybind11`'s conversion routines.
https://github.com/glotzerlab/hoomd-blue/blob/a2f63502adc467f3ff555616d0e27bb25d5ca9fa/hoomd/data/typeconverter.py#L37-L49

If we add more and more convenience conversions (e.g. a 7 element tuple (A, B, t_start, t_A, t_AB, t_B, t_BA) maps to variant.Ramp with those arguments), then those need to be documented somewhere. But then I'm not sure I'd call this duck typing and more a conversion step.

vyasr:
Ah sorry I misread your previous comment. You were specifically referring to the Constant when you said that pybind11 wouldn't automatically cast. In any case, I think my previous statement still holds. Implicit conversion to float via `__float__` is IMO sufficiently well documented as `float` in the type hint. No need to be explicit about that conversion.

> If we add more and more convenience conversions (e.g. a 7 element tuple (A, B, t_start, t_A, t_AB, t_B, t_BA) maps to variant.Ramp with those arguments), then those need to be documented somewhere. But then I'm not sure I'd call this duck typing and more a conversion step.

Right, I agree with that. That's where I think the analogy to numpy is important. Does variant-like just mean a short list of types, or does it mean anything that you can call `hoomd.asvariant` (yet to be defined) on, similar to `np.asarray` being the default preprocessing step for all numpy functions?

b-butler:
@tommy-waltmann Please review this when you have time.

hoomd/box.py
```diff
@@ -511,3 +505,67 @@ def __neq__(self, other):
...
+"""Objects that are or can be converted to `Box`.
+
+This includes
+
+* `hoomd.Box` objects
+* objects with attributes ``Lx, Ly, Lz xy, xz, yz``
+* lists like ``[Lx, Ly, Lz, xy, xz, yz]``
+* dictionaries with keys ``'Lx', 'Ly', 'Lz', 'xy', 'xz', 'yz'``
+* 3x3 NumPy arrays or object convertible to a 3x3 array (see
+  `hoomd.Box.from_matrix`)
```
tommy-waltmann:
```suggestion
* `hoomd.Box` objects.
* Objects with attributes ``Lx, Ly, Lz, xy, xz, yz``.
* Lists like ``[Lx, Ly, Lz, xy, xz, yz]``.
* Dictionaries with keys ``'Lx', 'Ly', 'Lz', 'xy', 'xz', 'yz'``.
* 3x3 NumPy arrays or objects convertible to a 3x3 array (see
  `hoomd.Box.from_matrix`).
```

tommy-waltmann (changes requested):
One last grammar thing, then I will approve.

tommy-waltmann (approved):

PR #1351, opened 2022-06-20
Refactor Autotuner

### LINKED ISSUE #562: API for auto-tuned parameters
## Description

Some simulations autotune slowly. It would be great to be able to restart those from a `.gsd` with optimal autotuning parameters.

## Proposed solution

Implement `[get/set]CurrentAutotunerParam()` interface on `C++` *GPU classes. Implement a property (#398) on the python side, pass to `dump.gsd()` object (by default?).

## Additional context

Autotuning in MD can take 10,000 time steps, or in HPMC (especially GPU JIT #549), up to 1-2 minutes. This change would save time at the start of the simulation and make profiling/benchmarking easier for developers.

## Developer

Yes, I can contribute. Other volunteers?

joaander:
## Description

* Refactor C++ Autotuner implementation to work on multi-dimensional parameters using std::array
* Refactor C++ base class structure to provide a common Python interface in `AutotunedObject`
* Refactor Python base class structure to provide a common interface to
    * Get kernel parameters
    * Set kernel parameters
    * Test whether kernel parameter tuning is complete
    * Start tuning kernel parameters
* Add convenience methods to do the above for all children in `Operations`.

## Motivation and context

* Allow users to save and restore optimal kernel parameters.
* Allow benchmarks to query when tuning is complete so they can proceed with timing measurements.
* Allow users to start a new tuning sequence when needed (e.g. after compressing the system).

Resolves #562

## How has this been tested?

Tested in interactive sessions:

```
>>> benchmark.sim.operations.integrator.tune_kernel_parameters()
notice(4): Autotuner hpmc_moves starting scan.
notice(4): Autotuner hpmc_update_pdata starting scan.
notice(4): Autotuner hpmc_excell_block_size starting scan.
notice(4): Autotuner hpmc_num_depletants starting scan.
notice(4): Autotuner hpmc_num_depletants_ntrial starting scan.
notice(4): Autotuner hpmc_narrow starting scan.
notice(4): Autotuner hpmc_convergence starting scan.
notice(4): Autotuner hpmc_depletants_accept starting scan.
notice(4): Autotuner hpmc_depletants starting scan.
notice(4): Autotuner hpmc_depletants_phase1 starting scan.
notice(4): Autotuner hpmc_depletants_phase2 starting scan.
notice(4): Autotuner cell_list starting scan.
notice(4): Autotuner cell_list_combine starting scan.
>>> benchmark.sim.operations.integrator.is_tuning_complete
False
>>> benchmark.sim.operations.integrator.kernel_parameters
{'hpmc_moves': (32,), 'hpmc_update_pdata': (32,), 'hpmc_excell_block_size': (32,), 'hpmc_num_depletants': (32,), 'hpmc_num_depletants_ntrial': (32,), 'hpmc_narrow': (32, 1, 4), 'hpmc_convergence': (32,), 'hpmc_depletants_accept': (32,), 'hpmc_depletants': (32, 4, 4), 'hpmc_depletants_phase1': (32, 4, 4), 'hpmc_depletants_phase2': (32, 4, 4)}
>>> benchmark.sim.run(10)
notice(5): IntegratorHPMCMono: updating nominal width to 1
notice(4): Autotuner hpmc_convergence found optimal parameter (992,) with a performance spread of 18%.
>>> benchmark.sim.operations.integrator.kernel_parameters
{'hpmc_moves': (288,), 'hpmc_update_pdata': (288,), 'hpmc_excell_block_size': (352,), 'hpmc_num_depletants': (32,), 'hpmc_num_depletants_ntrial': (32,), 'hpmc_narrow': (320, 1, 2), 'hpmc_convergence': (992,), 'hpmc_depletants_accept': (32,), 'hpmc_depletants': (32, 4, 4), 'hpmc_depletants_phase1': (32, 4, 4), 'hpmc_depletants_phase2': (32, 4, 4)}
>>> benchmark.sim.operations.integrator.is_tuning_complete
False
>>> benchmark.sim.run(100)
notice(5): IntegratorHPMCMono: updating nominal width to 1
notice(4): Autotuner hpmc_narrow found optimal parameter (32,1,1,) with a performance spread of 2236%.
notice(4): Autotuner hpmc_moves found optimal parameter (32,) with a performance spread of 66%.
notice(4): Autotuner hpmc_update_pdata found optimal parameter (256,) with a performance spread of 54%.
>>> benchmark.sim.operations.integrator.kernel_parameters
{'hpmc_moves': (32,), 'hpmc_update_pdata': (256,), 'hpmc_excell_block_size': (480,), 'hpmc_num_depletants': (32,), 'hpmc_num_depletants_ntrial': (32,), 'hpmc_narrow': (32, 1, 1), 'hpmc_convergence': (992,), 'hpmc_depletants_accept': (32,), 'hpmc_depletants': (32, 4, 4), 'hpmc_depletants_phase1': (32, 4, 4), 'hpmc_depletants_phase2': (32, 4, 4)}
>>> benchmark.sim.operations.integrator.is_tuning_complete
False
>>> benchmark.sim.run(1000)
notice(5): IntegratorHPMCMono: updating nominal width to 1
notice(4): Autotuner cell_list found optimal parameter (64,) with a performance spread of 80%.
notice(4): Autotuner hpmc_excell_block_size found optimal parameter (32,) with a performance spread of 142%.
>>> benchmark.sim.operations.integrator.kernel_parameters
{'hpmc_moves': (32,), 'hpmc_update_pdata': (256,), 'hpmc_excell_block_size': (32,), 'hpmc_num_depletants': (32,), 'hpmc_num_depletants_ntrial': (32,), 'hpmc_narrow': (32, 1, 1), 'hpmc_convergence': (992,), 'hpmc_depletants_accept': (32,), 'hpmc_depletants': (32, 4, 4), 'hpmc_depletants_phase1': (32, 4, 4), 'hpmc_depletants_phase2': (32, 4, 4)}
>>> benchmark.sim.operations.integrator.is_tuning_complete
False
```

The last line should be `True`, I will investigate later. It does become `True` for the initial tuning, but stays `False` when calling `tune_kernel_parameters`.

## Change log

```
Added:

* `AutotunedObject` class that provides an interface to read and write tuned kernel parameters, query whether tuning is complete, and start tuning again at the object level.
* `is_tuning_complete` method to `Operations`. Check whether kernel parameter tuning is complete for all operations.
* `tune_kernel_parameters` methods to `Operations` and many other classes. Start tuning kernel parameters in all operations.
```

## Checklist:

## TODO: 
- [x] add unit tests
- [x] Write change log entry
- [x] Fix `False` `is_tuning_complete` after calling `tune_kernel_parameters`
- [x] Port MD and other components.
- [x] Add overview documentation.


hoomd/Action.cc
```diff
@@ -0,0 +1,22 @@
...
+namespace detail
+    {
+void export_Action(pybind11::module& m)
+    {
+    pybind11::class_<Action, std::shared_ptr<Action>>(m, "Action")
+        .def(pybind11::init<std::shared_ptr<SystemDefinition>>())
+        .def("getAutotunerParameters", &Action::getAutotunerParameters)
+        .def("setAutotunerParameters", &Action::setAutotunerParameters)
+        .def("startAutotuning", &Action::startAutotuning)
+        .def("isAutotuningComplete", &Action::isAutotuningComplete);
```
b-butler: Why do we expose this at the `Action` and not `Autotuned` level?

hoomd/Autotuner.h
```diff
@@ -24,227 +34,620 @@
...
+    */
+    static std::vector<unsigned int>
+    getTppListPow2(const std::shared_ptr<const ExecutionConfiguration> exec_conf,
+                   int force_size = -1)
+        {
+        std::vector<unsigned int> v;
+        unsigned int warp_size = exec_conf->dev_prop.warpSize;
+        if (force_size > 0)
+            {
+            warp_size = (unsigned int)force_size;
```
b-butler:
```suggestion
            warp_size = static_cast<unsigned int>(force_size);
```

hoomd/Autotuner.h
```diff
@@ -24,227 +34,620 @@
...
+    /// Test if the scan is complete.
+    /*! \returns true when autotuning is complete.
      */
-    void setPeriod(unsigned int period)
+    virtual bool isComplete()
         {
-        m_exec_conf->msg->notice(6)
-            << "Set Autotuner " << m_name << " period = " << period << std::endl;
-        m_period = period;
+        if (m_state != SCANNING)
```
b-butler, line 307: Since `begin` changes `INACTIVE` to `SCANNING` shouldn't that also be considered not complete?

hoomd/Autotuner.h
```diff
@@ -24,227 +34,620 @@
...
+        s << "Autotuner " << m_name << " has no valid parameters.";
+        throw std::invalid_argument(s.str());
+        }
+
+    // Ensure that m_n_samples is non-zero and odd to simplify the median computation.
+    if ((m_n_samples & 1) == 0)
+        m_n_samples += 1;
+
+    // Initialize memory.
+    m_samples.resize(m_parameters.size());
```
b-butler, line 485: Should you resize `m_parameters` to possibly free up memory?

hoomd/Autotuner.h
```diff
@@ -24,227 +34,620 @@
...
+#endif
+        if (is_root)
+            {
+            if (m_mode == mode_avg)
+                {
+                // Compute average.
+                float sum = 0.0f;
+                for (std::vector<float>::iterator it = v.begin(); it != v.end(); ++it)
+                    sum += *it;
+                m_sample_median[i] = sum / float(v.size());
```
b-butler: Should we rename this variable considering it isn't always the median?

hoomd/Autotuner.h
```diff
@@ -24,227 +34,620 @@
...
+                size_t n = v.size() / 2;
+                nth_element(v.begin(), v.begin() + n, v.end());
+                m_sample_median[i] = v[n];
+                }
+            }
+        }
+
+    size_t min_idx = 0;
+
+    if (is_root)
```
b-butler, line 630:
```suggestion
    // Report performance characteristics of Autotuning
    if (is_root)
```

hoomd/Autotuner.h
```diff
@@ -24,227 +34,620 @@
...
+        }
+    }
+
+/*! \returns The index of the optimal parameter given the current data in m_samples.
+
+    computeOptimalParameter computes the median, average, or maximum time among all samples for all
+    elements. It then chooses the fastest time (with the lowest index breaking a tie) and returns
+    the index of the parameter that resulted in that time.
+*/
+template<size_t n_dimensions> size_t Autotuner<n_dimensions>::computeOptimalParameterIndex()
```
b-butler, line 561: If possible, I think this would be cleaner as two methods: one that computes the performance metric and another that then gets the best index and reports the  #performance.

hoomd/Autotuner.h
```diff
@@ -24,227 +34,620 @@
...
+
+    /// Get the autotuner's name
+    std::string getName()
+        {
+        return m_name;
+        }
+
+#ifndef __HIPCC__
+    /// Get the autotuner parameters as a Python tuple.
+    virtual pybind11::tuple getParameterPython()
```
b-butler, line 71: Wouldn't a method like this work better as a pure virtual method? The class is an interface, but we provide default (null) implementations for all the methods.

Or is this a requirement since we use this type to store autotuners in a `std::vector`?

hoomd/Autotuner.h
```diff
@@ -24,227 +34,620 @@
...
+                if (is_parameter_valid(parameter))
+                    {
+                    m_parameters.push_back(parameter);
+                    m_exec_conf->msg->notice(5) << "Autotuner " << m_name << " adding parameter "
+                                                << formatParam(parameter) << std::endl;
+                    }
+                }
+            else
+                {
+                initializeParameters(dimension_ranges,
```
b-butler, line 419:
```suggestion
                // populate parameter for higher dimensions
                initializeParameters(dimension_ranges,
```

hoomd/Autotuner.h
```diff
@@ -24,227 +34,620 @@
...
+                }
+            else
+                {
+                m_current_param = m_parameters[m_current_element];
+                }
+            }
+        else
+            {
+            m_current_param = m_parameters[m_current_element];
+            }
```
b-butler, line 551: Is this necessary?

hoomd/CellListGPU.cc
```diff
@@ -24,22 +24,15 @@ CellListGPU::CellListGPU(std::shared_ptr<SystemDefinition> sysdef)
...
-                                        this->m_exec_conf));
+    m_tuner.reset(new Autotuner<1>({AutotunerInterface::makeBlockSizeRange(m_exec_conf)},
+                                   this->m_exec_conf,
+                                   "cell_list"));
+    m_autotuners.push_back(m_tuner);
+
+    m_tuner_combine.reset(new Autotuner<1>({AutotunerInterface::makeBlockSizeRange(m_exec_conf)},
+                                           this->m_exec_conf,
+                                           "cell_list_combine"));
+    m_autotuners.push_back(m_tuner_combine);
```
b-butler: Looking at this makes me wonder if we should use an unordered map. Then we wouldn't need to store autotuners in a member variable and `m_autotuners`.

hoomd/System.cc
```diff
@@ -188,31 +188,33 @@ void System::updateTPS()
...
-*/
-void System::setAutotunerParams(bool enabled, unsigned int period)
+void System::startAutotuning()
     {
     // set the autotuner parameters on everything
     if (m_integrator)
-        m_integrator->setAutotunerParams(enabled, period);
+        m_integrator->startAutotuning();
+
+    //     //TODO: Integrator needs to call startAutotuning on all child classes, as do classes like
```
b-butler: Comment highlighting the TODO

sphinx-doc/features.rst
```diff
@@ -77,6 +77,11 @@ run time with the `device <hoomd.device>` module. Unless otherwise stated in the
 **all** operations and methods support GPU execution. At runtime, `hoomd.version.gpu_enabled` indicates
 whether the build supports GPU devices.
 
+Autotuned kernel launch parameters
+----------------------------------
+
+TODO
```
b-butler: Comment marking todo

hoomd/operation.py
```diff
@@ -288,7 +288,88 @@ def __getstate__(self):
...
+        making the change with `tune_kernel_parameters`.
+
+    See Also:
+        * `hoomd.Operations.is_tuning_complete`
+        * `hoomd.Operations.tune_kernel_parameters`
+    """
+
+    @property
+    def kernel_parameters(self):
+        """dict: Kernel parameters.
```
b-butler:
```suggestion
        """dict[str, tuple[float]]: Kernel parameters.
```

hoomd/operations.py
```diff
@@ -290,6 +291,41 @@ def computes(self):
...
+        if not self._scheduled:
+            raise DataAccessError("is_tuning_complete")
+
+        result = True
+        # TODO: Call C++ System object to do this, non-operations like
+        # ParticleData, Communicator, etc... may be autotuned as well
+        for op in self:
+            result = result and op.is_tuning_complete
+
+        return result
```
b-butler:
```suggestion
        # TODO: Call C++ System object to do this, non-operations like
        # ParticleData, Communicator, etc... may be autotuned as well
        return all(op.is_tuning_complete for op in self)
```

b-butler (commented):
- :+1: to the new `Action` C++ class.
- I think making `m_autotuners` a `std::unordered_map<std::string, std::shared_ptr<AutotunerInterface>>` would make sense and we could remove the individual tuner member variables without losing readability from the name.
- I think the dict, tuple interface in Python is the right approach.

In general, I think the code make many good changes to the autotuner API both in C++ and Python. I have left specific comments in a few places. All the big level changes I have made above or commented in my review.

hoomd/Action.cc
joaander: Because I implemented it a the `Action` level first, then later factored out the `Autotuned` class. This will need to be fixed for `Autotuned` classes (like MD integration methods) to have a functioning Python API.

hoomd/Autotuner.h
joaander: We could call `shrink_to_fit`: https://en.cppreference.com/w/cpp/container/vector/shrink_to_fit

hoomd/Autotuner.h
joaander: Yes.

hoomd/Autotuner.h
joaander: I don't know if it is a requirement of `std::vector` or not.

hoomd/Autotuner.h
joaander: No. It is an error, in fact. It will overwrite a `m_current_param` set by the user. It is probably left over from the old `Autotuner` code which had a more complex state machine.

hoomd/CellListGPU.cc
joaander: We could, yes. Would that set their string name as well? It would be the most logical connection between the Python dict of tuples interface. However if that were the case, how would they know their name to include in the notice messages?

hoomd/Autotuner.h
joaander: This is intentional. I want tuners that are created, but`begin` is never called to be "complete". Otherwise, they would prevent the `Autotuned` object as a whole from ever registering as complete.

hoomd/CellListGPU.cc
b-butler: Hmm, that is a good point. We could provide an interface for adding or removing tuners from the base class that set's it both places.

Otherwise, we could write an outside method/function that writes the logging messages (if possible I can't remember what messages are written).

hoomd/Autotuner.h
joaander: This class really isn't an interface as it does implement some common code. I will rename it.

hoomd/Autotuner.h
joaander: On looking at this more carefully, the code is correct as implemented. It does not override the parameter set by the user because this is in an `if (m_state == SCANNING)` block.

joaander: The reason this is necessary in the first place is because of the order in when the state machine moves over elements and samples.

hoomd/CellListGPU.cc
joaander: Though I do worry about the overhead of an `unordered_map` lookup every time we accesses one of these tuners. It seems unnecessarily complex. Is your goal to reduce the amount of code spent in initialization? I could collect all the ` m_autotuners.push_back` lines into one.

b-butler: Yeah, if the notice messages are necessary then the name has to live in two places. I was trying to actually reduce the code complexity (from a human perspective not performance). Having to store every auto tuner in two places seems a bad pattern. We can't just use a `std::vector` though because that would obfuscate the code which is why I suggested an `unordered_map`. The slow down would only be while tuning correct? If so that shouldn't effect performance as a whole by much.

joaander: > The slow down would only be while tuning correct? If so that shouldn't effect performance as a whole by much.

No, the tuner's `begin` and `end` methods are called for every kernel launch. The caller doesn't know whether the tuner is active or not.

hoomd/Autotuner.h
joaander: Due to a lack of time, I will not implement this suggestion.

hoomd/Autotuner.h
```diff
@@ -24,227 +34,628 @@
...
+            {
+            warp_size = static_cast<unsigned int>(force_size);
+            }
+
+        for (unsigned int s = 4; s <= warp_size; s *= 2)
+            {
+            v.push_back(s);
+            }
+        v.push_back(1);
+        v.push_back(2);
```
b-butler:
```suggestion
        std::vector<unsigned int> v {1, 2};
        unsigned int warp_size = exec_conf->dev_prop.warpSize;
        if (force_size > 0)
            {
            warp_size = static_cast<unsigned int>(force_size);
            }

        for (unsigned int s = 4; s <= warp_size; s *= 2)
            {
            v.push_back(s);
            }
```

hoomd/hpmc/pytest/test_compute_free_volume.py
```diff
@@ -76,6 +76,14 @@ def test_validation_systems(simulation_factory, two_particle_snapshot_factory,
                                free_volume_compute.free_volume,
                                rtol=2e-2)
 
+    # Tet the kernel parameter tuner.
+    def activate_tuner():
+        sim.run(1)
+        return free_volume_compute.free_volume
```
b-butler: No need for the return statement
```suggestion
        # We need to make the kernel be called.
        free_volume_compute.free_volume
```

hoomd/md/ActiveForceComputeGPU.h
```diff
@@ -31,22 +31,17 @@ class PYBIND11_EXPORT ActiveForceComputeGPU : public ActiveForceCompute
...
-    virtual void setAutotunerParams(bool enable, unsigned int period)
+    /// Start autotuning kernel launch parameters
+    virtual void startAutotining()
         {
-        ActiveForceCompute::setAutotunerParams(enable, period);
-        m_tuner_force->setPeriod(period);
-        m_tuner_force->setEnabled(enable);
-        m_tuner_diffusion->setPeriod(period);
-        m_tuner_diffusion->setEnabled(enable);
+        // ActiveForceCompute::startAutotuning();
```
b-butler:
```suggestion
```

hoomd/md/PotentialPair.h
```diff
@@ -843,6 +849,21 @@ CommFlags PotentialPair<evaluator>::getRequestedCommFlags(uint64_t timestep)
...
+
+    // Start autotuning the neighbor list.
+    m_nlist->startAutotuning();
+    }
+
+template<class evaluator> bool PotentialPair<evaluator>::isAutotuningComplete()
+    {
+    bool result = ForceCompute::isAutotuningComplete();
+    result = result && m_nlist->isAutotuningComplete();
+    return result;
```
b-butler:
```suggestion
    return result && m_nlist->isAutotuningComplete();
```

hoomd/md/PotentialTersoff.h
```diff
@@ -107,6 +107,23 @@ template<class evaluator> class PotentialTersoff : public ForceCompute
...
+        // Start autotuning the neighbor list.
+        m_nlist->startAutotuning();
+        }
+
+    /// Check if autotuning is complete.
+    virtual bool isAutotuningComplete()
+        {
+        bool result = ForceCompute::isAutotuningComplete();
+        result = result && m_nlist->isAutotuningComplete();
+        return result;
```
b-butler:
```suggestion
        return result && m_nlist->isAutotuningComplete();
```

hoomd/module.cc
```diff
@@ -153,6 +154,18 @@ std::string mpi_bcast_str(pybind11::object string,
 #endif
     }
 
+bool mpi_allreduce_bcast_and(bool v, std::shared_ptr<ExecutionConfiguration> exec_conf)
+    {
+#ifdef ENABLE_MPI
+    bool reduced_result = false;
+    MPI_Allreduce(&v, &reduced_result, 1, MPI_C_BOOL, MPI_LAND, exec_conf->getMPICommunicator());
+    bcast(reduced_result, 0, exec_conf->getMPICommunicator());
```
b-butler: Shouldn't `MPI_Allreduce` handle this?
```suggestion
```

b-butler (commented):
Overall looks really good. I left a few comments.

hoomd/module.cc
joaander: Yes. The microsoft documentation (first hit on DuckDuckGo) is incorrect.

hoomd/md/ActiveForceComputeGPU.h
joaander: Good catch. I should remove the whole method as this is just a reimplementation of `Autotuned::startAutotuning`.

hoomd/hpmc/pytest/test_compute_free_volume.py
joaander: I'll try this change. I didn't write it this way as I expected flake8 to complain.

hoomd/Autotuner.h
joaander: This ordering is intentional. Unit tests tend to trigger only the first few entries in the autotuner parameter list. By ensuring the more general multiple threads per particle path is the first executed the unit tests can find bugs in that code path. We made this change when the tpp=1 code path passed all unit tests but the tpp=2+ path failed. 

I'll add a comment to explain this reasoning.

joaander:
#1374 documents some additional work to be completed in a future PR.

joaander:
All tests passing. This is ready for the final review round.

hoomd/operation.py
```diff
@@ -288,7 +288,95 @@ def __getstate__(self):
...
+    Check `is_tuning_complete` to check if the search is complete. After the
+    search is complete, `AutotunedObject` holds the parameters constant.
+    Typical operations require thousands of time steps to complete tuning. Some
+    may take tens of thousands or more depending on the parameters you set.
+
+    Tip:
+        When you significantly change your system during the simulation (e.g.
+        compress to a higher density), then tune the parameters again after
+        making the change with `tune_kernel_parameters`.
+
```
Charlottez112, line 314: This is a very helpful tip, but it's not very clear if this is something we always have to do after changing the system. I imagine it's necessary if the change significantly affects the performance?

hoomd/operation.py
```diff
@@ -288,7 +288,95 @@ def __getstate__(self):
...
+    Note:
+        In MPI parallel execution, all methods and attributes of
+        `AutotunedObject` reference the rank local tuner objects. Different
+        ranks may tune different optimal kernel parameters and may complete
+        tuning at different times.
+
```
Charlottez112, line 322: It seems like this has some implications. Could you explain what's implied here?

ARCHITECTURE.md
```diff
@@ -205,6 +207,19 @@ of bare pointers to data arrays, array sizes, etc... The ABI for these calls is
...
 parameters (such as `BoxDim`).
 
+## Autotuning
+
+HOOMD-blue automatically tunes kernel block sizes, threads per particle, and other kernel launch
+paramters. The `Autotuner` class manages the sparse multi-dimensional of parameters for each kernel.
+It dynamically cycles through the possible parameters and records the performance of each using CUDA
+events. After scanning through all parameters, it selects the best performing one to continue
+executing. GPU code in HOOMD-blue should instantiate and use one `Autotuner` for each kernel.
+Classes that use the autotuner should inherit from `Autotuned` which tracks all the autotuners and
```
Charlottez112: How does one know which class is autotuned?

hoomd/operation.py
joaander: I added a statement that this is optional, but that it can increase performance.

hoomd/operation.py
joaander: Users can inspect `is_tuning_complete` for individual objects, but should not use them to control the simulation - otherwise deadlock could result. I added a statement indicating that one should use `Operations.is_tuning_complete` for control flow.

ARCHITECTURE.md
joaander: When a class has `Autotuner` member variables. I reworded the documentation to indicate this.

Charlottez112 (approved):

hoomd/hpmc/pytest/test_pair_user.py
```diff
@@ -89,14 +89,14 @@ def test_attaching(device, simulation_factory, two_particle_snapshot_factory):
 
 @pytest.mark.validate
 @pytest.mark.skipif(llvm_disabled, reason='LLVM not enabled')
-def test_kernel_parameters(simulation_factory, two_particle_snapshot_factory):
-    patch = hoomd.hpmc.pair.user.CPPPotential(r_cut=3,
+def test_kernel_parameters(simulation_factory, lattice_snapshot_factory):
```
b-butler, line 92: Is there a reason for switching to use `lattice_snapshot_factory` in more places?

b-butler (approved):
The changes look good to me.

hoomd/hpmc/pytest/test_pair_user.py
joaander: With the two particle snapshot factory, one of the ranks had no particles and failed to run the kernel. Thus the autotuner on that rank didn't change and the assert checking for changed values failed.

PR #1352, opened 2022-06-20
Reexport `Integrator`'s `setHalfStepHook` and `removeHalfStepHook`

p-zubieta:
## Description

Expose `setHalfStepHook` and `removeHalfStepHook` as part of python-side `Integrator` interface for HOOMD-blue 3.

## Motivation and context

In https://github.com/glotzerlab/hoomd-blue/commit/4f91146843be1e47443a152bf3fe536f585f0c90 `setHalfStepHook` and `removeHalfStepHook` were dropped from the python interface for `Integrator`. We rely on these in https://github.com/SSAGESLabs/PySAGES

## How has this been tested?

Not sure what's the best way to test directly here.

## Change log

```
*- Expose `setHalfStepHook` and `removeHalfStepHook` as part of python-side `Integrator` interface.
```

## Checklist:

### DIFF
--- a/hoomd/Integrator.cc
+++ b/hoomd/Integrator.cc
@@ -914,7 +914,9 @@ void export_Integrator(pybind11::module& m)
         .def_property("dt", &Integrator::getDeltaT, &Integrator::setDeltaT)
         .def_property_readonly("forces", &Integrator::getForces)
         .def_property_readonly("constraints", &Integrator::getConstraintForces)
-        .def("computeLinearMomentum", &Integrator::computeLinearMomentum);
+        .def("computeLinearMomentum", &Integrator::computeLinearMomentum)
+        .def("setHalfStepHook", &Integrator::setHalfStepHook)
+        .def("removeHalfStepHook", &Integrator::removeHalfStepHook);
     }
 
     } // end namespace detail
--- a/hoomd/md/HalfStepHook.cc
+++ b/hoomd/md/HalfStepHook.cc
@@ -1,0 +1,48 @@
+// Copyright (c) 2009-2022 The Regents of the University of Michigan.
+// Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+#include "hoomd/HalfStepHook.h"
+
+namespace hoomd
+    {
+namespace md
+    {
+//! Trampoline for HalfStepHook classes inherited in python
+class HalfStepHookPy : public HalfStepHook
+    {
+    public:
+    // Inherit the constructor
+    using HalfStepHook::HalfStepHook;
+
+    // Trampoline methods
+    void setSystemDefinition(SystemDefinitionSPtr sysdef) override
+        {
+        PYBIND11_OVERLOAD_PURE(void, HalfStepHook, setSystemDefinition, sysdef);
+        }
+
+    void update(uint64_t timestep) override
+        {
+        PYBIND11_OVERLOAD_PURE(void, HalfStepHook, update, timestep);
+        }
+    };
+
+namespace detail
+    {
+// Method to enable unit testing of C++ HalfStepHook::update from pytest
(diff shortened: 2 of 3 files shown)


joaander (commented):
I don't see how this change set is sufficient to enable the support needed for Python libraries to access the half step hook. The HalfStepHook class is currently C++ only, added for use by the C++ SSAGES library.

I would have thought that we would also need a Python layer along the lines of the [hoomd.custom.Action](https://hoomd-blue.readthedocs.io/en/v3.2.0/module-hoomd-custom.html#hoomd.custom.Action) class implemented by @b-butler.

p-zubieta:
Yes you're right in this is not sufficient for python libraries to access the half step hook. We have a C++ plugin https://github.com/SSAGESLabs/hoomd-dlext (that we use in https://github.com/SSAGESLabs/PySAGES) which exports a python class derived from `HalfStepHook`. For the latter, we call `setHalfStepHook` from the python side for HOOMD 2.

I understand if you'd prefer to keep this as is, but wanted to check if it was possible to do it here since that wouldn't require many changes to our plugin. For HOOMD 3 we should be able to move the bits that call `setHalfStepHook` to the C++ plugin if this PR doesn't come through.

joaander:
> I understand if you'd prefer to keep this as is, but wanted to check if it was possible to do it here since that wouldn't require many changes to our plugin. For HOOMD 3 we should be able to move the bits that call `setHalfStepHook` to the C++ plugin if this PR doesn't come through.

We can certainly make this change here. I wanted to understand how you are using it and I'm suggesting additions to make the hook usable by more than just PySSAGES users. At a minimum, I think we should move the pybind11 export the base `HalfStepHook` into HOOMD-blue. It should not be the plugin's responsibility to export base classes.

I'd like to also make the Integrator's hook available as a property (`half_step_hook`) in the public API so that PySSAGES and other users can work with documented APIs. We have an already established pattern of exposing C++ classes directly to users and documenting them through a proxy Python class (see `variant.py`, `trigger.py`, and `filter/*.py`.) so that they can be extended in Python (and C++).

I was incorrect earlier that this would need to be modeled after `custom.Action` as `HalfStepHook` is much more similar to a variant or trigger.

Would you be willing to make these changes on this PR?

1. Export `HalfStepHook` with pybind11 (move the relevant code from hoomd-dlext to here)
2. Add a `half_step_hook` property to `Integrator` and document it.  (you can then use this from the Python side to set your classes derived from `HalfStepHook`)
3. Add a Python level `HalfStepHook` proxy class to provide user-facing documentation for the class and a way for users to extend it in Python.

`HalfStepHook` is specific to MD. We should add it to a module in the `md` namespace: `hoomd.md.hook.HalfStepHook`.

p-zubieta:
Ok, great. Yes, I'd be willing to do the changes you propose in here.

 - [x] Export `HalfStepHook` with pybind11.
 - [ ] Add a `half_step_hook property` to Integrator and document it.
 - [ ] Add a Python level `HalfStepHook` proxy class to provide user-facing documentation for the class and a way for users to extend it in Python.

p-zubieta:
Should I put `HalfStepHook.cc` within `hoomd` or should I have it instead in `hoomd/md`? We would export from the `_md` submodule either way.

joaander:
> Should I put `HalfStepHook.cc` within `hoomd` or should I have instead in `hoomd/md`? we would export from the `_md` submodule either way.

Put it in `hoomd/md`. I think it is an oversight that `HalfStepHook` is available in `Integrator`, as it is specific to the MD `IntegratorTwoStep`.

joaander:
Feel free to reopen if you would like to complete this work. Also let me know if you have any questions. We can help.

p-zubieta:
We plan on finishing this, just caught up addressing other pending work on PySAGES. I'll get back to this within the next couple of weeks.

PR #1356, opened 2022-07-05
Implement brute force overlap search for SDF compute

tcmoore3:
## Description

This PR implements a brute force search for finding overlaps in `hpmc.compute.SDF`. In contrast to the binary search algorithm that is currently in place, the "first" overlap for a pair of particles is found by iteratively scaling down the vector connecting their centers and checking for overlaps at each scale value. The compute also weights the points added to the histogram by a factor that is a function of the change in energy upon overlap, where we now have a notion of "soft overlaps" that result from a change in the interaction energy for a pair of particles in addition to the "hard overlaps" that are a standard HPMC concept. By considering soft overlaps, we can accurately calculate the pressure in systems of patchy particles.

The user-facing API is unaffected by these changes. The binary search is valid for athermal systems of convex particles, and is much faster than the brute force method in that case. Therefore, upon computation of the SDF, we check if there is a patch energy evaluator and only use the brute force method if there is. Otherwise, we use the binary search method.

## Motivation and context

This change is needed to calculate the pressure in systems with discontinuous pair potentials where the change in the energy for a pair of particles can be a nonmonotonic function of their separation (e.g., patchy particles or non-convex hard particles).

## How has this been tested?

I have done fairly extensive tests for systems of patchy hexagons. I ran a series of NPT simulations at several pressures and temperatures and constructed equations of state isotherms. I then ran NVT systems at the same temperatures and the mean densities from the NPT simulations. The two sets of isotherms (obtained from NPT simulations and obtained from NVT simulations) overlap well within error bars. Pressures measured in NVT simulations are generally accurate to within 2 decimals places of the pressure that gave the corresponding density in NPT simulations.

I plan on doing similar tests (probably not as comprehensive but several state points) for several different systems that span dimensionality (2D vs 3D), shape (e.g., polygons, polyhedra, spheres), and density.

## Change log

```
- Enable pressure measurement via hpmc.compute.SDF for systems with pair potentials
```

## Checklist:

- [ ] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/CONTRIBUTING.md).


hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                        = m_mc->m_patch->energy(r_ij_scaled,
+                                                                typ_i,
+                                                                quat<float>(shape_i.orientation),
+                                                                float(h_diameter.data[i]),
+                                                                float(h_charge.data[i]),
+                                                                typ_j,
+                                                                quat<float>(orientation_j),
+                                                                float(h_diameter.data[j]),
+                                                                float(h_charge.data[j]));
+                                    // if enery has changed, there is a new soft overlap
```
AlainKadar, line 565: *energy

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+        = m_mc->getParams();
+
+    // loop through N particles
+    // At the top of this loop, we initialize min_bin_ptl_i to the size of the sdf histogram
+    // For each of particle i's neighbors, we find the scaling that produces the first overlap.
+    // For each neighbor, we do a brute force search from the scaling corresponding to bin 0
+    // up to the minimum bin that we've already found for particle i.
+    // Then we add to m_hist[min_bin_ptl_i] the negative Mayer-function corresponding to the type
+    // of overlap corresponding to particle i's first overlap.
+    for (unsigned int i = 0; i < m_pdata->getN(); i++)
```
AlainKadar, line 446: Why not store result of `getN()` instead of calling on every iteration, like you do with `n_images` below? (Does it change?)

hoomd/hpmc/ComputeSDF.h
```diff
@@ -293,15 +315,17 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
     ArrayHandle<Scalar4> h_orientation(m_pdata->getOrientationArray(),
                                        access_location::host,
                                        access_mode::read);
-
+    ArrayHandle<Scalar> h_diameter(m_pdata->getDiameters(),
+                                   access_location::host,
+                                   access_mode::read);
+    ArrayHandle<Scalar> h_charge(m_pdata->getCharges(), access_location::host, access_mode::read);
```
b-butler:
```suggestion

```
These are not used.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
                     }
                 } // end loop over AABB nodes
             }     // end loop over images
+        if (min_bin_ptl_i < m_hist.size())
+            {
+            m_hist[min_bin_ptl_i]++;
+            }
+        } // end loop over all particles
+    }     // end countHistogram()
```
b-butler:
```suggestion
    }     // end countHistogramBinarySearch()
```

hoomd/hpmc/IntegratorHPMC.h
```diff
@@ -446,6 +446,8 @@ class PYBIND11_EXPORT IntegratorHPMC : public Integrator
         return m_patch;
         }
 
+    std::shared_ptr<PatchEnergy> m_patch; //!< Patchy Interaction
+
```
b-butler: ? This shouldn't be public. It should be protected. Use `getPatchEnergy` to get it.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                    // add appropriate weight to appropriate bin and exit the
+                                    // loop over bins
+                                    hist_weight_ptl_i = 1.0; // = 1-e^(-\infty)
+                                    min_bin_ptl_i = bin_to_sample;
+                                    break;
+                                    } // end if overlap
+
+                                // if no hard overlap, check for a soft overlap if we have
+                                // patches
+                                if (m_mc->m_patch)
```
b-butler:
```suggestion
                                if (m_mc->getPatchEnergy())
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
-        // record the minimum bin
-        if ((unsigned int)min_bin < m_hist.size())
-            m_hist[min_bin]++;
+                            // put particles in coordinate system of particle i
+                            vec3<Scalar> r_ij = vec3<Scalar>(postype_j) - pos_i_image;
 
+                            double u_ij_0 = 0.0; // energy of pair interaction in unperturbed state
+                            if (m_mc->m_patch)
+                                {
+                                u_ij_0 = m_mc->m_patch->energy(r_ij,
```
b-butler:
```suggestion
                                u_ij_0 = m_mc->getPatchEnergy()->energy(r_ij,
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                            int typ_j = __scalar_as_int(postype_j.w);
 
-        // record the minimum bin
-        if ((unsigned int)min_bin < m_hist.size())
-            m_hist[min_bin]++;
+                            // put particles in coordinate system of particle i
+                            vec3<Scalar> r_ij = vec3<Scalar>(postype_j) - pos_i_image;
 
+                            double u_ij_0 = 0.0; // energy of pair interaction in unperturbed state
+                            if (m_mc->m_patch)
```
b-butler:
```suggestion
                            if (m_mc->getPatchEnergy())
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,23 +269,40 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
+   without any communication.
       - The integrator performs the ghost exchange (with the ghost width extra that we add)
-      - Only on writeOutput() do we need to sum the per-rank histograms into a global histogram
+
+    This function is a wrapper that calls the appropriate method depending on the value of
+   m_mc->m_patch
 */
 template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
+    {
+    if (m_mc->m_patch)
```
b-butler:
```suggestion
    if (m_mc->getPatchEnergy())
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,23 +269,40 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
+    the bin corresponding to the scale factor of the vector r_ij that induces the first overlap
+    between particles *i* and *j*, and then adds s_ij to the bin, where s_ij = 1 - e^{\beta\Delta
+   U_{ij}}. Note that this definition allows for the consideration of soft overlaps where the pair
+   potential changes value, and reduces 1 for hard particle overlaps.  countHistogram() operates
+   without any communication.
       - The integrator performs the ghost exchange (with the ghost width extra that we add)
-      - Only on writeOutput() do we need to sum the per-rank histograms into a global histogram
+
+    This function is a wrapper that calls the appropriate method depending on the value of
+   m_mc->m_patch
```
b-butler:
```suggestion
   m_mc->getPatchEnergy()
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+
+                                // if no hard overlap, check for a soft overlap if we have
+                                // patches
+                                if (m_mc->m_patch)
+                                    {
+                                    // compute the energy at this size of the perturbation and
+                                    // compare to the energy in the unperturbed state
+                                    vec3<Scalar> r_ij_scaled = r_ij * (Scalar(1.0) - scale_factor);
+                                    double u_ij_new
+                                        = m_mc->m_patch->energy(r_ij_scaled,
```
b-butler:
```suggestion
                                        = m_mc->getPatchEnergy()->energy(r_ij_scaled,
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                    // if enery has changed, there is a new soft overlap
+                                    // add the appropriate weight to the appropriate bin of the
+                                    // histogram and break out of the loop over bins
+                                    if (u_ij_new != u_ij_0)
+                                        {
+                                        min_bin_ptl_i = bin_to_sample;
+                                        hist_weight_ptl_i = 1.0 - fast::exp(-(u_ij_new - u_ij_0));
+                                        break;
+                                        }
+                                    } // end if (m_mc->m_patch)
```
b-butler:
```suggestion
                                    } // end if (m_mc->getPatchEnergy())
```

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -14,7 +14,7 @@
 def test_before_attaching():
     xmax = 0.02
     dx = 1e-4
-    sdf = hoomd.hpmc.compute.SDF(xmax=xmax, dx=dx)
+    sdf = hoomd.hpmc.compute.SDF(xmax=xmax, dx=dx, num_random_samples=0, beta=1)
```
b-butler:
```suggestion
    sdf = hoomd.hpmc.compute.SDF(xmax=xmax, dx=dx)
```
These aren't constructor arguments?

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -53,7 +53,7 @@ def test_after_attaching(valid_args, simulation_factory,
 
     xmax = 0.02
     dx = 1e-4
-    sdf = hoomd.hpmc.compute.SDF(xmax=xmax, dx=dx)
+    sdf = hoomd.hpmc.compute.SDF(xmax=xmax, dx=dx, num_random_samples=0, beta=1)
```
b-butler:
```suggestion
    sdf = hoomd.hpmc.compute.SDF(xmax=xmax, dx=dx)
```

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -181,7 +181,10 @@ def test_values(simulation_factory, lattice_snapshot_factory):
     }
     sim.operations.add(mc)
 
-    sdf = hoomd.hpmc.compute.SDF(xmax=0.02, dx=1e-4)
+    sdf = hoomd.hpmc.compute.SDF(xmax=0.02,
+                                 dx=1e-4,
+                                 num_random_samples=0,
+                                 beta=1)
```
b-butler:
```suggestion
    sdf = hoomd.hpmc.compute.SDF(xmax=0.02, dx=1e-4)
```

hoomd/hpmc/ComputeSDF.h
b-butler: Actually isn't this redundant given we only call this if `m_patch != nullptr`?

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                }     // end loop over histogram bins
+                            }
+                        }
+                    }
+                else
+                    {
+                    // skip ahead
+                    cur_node_idx += aabb_tree.getNodeSkip(cur_node_idx);
+                    }
```
b-butler, line 583: To reduce nesting place this at the top with `aabb_tree.isNodeLeaf(cur_node_idx)`

```python
if (!detail::overlap(aabb_tree.getNodeAABB(cur_node_idx), aabb))
    {
    cur_node_idx += aabb_tree.getNodeSkip(cur_node_idx);
    continue;
    }
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+            aabb.translate(pos_i_image);
+
+            // stackless search
+            for (unsigned int cur_node_idx = 0; cur_node_idx < aabb_tree.getNumNodes();
+                 cur_node_idx++)
+                {
+                if (detail::overlap(aabb_tree.getNodeAABB(cur_node_idx), aabb))
+                    {
+                    if (aabb_tree.isNodeLeaf(cur_node_idx))
+                        {
```
b-butler, line 486: Why not reduce nesting with `if (!aabb_tree.isNodeLeaf(cur_node_idx)) continue;`?

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                        for (unsigned int cur_p = 0;
+                             cur_p < aabb_tree.getNodeNumParticles(cur_node_idx);
+                             cur_p++)
+                            {
+                            // read in its position and orientation
+                            unsigned int j = aabb_tree.getNodeParticle(cur_node_idx, cur_p);
+
+                            // skip i==j in the 0 image
+                            if (cur_image == 0 && i == j)
+                                continue;
```
b-butler, line 497:
```suggestion
                                {
                                continue;
                                }

```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                                                float(h_diameter.data[i]),
+                                                                float(h_charge.data[i]),
+                                                                typ_j,
+                                                                quat<float>(orientation_j),
+                                                                float(h_diameter.data[j]),
+                                                                float(h_charge.data[j]));
+                                    // if enery has changed, there is a new soft overlap
+                                    // add the appropriate weight to the appropriate bin of the
+                                    // histogram and break out of the loop over bins
+                                    if (u_ij_new != u_ij_0)
```
b-butler, line 568: Is this the condition we want or would we want some small numerical tolerance here, such as 1e-14 or such.

b-butler (changes requested):
I have tried to leave some helpful comments regarding some refactoring. Thanks @tcmoore3 for the PR.

hoomd/hpmc/ComputeSDF.h
joaander: This is just a copy and paste from all the other places that use the AABB tree search in HPMC.

hoomd/hpmc/ComputeSDF.h
joaander: In a future pull request, this will also call the linear search for concave shapes.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -154,16 +155,19 @@ template<class Shape> class ComputeSDF : public Compute
     double m_xmax;                                   //!< Maximum lambda value
     double m_dx;                                     //!< Histogram step size
 
-    std::vector<unsigned int> m_hist; //!< Raw histogram data
-    std::vector<double> m_sdf;        //!< Computed SDF
+    std::vector<double> m_hist; //!< Raw histogram data
+    std::vector<double> m_sdf;  //!< Computed SDF
 
     Scalar m_last_max_diam; //!< Last recorded maximum diameter
+    unsigned int m_instance = 0;
```
joaander:
```suggestion
```
`m_instance` is unused.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -154,16 +155,19 @@ template<class Shape> class ComputeSDF : public Compute
...
 
     //! Zero the histogram counts
     void zeroHistogram();
 
     //! Add to histogram counts
     void countHistogram(uint64_t timestep);
+    void countHistogramBinarySearch(uint64_t timestep);
+    void countHistogramBruteForce(uint64_t timestep);
```
joaander:
```suggestion
    void countHistogramLinearSearch(uint64_t timestep);
```
Let's name this based on a more descriptive title.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -183,7 +187,8 @@ ComputeSDF<Shape>::ComputeSDF(std::shared_ptr<SystemDefinition> sysdef,
                               double dx)
     : Compute(sysdef), m_mc(mc), m_xmax(xmax), m_dx(dx)
     {
-    m_exec_conf->msg->notice(5) << "Constructing ComputeSDF: " << xmax << " " << dx << std::endl;
+    m_exec_conf->msg->notice(5) << "Constructing ComputeSDF: xmax = " << xmax << ", dx = " << dx
```
joaander: Revert this unnecessary change.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -174,7 +178,7 @@ template<class Shape> class ComputeSDF : public Compute
 
     //! Return the sdf
     virtual void computeSDF(uint64_t timestep);
-    };
+    }; // end class ComputeSDF
```
joaander: Revert this unnecessary change.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -293,15 +315,17 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+    ArrayHandle<Scalar> h_charge(m_pdata->getCharges(), access_location::host, access_mode::read);
     const std::vector<param_type, hoomd::detail::managed_allocator<param_type>>& params
         = m_mc->getParams();
 
     // loop through N particles
     for (unsigned int i = 0; i < m_pdata->getN(); i++)
         {
-        size_t min_bin = m_hist.size();
-
+        size_t min_bin_ptl_i = m_hist.size();
```
joaander: You should revert unnecessary changes to make the pull request easier to read.

hoomd/hpmc/ComputeSDF.h
joaander: `getN` is a const inlined function. The compiler should be able to optimize this away.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+        double hist_weight_ptl_i = 2.0;
+
+        // read in the current position and orientation
+        Scalar4 postype_i = h_postype.data[i];
+        Scalar4 orientation_i = h_orientation.data[i];
+        int typ_i = __scalar_as_int(postype_i.w);
+        Shape shape_i(quat<Scalar>(orientation_i), params[__scalar_as_int(postype_i.w)]);
+        vec3<Scalar> pos_i = vec3<Scalar>(postype_i);
+
+        // TODO: do we need to account for patch r_cut in here somewhere?
```
joaander: Yes, you need to account for the patch r_cut. If you don't, you will miss potentially interacting pairs of particles.
See `IntegratorHPMC` for details on how to include the r_cut into the AABB tree `R_query`.

hoomd/hpmc/ComputeSDF.h
joaander: This is just a copy and paste of the same loop used elsewhere in HPMC. You could open a code style issue/pr and solve them all at the same time if you like. Test to ensure that performance remains the same.

hoomd/hpmc/ComputeSDF.h
joaander: A future PR will use this method for concave hard particles (with or without patches).

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +390,184 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                                               float(h_charge.data[j]));
+                                }
+                            // loop over bins; only going up to min_bin_ptl_i-1 since we only
+                            // want the _first_ overlap of particle i with its neighbors
+                            for (size_t bin_to_sample = 0; bin_to_sample < min_bin_ptl_i;
+                                 bin_to_sample++)
+                                {
+                                double scale_factor = m_dx * static_cast<double>(bin_to_sample + 1);
+
+                                // TODO: think about order of overlap checks: there may be a
```
joaander: Have you tested this? Which way performs best?

Remove the TODO once you have chosen which order to evaluate the interactions.

hoomd/hpmc/ComputeSDF.h
joaander: This method is valid for patch energies with step functions. So yes, this is the right condition.

hoomd/hpmc/IntegratorHPMC.h
joaander: Yes.

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -14,7 +14,7 @@
 def test_before_attaching():
```
joaander, line 16: Add a unit test to ensure that the linear search code path executes and computes the correct values.

joaander (changes requested):
I focused on the code in this review. Once that is improved, I will review the documentation.

hoomd/hpmc/ComputeSDF.h
tcmoore3: They are used for the patch energy calculation.

b-butler: They are but not in this function that I can see.

joaander:
@tcmoore3 Let us know if you have any questions.

hoomd/hpmc/ComputeSDF.h
tcmoore3: You're right. That's left over from my initial implementation where I didn't put the linear search in a separate method.

hoomd/hpmc/pytest/test_compute_sdf.py
tcmoore3: Correct.

hoomd/hpmc/ComputeSDF.h
tcmoore3: Done.

hoomd/hpmc/ComputeSDF.h
tcmoore3: Done.

hoomd/hpmc/ComputeSDF.h
tcmoore3: I have not tested this, but I think we do need to keep this order of overlap checks. This order ensures that if the scaling value results in both a new hard and soft overlap that the hard overlap (i.e., the one that dominates the Boltzmann weight) is the one that gets counted for the histogram. This scenario could may arise for systems with very short-ranged patchy interactions that are shorter-ranged than the compute's `dx`, but in that case the user should select SDF parameters that are more suited for the specific system.

Is this something worth looking more into?

joaander: It is the user's responsibility to choose parameters.

Whichever order you choose to evaluate the overlaps, you should compute the same result. If there is a performance benefit to evaluating soft overlaps first, then you can use conditionals to add the correct value to the histogram.

In either case, I don't expect a performance benefit. In most loop iterations, you need to check for both hard and soft overlaps. Optimizations that evaluate only one overlap in the last loop iteration don't change that.

tcmoore3:
This is ready for another set of reviews. It especially needs a careful review of whether I correctly account for the patch cutoff when building the AABB around particle `i` in the AABB tree search.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,23 +267,40 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
 
     countHistogram() loops through all particle pairs *i,j* where *i* is on the local rank, computes
-    the bin in which that pair should be and adds 1 to the bin. countHistogram() can be called
-    multiple times to increment the counters for averaging, and it operates without any
-    communication.
+    the bin corresponding to the scale factor of the vector r_ij that induces the first overlap
+    between particles *i* and *j*, and then adds s_ij to the bin, where s_ij = 1 - e^{\beta\Delta
+   U_{ij}}. Note that this definition allows for the consideration of soft overlaps where the pair
+   potential changes value, and reduces 1 for hard particle overlaps.  countHistogram() operates
+   without any communication.
```
joaander:
```suggestion
    U_{ij}}. Note that this definition allows for the consideration of soft overlaps where the pair
    potential changes value, and reduces to 1 for hard particle overlaps.  countHistogram() operates
    without any communication.
```
Indent with 4 spaces.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,23 +267,40 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
       - The integrator performs the ghost exchange (with the ghost width extra that we add)
-      - Only on writeOutput() do we need to sum the per-rank histograms into a global histogram
+
+    This function is a wrapper that calls the appropriate method depending on the value of
+   m_mc->getPatchEnergy()
```
joaander:
```suggestion
    m_mc->getPatchEnergy()
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +384,181 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+        Scalar4 orientation_i = h_orientation.data[i];
+        int typ_i = __scalar_as_int(postype_i.w);
+        Shape shape_i(quat<Scalar>(orientation_i), params[__scalar_as_int(postype_i.w)]);
+        vec3<Scalar> pos_i = vec3<Scalar>(postype_i);
 
+        // construct the AABB around the particle's circumsphere
+        // pad with enough extra width so that when scaled by xmax, found particles might touch
+        hoomd::detail::AABB aabb_i_local(vec3<Scalar>(0, 0, 0),
+                                         shape_i.getCircumsphereDiameter() / Scalar(2)
+                                             + extra_width);
```
joaander: This is not correct. We should have a real time discussion with a white board to determine the modifications needed.

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -201,6 +203,58 @@ def test_values(simulation_factory, lattice_snapshot_factory):
...
+                                              code=square_well,
+                                              param_array=[])
+    mc.pair_potential = patch
+
+    # assert we have an overlap in the first bin with the expected weight and
+    # that the SDF is zero everywhere else
+    sim.run(1)
+    neg_mayerF = 1 - np.exp(epsilon)
+    assert (np.count_nonzero(sdf.sdf) == 1)
+    assert (sdf.sdf[0] == neg_mayerF * norm_factor)
```
joaander: These checks fail in MPI. Per the documentation, the `sdf` array is only available on rank 0.

hoomd/hpmc/compute.py
```diff
@@ -222,8 +236,10 @@ class SDF(Compute):
     smaller ``xmax``. Check that :math:`\sum_k s(x_k) \cdot dx \approx 0.5`.
 
     Warning:
-        `SDF` does not compute correct pressures for simulations with
-        concave particles or enthalpic interactions.
+        `SDF` only considers negative volume perturbations, and therefore may
+        not compute the correct pressure in simulations where positive volume
+        perturbations may change the system's potential energy, e.g., systems of
+        concave particles or with enthalpic interactions.
```
joaander:
```suggestion
        concave particles or with non-monotonic enthalpic interactions.
```

hoomd/hpmc/compute.py
```diff
@@ -222,8 +236,10 @@ class SDF(Compute):
     smaller ``xmax``. Check that :math:`\sum_k s(x_k) \cdot dx \approx 0.5`.
 
     Warning:
-        `SDF` does not compute correct pressures for simulations with
-        concave particles or enthalpic interactions.
+        `SDF` only considers negative volume perturbations, and therefore may
```
joaander:
```suggestion
        `SDF` only considers negative volume perturbations, and therefore does
```

hoomd/hpmc/compute.py
```diff
@@ -161,15 +161,25 @@ class SDF(Compute):
...
+                S_j(\mathbf{q}_j, (1-x)(\vec{r}_j - (\vec{r}_i + \vec{A})))
+            \right) \ne \emptyset
+            \\
+            \lor U_{\mathrm{pair},ij}((1-x)(\vec{r}_j - (\vec{r}_i + \vec{A})),
+                                 \mathbf{q}_i,
+                                 \mathbf{q}_j)
+                \ne
+            U_{\mathrm{pair},ij}(\vec{r}_j - (\vec{r}_i + \vec{A}),
+                                 \mathbf{q}_i,
+                                 \mathbf{q}_j) \}
```
joaander: This run-on equation is hard to read. I suggest spacing it onto more lines and using alignment so the terms of the "or" operation line up. See for example, the U_pair and U_shape equations at https://hoomd-blue.readthedocs.io/en/v3.3.0/module-hpmc-integrate.html

joaander (changes requested):
Thanks! This is much cleaner. There are some additional small changes needed. Let's meet in real time to discuss the AABB query modifications.

hoomd/hpmc/compute.py
tcmoore3: I broke it into several lines now. It's definitely easier to read, but perhaps takes up more vertical space than needed. Feel free to change.

hoomd/hpmc/pytest/test_compute_sdf.py
tcmoore3: Fixed.

hoomd/hpmc/ComputeSDF.h
tcmoore3: I implemented the `getMaxInteractionDiameter()` method we discussed. I _think_ it does what it should now, but a close second look would be good.

hoomd/hpmc/pytest/test_compute_sdf.py
tcmoore3: Done

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,27 +269,64 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
         {
-        m_hist[i] = 0;
+        m_hist[i] = 0.0;
         }
     }
 
+template<class Shape> Scalar ComputeSDF<Shape>::getMaxInteractionDiameter()
+    {
+    Scalar max_core_diameter = m_mc->getMaxCoreDiameter();
+    Scalar max_r_cut_patch(0.0);
```
joaander:
```suggestion
    Scalar max_r_cut_patch = 0.0;
```
We don't normally use constructors when initializing basic types.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,27 +269,64 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
+
+    if (m_mc->getPatchEnergy())
+        {
+        const std::vector<param_type, hoomd::detail::managed_allocator<param_type>>& params
+            = m_mc->getParams();
```
joaander:
```suggestion
        const auto& params = m_mc->getParams();
```
Let the compiler do the work.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,27 +269,64 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
+        for (unsigned int typ_i = 0; typ_i < m_pdata->getNTypes(); typ_i++)
+            {
+            Shape shape_i(quat<Scalar>(), params[typ_i]);
```
joaander:
```suggestion
```
Unused.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,27 +269,64 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
+template<class Shape> void ComputeSDF<Shape>::countHistogramBinarySearch(uint64_t timestep)
     {
     // update the aabb tree
     const hoomd::detail::AABBTree& aabb_tree = m_mc->buildAABBTree();
+
     // update the image list
     const std::vector<vec3<Scalar>>& image_list = m_mc->updateImageList();
 
-    Scalar extra_width = m_xmax / (1 - m_xmax) * m_mc->getMaxCoreDiameter();
+    Scalar extra_width = m_xmax / (1 - m_xmax) * this->getMaxInteractionDiameter();
```
joaander: We could save a few cycles by making `max_diam` a member variable (`m_max_diam`) and reusing the computation here.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +406,180 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+
+template<class Shape> void ComputeSDF<Shape>::countHistogramLinearSearch(uint64_t timestep)
+    {
+    // update the aabb tree
+    const hoomd::detail::AABBTree& aabb_tree = m_mc->buildAABBTree();
+
+    // update the image list
+    const std::vector<vec3<Scalar>>& image_list = m_mc->updateImageList();
+
+    Scalar extra_width = m_xmax / (1 - m_xmax) * this->getMaxInteractionDiameter();
```
joaander: We could save a few cycles by making `max_diam` a member variable (`m_max_diam`) and reusing the computation here.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +406,180 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                   access_mode::read);
+    ArrayHandle<Scalar4> h_orientation(m_pdata->getOrientationArray(),
+                                       access_location::host,
+                                       access_mode::read);
+    ArrayHandle<Scalar> h_diameter(m_pdata->getDiameters(),
+                                   access_location::host,
+                                   access_mode::read);
+    ArrayHandle<Scalar> h_charge(m_pdata->getCharges(), access_location::host, access_mode::read);
+    const std::vector<param_type, hoomd::detail::managed_allocator<param_type>>& params
+        = m_mc->getParams();
```
joaander:
```suggestion
    const auto& params = m_mc->getParams();
```

hoomd/hpmc/ComputeSDF.h
joaander: You still need to account for the patch `r_cut` and additive cutoff in the query AABB.

From `IntegratorHPMCMono.h` (with `extra_width`) added.
```
            OverlapReal r_cut_patch = 0;
            if (m_patch)
                {
                r_cut_patch = static_cast<OverlapReal>(m_patch->getRCut()) +
                    static_cast<OverlapReal>(0.5) * static_cast<OverlapReal>(m_patch->getAdditiveCutoff(typ_i));
                }


            // subtract minimum AABB extent from search radius
            OverlapReal R_query = std::max(shape_i.getCircumsphereDiameter()/OverlapReal(2.0),
                r_cut_patch-getMinCoreDiameter()/(OverlapReal)2.0);
            hoomd::detail::AABB aabb_i_local = hoomd::detail::AABB(vec3<Scalar>(0,0,0),R_query + extra_width);
```

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -746,8 +746,8 @@ void IntegratorHPMCMono<Shape>::update(uint64_t timestep)
                 {
                 r_cut_patch = static_cast<OverlapReal>(m_patch->getRCut()) + static_cast<OverlapReal>(0.5) *
                     static_cast<OverlapReal>(m_patch->getAdditiveCutoff(typ_i));
-                r_cut_patch = static_cast<OverlapReal>(m_patch->getRCut()) +
-                    static_cast<OverlapReal>(0.5) * static_cast<OverlapReal>(m_patch->getAdditiveCutoff(typ_i));
+                r_cut_patch = static_cast<OverlapReal>(m_patch->getRCut()) + static_cast<OverlapReal>(0.5) *
+                    static_cast<OverlapReal>(m_patch->getAdditiveCutoff(typ_i));
```
joaander:
```suggestion
```
We only need to compute `r_cut_patch` once.

hoomd/hpmc/compute.py
```diff
@@ -161,15 +161,27 @@ class SDF(Compute):
 
     For each pair of particles :math:`i` and :math:`j` `SDF` scales the particle
     separation vector by the factor :math:`(1-x)` and finds the smallest
-    positive value of :math:`x` leading to an overlap of the particle shapes:
+    positive value of :math:`x` leading to either an overlap of the particle
+    shapes (a "hard overlap") or a change in the pair energy
+    :math:`U_{\mathrm{pair},ij}` (a "soft overlap"):
```
joaander:
```suggestion
    shapes (a "hard overlap") or a discontinuous change in the pair energy
    :math:`U_{\mathrm{pair},ij}` (a "soft overlap"):
```

hoomd/hpmc/compute.py
```diff
@@ -222,8 +238,10 @@ class SDF(Compute):
     smaller ``xmax``. Check that :math:`\sum_k s(x_k) \cdot dx \approx 0.5`.
 
     Warning:
-        `SDF` does not compute correct pressures for simulations with
-        concave particles or enthalpic interactions.
+        `SDF` only considers negative volume perturbations, and therefore does
+        not compute the correct pressure in simulations where positive volume
+        perturbations may change the system's potential energy, e.g., systems of
+        concave particles or with non-monotonic enthalpic interactions.
```
joaander, line 244:
```suggestion
        concave particles or with non-monotonic enthalpic interactions.
```
I suggested the word non-monotonic but now realize this isn't general enough. "Concave" interactions would also need expansion moves (consider a multi-site patch model). What word would better describe this?

We also need a second warning box that the pressure is correct only for potentials with constant values and step discontinuities (like square well and Kern-Frenkel).

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,27 +269,64 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
+
+    if (m_mc->getPatchEnergy())
+        {
+        const std::vector<param_type, hoomd::detail::managed_allocator<param_type>>& params
+            = m_mc->getParams();
+        for (unsigned int typ_i = 0; typ_i < m_pdata->getNTypes(); typ_i++)
+            {
+            Shape shape_i(quat<Scalar>(), params[typ_i]);
+            Scalar r_cut_patch_i = m_mc->getPatchEnergy()->getRCut()
+                                   + 0.5 * m_mc->getPatchEnergy()->getAdditiveCutoff(typ_i);
```
joaander:
```suggestion
                                   + m_mc->getPatchEnergy()->getAdditiveCutoff(typ_i);
```
This method computes a diameter, not a radius. It needs to find the largest distance between two particles that would be interacting. The largest distance is the correct value to use when computing the system wide extra ghost width (see the call to `setExtraGhostWidth`). It is an over-conservative estimate when computing `extra_width` as we could use the known additive cutoff for type `i` to reduce the distance somewhat. *However*, I don't think that is necessary as it makes the code even more complex, and it is not clear how much performance benefit there would be. Let's make a note of it and can revisit if anyone ever comes to me with an additive cutoff system with the SDF performance as the main bottleneck.

hoomd/hpmc/ComputeSDF.h
joaander:
```suggestion
    // Note - If needed for future simulations with a large disparity in additive cutoffs, compute
    // extra_width_i with knowledge of the additive cutoff of type i and half the largest additive
    // cutoff.
    Scalar extra_width = m_xmax / (1 - m_xmax) * this->getMaxInteractionDiameter();
```

joaander (changes requested):
Thanks! I have some suggestions to improve style and there are a few changes needed to find all the potentially interacting neighbors.

hoomd/hpmc/IntegratorHPMCMono.h
tcmoore3: I thought I had removed this repeated line in a previous commit, but I must have not actually removed it.

Ah, [this commit](https://github.com/glotzerlab/hoomd-blue/commit/3eca50ef8e29cd472f212e3947871545d7b36a84) is what I was thinking of, but I guess I just reformatted that line in that commit.

hoomd/hpmc/compute.py
tcmoore3: I think "non-monotonic" still covers "concave" interactions, if I'm picturing the what you mean by concave correctly. Consider a concave patchy interaction: the pair energy could change from 0 to `u_ij ≠ 0` back to 0 as the pair separation changes. So even if `u_ij` is constant when the particles are interacting, the pair energy is a non-monotonic function of the particle separation.

joaander: "monotonic" applies to an expansion or compression along a single direction. What word would better indicate that U_ij is monotonic in all directions?

tcmoore3: Spherically symmetric?

joaander: That's too restrictive. Maybe "non-monotonic" is close enough. We'll need to revisit this with the next PR that adds expansion perturbations.

joaander (approved):
Thanks!
@b-butler please review.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,27 +269,61 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
+template<class Shape> Scalar ComputeSDF<Shape>::getMaxInteractionDiameter()
+    {
+    Scalar max_core_diameter = m_mc->getMaxCoreDiameter();
+    Scalar max_r_cut_patch = 0.0;
+
+    if (m_mc->getPatchEnergy())
+        {
+        for (unsigned int typ_i = 0; typ_i < m_pdata->getNTypes(); typ_i++)
+            {
+            Scalar r_cut_patch_i = m_mc->getPatchEnergy()->getRCut()
```
b-butler:
```suggestion
            const Scalar r_cut_patch_i = m_mc->getPatchEnergy()->getRCut()
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -264,27 +269,61 @@ template<class Shape> void ComputeSDF<Shape>::zeroHistogram()
...
     for (size_t i = 0; i < m_hist.size(); i++)
         {
-        m_hist[i] = 0;
+        m_hist[i] = 0.0;
         }
     }
 
+template<class Shape> Scalar ComputeSDF<Shape>::getMaxInteractionDiameter()
+    {
+    Scalar max_core_diameter = m_mc->getMaxCoreDiameter();
```
b-butler:
```suggestion
    const Scalar max_core_diameter = m_mc->getMaxCoreDiameter();
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +403,189 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+        {
+        size_t min_bin = m_hist.size();
+        double hist_weight_ptl_i = 2.0;
+
+        // read in the current position and orientation
+        Scalar4 postype_i = h_postype.data[i];
+        Scalar4 orientation_i = h_orientation.data[i];
+        int typ_i = __scalar_as_int(postype_i.w);
+        Shape shape_i(quat<Scalar>(orientation_i), params[__scalar_as_int(postype_i.w)]);
+        vec3<Scalar> pos_i = vec3<Scalar>(postype_i);
```
b-butler:
```suggestion
        const Scalar4 postype_i = h_postype.data[i];
        const Scalar4 orientation_i = h_orientation.data[i];
        const int typ_i = __scalar_as_int(postype_i.w);
        const Shape shape_i(quat<Scalar>(orientation_i), params[__scalar_as_int(postype_i.w)]);
        const vec3<Scalar> pos_i = vec3<Scalar>(postype_i);
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +403,189 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+        // construct the AABB around the particle's circumsphere
+        // pad with enough extra width so that when scaled by xmax, found particles might touch
+        OverlapReal r_cut_patch = 0;
+        if (m_mc->getPatchEnergy())
+            {
+            r_cut_patch = static_cast<OverlapReal>(
+                m_mc->getPatchEnergy()->getRCut()
+                + 0.5 * m_mc->getPatchEnergy()->getAdditiveCutoff(typ_i));
+            }
+        OverlapReal R_query = std::max(shape_i.getCircumsphereDiameter() / OverlapReal(2.0),
```
b-butler:
```suggestion
        const OverlapReal R_query = std::max(shape_i.getCircumsphereDiameter() / OverlapReal(2.0),
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +403,189 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                       r_cut_patch - m_mc->getMinCoreDiameter() / (OverlapReal)2.0);
+        hoomd::detail::AABB aabb_i_local(vec3<Scalar>(0, 0, 0), R_query + extra_width);
+
+        size_t n_images = image_list.size();
```
b-butler:
```suggestion
        const size_t n_images = image_list.size();
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +403,189 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                {
+                                continue;
+                                }
+
+                            Scalar4 postype_j = h_postype.data[j];
+                            Scalar4 orientation_j = h_orientation.data[j];
+                            int typ_j = __scalar_as_int(postype_j.w);
+
+                            // put particles in coordinate system of particle i
+                            vec3<Scalar> r_ij = vec3<Scalar>(postype_j) - pos_i_image;
```
b-butler:
```suggestion
                            const Scalar4 postype_j = h_postype.data[j];
                            const Scalar4 orientation_j = h_orientation.data[j];
                            const int typ_j = __scalar_as_int(postype_j.w);

                            // put particles in coordinate system of particle i
                            const vec3<Scalar> r_ij = vec3<Scalar>(postype_j) - pos_i_image;
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +403,189 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                    break;
+                                    } // end if overlap
+
+                                // if no hard overlap, check for a soft overlap if we have
+                                // patches
+                                if (m_mc->getPatchEnergy())
+                                    {
+                                    // compute the energy at this size of the perturbation and
+                                    // compare to the energy in the unperturbed state
+                                    vec3<Scalar> r_ij_scaled = r_ij * (Scalar(1.0) - scale_factor);
```
b-butler:
```suggestion
                                    const vec3<Scalar> r_ij_scaled = r_ij * (Scalar(1.0) - scale_factor);
```

hoomd/hpmc/ComputeSDF.h
```diff
@@ -364,13 +403,189 @@ template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
...
+                                    typ_j,
+                                    quat<float>(orientation_j),
+                                    float(h_diameter.data[j]),
+                                    float(h_charge.data[j]));
+                                }
+                            // loop over bins; only going up to min_bin-1 since we only
+                            // want the _first_ overlap of particle i with its neighbors
+                            for (size_t bin_to_sample = 0; bin_to_sample < min_bin; bin_to_sample++)
+                                {
+                                double scale_factor = m_dx * static_cast<double>(bin_to_sample + 1);
```
b-butler:
```suggestion
                                const double scale_factor = m_dx * static_cast<double>(bin_to_sample + 1);
```

b-butler (commented):
Looks good. I littered the code with many `const` suggestions, but other than that I think this is good to go.

b-butler (approved):

tcmoore3:
I addressed Brandon's last round of suggestions, so I think this is good to go now.

joaander:
Thanks!

PR #1362, opened 2022-07-14
Feature/notice API

### LINKED ISSUE #1329: User-facing API to print notice messages.
## Description

Add a user-facing API for `Messenger.notice`.

## Proposed solution

`device.notice(3, 'Message')`
`device.print('Message')`

## Additional context

Unlike the Python I/O functions, `Device.print` would be MPI-aware. Code like this:
```
tps = sim.tps
if device.communicator.rank == 0:
    print(tps)
```

Would turn into:
```
device.print(sim.tps)
```

This will also allow users running large partitioned simulations to easily direct output to the partition specific `msg_file`: https://hoomd-blue.readthedocs.io/en/v3.1.0/module-hoomd-device.html#hoomd.device.Device.msg_file

## Developer

Seeking a Python export who can make `print` have a useful Pythonic API.

kodyt:
## Description

Adds a print API for `Messenger.notice` to `hoomd.device` denoted as `device.notice("Message")`.

## Motivation and context
Easily allows users to output information to a directed `msg_file` or stdout.
Resolves #1329.

## How has this been tested?

I added a unit test to check whether `"Message"` is properly outputted to a `msg_file` when explicitly declared.

## Change log

```
Added: An user-facing API for `Messenger.notice` 
```

## Checklist:

- [X] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/CONTRIBUTING.md).

### DIFF
--- a/hoomd/device.py
+++ b/hoomd/device.py
@@ -28,7 +28,10 @@ from hoomd import _hoomd
 class Device:
     """Base class device object.
 
-    Provides methods and properties common to `CPU` and `GPU`.
+    Provides methods and properties common to `CPU` and `GPU`, including those
+    that control where status messages are stored (`msg_file`) how many status
+    messages HOOMD-blue prints (`notice_level`) and a method for user provided
+    status messages (`notice`).
 
     Warning:
         `Device` cannot be used directly. Instantate a `CPU` or `GPU` object.
--- a/hoomd/device.py
+++ b/hoomd/device.py
@@ -140,6 +140,15 @@ class Device:
         else:
             self._cpp_exec_conf.setNumThreads(int(num_cpu_threads))
 
+    #Want to add a function here for print
+    def notice_print(self, mpi_config, notice_level, msg_file):
+        msg = _hoomd.Messenger(mpi_config)
+        if not hoomd.version.tbb_enabled:
+            msg.notice(notice_level)
+        
+        # This will open the given msg_file if not None
+        if msg_file is not None:
+            msg.openFile(msg_file)
 
 def _create_messenger(mpi_config, notice_level, msg_file):
     msg = _hoomd.Messenger(mpi_config)
--- a/hoomd/device.py
+++ b/hoomd/device.py
@@ -144,7 +144,9 @@ class Device:
     def notice_print(self, mpi_config, notice_level, msg_file):
         msg = _hoomd.Messenger(mpi_config)
         if not hoomd.version.tbb_enabled:
-            msg.notice(notice_level)
+            self._cpp_msg.notice(
+                "Input a message here"
+            )
         
         # This will open the given msg_file if not None
         if msg_file is not None:
--- a/hoomd/device.py
+++ b/hoomd/device.py
@@ -146,11 +146,12 @@ class Device:
         if notice_level > self.notice_level:
             return
         else:
-            return self._cpp_msg.noticeStr(notice_level, str(msg) + "\n")
+            return self._cpp_msg.notice(notice_level, str(msg) + "\n")
 
     def print(self, msg):
-        """Easily directed to msg_file"""
-        return self._cpp_msg.notice(str(msg) + "\n")
+        """Easily directed to msg_file regardless of notice level"""
+        return self._cpp_msg.notice(5, str(msg) + "\n")
+        # Only works if notice_level is > 3?
 
 
 def _create_messenger(mpi_config, notice_level, msg_file):
--- a/hoomd/device.py
+++ b/hoomd/device.py
@@ -140,17 +140,12 @@ class Device:
         else:
             self._cpp_exec_conf.setNumThreads(int(num_cpu_threads))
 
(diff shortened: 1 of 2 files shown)


hoomd/device.py
```diff
@@ -140,6 +140,11 @@ def num_cpu_threads(self, num_cpu_threads):
         else:
             self._cpp_exec_conf.setNumThreads(int(num_cpu_threads))
 
+    def notice(self, msg, notice_level=1):
+        """If msg_file is specified then msg will output in msg_file,
+           else msg will output to stdout"""
```
b-butler:
```suggestion
        """If msg_file is specified then msg will output in msg_file, \
           else msg will output to stdout"""
```

hoomd/pytest/test_device.py
```diff
@@ -108,3 +108,26 @@ def test_cpu_build_specifics():
...
+def test_device_notice(device, tmp_path):
+    # Message file declared. Should output in specified file
+    device.notice_level = 4
+    device.msg_file = str(tmp_path / "example.txt")
+    device.notice("This message should output.")
+    file = open(device.msg_file)
+    # Check the msg file if the output is correctly placed
+    assert file.read() == "This message should output.\n"
+
+    file.close()
```
b-butler:
```suggestion
    msg = "This message should output."
    device.notice(msg)
    with open(device.msg_file) as fh:
        # Check the msg file if the output is correctly placed
        assert fh.read() == msg + "\n"
```

hoomd/pytest/test_device.py
```diff
@@ -108,3 +108,26 @@ def test_cpu_build_specifics():
...
+
+    snapshot = gsd.hoomd.Snapshot()
+    snapshot.particles.N = 4
+
```
b-butler:
```suggestion
```

hoomd/pytest/test_device.py
```diff
@@ -108,3 +108,26 @@ def test_cpu_build_specifics():
...
+    device.msg_file = str(tmp_path/"example2.txt")
+    # Test notice with a message that is not a string
+    device.notice(snapshot.particles.N)
+    file = open(device.msg_file)
+    assert file.read() == "4\n"
+
+    file.close()
```
b-butler:
```suggestion
    device.msg_file = str(tmp_path/"example2.txt")
    # Test notice with a message that is not a string
    msg = 123456
    device.notice(msg)
    with open(device.msg_file) as fh:
        assert fh.read() == str(msg) + "\n"
```

b-butler (changes requested):
I made some changes to the test code and documentation.

We should test that creating a message with too low a notice_level (a larger value is lower priority), does not print out.

hoomd/pytest/test_device.py
```diff
@@ -3,7 +3,7 @@
 
 import hoomd
 import pytest
-
+import gsd.hoomd
```
joaander:
```suggestion
```

hoomd/device.py
joaander: I will revise this documentation further.

joaander:
Any thoughts @cbkerr ?

b-butler (approved):
Thanks @joaander for the fixes to the PR.

hoomd/pytest/test_device.py
```diff
@@ -108,3 +108,33 @@ def test_cpu_build_specifics():
...
+def test_device_notice(device, tmp_path):
+    # Message file declared. Should output in specified file.
+    device.notice_level = 4
+    device.msg_file = str(tmp_path / "str_message")
+    msg = "This message should output."
+    device.notice(msg)
+
+    if device.communicator.rank == 0:
+        with open(device.msg_file) as fh:
+            assert fh.read() == msg + "\n"
```
cbkerr, line 122: My only concern is different end of file conventions on different platforms. Would that sometimes cause the test to fail if it doesn't end with a newline?

cbkerr (approved):
I would like to see a short paragraph added to the top of the Device class saying that the user may write their own notices using this function. This would make it more discoverable. Otherwise looks good!

hoomd/pytest/test_device.py
joaander: This is not a problem. See https://docs.python.org/3/library/functions.html#open
> When reading input from the stream, if newline is None, universal newlines mode is enabled. Lines in the input can end in '\n', '\r', or '\r\n', and these are translated into '\n' before being returned to the caller. If it is '', universal newlines mode is enabled, but line endings are returned to the caller untranslated. If it has any of the other legal values, input lines are only terminated by the given string, and the line ending is returned to the caller untranslated.

joaander:
> I would like to see a short paragraph added to the top of the Device class saying that the user may write their own notices using this function. This would make it more discoverable. Otherwise looks good!

Thanks for the review. I added some text mentioning the messaging system to the `Device` class docstring.

PR #1376, opened 2022-08-10
Add needed thrust includes.

joaander:
## Description

Add needed thrust includes.

## Motivation and context

This fixes compile errors with the next CUDA release.

## How has this been tested?

Peng Wang at NVIDIA tested these changes with an unspecified newer version of CUDA.

## Checklist:

### DIFF
--- a/hoomd/BondedGroupData.cu
+++ b/hoomd/BondedGroupData.cu
@@ -9,8 +9,9 @@
 #include <hip/hip_runtime.h>
 #include <thrust/device_ptr.h>
 #include <thrust/execution_policy.h>
-#include <thrust/sort.h>
 #include <thrust/iterator/constant_iterator.h>
+#include <thrust/scan.h>
+#include <thrust/sort.h>
 #pragma GCC diagnostic pop
 
 /*! \file BondedGroupData.cu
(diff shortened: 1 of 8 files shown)


bdice:
@joaander Thrust 1.16 (released a while ago but not yet in a CUDA toolkit release) removed some internal header inclusions that some applications had implicitly relied on (as transitive includes). You can probably build with Thrust 1.16 or 1.17 from GitHub and see the errors.

tommy-waltmann (approved):

bdice:
@joaander I ran a "thrust include finder" script that I was using for another project, which located a few more missing includes. I will push those to this PR. It's not a large changeset but should make this a little more complete.

bdice (approved):
Approving -- feel free to accept or reject the commit I added here. The includes I added may not be needed to compile because of transitive includes in Thrust's internals, but that's not desirable and should be replaced with explicit includes (include what you use). My additional commit makes the includes explicit for all the used Thrust features.

joaander:
> Approving -- feel free to accept or reject the commit I added here. The includes I added may not be needed to compile because of transitive includes in Thrust's internals, but that's not desirable and should be replaced with explicit includes (include what you use). My additional commit makes the includes explicit for all the used Thrust features.

Thanks for the additional includes. Those will help future proof this code against compiler errors with new thrust releases.

PR #1377, opened 2022-08-11
Raise an error when there are duplicate types.

### LINKED ISSUE #1368: HOOMD fails to error when users define multiple types of the same name
## Description

HOOMD fails to error when users define multiple types of the same name.

## Script

```python
snapshot.particles.types = ['A', 'B', 'A']
```

## Developer

On initializing from a snapshot, check for multiple types of the same name and raise an error.

joaander:
## Description

Raise an error when there are duplicate types.

## Motivation and context

When the user provides duplicate types, for example ['A', 'B', 'A'], the first shadows the 2nd. Users can create particles with typeid 2, but cannot set type parameters for 2 as the 'A' lookup will return typeid 0.

This fix gives the user a more understandable error.

Resolves #1368

## How has this been tested?

Added unit tests.

## Change log

```
Changed:

* Raise an error when initializing with duplicate types.
```

## Checklist:

### DIFF
--- a/hoomd/BondedGroupData.cc
+++ b/hoomd/BondedGroupData.cc
@@ -230,11 +230,9 @@ void BondedGroupData<group_size, Group, name, has_type_mapping>::initializeFromS
     const Snapshot& snapshot)
     {
     // check that all fields in the snapshot have correct length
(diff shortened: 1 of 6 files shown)


b-butler (approved):
Clean.

PR #1378, opened 2022-08-12
Update gpu memory management.

### LINKED ISSUE #1272: Remove exceptions from allocators/deallocators as appropriate
## Description

Check the C++ spec and determine whether allocators or deallocators (e.g. in `managed_allocator.h`) are allowed to throw exceptions. Then remove the exceptions from those that aren't allowed.

## Motivation and context

The deallocators get called during teardown. When the deallocator throws an exception, it confuses the teardown process to the point that is is nearly impossible to determine the original exception.

joaander:
## Description

* Throw `std::bad_alloc` when memory allocations fail.
* Remove CUDA error checks from destructors.
* Remove `MemoryTraceback` and deprecate `Device.memory_traceback`.

## Motivation and context

* `std::bad_alloc` will automatically translate to `MemoryError` in Python: https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#built-in-c-to-python-exception-translation
* Bad things happen when destructors throw exceptions: https://stackoverflow.com/questions/41174167/c-exception-in-destructor. I recently spent many hours debugging an allocation problem. The exceptions thrown in the destructors during the stack unwinding masked the original error.
* `MemoryTraceback` is currently not accessible to the user and provides limited to no information to developers. I considered expanding it to be more useful, but it really isn't feasible to tie in to **all** memory allocations in HOOMD. Developers would be better to use a dedicated external debugging tool than invest time in developing a poor replacement.

I likely missed some `CHECK_CUDA_ERROR` calls in destructors. I can't think of a good way to grep for them.

Resolves #1272

## How has this been tested?

I ran a local test that configured a MD simulation with a massive box and a cell list and got:
```
$ python test.py
  File "/home/joaander/test/test-memory/test.py", line 43, in <module>
    sim.run(0)
  File "/home/joaander/build/hoomd/hoomd/simulation.py", line 462, in run
MemoryError: std::bad_alloc
```

## Change log

```
Fixed:

* Memory allocation errors in C++ now result in `MemoryError` exceptions in Python.

Deprecated:

* ``Device.memory_traceback`` attribute. This attribute has no effect.
```

## Checklist:


tommy-waltmann (approved):
The changes look good. Other than the brute force method of grepping for `CHECK_CUDA_ERROR` everywhere and then seeing if each one is in a destructor manually, I don't have any ideas. Maybe an IDE has fancier grep tools?

joaander:
> The changes look good. Other than the brute force method of grepping for `CHECK_CUDA_ERROR` everywhere and then seeing if each one is in a destructor manually, I don't have any ideas. Maybe an IDE has fancier grep tools?

I found one in the `Autotuner` destructor that showed up in testing. Just pushed that an a few other changes I found I missed in the original PR submission.

SchoeniPhlippsn (approved):
Looks good to me!

PR #1382, opened 2022-08-16
Refactor updater plugin example to work with version 3, add example MD potential pair

### LINKED ISSUE #1161: Add example of MD potential plugin
## Description

While the example plugin included in this repository is helpful for users to get started writing their own plugins, it is only one example and it only showcases how to implement an Updater. Another common type of plugin that users will want to implement is an MD potential. This could be a pair potential already implemented in the `hoomd.md.pair` module (or not); the point is more to just give users the skeleton of how to construct these types of plugins without having to delve into the source code.

## Proposed solution

- Change the name of "example_plugin" to "example_plugin**s**" in the root of the project
- Change all other references of  "example_plugin" to "updater_plugin"
- Add a new module "pair_plugin" to "example_plugins". This could look like (or just be) what I've already written [here](https://github.com/ianrgraham/mlj-plugin), which is (sort of) the hoomd implementation of [this](https://docs.lammps.org/pair_lj_expand.html) pair from lammps (slightly different from [SLJ](https://hoomd-blue.readthedocs.io/en/latest/module-md-pair.html#hoomd.md.pair.SLJ) in that it's not tied to the particle diameter member)

## Additional context

NA

## Developer

I'd be happy to make the changes.

ianrgraham:
## Description

Refactored the `example_plugin` folder to `example_plugins` and added an additional `pair_plugin` implementing a simple harmonic repulsion. The updater plugin from before was refactored to work with v3 code. It appeared to still have old v2 code being called from `update.py`.

## Motivation and context

The example pair demonstrates for users how to extend the PairPotential force compute as a plugin. 

Resolves #1161

## How has this been tested?

Compiled with ENABLE_GPU, executed a simple test script using both the example pair and updater on both the CPU and GPU. Verified both worked as expected by viewing the results with Ovito.

No changes were made to spinx documentation. Builds and was viewable as expected.

## Change log

```
Refactored plugin examples.
```

## Checklist:


example_plugins/CMakeLists.txt
```diff
@@ -1,15 +1,19 @@
...
 message(STATUS "Found HOOMD ${HOOMD_VERSION}: ${HOOMD_INSTALL_PREFIX}/${PYTHON_SITE_INSTALL_DIR}")
 
 # Force installation to the HOOMD installation location
 set(CMAKE_INSTALL_PREFIX ${HOOMD_INSTALL_PREFIX} CACHE PATH "Installation prefix" FORCE)
 
 # add subdirectories
-add_subdirectory(${PROJECT_NAME})
+add_subdirectory(updater_plugin)
+add_subdirectory(pair_plugin)
+# add_subdirectory(force_compute_plugin)
```
joaander:
```suggestion
```

example_plugins/pair_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+install(FILES ${files}
+        DESTINATION ${PYTHON_SITE_INSTALL_DIR}/${COMPONENT_NAME}
+       )
+
+copy_files_to_build("${files}" "${COMPONENT_NAME}" "*.py")
+
+if (BUILD_TESTING)
+    # add_subdirectory(test-py)
```
joaander: This is optional, but it would be great if you added minimal pytest tests to each example plugin. This will ensure that the examples remain in good working order and promote good testing practices to those who copy the plugins.

The plugins would need to be installed before tests could be run so the hoomd/plugin namespace is consistent.

example_plugins/pair_plugin/EvaluatorPairExample.h
```diff
@@ -0,0 +1,190 @@
...
+            k = v["k"].cast<Scalar>();
+            sigma = v["sigma"].cast<Scalar>();
+            }
+
+        // this constructor facilitates unit testing
+        param_type(Scalar k, Scalar sigma, bool managed = false)
+            {
+            this->k = k;
+            this->sigma = sigma;
+            }
```
joaander:
```suggestion
```
We only use this style of constructor in potentials tested by legacy C++ unit tests.

example_plugins/pair_plugin/pair.py
```diff
@@ -0,0 +1,61 @@
...
+    _accepted_modes = ("none", "shift")
+
+    def __init__(self, nlist, default_r_cut=None, default_r_on=0., mode='none'):
+        super().__init__(nlist, default_r_cut, default_r_on, mode)
+        params = TypeParameter(
+            'params', 'particle_types',
+            TypeParameterDict(k=float, sigma=float, len_keys=2))
+        self._add_typeparam(params)
+
+    # this reimplementation is necessary at the moment because the base class
```
joaander: We could modify the base class with `_ext_module = _md` and then use `_ext_module` in `_attach`. That way, the reimplementation isn't necessary.

joaander (changes requested):
Great, thanks a lot for this! I have a few small suggestions for improvement.

@Tobias-Dwyer as you review, ask yourself this question: "Would I be able to copy this template and implement a new pair potential?" Understand that the form of the code is fixed, but you could offer suggestions on where we need more or different documentation.

@ianrgraham please also update `sphinx-doc/components.rst` to reflect the new `example_plugins` directory. Suggestion:
> The **HOOMD-Blue** source provides example component templates in the ``example_plugins`` subdirectory. ``updater_plugin`` demonstrates how to add a new ``update`` command with both CPU and GPU implementations, and ``pair_plugin`` shows how to add a new MD pair potential.

example_plugins/pair_plugin/pair.py
ianrgraham: I was going to suggest this! Yes, that would be great!

example_plugins/pair_plugin/CMakeLists.txt
ianrgraham: Sure thing! Do you think just python tests are sufficient? No need to write C++ tests for these simple cases?

example_plugins/CMakeLists.txt
ianrgraham: Yes, I wanted to suggest also implementing an example force compute since I have some experience doing this in my own projects. I have some uncommitted changes regarding that. But that can all be a separate pull request if you'd like to see it too. Maybe the example could be a replication of the example python ForceCompute from the docs, but in C++?

example_plugins/pair_plugin/CMakeLists.txt
joaander: Yes, Python tests are sufficient.

example_plugins/CMakeLists.txt
joaander: Certainly, more examples are a welcome addition for the community. I don't have a preference for one pull request vs two on this. Either is fine, I leave it up to you.

example_plugins/pair_plugin/CMakeLists.txt
ianrgraham: Hi Joshua, I added the pytests. But like you said, the namespace is currently inconsistent in the build directory. For external components this behaviour is okay, but at the moment this is breaking in-build testing when building hoomd itself since example_plugins is being symlinked into the hoomd dir. Do you have any suggestions?

joaander: Let's ignore the plugin tests in the build directory to get around this.

Add a `pytest.ini` file that gets copied into the directory `build/hoomd` with the contents:
```
[pytest]
norecursedirs = *_plugin
```
To copy it to the build directory, add `pytest.ini` to the `set(files` list in `hoomd/CMakeLists.txt`.

ianrgraham:
It looks like the last thing to fix is to throw a guard against building the pair_plugin if BUILD_MD=OFF.

example_plugins/pair_plugin/pytest/test_example_pair.py
```diff
@@ -0,0 +1,109 @@
...
+
+    snap = hoomd.Snapshot()
+    snap.particles.N = 2
+    snap.particles.types = ["A"]
+    snap.particles.position[:] = [[0, 0, 0], pos]
+    snap.configuration.box = [4, 4, 4, 0, 0, 0]
+
+    sim.create_state_from_snapshot(snap)
+
+    return sim
```
joaander: Could we use the existing `two_particle_snapshot_factory` instead of this method? If not, can we add the necessary optional arguments to `two_particle_snapshot_factory`, or extend it with another fixture?

`two_particle_snapshot_factory` is a pytest fixture, will reuse the existing device fixture, reduce code duplication, and already works with MPI (this method is currently failing MPI tests).
https://github.com/glotzerlab/hoomd-blue/blob/73c79a05f517c1c97b95f92702204c4b7f8405f5/hoomd/conftest.py#L126-L161

example_plugins/pair_plugin/pytest/test_example_pair.py
```diff
@@ -0,0 +1,109 @@
...
+
+    return f, e
+
+
+rng = np.random.default_rng(seed=0)
+positions = 0.5 * rng.random((10, 3)) - 0.25
+positions = list(positions) + list(2.0 * rng.random((10, 3)) - 1.0)
+devices = [hoomd.device.CPU()]
+if hoomd.device.GPU.is_available():
+    devices.append(hoomd.device.GPU())
```
joaander: This will be not needed when using `two_particle_snapshot_factory`. The existing `device` fixture adds the GPU device to the test parameters when needed and desired.

example_plugins/pair_plugin/pytest/test_example_pair.py
```diff
@@ -0,0 +1,109 @@
...
+    e = 0.5 * k * (dr - sigma)**2
+    if shift:
+        e -= 0.5 * k * (r_cut - sigma)**2
+
+    return f, e
+
+
+rng = np.random.default_rng(seed=0)
+positions = 0.5 * rng.random((10, 3)) - 0.25
+positions = list(positions) + list(2.0 * rng.random((10, 3)) - 1.0)
```
joaander: Please set positions to an explicit list (can be copied and pasted output from this code). This will ensure that future changes in numpy don't cause this test to fail.

example_plugins/pair_plugin/pytest/test_example_pair.py
```diff
@@ -0,0 +1,109 @@
...
+ks = [0.5, 1.0, 2.0, 5.0]
+sigmas = [0.5, 1.0, 1.5]
+
+testdata = list(itertools.product(positions, devices, ks, sigmas))
+
+
+@pytest.mark.parametrize("pos,device,k,sigma", testdata)
+def test_force_and_energy_eval(pos, device, k, sigma):
+
+    sim = build_binary_system(list(pos), device)
```
joaander: Also, use `simulation_factory` to make the `Simulation` object. `simulation_factory` sets some parameters that help unit tests run faster.

example_plugins/updater_plugin/pytest/test_example_updater.py
```diff
@@ -0,0 +1,78 @@
+# Copyright (c) 2009-2022 The Regents of the University of Michigan.
```
joaander, line 1: Some comments on this test as the first.

example_plugins/updater_plugin/pytest/test_example_updater.py
```diff
@@ -0,0 +1,78 @@
...
+import hoomd
+from hoomd import operation
+from hoomd.device import Device
+
+from hoomd import updater_plugin
+
+import numpy as np
+
+
+def build_system(vel: Optional[List[float]] = None,
```
joaander: You could base this on the `one_particle_snapshot_factory` fixture. Here is an example of making a fixture to extend another of the snapshot factories:
https://github.com/glotzerlab/hoomd-blue/blob/73c79a05f517c1c97b95f92702204c4b7f8405f5/hoomd/md/pytest/test_bond.py#L99-L112

joaander (changes requested):
Thanks a lot for all the work you are doing on this. I have some suggestions to improve the unit tests. I like what you did, but we have existing infrastructure that takes care of the same tasks (e.g. testing on CPU and GPU with fixtures instead of parameters). Let me know if you have any questions about my suggestions or if you need pointers to more examples to work from. Using these fixtures should only require minimal changes to what you already wrote.

example_plugins/updater_plugin/pytest/test_example_updater.py
ianrgraham: Yea, I noticed these when I was looking through the existing tests. No problem!

example_plugins/pair_plugin/pytest/test_example_pair.py
ianrgraham: I just refactored the tests to use the the two snapshot factories like you suggested. The whole `fixture(scope='session')` business was a little surprising to me, but I think some comments will help clarify what's going on to users.

example_plugins/pair_plugin/pytest/test_example_pair.py
ianrgraham: I just replaced these with an `np.linspace` call.

ianrgraham:
It's no problem! And thank you for the advice and suggestions! I have the most recent changes to the tests passing with MPI enabled. Just need to add comments to the tests and that should be it!

example_plugins/pair_plugin/pytest/test_example_pair.py
```diff
@@ -0,0 +1,84 @@
...
+    sim.operations.integrator = integrator
+
+    sim.run(0)
+    snap = sim.state.get_snapshot()
+    vec_dist = snap.particles.position[1] - snap.particles.position[0]
+
+    # Compute force and energy from Python
+    shift = mode == "shift"
+    f, e = harm_force_and_energy(vec_dist, k, sigma, sigma, shift)
+    e /= 2.0
```
joaander:
```suggestion
    if snap.communicator.rank == 0:
        vec_dist = snap.particles.position[1] - snap.particles.position[0]

        # Compute force and energy from Python
        shift = mode == "shift"
        f, e = harm_force_and_energy(vec_dist, k, sigma, sigma, shift)
        e /= 2.0
```

example_plugins/pair_plugin/pytest/test_example_pair.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Test that the forces and energies match that predicted by the Python impl.
+    forces = example_pair.forces
+    np.testing.assert_array_almost_equal(forces, [-f, f], decimal=6)
```
joaander:
```suggestion
    if snap.communicator.rank == 0:
        np.testing.assert_array_almost_equal(forces, [-f, f], decimal=6)
```

example_plugins/pair_plugin/pytest/test_example_pair.py
```diff
@@ -0,0 +1,84 @@
...
+
+    energies = example_pair.energies
+    np.testing.assert_array_almost_equal(energies, [e, e], decimal=6)
```
joaander:
```suggestion
    if snap.rank == 0:
        np.testing.assert_array_almost_equal(energies, [e, e], decimal=6)
```

example_plugins/updater_plugin/pytest/test_example_updater.py
```diff
@@ -0,0 +1,49 @@
...
+
+# Use pytest decorator to automate testing over the sequence of parameters.
+@pytest.mark.parametrize("vel", velocities)
+def test_updater(simulation_factory, one_particle_snapshot_factory, vel):
+
+    # `one_particle_snapshot_factory` and `simulation_factory` are pytest
+    # fixtures defined in hoomd/conftest.py. These factories automatically
+    # handle iterating tests over different CPU and GPU devices.
+    snap = one_particle_snapshot_factory()
+    snap.particles.velocity[0] = vel
```
joaander:
```suggestion
    if snap.communicator.rank == 0:
        snap.particles.velocity[0] = vel
```

example_plugins/updater_plugin/pytest/test_example_updater.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Add our plugin to the simulation.
+    updater: operation.Updater = updater_plugin.update.ExampleUpdater(
+        hoomd.trigger.On(sim.timestep))
+    sim.operations.updaters.append(updater)
+
+    # Test that the initial velocity matches our input.
+    sim.run(0)
+    velocity = sim.state.get_snapshot().particles.velocity[0]
+    np.testing.assert_array_almost_equal(velocity, vel, decimal=6)
```
joaander:
```suggestion
    snap = sim.state.get_snapshot()
    if snap.communicator.rank == 0:
        velocity = snap.particles.velocity[0]
        np.testing.assert_array_almost_equal(velocity, vel, decimal=6)
```

example_plugins/updater_plugin/pytest/test_example_updater.py
```diff
@@ -0,0 +1,49 @@
...
+
+    # Test that the velocity is properly zeroed after the update.
+    sim.run(1)
+    velocity = sim.state.get_snapshot().particles.velocity[0]
+    np.testing.assert_array_almost_equal(velocity,
+                                         np.array([0.0, 0.0, 0.0]),
+                                         decimal=6)
```
joaander:
```suggestion
    snap = sim.state.get_snapshot()
    if snap.communicator.rank == 0:
        velocity = sim.state.get_snapshot().particles.velocity[0]
        np.testing.assert_array_almost_equal(velocity,
                                             np.array([0.0, 0.0, 0.0]),
                                             decimal=6)
```

joaander (changes requested):
There are some idiosyncrasies in how snapshots work with MPI. `get_snapshot` must be called on all ranks, but the data is only available on rank 0. I suggested the edits that should fix the failing tests.

ianrgraham:
Does it matter to you if the communicator is accessed from sim.device or the snapshot?

joaander:
> Does it matter to you if the communicator is accessed from sim.device or the snapshot?

No, they will be the same communicator.

ianrgraham:
@joaander do you have any tips on running the CI workflow locally? While I think I see where the last mistake is, I'd like to try running more thorough tests before I push in the future. I don't write very much MPI code, so these gotchas have been sneaking up on me.

The calls to `example_pair.forces` and `example_pair.energies` are inside the check for rank 0. Let me know if this looks to be the issue, or if there is anything else that's suspect.

joaander:
> The calls to `example_pair.forces` and `example_pair.energies` are inside the check for rank 0. Let me know if this looks to be the issue, or if there is anything else that's suspect.

Yeah, the MPI data access API is confusing and it is easy to create deadlocks. I'm testing changes locally now.

ianrgraham:
Thanks!

joaander:
You found the fix for the deadlock. The reason this works is `forces = example_pair.forces` must be called on all ranks - it involves the collection of force data that is distributed among the ranks. However, the resulting forces object is available only on rank 0, so the `forces` object can only be accessed inside the `if rank==0` check.

joaander:
Looks like one more test is failing. I don't have time to build and check right now, but one guess is that the 
```
hipLaunchKernelGGL(gpu_zero_velocities_kernel, dim3(grid), dim3(threads), 0, 0, d_vel, N);
```
is launching 0 threads and that is an error.

We could ensure that this call launches at least one thread. Or, we could disable the test in MPI with the decorator:
```
@pytest.mark.serial
```

ianrgraham:
I'll give it a look later today. I don't have a strong opinion here, but it's probably best to just fix it? Skipping MPI-CUDA support in the tests for this simple case doesn't set a good precedent for other that come across the plugin examples? And I actually haven't touched the updater example C++ code at all, so this MPI-CUDA error has likely been around silently since it was committed.

joaander:
Yes, I know the original plugin code was not tested.

joaander:
I confirmed that this patch fixes the currently failing test.
```
--- a/example_plugins/updater_plugin/ExampleUpdater.cu
+++ b/example_plugins/updater_plugin/ExampleUpdater.cu
@@ -45,6 +45,10 @@ hipError_t gpu_zero_velocities(Scalar4* d_vel, unsigned int N)
     // setup the grid to run the kernel
     int block_size = 256;
     dim3 grid((int)ceil((double)N / (double)block_size), 1, 1);
+    if (grid.x == 0)
+        {
+        grid.x = 1;
+        }
     dim3 threads(block_size, 1, 1);
```

ianrgraham:
Sounds good! Thank you!

joaander (approved):
Thanks!

joaander:
#1393 tests that these plugins compile and pas tests in CI. It also fixes issues in the templates so that they function correctly with MPI and CUDA enabled on currently supported Ubuntu installations.

PR #1391, opened 2022-08-31
Feature/expansive sdf

tcmoore3:
## Description

Adds expansive volume perturbations to `ComputeSDF`. 

## Motivation and context

This is a follow-up to #1356 that enables pressure computation for systems of concave particles. 

## How has this been tested?

I tested this code in a separate branch before adding it to this branch by running simulations of a 2D sphere union in and confirming pressure–density consistency between NVT and NPT simulations. I will re-run that test using this branch and post the results when complete.

## Change log

```
- Enable computation of pressure in HPMC simulations of nonconvex particles
```

## Checklist:

- [ ] Post validation test results
- [x] Update python documentation
- [x] Figure out what unit tests are needed and implement them


hoomd/hpmc/ComputeSDF.h
```diff
@@ -149,14 +149,17 @@ template<class Shape> class ComputeSDF : public Compute
 
     //! Return an sdf
     virtual pybind11::array_t<double> getSDF();
+    virtual pybind11::array_t<double> getSDF_expansion();
```
joaander: Follow the naming conventions outline in https://hoomd-blue.readthedocs.io/en/v3.4.0/style.html

hoomd/hpmc/ComputeSDF.h
```diff
@@ -171,6 +174,7 @@ template<class Shape> class ComputeSDF : public Compute
     void countHistogramLinearSearch(uint64_t timestep);
 
     //! Determine the s bin of a given particle pair
+    //! Only used for the binary search mode
```
joaander: It is not valid to provide two brief descriptions `//!` for the same method.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -149,14 +149,17 @@ template<class Shape> class ComputeSDF : public Compute
...
 
     protected:
     std::shared_ptr<IntegratorHPMCMono<Shape>> m_mc; //!< The parent integrator
     double m_xmax;                                   //!< Maximum lambda value
     double m_dx;                                     //!< Histogram step size
 
-    std::vector<double> m_hist; //!< Raw histogram data
-    std::vector<double> m_sdf;  //!< Computed SDF
+    std::vector<double> m_hist;           //!< Raw histogram data
+    std::vector<double> m_sdf;            //!< Computed SDF
```
joaander: Will need to name these and related variables with `_compression` to avoid ambiguity.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -305,7 +335,7 @@ template<class Shape> Scalar ComputeSDF<Shape>::getMaxInteractionDiameter()
 */
 template<class Shape> void ComputeSDF<Shape>::countHistogram(uint64_t timestep)
     {
-    if (m_mc->getPatchEnergy())
+    if (m_mc->getPatchEnergy() || m_mc->requiresExpansiveSDF())
```
joaander: There is no need to introduce a new method in every shape. Use C++ constexpr to determine the expansive flag based on the type of `Shape`. For safety, default to `true` and only set to `false` for known convex `Shape` classes.

hoomd/hpmc/ComputeSDF.h
```diff
@@ -518,61 +550,123 @@ template<class Shape> void ComputeSDF<Shape>::countHistogramLinearSearch(uint64_
...
-                            // loop over bins; only going up to min_bin-1 since we only
-                            // want the _first_ overlap of particle i with its neighbors
-                            for (size_t bin_to_sample = 0; bin_to_sample < min_bin; bin_to_sample++)
+                            // do compressive and expansive perturbations in the same loop, keeping
+                            // track of which overlap checks we do not need to perform loop over
+                            // bins; only going up to min_bin-1 since we only want the _first_
+                            // overlap of particle i with its neighbors
+                            bool do_compression = true;
+                            bool do_expansion = true;
+                            for (size_t bin_to_sample = 0;
```
joaander: This would be more readable (and a smaller diff) if split into two separate for loops instead of one for loop that conspicuously goes to the max(i) and manages the different paths with bool flags.

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -346,6 +346,16 @@ class IntegratorHPMCMono : public IntegratorHPMC
                     break;
                     }
                 }
+            m_requiresExpansiveSDF = false;
```
joaander: There are no changes needed to this file or the `Shape*` files. As discussed above, detect based on the type of the `Shape` class.

hoomd/hpmc/compute.py
```diff
@@ -311,16 +311,32 @@ def sdf(self):
...
+    def sdf_expansion(self):
+        """(*N_bins*,) `numpy.ndarray` of `float`): :math:`s[k]` - The scale \
+        distribution function :math:`[\\mathrm{probability\\ density}]`.
+
+        The :math:`x` at the center of bin :math:`k` is:
+        :math:`x = k \\cdot \\delta x + \\delta x/2`.
+
+        Attention:
+            In MPI parallel execution, the array is available on rank 0 only.
+            `sdf` is `None` on ranks >= 1.
```
joaander: This appears to be copied and pasted from `sdf` without modification. Make the necessary changes to correctly document `sdf_expansion`.

hoomd/hpmc/compute.py
```diff
@@ -311,16 +311,32 @@ def sdf(self):
         self._cpp_obj.compute(self._simulation.timestep)
```
joaander, line 378: Rename `sdf` to `sdf_compression`. Create a deprecated `sdf` that aliases it. Document that deprecation in `deprecated.rst`.

hoomd/hpmc/compute.py
```diff
@@ -311,16 +311,32 @@ def sdf(self):
...
         return self._cpp_obj.sdf
```
joaander: I can't comment further up. In the class level docstring, document the expansion histogram and how `\beta P` relates to both the compression and expansion sdfs.

hoomd/hpmc/compute.py
```diff
@@ -329,19 +345,37 @@ def betaP(self):  # noqa: N802 - allow function name
...
-            box = self._simulation.state.box
-            N = self._simulation.state.N_particles
-            rho = N / box.volume
-            return rho * (1 + numpy.polyval(p, 0.0) / (2 * box.dimensions))
-        else:
-            return None
+            p0_compression = numpy.polyval(p, 0.0)
+            compression_contribution = rho * p0_compression / (2
+                                                               * box.dimensions)
+        if not numpy.isnan(self.sdf_expansion).all():
```
joaander: Why are any of the sdf values `nan` in the first place? They are initialized to 0.

hoomd/hpmc/compute.py
```diff
@@ -329,19 +345,37 @@ def betaP(self):  # noqa: N802 - allow function name
...
+            # get the values to fit
+            # reverse the sdf so that it starts at 0 and goes negative
+            sdf_fit_expansion = self.sdf_expansion[0:n_fit][::-1]
```
joaander: Should this be the natural representation that we provide to users? Or do you expect that users may want to plot `[sdf_expansion] + [sdf_compression]` vs some x that goes from `-xmax` to `+xmax`?

joaander (changes requested):
Thanks! I have many comments, questions, and suggestions for improvement.

hoomd/hpmc/compute.py
joaander: Here is another thought. Instead of providing `sdf_compression` and `sdf_expansion` to the user, we could provide one `sdf` that covers both. We should consider what is the better API for users.

tcmoore3:
My validation tests indicate that I introduced a bug in between my initial implementation and this pull request. Currently investigating.

hoomd/hpmc/compute.py
tcmoore3: That was already in the code, so I can't say for sure why it was added. But you're right, they're initialized to 0 and the only way I see they could become `nan` is if `m_pdata->getNGlobal() * m_dx == 0`.

hoomd/hpmc/compute.py
```diff
@@ -295,28 +348,80 @@ def _attach_hook(self):
     @log(category='sequence', requires_run=True)
     def sdf(self):
         """(*N_bins*,) `numpy.ndarray` of `float`): :math:`s[k]` - The scale \
-        distribution function :math:`[\\mathrm{probability\\ density}]`.
+        distribution function for compression moves \
+        :math:`[\\mathrm{probability\\ density}]`.
+
+        .. deprecated:: v3.8.0
```
joaander:
```suggestion
        .. deprecated:: v3.11.0
```
The next feature release will be v3.11.0.

hoomd/hpmc/compute.py
```diff
@@ -295,28 +348,80 @@ def _attach_hook(self):
...
+            Use `sdf_compression`.
+        """
+        self._cpp_obj.compute(self._simulation.timestep)
+        return self._cpp_obj.sdf_compression
```
joaander, line 363:
```suggestion
        return self.sdf_compression
```
Reuse existing code.

hoomd/hpmc/compute.py
```diff
@@ -325,18 +430,37 @@ def betaP(self):  # noqa: N802 - allow function name
             In MPI parallel execution, `betaP` is available on rank 0 only.
             `betaP` is `None` on ranks >= 1.
         """
-        if not numpy.isnan(self.sdf).all():
+        if not (numpy.isnan(self.sdf_compression).all()
+                or numpy.isnan(self.sdf_expansion).all()):
```
joaander:
```suggestion
        if self.sdf_compression is not None and self.sdf_expansion is not None:
```
Better yet,
```
sdf_compression = self.sdf_compression;
sdf_expansion = self.sdf_expansion;
if self.sdf_compression is not None and self.sdf_expansion is not None:
```
and then use `sdf_*` below instead of `self.sdf_*` to avoid making unnecessary data copies.

hoomd/hpmc/compute.py
```diff
@@ -325,18 +430,37 @@ def betaP(self):  # noqa: N802 - allow function name
...
+            rho = N / box.volume
+
+            # compressive contribution
             # get the values to fit
-            sdf_fit = numpy.array(self.sdf)
-            # construct the x coordinates
-            x_fit = numpy.arange(0, len(sdf_fit), 1) * self.dx
+            sdf_fit_compression = numpy.array(self.sdf_compression)
+            x_fit = numpy.arange(0, len(sdf_fit_compression), 1) * self.dx
             x_fit += self.dx / 2
```
joaander: Reuse code, use `self.x_compression`.

hoomd/hpmc/compute.py
```diff
@@ -325,18 +430,37 @@ def betaP(self):  # noqa: N802 - allow function name
...
+            # construct the x coordinates
+            # perform the fit and extrapolation
+            p = numpy.polyfit(x_fit, sdf_fit_compression, 5)
+            p0_compression = numpy.polyval(p, 0.0)
+            compression_contribution = rho * p0_compression / (2
+                                                               * box.dimensions)
+
+            # expansive contribution
+            # reverse the x-coordinates
+            x_fit = -x_fit[::-1]
```
joaander: This assumes that `len(sdf_compression) == len(sdf_expansion)`. Perhaps add an assert here to be certain.

hoomd/hpmc/compute.py
```diff
@@ -295,28 +348,80 @@ def _attach_hook(self):
...
+        return self._cpp_obj.sdf_expansion
+
+    @log(category='sequence', requires_run=False)
+    def x_compression(self):
+        """(*N_bins*,) `numpy.ndarray` of `float`): The x \
+        values at the center of each bin corresponding to the scale \
+        distribution function for the compressive perturbations \
+        :math:`[\\mathrm{length}]`."""
+        sdf_compression = numpy.array(self.sdf_compression)
+        x = numpy.arange(0, len(sdf_compression), 1) * self.dx
```
joaander: If it doesn't already exist, add a new C++ method to get the number of bins. Call that instead of `len(self.sdf_compression)`. This avoids making numerous extra copies of the histogram (see below).

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -286,5 +286,13 @@ def test_logging():
...
-            }
+            },
+            'sdf_compression': {
+                'category': LoggerCategories.sequence,
+                'default': True
+            },
+            'sdf_expansion': {
+                'category': LoggerCategories.sequence,
+                'default': True
+            },
```
joaander, line 366: Add `x_compression` and `x_expansion` to this list.

hoomd/hpmc/pytest/test_compute_sdf.py
```diff
@@ -270,7 +270,7 @@ def test_linear_search_path(simulation_factory, two_particle_snapshot_factory):
     # that the SDF is zero everywhere else
```
joaander, line 292: Add unit tests for `sdf_expansion`.

joaander (changes requested):
Thanks! This is looking a lot better. I look forward to seeing the validation test results and a new test workflow in https://github.com/glotzerlab/hoomd-validation/. There are some small changes needed to polish this code off (presuming the validation tests pass).

hoomd/hpmc/compute.py
```diff
@@ -182,11 +202,31 @@ class SDF(Compute):
...
+                                 \mathbf{q}_i,
+                                 \mathbf{q}_j)
+                \ne
+            U_{\mathrm{pair},ij}(\vec{r}_j - (\vec{r}_i + \vec{A}),
+                                 \mathbf{q}_i,
+                                 \mathbf{q}_j) \\
+            \} &
+
+
+    For particle :math:`i`, `SDF` finds th the minimum (maximum for expansive
```
tcmoore3:
```suggestion
    For particle :math:`i`, `SDF` finds the the minimum (maximum for expansive
```

joaander (approved):
Thanks!

joaander:
Validation tests will be provided in separate work on `hoomd-validation`: https://github.com/glotzerlab/hoomd-validation/issues/43

PR #1393, opened 2022-08-31
Test example plugins with external builds.

### LINKED ISSUE #1389: Add external plugins to CI tests.
### Description

Add a set of CI tests that build the example plugins externally and runs their tests.

### Motivation and context

These tests will detect more errors like those fixed in #1388. They will also serve to test the example CMake code when used for external plugins.

joaander:
## Description

Build the example plugins external to HOOMD.

## Motivation and context

This ensures that the HOOMD build process installs the header files necessary for plugins to build without error.

Resolves #1389

## How has this been tested?

The CI checks in Github Actions will test these changes.

## Checklist:


joaander:
Confirmation the installed plugin tests are running: https://github.com/glotzerlab/hoomd-blue/runs/8121421002?check_suite_focus=true#step:6:2717

example_plugins/pair_plugin/pytest/test_example_pair.py
```diff
@@ -9,18 +9,11 @@
...
 
 
 # Python implementation of the pair force and energy.
-def harm_force_and_energy(
-        dx: ArrayLike,
-        k: float,
-        sigma: float,
-        r_cut: float,
-        shift: Optional[bool] = False) -> Tuple[NDArray[np.float64], float]:
+def harm_force_and_energy(dx, k, sigma, r_cut, shift=False):
```
b-butler, line 16: Why this change?

b-butler (commented):
Just one comment on the change to the plugin Python code,

example_plugins/pair_plugin/pytest/test_example_pair.py
joaander: The above `from numpy.typing import NDArray, ArrayLike` fails in the Python 3.6 test image: https://github.com/glotzerlab/hoomd-blue/runs/8122103142?check_suite_focus=true#step:6:28

This error was not caught in #1382 because the Python 3.6 tests were not run on that pull request. I changed the build configurations to always include the older python versions in c256cb9 to catch these types of errors sooner.

b-butler: Ah okay. That makes sense. Also do we document our supported Python policy (I know the requirements is in the building from source docs)? NumPy stopped supporting 3.6 since Jun 23, 2020, so we have to grab pretty old packages to use 3.6.

b-butler (approved):

example_plugins/pair_plugin/pytest/test_example_pair.py
joaander: In [ARCHITECTURE.md](https://github.com/glotzerlab/hoomd-blue/blob/trunk-patch/ARCHITECTURE.md#hardware-and-software-support) I state that we support 
> Compilers and software dependencies available on the oldest maintained Ubuntu LTS release. 
That is currently Ubuntu 18.04 which installs Python 3.6 by default.

Following this documentation, we can require Python 3.7 and support Ubuntu 20.04+ starting in April 2023.

PR #1408, opened 2022-09-20
Fix pair table performance.

joaander:
## Description

Store pair potential parameters by reference.

## Motivation and context

This improves the performance of `md.pair.Table` in particular. Prior to this change, the copy constructors for the energy and force arrays were being called *twice* per pair interaction. After this change, `md.pair.Table` is 10-40x faster.

## How has this been tested?

I evaluated the performance increase with a new `hoomd-benchmarks` benchmark: https://github.com/glotzerlab/hoomd-benchmarks/pull/33

## Change log

```
Fixed:

* Increased performance of `md.pair.Table` on the CPU.
```

## Checklist:

### DIFF
--- a/hoomd/md/PotentialTersoffGPU.cuh
+++ b/hoomd/md/PotentialTersoffGPU.cuh
@@ -277,7 +277,7 @@ __global__ void gpu_compute_triplet_forces_kernel(Scalar4* d_force,
             unsigned int typpair
                 = typpair_idx(__scalar_as_int(postypei.w), __scalar_as_int(postypej.w));
             Scalar rcutsq = s_rcutsq[typpair];
-            typename evaluator::param_type& param = s_params[typpair];
+            const typename evaluator::param_type& param = s_params[typpair];
 
             // compute the base repulsive and attractive terms of the potential
             Scalar invratio = 0.0;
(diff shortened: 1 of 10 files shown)


vyasr (approved):
Nice find! The change to `PotentialPair.h` should improve performance for all pair potentials, right? Although maybe it matters less for non-table potentials since the parameters for other potentials are just a small set of scalars rather than a table?

joaander:
> Nice find! The change to `PotentialPair.h` should improve performance for all pair potentials, right? Although maybe it matters less for non-table potentials since the parameters for other potentials are just a small set of scalars rather than a table?

Yes, it has the potential to improve performance for other pair potentials as well. I'm not able to measure a performance difference for LJ. I suspect the compiler is able to optimize out the copy of the trivial LJ parameter structure. I did not inspect the assembly output to confirm. The optimizer heuristics might choose to not optimize out the copies of larger parameter structures - the const reference ensures that there are no extra copies.

I grepped the codebase and fixed several other instances where parameter structures were stored by coping.

hoomd/md/PotentialTersoffGPU.cuh
```diff
@@ -277,7 +277,7 @@ __global__ void gpu_compute_triplet_forces_kernel(Scalar4* d_force,
             unsigned int typpair
                 = typpair_idx(__scalar_as_int(postypei.w), __scalar_as_int(postypej.w));
             Scalar rcutsq = s_rcutsq[typpair];
-            typename evaluator::param_type param = s_params[typpair];
+            typename evaluator::param_type& param = s_params[typpair];
```
b-butler: Are these not const in this context?

b-butler (commented):
I am not sure if some `const` were missed or not. If not, I will approve. Otherwise, I will approve after the change.

hoomd/md/PotentialTersoffGPU.cuh
joaander: They are const. Fixed.

PR #1413, opened 2022-10-04
feat: Support adding force directly to simulation.

b-butler:
## Description
This makes the force a Compute on the Python side so it can be querried for force and energy even when not applied to the system.

## Motivation and context
This was motivated by a use case for querying a non-existent force by @Charlottez112.

## How has this been tested?
~Has not been tested currently, though I plan to add tests soon.~ Tests added.

## Change log

```
Added
- MD forces can now be added to the Operations list allowing for force, energy, virial querying without the forces effecting the system.
```

## Checklist:

### DIFF
--- a/hoomd/data/syncedlist.py
+++ b/hoomd/data/syncedlist.py
@@ -28,16 +28,6 @@ def identity(obj):
     return obj
 
 
-class _SimulationPlaceHolder:
-    """Used to ensure objects are not added to two locations at once."""
-
-    def __init__(self, obj):
-        self._id = id(obj)
-
-    def __eq__(self, other):
-        return isinstance(other, type(self)) and self._id == other._id
-
-
 class SyncedList(MutableSequence):
     """Provides syncing and validation for a Python and C++ list.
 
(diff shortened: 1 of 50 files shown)


hoomd/data/syncedlist.py
```diff
@@ -199,7 +199,12 @@ def _attach_value(self, value, raise_if_added=True):
             raise RuntimeError(f"Object {value} cannot be added to two lists.")
         value._add(self._simulation)
         if self._synced:
-            value._attach()
+            # To allow for dependencies within operations (like computes e.g.
+            # forces and neighbor list) we need to check it the object is
+            # attached first. Assuming the code for adding a simulation to an
+            # operation is correct, this is safe.
+            if not value._attached:
+                value._attach()
```
joaander, line 190: Could we test this in a unit test that adds lj to both the `integrator.forces` and `operations.computes`? I suppose the test could use internal members to ensure that the `_cpp_obj` is the same for both and/or that it doesn't change when adding it the 2nd time.

Should we also test unsetting the integrator? Need to make sure this doesn't forcibly detach the force that is present in `computes`.

joaander (commented):
Looks good! I have some questions on testing the dependent add fix.

b-butler:
Wait on reviewing this. I still have to fix one problem in adding an object to two `SyncedList` objects.

b-butler:
Good to review again.

joaander (approved):
Thanks! This looks great!

PR #1415, opened 2022-10-18
Fix round off errors in BoxMC with patch potentials.

joaander:
## Description

Prevent round off errors in BoxMC.

## Motivation and context

When the potential energy of the system is large, BoxMC would previously add a small number to a large number, then subtract another large number. This is the potential wash out the effect of the small number (the P * delta V term).

## How has this been tested?

I tested the changes in another repository that checks the physical validity of HOOMD-blue simulations. The change doesn't significantly improve accuracy in my test cases, but it could in others where the potential energy of the system is very large in magnitude.

## Change log

```
* Improve accuracy of ``hoomd.hpmc.update.BoxMC`` when used with patch potentials.
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/UpdaterBoxMC.cc
+++ b/hoomd/hpmc/UpdaterBoxMC.cc
@@ -226,7 +226,7 @@ inline bool UpdaterBoxMC::box_resize_trial(Scalar Lx,
                                            Scalar xz,
                                            Scalar yz,
                                            uint64_t timestep,
-                                           Scalar deltaE,
+                                           double delta_beta_H,
                                            hoomd::RandomGenerator& rng)
     {
     // Make a backup copy of position data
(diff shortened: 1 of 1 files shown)


hoomd/hpmc/UpdaterBoxMC.cc
```diff
@@ -259,24 +260,23 @@ inline bool UpdaterBoxMC::box_resize_trial(Scalar Lx,
...
-        // The exponential is a very fast function and we may do better to add pseudo-Hamiltonians
-        // and exponentiate only once...
-        deltaE += ext_energy;
+        delta_U_external = ext_energy;
         }
 
     double p = hoomd::detail::generate_canonical<double>(rng);
 
-    if (allowed && p < fast::exp(-deltaE))
+    if (allowed && p < exp(-delta_beta_H) * exp(-delta_U_pair) * exp(-delta_U_external))
```
tommy-waltmann, line 279: Why remove the `fast::` exponential?

joaander:
FYI, the acceptance criterion for NPT MC is here: http://www.sklogwiki.org/SklogWiki/index.php/Constant-pressure_Monte_Carlo

hoomd/hpmc/UpdaterBoxMC.cc
joaander: This is an accurate exponential computed with doubles. It is computed only once per timestep, so performance is not a concern.

tommy-waltmann (approved):

cbkerr (approved):
The changes make sense to calculate the exponents of numbers of different magnitude first.

PR #1417, opened 2022-10-18
Revise ALJ documentation.

joaander:
## Description

Clarify values in the ALJ documentation and explicitly document the required r_cut values.

## Motivation and context

The documentation does not currently describe the minimum value of r_cut needed for the simulation to run.

## How has this been tested?

I reviewed the generated documentation locally.

## Change log

```
* Revise ``hoomd.md.pair.aniso.ALJ`` documentation.
```

## Checklist:

### DIFF
--- a/hoomd/md/pair/aniso.py
+++ b/hoomd/md/pair/aniso.py
@@ -244,14 +244,10 @@ class ALJ(AnisotropicPair):
     Args:
         nlist (hoomd.md.nlist.NeighborList): Neighbor list
         default_r_cut (float): Default cutoff radius :math:`[length]`.
-        mode (`str`, optional): the energy shifting mode, defaults to "none"
-          (ignored).
+        mode (`str`, optional): the energy shifting mode, defaults to "none".
+          Computes the same energy regardless of value passed.
 
           .. deprecated:: v3.1.0
-
-    Note:
-        `ALJ` accepts the ``mode`` parameter, but computes the same energy
-        regardless of the value of ``mode`` (starting with v3.0.0).
 
     `ALJ` computes the Lennard-Jones force between anisotropic particles as
     described in `Ramasubramani, V.  et al. 2020`_, using the formula:
--- a/hoomd/md/pair/aniso.py
+++ b/hoomd/md/pair/aniso.py
@@ -338,7 +338,8 @@ class ALJ(AnisotropicPair):
     that `ALJ` can compute interactions for all possible relative placements
     and orientations of the particles. The furthest apart two particles can be
     while still interacting depends on the value of ``alpha`` (where :math:`R_i`
-    is the circumsphere radius of the particle with type :math:`i`):
+    is the radius of the shape's minimal origin centered bounding sphere of the
+    particle with type :math:`i`):
 
     * For alpha=0:
 
--- a/hoomd/md/pair/aniso.py
+++ b/hoomd/md/pair/aniso.py
@@ -473,11 +473,11 @@ class ALJ(AnisotropicPair):
         * ``alpha`` (`int`, **required**) - Integer 0-3 indicating whether or
           not to include the attractive component of the interaction (see
           above for details).
-        * ``contact_ratio_i`` (`float`, **optional**) - the ratio of the contact
-          sphere diameter of the first type with ``sigma_i`` :math:`\beta_i`.
+        * ``contact_ratio_i`` (`float`, **optional**) - :math:`\beta_i`, the
+          ratio of the contact sphere diameter of the first type with ``sigma_i``.
           Defaults to 0.15.
-        * ``contact_ratio_j`` (`float`, **optional**) - the ratio of the contact
-          sphere diameter of the second type with ``sigma_j`` :math:`\beta_j`.
+        * ``contact_ratio_j`` (`float`, **optional**) - :math:`\beta_j`, the
+          ratio of the contact sphere diameter of the second type with ``sigma_j``.
           Defaults to 0.15.
         * ``average_simplices`` (`bool`, **optional**) - Whether to average over
           simplices. Defaults to ``True``. See class documentation for more
--- a/hoomd/md/pair/aniso.py
+++ b/hoomd/md/pair/aniso.py
@@ -338,7 +338,7 @@ class ALJ(AnisotropicPair):
     that `ALJ` can compute interactions for all possible relative placements
     and orientations of the particles. The furthest apart two particles can be
     while still interacting depends on the value of ``alpha`` (where :math:`R_i`
-    is the circumsphere radius of the particle with type :math:`i`:):
+    is the circumsphere radius of the particle with type :math:`i`):
 
     * For alpha=0:
 
--- a/hoomd/md/pair/aniso.py
+++ b/hoomd/md/pair/aniso.py
@@ -299,7 +299,7 @@ class ALJ(AnisotropicPair):
     Like a standard LJ potential, each term has an independent cutoff beyond
     which it decays to zero. The behavior of these cutoffs is dependent on
     whether a user requires LJ or Weeks-Chandler-Anderson (WCA)-like
-    (repulsive-only) behavior. This behavior is controlled using the ``alpha``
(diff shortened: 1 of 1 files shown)


hoomd/md/pair/aniso.py
```diff
@@ -328,6 +328,54 @@ class ALJ(AnisotropicPair):
...
+            R_{\mathrm{rounding},j}+ \frac{2^\frac{1}{6}}{2}
+            (\beta_i \cdot \sigma_i + \beta_j \cdot \sigma_j) \right)
+
+    * For alpha=2:
+
+      .. math::
+
+            r_{\mathrm{cut},ij} = \max \left( \frac{2^\frac{1}{6}}{2}
+            (\sigma_i + \sigma_j)),  R_i + R_j + R_{\mathrm{rounding},i} +
+            R_{\mathrm{rounding},j} + \frac{2.5}{2}
```
cbkerr: What determines whether the factor is 2^(1/6) or 2.5, and where does the 2.5 come from? We may encourage more understanding by factoring out the parts of the equation that stay constant.

hoomd/md/pair/aniso.py
```diff
@@ -328,6 +328,54 @@ class ALJ(AnisotropicPair):
...
+    and orientations of the particles. The furthest apart two particles can be
+    while still interacting depends on the value of ``alpha`` (where :math:`R_i`
+    is the circumsphere radius of the particle with type :math:`i`:):
+
+    * For alpha=0:
+
+      .. math::
+
+        r_{\mathrm{cut},ij} = \max \left( \frac{2^\frac{1}{6}}{2}
+        (\sigma_i + \sigma_j), R_i + R_j + R_{\mathrm{rounding},i} +
```
cbkerr: Rendering locally, I think we need more space after the comma.

I also like how 1/6 takes up less vertical space than \frac{1}{6}.

```suggestion
        r_{\mathrm{cut},ij} = \max \left( \frac{2^{1/6} (\sigma_i + \sigma_j) }{2}
        \: , R_i + R_j + R_{\mathrm{rounding},i} +
```

hoomd/md/pair/aniso.py
joaander: `alpha` sets whether only the repulsive part of the potential is included, or the attractive part. Internally, this uses a cutoff of `2**(1/6)` for the repulsive only potentials (WCA) - cutoff at the minimum of the potential. Internally, it uses *no* cutoff for the attractive part. 2.5 \sigma is the traditional `r_cut` used for LJ simulations.

joaander: > We may encourage more understanding by factoring out the parts of the equation that stay constant.

What would you suggest factoring out?

hoomd/md/pair/aniso.py
```diff
@@ -328,6 +328,54 @@ class ALJ(AnisotropicPair):
...
+        `ALJ` implicitly rounds the given shape by :math:`\sigma_c` via the WCA
+        or Lennard-Jones potential.
+
+    .. rubric:: Choosing `r_cut <hoomd.md.pair.Pair.r_cut>`:
+
+    Set `r_cut <hoomd.md.pair.Pair.r_cut>` for each pair of particle types so
+    that `ALJ` can compute interactions for all possible relative placements
+    and orientations of the particles. The furthest apart two particles can be
+    while still interacting depends on the value of ``alpha`` (where :math:`R_i`
+    is the circumsphere radius of the particle with type :math:`i`):
```
joaander: Replace circumsphere with "distance from the origin to the furthest point". This is only the circumsphere radius when the particle's origin is the center of the circumsphere.

tommy-waltmann: What if we used coxeter language and described R_i as the "radius of the shape's minimal centered bounding sphere"?

tommy-waltmann (approved):

hoomd/md/pair/aniso.py
cbkerr: It would be nice to reduce duplicated stuff like the sum of all the R's to highlight the term that is changing.

I experimented with putting this in a table, but even that is a little bit too large.

hoomd/md/pair/aniso.py
```diff
@@ -316,6 +316,20 @@ class ALJ(AnisotropicPair):
     * 3:
       All interactions include attractive and repulsive components.
 
+    .. list-table:: Set alpha based on range of the center-center and contact-contact interactions.
```
cbkerr: What do you think of this format? I think it's easier to read. Should it replace the list form above this?
image

hoomd/md/pair/aniso.py
```diff
@@ -410,16 +481,18 @@ class ALJ(AnisotropicPair):
...
+          first particle type, :math:`\sigma_i` :math:`[length]`.
         * ``sigma_j`` (`float`, **required**) - the insphere diameter of the
-          second particle type, :math:`[length]`.
+          second particle type, :math:`\sigma_j` :math:`[length]`.
         * ``alpha`` (`int`, **required**) - Integer 0-3 indicating whether or
           not to include the attractive component of the interaction (see
           above for details).
         * ``contact_ratio_i`` (`float`, **optional**) - the ratio of the contact
-          sphere diameter of the first type with ``sigma_i``. Defaults to 0.15.
+          sphere diameter of the first type with ``sigma_i`` :math:`\beta_i`.
```
cbkerr: What does the extra \beta_i mean?

hoomd/md/pair/aniso.py
cbkerr: I pushed a commit defining where the 2.5 and 2^1/6 come from

hoomd/md/pair/aniso.py
joaander: Yes, this looks good.

hoomd/md/pair/aniso.py
joaander: There are a lot of common terms, but giving them names and reusing them only adds more layers of complexity for the user to understand.

hoomd/md/pair/aniso.py
joaander: `\beta_i` is the math symbol we use in the above equations for the parameter named `contact_ratio_i` in code.

hoomd/md/pair/aniso.py
joaander: It's not clear to me whether users would find that more readable or not - e.g. if we gave names to R_i + R_j + R_rounding_i + R_rounding_j and (\sigma_i * \beta_i + \sigma_j * \beta_j).

As the equations are now, we would need to break them at some point, they run past the end of the rtd theme page width.

hoomd/md/pair/aniso.py
cbkerr: Ah, it looked like multiplication or something. I'll move it to the beginning of the line

hoomd/md/pair/aniso.py
```diff
@@ -328,6 +328,65 @@ class ALJ(AnisotropicPair):
     only ``vertices`` creates a convex polytope (set ``vertices`` and ``faces``
     to empty lists to create the ellipsoid).
 
+    Important:
+        `ALJ` implicitly rounds the given shape by :math:`\sigma_c` via the WCA
+        or Lennard-Jones potential.
```
cbkerr: I'm not sure what "implicitly" means here. Could we say "has the same effect as rounding radius" ?

joaander: It does not have the same effect as the rounding radius. The rounding radius changes the underlying shape of the particle from a convex polyhedron to a sphero-convex polyhedron. The contact distance *r* is computed between the two shapes.

I say that the contact sigma "implicltly" rounds the shape because the LJ potential has high energy values for *r* < 1.0 * contact sigma, preventing particles from achieving closer contacts in normal thermodynamic conditions.

cbkerr: Resolved offline

hoomd/md/pair/aniso.py
cbkerr: The natural spot is to break after the comma in the max function

cbkerr (approved):
Big improvement

PR #1399, opened 2022-09-07
PairList accessor for NeighborLists

ianrgraham:
## Description

Allows the `NeighborList` Python class to act as a standalone `Compute`, and adds an accessor method to retrieve a pair list. The implementation isn't completely finished, but it's working for CPU (and likely GPU) builds, MPI is probably broken. I started the pull request so I could get some feedback on the Python API, as well as how the C++ implementation for `getPairList` should behave.

From the Python side I added one new parameter to all of the nlists, `default_r_cut`, which defaults to `None`. The behavior here differs from pair potentials in that if the parameter remains `None`, it sets the default `r_cut` to 0 for all type pairs. Users can now modify the nlist's `r_cut`s after initialization using the TypeParameter. The pair list at the current timestep is accessed with the `pair_list` property, which is loggable (untested). Does this API look suitable?

The C++ impl of `getPairListPython` updates the nlist, loops through all neighbors (unfiltered, including neighbors within r_cut + buffer), and returns a vector of unsigned int pairs. Here I'm curious what the behavior should be when multiple nlist consumers are registered to the NeighborList. Should it select neighbors based upon `rcut_max`, `rcut_base`, or just leave the filtering job up to the caller? Should there be some logic to handle multiple cases? Should we cache the pair list somewhere if the same case is encountered at the same timestep? Is an SOA return type preferred to AOS? At the moment the pair list is respecting the current `StorageMode` of the nlist. Should it always yield the "full" or "half" nlist instead?

I'd also like to implement a zero-copy accessor the the 3 primary neighbor list arrays (NNeighArray, NListArray, and HeadList). I'm going to try copying from the context manager implementation of `{device}_local_force_arrays`. Any advice there would be appreciated.

## Motivation and context

Resolves #1259

## How has this been tested?

I'm intending on implement a few additional pytests in the `md` module.
- [x] Test behavior before and after attaching/detaching nlist from simulation
- [x] Test `pair_list` correctness and API usage
  - [x] Specified `default_r_cut`, no other nlist consumers
  - [x] `default_r_cut` + additional nlist consumers with different r_cuts
  - [x] Manually specified `r_cut` after nlist instantiation
- [x] ~~Test logging pair list to file~~

I haven't modified docs yet, but once the Python API is fixed I'll do so.
- [x] Update docs to reflect `pair_list` API

## Change log

> Added: 
> - Neighborlists may now be added to the simulation as standalone computes
> - Pair list property to access global neighbor information
[...]

PR #1418, opened 2022-10-19
refactor: Remove _remove and _add from data model

b-butler:
## Description
Removes `_remove` and `_add` from the HOOMD object data model. Given the changes on #1413, we can no longer detect an object being added to multiple lists early which was the main purpose of the methods to aid users by erroring early.

This also removes excess and (now) faulty logic in `Pair` and similar classes regarding neighbor lists.

## Motivation and context
This is required/useful to simplify logic and enable #1399.
## How has this been tested?
Existing test pass.

## Change log

Don't know if this should go into changelog.

```
Changed: Removed internal methods ``_remove`` and ``_add`` from HOOMD data model.
```

## Checklist:

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -234,13 +234,10 @@ except where necessary.
 ### Base Data Model
 
 #### Definitions
-- added: An object is associated with a `Simulation`.
 - attached: An object has its corresponding C++ class instantiated and is
   connected to a `Simulation`.
-- unattached: An object's data resides in pure Python and may be or not be
-  added.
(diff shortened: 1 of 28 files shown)


joaander (approved):
Thanks! This looks like a good simplification of the data model with less need to override methods.

Can you give an example of what kind of deferred error message users will now get?

b-butler:
Users will get errors upon attempting to attach an object to a second simulation (when run is called on the second job). Neighbor list and the Mesh data structure both will deepcopy themselves to avoid this error as they currently do.

PR #1433, opened 2022-11-09
Refactor Rigid Bodies Body Definition

### LINKED ISSUE #1350: `md.constrain.Rigid` uses `charges` and `diameters` only in `create_bodies`
## Description

In undocumented behavior, `md.constrain.Rigid` uses `charges` and `diameters` only in `create_bodies`. This is unlike `constituent_types`, which is used checked the body validation process and `positions` and `orientation` which are applied every time step.

Also, changing only the `charges` or `diameters` is not counted as a changed body (e.g. making a change to `charges` only does not trigger a new validation): https://github.com/glotzerlab/hoomd-blue/blob/319c199d18d449a7f1cfca45ea0e8db99f48877e/hoomd/md/ForceComposite.cc#L167-L171 

## Script

```python
import hoomd

snapshot = hoomd.Snapshot()
snapshot.configuration.box = (10,10,10,0,0,0)
snapshot.particles.N = 2
snapshot.particles.position[:] = [(0, 0, 0), (1, 0, 0)]
snapshot.particles.types = ['A', 'B']
snapshot.particles.typeid[:] = [0, 1]
snapshot.particles.charge[:] = [-1.0, 1.0]
snapshot.particles.diameter[:] = [1.0, 2.0]
snapshot.particles.body[:] = [0.0, 0.0]

sim = hoomd.Simulation(device=hoomd.device.CPU(), seed=10)

rigid = hoomd.md.constrain.Rigid()
rigid.body['A'] = {
    "constituent_types": ['B'],
    "positions": [(0,0,1)],
    "orientations": [(1.0, 0.0, 0.0, 0.0)],
    "charges": [0.0],
    "diameters": [1.0]
    }

integrator = hoomd.md.Integrator(dt=0, integrate_rotational_dof=True)
integrator.rigid = rigid

sim.run(0)
```

## Output

*no output*
## Expected output

I expected an error that the charge and/or diameter given in `rigid.body['A']` fails to match that in the simulation state. Like this error given when `constituent_types` doesn't match:

```
RuntimeError: Error validating rigid bodies: Constituent particle types must be consistent with the rigid body definition.  Rigid body definition is in order of particle tag.
```

ALTERNATIVELY, I expect the API to not accept `charges` and `diameters` as body type parameters (see discussion below).

## Configuration

Versions:
- Python version: all
- **HOOMD-blue** version: 3.2.0

## Developer

Before implementing a fix, we should discuss what the correct API should be. Many users have expressed extreme confusion over the charge and diameter fields. There is less confusion over `constituent_types`, however it has the same behavior (used for `create_bodies` and validation only).

I see two rational options:

* Validate `charges` and `diameters` the same way as `constituent_types`. Document how these are used.
* Remove `charges`, `diameters` and `consituent_types` from the `bodies` type parameter and make them parameters of `create_bodies` (optional in the case of `charges` and `diameters`). Document how these are used.

The 2nd option will involve a deprecation or a breaking change. However, it cleans up the API as it longer requires that the user provide the same information in two places (simulation state and body definition). Users will still be annoyed that they need to define the local positions and orientations, but that is not avoidable as the data available in the simulation state is both not sufficient and overdetermined.

b-butler:
## Description
Remove charge/diameter from rigid body definition. Allows setting charges in createRigidBodies, but not diameters since we are removing that.

## Motivation and context
This is one of the main tasks for 4.0.

Resolves #1350

## How has this been tested?
Existed _modified_ tests are sufficient.

## Change log

```
Changed: `hoomd.md.constrain.Rigid` no longer takes `diameters` or `charges` as body arguments. The `create_bodies` method now takes an optional `charges` argument to set charges.
```

## Checklist:

### DIFF
--- a/hoomd/md/ForceComposite.cc
+++ b/hoomd/md/ForceComposite.cc
@@ -8,9 +8,7 @@
 #include <sstream>
 #include <string.h>
 
-#ifndef __HIPCC__
 #include <pybind11/stl.h>
-#endif
 
 /*! \file ForceComposite.cc
     \brief Contains code for the ForceComposite class
--- a/hoomd/md/ForceComposite.cc
+++ b/hoomd/md/ForceComposite.cc
@@ -528,6 +528,7 @@ void ForceComposite::createRigidBodies(
     bool remove_existing_bodies = false;
     unsigned int n_constituent_particles = 0;
     unsigned int n_free_bodies = 0;
+    if (m_exec_conf->getRank() == 0)
         {
         // We restrict the scope of h_body_len to ensure if we remove_existing_bodies or in any way
         // reallocated m_body_len to a new memory location that we will be forced to reaquire the
@@ -610,6 +611,7 @@ void ForceComposite::createRigidBodies(
                 snap.type[constituent_particle_tag]
                     = h_body_type.data[m_body_idx(body_type, current_body_index)];
                 snap.body[constituent_particle_tag] = particle_tag;
+                snap.pos[constituent_particle_tag] = snap.pos[particle_tag];
                 if (!charges.empty())
                     {
                     snap.charge[constituent_particle_tag]
--- a/hoomd/md/ForceComposite.cc
+++ b/hoomd/md/ForceComposite.cc
@@ -8,6 +8,10 @@
 #include <sstream>
 #include <string.h>
 
+#ifndef __HIPCC__
+#include <pybind11/stl.h>
+#endif
+
 /*! \file ForceComposite.cc
     \brief Contains code for the ForceComposite class
 */
@@ -54,9 +58,6 @@ ForceComposite::ForceComposite(std::shared_ptr<SystemDefinition> sysdef)
         h_body_len.data[i] = 0;
         }
 
-    m_body_charge.resize(m_pdata->getNTypes());
-    m_body_diameter.resize(m_pdata->getNTypes());
-
     m_d_max.resize(m_pdata->getNTypes(), Scalar(0.0));
     m_d_max_changed.resize(m_pdata->getNTypes(), false);
 
@@ -96,15 +97,11 @@ ForceComposite::~ForceComposite()
 void ForceComposite::setParam(unsigned int body_typeid,
                               std::vector<unsigned int>& type,
                               std::vector<Scalar3>& pos,
-                              std::vector<Scalar4>& orientation,
-                              std::vector<Scalar>& charge,
-                              std::vector<Scalar>& diameter)
+                              std::vector<Scalar4>& orientation)
     {
     assert(m_body_types.getPitch() >= m_pdata->getNTypes());
     assert(m_body_pos.getPitch() >= m_pdata->getNTypes());
     assert(m_body_orientation.getPitch() >= m_pdata->getNTypes());
-    assert(m_body_charge.size() >= m_pdata->getNTypes());
-    assert(m_body_diameter.size() >= m_pdata->getNTypes());
 
     if (body_typeid >= m_pdata->getNTypes())
         {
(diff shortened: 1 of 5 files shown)


b-butler:
@joaander the containers are failing to initialize which results in test failing.

It looks like the GPU driver is causing the error?

joaander:
@b-butler the tests are now running.

hoomd/md/ForceComposite.cc
```diff
@@ -8,6 +8,10 @@
 #include <sstream>
 #include <string.h>
 
+#ifndef __HIPCC__
+#include <pybind11/stl.h>
+#endif
```
joaander:
```suggestion
#include <pybind11/stl.h>
```
Check not needed in `.cc` files.

hoomd/md/ForceComposite.cc
```diff
@@ -513,7 +489,39 @@ void ForceComposite::validateRigidBodies()
...
+            }
+        const auto charges_list = item.second.cast<pybind11::list>();
+        if (pybind11::len(charges_list) != h_body_len.data[type])
+            {
+            throw std::runtime_error("Charges provided not consistent with rigid body size.");
+            }
+        std::vector<float> charges_vector;
+        for (auto& charge : charges_list)
+            {
+            charges_vector.emplace_back(charge.cast<float>());
```
joaander:
```suggestion
            charges_vector.emplace_back(charge.cast<Scalar>());
```

hoomd/md/ForceComposite.cc
```diff
@@ -513,7 +489,39 @@ void ForceComposite::validateRigidBodies()
...
+        if (h_body_len.data[type] == 0)
+            {
+            throw std::runtime_error("Charge provided for non-central particle type.");
+            }
+        const auto charges_list = item.second.cast<pybind11::list>();
+        if (pybind11::len(charges_list) != h_body_len.data[type])
+            {
+            throw std::runtime_error("Charges provided not consistent with rigid body size.");
+            }
+        std::vector<float> charges_vector;
```
joaander:
```suggestion
        std::vector<Scalar> charges_vector;
```

hoomd/md/constrain.py
```diff
@@ -290,19 +290,20 @@ def __init__(self):
...
         self._add_typeparam(body)
         self.body.default = None
 
-    def create_bodies(self, state):
+    def create_bodies(self, state, charges=None):
         r"""Create rigid bodies from central particles in state.
 
         Args:
             state (hoomd.State): The state in which to create rigid bodies.
+            charges (dict[str, float], optional): The charges for each of the
```
joaander:
```suggestion
            charges (dict[str, list[float]], optional): The charges for each of the
```

joaander (changes requested):
Thanks! This turned out to be a minimal change to fix the API.

Update `test_rigid_filter` to work with this change: https://github.com/glotzerlab/hoomd-blue/actions/runs/3438664815/jobs/5803196403#step:6:4513

joaander:
`test_filter_md.py` is still failing.

b-butler:
Sorry missed that.

joaander:
We also need to make a `trunk-major` branch on https://github.com/glotzerlab/hoomd-examples, set it to build the `trunk-major` branch of HOOMD for testing, and update the rigid body tutorial to match the new API.

hoomd/md/constrain.py
```diff
@@ -290,19 +290,20 @@ def __init__(self):
...
 
-    def create_bodies(self, state):
+    def create_bodies(self, state, charges=None):
         r"""Create rigid bodies from central particles in state.
 
         Args:
             state (hoomd.State): The state in which to create rigid bodies.
+            charges (dict[str, list[float]], optional): The charges for each of the
+                constituent particles, defaults to ``None``. If ``None``, all
+                charges are zero. The keys should be the central particles.
```
joaander:
```suggestion
            charges (dict[str, list[float]], optional): The charges for each of
                the constituent particles, defaults to ``None``. If ``None``,
                all charges are zero. The keys should be the central particles.
```
fix: `hoomd/md/constrain.py:304:81: E501 line too long (83 > 80 characters)`

joaander (approved):
Thanks! This looks good.

One small change is needed so that flake8 will pass.

joaander:
Reminder to update the tutorial: https://github.com/glotzerlab/hoomd-examples/issues/72

PR #1434, opened 2022-11-11
Refactor Local Access

b-butler:
## Description

- refactor: Remove `getN{...}` from `LocalDataAccess`.
    - Creates a new `SnapshotLocalDataAccess` class which inherits from
      `LocalDataAccess`, and takes in the local number of particles, ghost
      particles, and global particles for use in accessing buffer. **This
      means there is no need for the `Data` class to have `getN` and
      associated methods.**
    - Base class now has a single buffer access method that requires an
      almost full specification of the buffer (but now supports N
      dimensional arrays as well).
    - Modifies the Python/C++ constructors as necessary to get the particle
      number information (usually from particle data).

- ~fix: `Force` local access buffer sizes.~
    - ~angles, dihedrals, etc forces will return properly sized buffers.~
    - ~The local access class requires the 3 N's (N, N ghost, N global) on
      the Python side.~
    - ~A new attribute was added to `hoomd.md.force.Force`, `_cpp_data_method`
      to denote what method to call from the Python side `SystemDefinition` to
      get the correct sizes.~
- doc: Update documentation on local access 

## Motivation and context

This was discussed in #1399.

## How has this been tested?

Existing tests pass. We may want to test the fix to the force buffers.

## Change log

```
Fixed: `Force` subclasses other than pair forces return properly sized arrays for CPU and GPU local forces.
```

## Checklist:

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -261,8 +261,7 @@ two objects. The class is inherited by `_HOOMDBaseObject` to handle dependent
 relationships between operations in Python. The class defines *dependents* and
 *dependencies* of an object whose removal from a simulation (detaching) can be
 handled by overwriting specific methods defined in `_DependencyRelation` in
-`hoomd/operation.py`. See the interface of neighbor lists to pair potentials as
-an example of this in `hoomd/md/nlist.py` and `hoomd/md/pair/pair.py`.
+`hoomd/operation.py`.
 
 #### `_HOOMDGetSetAttrBase`
 
@@ -462,6 +461,14 @@ object then any necessary changes should be made (if the constructor is used
 then any new arguments to constructor must have defaults via semantic
 versioning and no changes should be needed to support pickling). The removal of
 internal attributes should not cause problems as well.
+
+## Zero Copy Buffer Access
+
+HOOMD allows for C++ classes to expose their GPU and CPU data buffers directly
+in Python using the `__cuda_array_interface__` and `__array_interface__`. This
+behavior is controlled using the `hoomd.data.local_access._LocalAcces` class in
+Python and the classes found in `hoomd/PythonLocalDataAccess.h`. See these files
+for more details. For example implementations look at `hoomd/ParticleData.h`.
 
 ## Directory structure
 
--- a/hoomd/ForceCompute.h
+++ b/hoomd/ForceCompute.h
@@ -237,13 +237,12 @@ class PYBIND11_EXPORT ForceCompute : public Compute
  *
  * */
 template<class Output>
-class PYBIND11_EXPORT LocalForceComputeData : public PerGroupLocalDataAccess<Output, ForceCompute>
+class PYBIND11_EXPORT LocalForceComputeData : public GhostLocalDataAccess<Output, ForceCompute>
     {
     public:
     LocalForceComputeData(ForceCompute& data, ssize_t N, ssize_t N_ghost, ssize_t N_global)
-        : PerGroupLocalDataAccess<Output, ForceCompute>(data, N, N_ghost, N_global),
-          m_force_handle(), m_torque_handle(), m_virial_handle(),
-          m_virial_pitch(data.getVirialArray().getPitch()),
+        : GhostLocalDataAccess<Output, ForceCompute>(data, N, N_ghost, N_global), m_force_handle(),
+          m_torque_handle(), m_virial_handle(), m_virial_pitch(data.getVirialArray().getPitch()),
           m_buffers_writeable(data.getLocalBuffersWriteable())
         {
         }
--- a/hoomd/ForceCompute.h
+++ b/hoomd/ForceCompute.h
@@ -240,11 +240,8 @@ template<class Output>
 class PYBIND11_EXPORT LocalForceComputeData : public SnapshotLocalDataAccess<Output, ForceCompute>
     {
     public:
-    LocalForceComputeData(ForceCompute& data, ParticleData& pdata)
-        : SnapshotLocalDataAccess<Output, ForceCompute>(data,
-                                                        pdata.getN(),
-                                                        pdata.getNGhosts(),
-                                                        pdata.getNGlobal()),
+    LocalForceComputeData(ForceCompute& data, ssize_t N, ssize_t N_ghost, ssize_t N_global)
+        : SnapshotLocalDataAccess<Output, ForceCompute>(data, N, N_ghost, N_global),
           m_force_handle(), m_torque_handle(), m_virial_handle(),
           m_virial_pitch(data.getVirialArray().getPitch()),
           m_buffers_writeable(data.getLocalBuffersWriteable())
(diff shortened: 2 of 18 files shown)


hoomd/PythonLocalDataAccess.h
```diff
@@ -236,108 +230,121 @@ template<class Output, class Data> class LocalDataAccess
...
-    template<class T, template<class> class U = GlobalArray>
-    Output getGlobalBuffer(std::unique_ptr<ArrayHandle<T>>& handle,
-                           const U<T>& (Data::*get_array_func)() const,
-                           bool read_only = true)
+    // clear should remove any references to ArrayHandle objects so the
+    // handle can be released for other objects.
+    virtual void clear() = 0;
+
+    /// object to access array data from
+    Data& m_data;
```
tommy-waltmann: Usually member variables are all defined in the same location in a class definition.

hoomd/PythonLocalDataAccess.h
```diff
@@ -236,108 +230,121 @@ template<class Output, class Data> class LocalDataAccess
...
+ *  or HOOMDHostBuffer
+ *  Data - the class of the object we wish to expose data from
+ *
+ *  The main methods of LocalDataAccess are getLocalBuffer and getGlobalBuffer
+ *  which provide a way to automatically convert an Global/GPUArray into an
+ *  object of type Output of a MPI local, MPI local with ghosts, or global size
+ *  (for particles bonds, etc.).
+ */
+template<class Output, class Data>
+class SnapshotLocalDataAccess : public LocalDataAccess<Output, Data>
```
tommy-waltmann: This class is used by force classes to grab local/global data and forces are not part of the snapshot. Can we call this class something like `RankLocalDataAccess` or `MPILocalDataAccess` instead of `SnapshotLocalDataAccess`? Just name it something that indicates this class can grab both local and global buffers.

ARCHITECTURE.md
```diff
@@ -261,8 +261,7 @@ two objects. The class is inherited by `_HOOMDBaseObject` to handle dependent
 relationships between operations in Python. The class defines *dependents* and
 *dependencies* of an object whose removal from a simulation (detaching) can be
 handled by overwriting specific methods defined in `_DependencyRelation` in
-`hoomd/operation.py`. See the interface of neighbor lists to pair potentials as
-an example of this in `hoomd/md/nlist.py` and `hoomd/md/pair/pair.py`.
```
tommy-waltmann, line 265: If this example is no longer relevant, can it be replaced with a different one?

tommy-waltmann (changes requested):
Only requesting small changes and had a few questions, otherwise this looks pretty good.

tommy-waltmann:
Also, can tests for the changes to the force buffers (as you mention in the PR description) be added?

ARCHITECTURE.md
b-butler: Unfortunately no. I don't think we use it anywhere else anymore. There are places we should/could use it, but we don't.

hoomd/PythonLocalDataAccess.h
b-butler: Well, `local` and `global` refer to the size of the array not the location, so all these classes provide MPI local access. The difference here is the use of per-X quantities.

hoomd/ForceCompute.h
```diff
@@ -247,52 +237,53 @@ class PYBIND11_EXPORT ForceCompute : public Compute
...
 template<class Output>
-class PYBIND11_EXPORT LocalForceComputeData : public LocalDataAccess<Output, ForceCompute>
+class PYBIND11_EXPORT LocalForceComputeData : public PerGroupLocalDataAccess<Output, ForceCompute>
     {
     public:
-    LocalForceComputeData(ForceCompute& data)
-        : LocalDataAccess<Output, ForceCompute>(data), m_force_handle(), m_torque_handle(),
-          m_virial_handle(), m_virial_pitch(data.getVirialArray().getPitch()),
-          m_buffers_writeable(data.getBuffersWriteable())
+    LocalForceComputeData(ForceCompute& data, ssize_t N, ssize_t N_ghost, ssize_t N_global)
```
joaander: Why a signed `ssize_t`? `GPUArray` and `GlobalArray` return a `size_t` number of elements.

hoomd/md/force.py
```diff
@@ -64,6 +64,10 @@ class Force(Compute):
         for `isinstance` or `issubclass` checks.
     """
 
+    # Which method to call on pybind11 exposed SystemDefinition to get correct
+    # numbers (e.g. getN) for the force. Must be overwritten by other classes.
+    _cpp_data_method = ""
```
joaander: This "fix" introduces a bug. All `ForceCompute` objects compute forces on each particle and store it in a `N_particles` sized array.

hoomd/PythonLocalDataAccess.h
```diff
@@ -236,108 +230,120 @@ template<class Output, class Data> class LocalDataAccess
...
+ *  or HOOMDHostBuffer
+ *  Data - the class of the object we wish to expose data from
+ *
+ *  The main methods of LocalDataAccess are getLocalBuffer and getGlobalBuffer
+ *  which provide a way to automatically convert an Global/GPUArray into an
+ *  object of type Output of a MPI local, MPI local with ghosts, or global size
+ *  (for particles bonds, etc.).
+ */
+template<class Output, class Data>
+class PerGroupLocalDataAccess : public LocalDataAccess<Output, Data>
```
joaander: We use the word `group` to refer to bonded groups, so `PerGroup` sounds like it applies only to those classes. This class provides helpers for arrays with ghost data, so name it `GhostLocalDataAccess`.

hoomd/md/data/local_access.py
```diff
@@ -22,10 +22,13 @@ def _cpp_cls(self):
         'virial': 'getVirial'
     }
 
-    def __init__(self, force_obj):
+    def __init__(self, force_obj, state):
         super().__init__()
         self._force_obj = force_obj
-        self._cpp_obj = self._cpp_cls(force_obj._cpp_obj)
+        data_cls = getattr(state._cpp_sys_def, force_obj._cpp_data_method)()
+        self._cpp_obj = self._cpp_cls(force_obj._cpp_obj, data_cls.getN(),
```
joaander: So `N`, `N_ghosts`, and `N_global` are now passed through on construction. Previously, these were dynamically determined in the `getBuffer` call. I assume this change is OK because the context manager cannot be reused again? For example, the user couldn't construct a context manager, then change the number of particles, the enter the context manager (and seg fault because `N` is no longer valid).

joaander (commented):
Thanks! I have a few questions and comments.

hoomd/md/force.py
b-butler: Good point. For some reason, my mind was thinking that the forces must be per-group, but of course these apply to individual particles ultimately. I will revert this "fix" as you said.

hoomd/PythonLocalDataAccess.h
b-butler: I'll rename it, but  this class also does per-particle buffers regardless of ghosts too.

hoomd/md/data/local_access.py
b-butler: Not without accessing multiple internal (underscored) attributes and essentially ignoring our API.

hoomd/ForceCompute.h
b-butler: There was a narrowing conversion warning somewhere. I forget where now that I am two weeks removed, but I can check.

b-butler:
@tommy-waltmann Force tests are not necessary as the code change was a bug on my part.

joaander (approved):
Looks good, thanks!

tommy-waltmann (approved):
LGTM

PR #1440, opened 2022-11-28
refactor: Automatically update types in ForceComposite::update

b-butler:
## Description

Update types in `ForceComposite::update`.

## Motivation and context

@joaander mentioned in a PR (where I assumed this already happened) that this would be a good change to introduce.

Resolves #???

## How has this been tested?
Existing tests are likely sufficient.

## Change log

```
Changed: `hoomd.md.constrain.Rigid` now updates constituent particle types each step.
```

## Checklist:


tommy-waltmann:
If I understand this PR correctly, it will allow users to change rigid body tags dynamically throughout the simulation. If that is something that was not previously possible, then the current tests likely aren't sufficient to cover the new feature.

b-butler:
Nope, this PR just sets types on the `updateCompositeParticle` step like it already does for positions and orientations. In other words, this means that if the constituent types are `[A, A, B]` if for some reason the type was changed by a users, we would change it back to the rigid body definition.

joaander:
Previously, the code only validated the types were correct once on attaching (and again only when body definitions changed). Now it ensures that the types are always set properly to avoid user errors. This also makes the behavior more consistent: **ALL** keys in `Rigid.body` are enforced by `Rigid` on every timestep.

tommy-waltmann (approved):

PR #1443, opened 2022-11-30
Test Python 3.11 and CUDA 11.8.

joaander:
## Description

Test HOOMD-blue with Python 3.11 and CUDA 11.8.

## Motivation and context

Ensure that HOOMD builds and passes test with the latest versions of these tools.

## How has this been tested?

The pull request checks.

## Change log

```
* Support Python 3.11.
* Support CUDA 11.8.
```

## Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,7 @@ jobs:
     name: Build release tarball
     runs-on: ubuntu-latest
     container:
-      image: glotzerlab/ci:2022.08-ubuntu20.04
+      image: glotzerlab/ci:2022.11-ubuntu20.04
 
     steps:
       - name: Checkout
--- a/.github/workflows/templates/configurations.yml
+++ b/.github/workflows/templates/configurations.yml
@@ -3,14 +3,14 @@
 # - config: a string encoded list containing the docker image name and any number of build options.
 # - The build options are `mpi`, `tbb`, `llvm`, `nomd`, and `nohpmc`.
 unit_test_configurations:
-- config: "[clang14_py310, mpi, tbb, llvm]"
-- config: "[gcc12_py310]"
-- config: "[gcc12_py310, nomd]"
-- config: "[gcc12_py310, nohpmc]"
-- config: "[gcc12_py310, nomd, nohpmc]"
-- config: "[cuda117_gcc11_py310, mpi, llvm, debug]"
-- config: "[cuda117_gcc11_py310, mpi, llvm]"
-- config: "[cuda117_gcc11_py310]"
+- config: "[clang14_py311, mpi, tbb, llvm]"
+- config: "[gcc12_py311]"
+- config: "[gcc12_py311, nomd]"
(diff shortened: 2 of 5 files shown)


joaander:
Closing and reopening to reset GitHub Actions.

.github/workflows/release.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     name: Build release tarball
     runs-on: ubuntu-latest
     container:
-      image: glotzerlab/ci:2022.08-ubuntu20.04
+      image: glotzerlab/ci:2022.11-ubuntu20.04
```
tommy-waltmann, line 35: Why was this change necessary? I see this in a few other places in this PR too.

.github/workflows/templates/configurations.yml
```diff
@@ -3,25 +3,25 @@
...
 # `unit_test_configurations`
 validate_configurations:
-- config: "[clang14_py310, mpi, tbb, llvm]"
-- config: "[gcc12_py310]"
-- config: "[cuda117_gcc11_py310, mpi, llvm]"
-- config: "[cuda117_gcc11_py310]"
+- config: "[clang14_py311, mpi, tbb, llvm]"
+- config: "[gcc12_py311]"
+- config: "[cuda118_gcc11_py310, mpi, llvm]"
+- config: "[cuda118_gcc11_py310]"
```
tommy-waltmann, line 24: Is cuda 11.8 compatible with gcc 12? If so, it might be a good idea to have at least one configuration with the newest versions of cuda, python, and gcc.

tommy-waltmann (commented):
Just a couple of quick questions, otherwise it looks good.

.github/workflows/release.yml
joaander: This pulls in the new Docker containers I built today with Python 3.11 installed: https://github.com/glotzerlab/ci-containers

.github/workflows/templates/configurations.yml
joaander: I don't know if CUDA 11.8 is compatible with gcc 12. 

I use NVIDIA provided CUDA development images (https://hub.docker.com/r/nvidia/cuda/tags) which install CUDA into an Ubuntu with the *default* compiler (gcc10 in this case). It would take some careful package management in https://github.com/glotzerlab/ci-containers to switch these containers to a different gcc and that is not an effort that I have made.

tommy-waltmann: How do you test in environments with cuda11.8 and gcc11 then?

joaander: gcc11 is default in ubuntu 22.04. gcc9 is default in ubuntu 20.04. I mistyped when I mentioned gcc10 was default.
NVIDIA provides CUDA docker images with different ubuntu LTS versions as the base: https://hub.docker.com/r/nvidia/cuda/tags. I typically choose the latest available at the time I build the images.

tommy-waltmann (approved):

PR #1444, opened 2022-11-30
Mixed precision infrastructure

### LINKED ISSUE #355: Mixed precision MD

joaander:
## Description

Add `ShortReal` and `LongReal` typedefs. The user can set the size of these types at compile time.

## Motivation and context

Enable mixed precision in the whole codebase. See https://github.com/glotzerlab/hoomd-blue/issues/355#issuecomment-1332577224 for a discussion on why this doesn't implement mixed precision MD calculations.

Resolves #355

## How has this been tested?

I build, tested, and benchmarked HOOMD locally with sizes `32/32`, `64/32`, and `64/64`.

When `HOOMD_LONGREAL_SIZE==32`:
* HPMC fails to build. (Previously, we forced BUILD_HPMC=off when building with SINGLE_PRECISION=on)
* Numerous pytest tests fail.
* MD simulations run very fast on A40 GPUs.
* I have no intention of fixing these, as I do not support this build configuration.

When `HOOMD_LONGREAL_SIZE==64` and `HOOMD_SHORTREAL_SIZE==32` (the default):
* Behavior is identical to `trunk-major` with `SINGLE_PRECISON=off`.
* CI tests will confirm everything is working.

When `HOOMD_LONGREAL_SIZE==64` and `HOOMD_SHORTREAL_SIZE==64`:
* Some compiler warnings when building.
* HPMC tests seg fault (in cluster moves).
* All other tests pass.
* I will leave fixing the HPMC seg fault for a future PR by an interested stakeholder.

## Change log

```
Changed:

* Control the precision with the CMake options `HOOMD_LONGREAL_SIZE` (default: 64) and `HOOMD_SHORTREAL_SIZE` (default: 32).
* [developers] `ShortReal` and `LongReal` types enable mixed precision implementations.

Removed:

* CMake option `SINGLE_PRECISION`.
* [developers] `SINGLE_PRECISION` compiler definition.

Deprecated:

* [developers] `Scalar`, `Scalar2`, `Scalar3`, and `Scalar4` data types. Use `LongReal[N]` instead in new code.
```

## Checklist:


AlainKadar (approved):

hoomd/HOOMDMath.cc
```diff
@@ -61,7 +61,7 @@ void export_hoomd_math_functions(pybind11::module& m)
             "w",
             [](const double4& s) { return (double)s.w; },
             [](double4& s, double v) { s.w = v; });
-#else
+
     pybind11::class_<float2, std::shared_ptr<float2>>(m, "float2")
         .def(pybind11::init<>())
         .def_property(
```
b-butler: We don't use `float2`, `float3`, or `float4` in Python either.

hoomd/hpmc/HPMCPrecisionSetup.h
```diff
@@ -23,31 +23,9 @@ namespace hoomd
...
-typedef double OverlapReal;
-typedef double3 OverlapReal3;
-typedef double4 OverlapReal4;
-
-#endif
-
-#endif
+typedef ShortReal OverlapReal;
+typedef ShortReal3 OverlapReal3;
+typedef ShortReal4 OverlapReal4;
```
b-butler, line 28: Is there a reason to keep this typedef?

hoomd/BoxDim.h
```diff
@@ -306,22 +306,22 @@ struct
 #ifdef __HIPCC__
         if (m_periodic.z)
             {
-            Scalar img = rint(w.z * m_Linv.z);
+            Scalar img = slow::rint(w.z * m_Linv.z);
```
b-butler, line 309: I know we talked about `rint` why are we using the slow version again?

hoomd/HOOMDMath.cc
```diff
@@ -104,50 +104,6 @@ void export_hoomd_math_functions(pybind11::module& m)
...
-        .def(pybind11::init<>())
-        .def_property(
-            "x",
-            [](const Scalar2& s) { return (Scalar)s.x; },
-            [](Scalar2& s, Scalar v) { s.x = v; })
-        .def_property(
-            "y",
-            [](const Scalar2& s) { return (Scalar)s.y; },
-            [](Scalar2& s, Scalar v) { s.y = v; });
-    pybind11::class_<Scalar3, std::shared_ptr<Scalar3>>(m, "Scalar3")
```
b-butler, line 119: How can we delete this and `make_scalar3` works fine?

hoomd/HOOMDMath.cc
b-butler: Or any of the `char` or `uint`. We use `int3` in `box.py` but given you deleted `Scalar3` this should be fine to remove as well.

b-butler (commented):
In general the PR looks good, I just have some small comments/questions.

hoomd/HOOMDMath.cc
joaander: Python and pybind11 do not understand typedefs - they see only the base type. Previously, the code exported only `double[N]` or `float[N]` depending on the preprocessor definition `SINGLE_PRECISON`. Now the code always exports both. I made this change so that any future code that interoperates with Python using `ShortReal[N]` would function when `HOOMD_SHORTREAL_SIZE=32`.

You are correct that many of these are unused. They are leftover from HOOMD v1 and v2 where the Python interface layer worked more closely with the native C++ types. Now (in many cases) we provide specialized methods for Python that work with `pybind::list` and `pybind11::tuple` in place of these.

I removed all the exports I could without failing tests. It appears that `Box` may be the last location in HOOMD where Python directly works with these C++ vector types. In a future PR we could refactor the Box exports so that uses tuples instead.

hoomd/hpmc/HPMCPrecisionSetup.h
joaander: To temporarily avoid changing 3037 instances of `OverlapReal` in 47 files. We don't know how long until the v4 release, so `trunk-major` will be a long lived branch. Better to not force merge conflicts on any upcoming HPMC development in the 3.x series. We can make this change just prior to a v4.0.0 release.

hoomd/BoxDim.h
joaander: `slow` is misnamed here. Without qualifiers, `rint` calls the *double* precision function [`double rint(double)`](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__DOUBLE.html#group__CUDA__MATH__DOUBLE_1g3b8026edb2f2e441669845f0f3fa3bf7). In builds with `HOOMD_LONGREAL_TYPE=32`, this forces the delta r values to be converted to double, the `rint` called in double, then the result converted back to float. I made this change after I found this performance degradation while profiling.

`slow::rint` calls HOOMD's version of `rint` which uses template specialization to choose between `::rint` and [`::rintf`](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html#group__CUDA__MATH__SINGLE_1g7791cd93108ffc6d24524f2e8635ccfd) (that works on floats). Both of these compile to a single SASS instruction and are fast. However, they are in `hoomd::math::slow` because they are *accurate*. The methods in `hoomd::math::fast` are [approximations on the GPU](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#intrinsic-functions) that are faster, but have reduced precision or are valid only for limited argument ranges.

Perhaps we should rename the hoomd math namespaces to `accurate` and `approximate`, but not in this pull request.

hoomd/HOOMDMath.cc
joaander: Because both `float3` and `double3` are exported. `make_scalar3` takes one of those or the other as an argument, depending on the value of `HOOMD_LONGREAL_SIZE` at build time. `pybind11` resolves typedefs to their actual type and finds that whatever `Scalar3` has been typedefed to exists.

hoomd/HOOMDMath.cc
joaander: [`uint3` is required as well](https://github.com/glotzerlab/hoomd-blue/actions/runs/3621200364/jobs/6104594641#step:6:2807).

joaander: `vec3<Scalar>` is needed for `Integrator::computeLinearMomentum`.

hoomd/hpmc/HPMCPrecisionSetup.h
joaander: #1447 reminds us to make this change.

hoomd/BoxDim.h
b-butler: That sounds good to me.

b-butler (approved):
LGTM

PR #1462, opened 2023-01-13
Box-resize implementation for GPUs

ianrgraham:
## Description

Implementation of the box-resize updater for GPUs.

## Motivation and context

Previously, only a CPU code path existed, leading to a noticeable performance loss on GPUs when using a high frequency trigger. 

## How has this been tested?

Existing tests for the box-resize updater still pass.

## Change log

```
GPU implementation for BoxResizeUpdater
```

## Checklist:


hoomd/BoxResizeUpdaterGPU.h
```diff
@@ -0,0 +1,50 @@
...
+#ifndef __BOX_RESIZE_UPDATER_GPU_H__
+#define __BOX_RESIZE_UPDATER_GPU_H__
+
+namespace hoomd
+    {
+/// Updates the simulation box over time using the GPU
+/** This simple updater gets the box lengths from specified variants and sets
+ * those box sizes over time. As an option, particles can be rescaled with the
+ * box lengths or left where they are. Note: rescaling particles does not work
+ * properly in MPI simulations.
```
joaander:
```suggestion
 * properly in MPI simulations with HPMC.
```

hoomd/BoxResizeUpdaterGPU.cu
```diff
@@ -0,0 +1,77 @@
...
+
+hipError_t gpu_box_resize_updater(const unsigned int N,
+                                  Scalar4* d_pos,
+                                  const BoxDim& cur_box,
+                                  const BoxDim& new_box,
+                                  const unsigned int* d_group_members,
+                                  const unsigned int group_size,
+                                  int3* d_image)
+    {
+    int block_size = 256;
```
joaander: This may be only two small kernels, but we should still autotune the block sizes to ensure the best performance possible:
Example:
https://github.com/glotzerlab/hoomd-blue/blob/0b4b6b54f2ec6acbfdc817215e89d9e9e863655c/hoomd/md/TwoStepNVEGPU.cc#L31-L48

https://github.com/glotzerlab/hoomd-blue/blob/0b4b6b54f2ec6acbfdc817215e89d9e9e863655c/hoomd/md/TwoStepNVEGPU.cc#L93-L111

https://github.com/glotzerlab/hoomd-blue/blob/0b4b6b54f2ec6acbfdc817215e89d9e9e863655c/hoomd/md/TwoStepNVEGPU.cu#L141-L146

You will need to separate `gpu_box_resize_updater` and `gpu_box_wrap_kernel` into two separate driver functions as the `Autotuner` class is only available in .cc files, not .cu.

hoomd/BoxResizeUpdaterGPU.cc
```diff
@@ -0,0 +1,97 @@
...
+    BoxDim new_box = getCurrentBox(timestep);
+
+    // check if the current box size is the same
+    BoxDim cur_box = m_pdata->getGlobalBox();
+
+    // only change the box if there is a change in the box dimensions
+    if (new_box != cur_box)
+        {
+        // set the new box
+        m_pdata->setGlobalBox(new_box);
```
joaander: Avoid duplicating this code by refactoring the base class `update` to separate out the kernel code into a `scaleAndWrapParticles` method. The GPU class will then override `scaleAndWrapParticles` instead of `update`.

joaander (changes requested):
Thanks! This will lead to a good performance boost for certain use-cases. I made several inline suggestions for improvement.

I also added a `BoxResize` microbenchmark to the `hoomd-benchmarks` suite, confirming that your code performs as expected: https://github.com/glotzerlab/hoomd-benchmarks/pull/46

hoomd/BoxResizeUpdaterGPU.cu
ianrgraham: Is it recommended for all new code to use the `GPUPartition` to split up work? Or is this only used in beefy kernels like the integrator?

ianrgraham: Also, do you recommend keeping or omitting the `hipDeviceSynchronize()` call at the end. I added it in caution, and to return the error state, but it will block all kernels?

joaander: I would skip doing GPU partitions. This is a code path that a previous developer added to some kernels to increase performance on the dense OLCF Summit nodes with NVLINK interconnects. I'm not aware of anyone actively using this code path any longer and I don't personally use or test it. Summit will be decommissioned at the end of the year and very few other HPC systems are built with high speed GPU<->GPU interconnects. 

Don't call `hipDeviceSynchronize() or check for error codes` unless `isCUDAErrorCheckingEnabled()` is true. Kernel launch latency and synchronization is expensive and the only way to keep performance high is to launch as many kernels asynchronously as possible. `CHECK_CUDA_ERROR` will call `hipDeviceSynchronize` to check for kernel launch errors.

ianrgraham:
Uhh, I guess my branch wanted to keep the `cub` sub module? Weird that I don't see that from my editor.

joaander:
> Uhh, I guess my branch wanted to keep the `cub` sub module? Weird that I don't see that from my editor.

No worries, there are many cases where it is easy to unintentionally commit changes to submodules, especially when switching branches. The safest thing to do is to remember to `git submodule update` after every `git switch`. Give it a few weeks and then none of the active branches will reference the submodule any longer and it won't be a problem.

joaander (approved):
Thanks!

@yuanzhou0827 please review. Let me know if you have any questions.

yuanzhou0827:
I tested the box resize updater in a test simulation running on GPU and it looks good to me!

yuanzhou0827 (approved):

PR #1464, opened 2023-01-17
Mesh data structure rewrite

SchoeniPhlippsn:
## Description

Major rewrite of the mesh data structure to allow for faster code and multiple triangle types. This will break the current version 3 API.

## Motivation and context

This rewrite is necessary to make the mesh part of Hoomd easier to extend in terms of applied potentials. 

## How has this been tested?

All tests have been updated accordingly. 

## Change log

```
Mesh data structure rewrite
```

## Checklist:


hoomd/MeshDefinition.cc
```diff
@@ -21,19 +21,22 @@ MeshDefinition::MeshDefinition() { }
...
     }
 
 void MeshDefinition::setTypes(pybind11::list types)
     {
-    m_meshbond_data->setTypeName(0, pybind11::cast<string>(types[0]));
-    m_meshtriangle_data->setTypeName(0, pybind11::cast<string>(types[0]));
+    for (unsigned int i = 0; i < len(types); i++)
+        {
+        m_meshbond_data->setTypeName(i, pybind11::cast<string>(types[i]));
+        m_meshtriangle_data->setTypeName(i, pybind11::cast<string>(types[i]));
```
b-butler:
```suggestion
        m_meshbond_data->setTypeName(i, types[i].cast<string>());
        m_meshtriangle_data->setTypeName(i, types[i].cast<string>());
```
a bit less verbose.

hoomd/MeshDefinition.cc
```diff
@@ -59,25 +62,31 @@ TriangleData::Snapshot MeshDefinition::getTriangleData()
...
 //! Triangle array setter
-void MeshDefinition::setTriangleData(pybind11::array_t<int> triangles)
+void MeshDefinition::setTriangleData(pybind11::array_t<int> triangles,
+                                     pybind11::array_t<int> type_ids)
     {
     TriangleData::Snapshot triangle_data = getTriangleData();
-    pybind11::buffer_info buf = triangles.request();
-    int* ptr = static_cast<int*>(buf.ptr);
+    pybind11::buffer_info buf1 = triangles.request();
+    int* ptr1 = static_cast<int*>(buf1.ptr);
```
b-butler:
```suggestion
    const int* ptr1 = triangles.data(0, 0);
```

hoomd/MeshDefinition.cc
```diff
@@ -59,25 +62,31 @@ TriangleData::Snapshot MeshDefinition::getTriangleData()
...
+
+    pybind11::buffer_info buf2 = type_ids.request();
+    int* ptr2 = static_cast<int*>(buf2.ptr);
```
b-butler:
```suggestion
    const int* ptr2 = type_ids.data(0);
```

hoomd/MeshDefinition.cc
```diff
@@ -59,25 +62,31 @@ TriangleData::Snapshot MeshDefinition::getTriangleData()
     }
 
 //! Triangle array setter
-void MeshDefinition::setTriangleData(pybind11::array_t<int> triangles)
+void MeshDefinition::setTriangleData(pybind11::array_t<int> triangles,
```
b-butler: Since you are not checking size this could easily result in a SEGFAULT.

hoomd/MeshGroupData.h
```diff
@@ -75,21 +75,12 @@ class MeshGroupData : public BondedGroupData<group_size, Group, name, true>
...
-    /*! \param tag Tag of particle that moves between domains
-        \param old_rank Old MPI rank for particle
-        \param new_rank New MPI rank
-     */
-    virtual void moveParticleGroups(unsigned int tag, unsigned int old_rank, unsigned int new_rank);
-#endif
-
     private:
-    virtual void rebuildGPUTable();
+    // virtual void rebuildGPUTable();
```
b-butler: Remove if unused.

hoomd/MeshGroupData.h
```diff
@@ -75,21 +75,12 @@ class MeshGroupData : public BondedGroupData<group_size, Group, name, true>
...
 
 #ifdef ENABLE_HIP
     //! Helper function to rebuild lookup by index table on the GPU
-    virtual void rebuildGPUTableGPU();
+    // virtual void rebuildGPUTableGPU();
```
b-butler: Remove if unused.

hoomd/MeshGroupData_old.cc
```diff
@@ -0,0 +1,1033 @@
+// Copyright (c) 2009-2022 The Regents of the University of Michigan.
```
b-butler: My guess is that this shouldn't exist right?

hoomd/md/mesh/potential.py
```diff
@@ -58,6 +58,17 @@ def _apply_typeparam_dict(self, cpp_obj, simulation):
...
+    def all_params(self):
+        """Type parameter setter for all mesh types."""
+        raise RuntimeError("all_params can only be used as a setter.")
+        return []
+
+    @all_params.setter
+    def all_params(self, value):
+        for mt in self.mesh.types:
+            self.params[mt] = value
+
```
b-butler: `TypeParameter` objects support a default value `tp.default = ...` which will be used for all types not explicitly set. This also applies for types where one parameter was set but not another.

hoomd/md/pytest/test_meshpotential.py
```diff
@@ -157,15 +185,55 @@ def test_after_attaching(triplet_snapshot_factory, simulation_factory,
         mesh_bond_potential.mesh = mesh1
 
 
+@pytest.mark.parametrize("mesh_bond_cls, potential_kwargs",
+                         get_mesh_bond_and_args())
+def test_multiple_types(triplet_snapshot_factory, simulation_factory,
+                        mesh_bond_cls, potential_kwargs):
+
+    sim = simulation_factory(triplet_snapshot_factory(d=0.969, L=5))
```
b-butler, line 190:
```suggestion
    sim = simulation_factory(mesh_snapshot_factory(d=0.969, L=5))
```
?

hoomd/mesh.py
```diff
@@ -102,13 +105,33 @@ def triangles(self):
...
     def triangles(self, triag):
         if self._attached:
-            self._cpp_obj.setTypes(list(self._param_dict['types']))
 
-            self._cpp_obj.setTriangleData(triag)
+            if len(triag) != len(self._type_ids):
+                raise ValueError(
+                    "Number of type_ids do not match number of triangles.")
+            else:
+                self._cpp_obj.setTypes(list(self._param_dict['types']))
```
b-butler: Why not pass the NumPy array?

hoomd/mesh.py
```diff
@@ -102,13 +105,33 @@ def triangles(self):
...
+    @log(category='sequence')
+    def type_ids(self):
+        """((*N*) `numpy.ndarray` of ``uint32``): Triangle type ids."""
+        if self._attached:
+            return self._cpp_obj.getTriangleData().typeid
+        return self._type_ids
+
+    @type_ids.setter
+    def type_ids(self, tid):
+        if self._attached and self.size == len(tid):
```
b-butler: Why is the second condition necessary?

hoomd/mesh.py
```diff
@@ -119,7 +142,6 @@ def bonds(self):
         return self._cpp_obj.getBondData().group
 
     def _preprocess_type(self, typename):
-        if not isinstance(typename, str) and isinstance(typename, Sequence):
-            if len(typename) != 1:
-                raise ValueError("Only one meshtype is allowed.")
+        if not isinstance(typename, Sequence):
+            raise ValueError("Expected a sequence of strings.")
```
b-butler: Unnecessary now.

b-butler (changes requested):
Code looks good overall. I requested some changes and had a few questions.

hoomd/MeshDefinition.cc
SchoeniPhlippsn: I think this doesn't work for empty meshes

hoomd/MeshDefinition.cc
SchoeniPhlippsn: I think this doesn't work for empty meshes

hoomd/MeshDefinition.cc
SchoeniPhlippsn: I check for size in `mesh.py`

hoomd/md/pytest/test_meshpotential.py
SchoeniPhlippsn: I think `triplet_snapshot_factory ` is correct

hoomd/mesh.py
SchoeniPhlippsn: Here I check if the number of triangles is equal the number of type_ids

hoomd/mesh.py
SchoeniPhlippsn: I chose it like this to make it consistent with the rest of the code. For example in `ParticleData.cc`.

b-butler: I mean you store it locally as a NumPy array. Consistency would be to expose the attributes and set them as lists. This is internal to us, and seems like it would be simpler to do without the list conversion.

hoomd/mesh.py
b-butler: You don't error in that case you still just set the value in Python which was throwing me off. Also why is this not part of the `ParameterDict`?

hoomd/MeshDefinition.cc
b-butler:
```suggestion
    const int* ptr2 = static_cast<const int*>(type_ids.request().ptr);
```

hoomd/MeshDefinition.cc
b-butler:
```suggestion
    const int* ptr1 = static_cast<const int*>(triangles.request().ptr);
```

hoomd/mesh.py
```diff
@@ -63,9 +62,7 @@ class Mesh(_HOOMDBaseObject):
     def __init__(self):
 
         param_dict = ParameterDict(size=int,
-                                   types=OnlyIf(
-                                       to_type_converter([str]),
-                                       preprocess=self._preprocess_type))
+                                   types=OnlyIf(to_type_converter([str])))
```
b-butler:
```python
types=[str])
```

b-butler (commented):
Thanks for addressing my questions and comments. I left a few more.

hoomd/mesh.py
SchoeniPhlippsn: About not erroring when we set `type_ids`: I don't error it so the user can add a suitable triangulation afterwards. Imagine yu want to delete triangles from your mesh. In that case you have to first set the new triangulation ids without these specific to the removed triangles. Afterwards the user passes the updated triangulation. If I introduce an error, step 1 wouldn't work. I could add a warning though.  

It is also not in `ParameterDict` as the array both dictates the ids of the triangles and edges in the mesh.

Regarding exposing the attribute. I think this is still the best way to avoid to store a `TriangleData::Snapshot`.

hoomd/mesh.py
```diff
@@ -53,28 +51,28 @@ class Mesh(_HOOMDBaseObject):
...
 
     """
 
     def __init__(self):
 
-        param_dict = ParameterDict(size=int,
-                                   types=OnlyIf(
-                                       to_type_converter([str]),
-                                       preprocess=self._preprocess_type))
+        param_dict = ParameterDict(size=int, types=[str])
```
b-butler: Maybe it would be worth having something like
```python
param_dict = ParameterDict(
    types=[str],
    definition=OnlyIf(
        to_type_converter(
            {"typeid": NDArrayValidator(...), "triangles": NDArrayValidator(...)}
        ), postprocessing=ensure_same_size
    )
)
This means that users can change size if they wish but we would error if they did it one at a time. What do you think?

Comment (SchoeniPhlippsn) [reply hoomd/mesh.py]: I like the idea. But does putting the triangle data into `ParameterDict` keep them loggable?

Comment (b-butler) [reply hoomd/mesh.py]: You would need to then add a loggable entry either directly upon construction or through a property. I forget these were loggable. Having read only properties though would allow logging, and I think we do something similar elsewhere but I forget where exactly.

We are sort of in a grey zone because the quantities here are user defined and don't change during the simulation (I think) so it isn't a normal logged quantity.

Comment (SchoeniPhlippsn) [reply hoomd/mesh.py]: Well, not here but later with the dynamical bonding updaters the triangles do change.  But I might add two additional properties to make them loggable.

File: hoomd/mesh.py [hunk 15]
```diff
@@ -53,19 +54,25 @@ class Mesh(_HOOMDBaseObject):
...
 
     def __init__(self):
 
-        param_dict = ParameterDict(size=int, types=[str])
+        param_dict = ParameterDict(
+            types=[str],
+            size=int,
+            triangulation=OnlyIf(to_type_converter({
+                "type_ids": NDArrayValidator(np.uint),
+                "triangles": NDArrayValidator(np.uint, shape=[None, 3])
```
b-butler:
```suggestion
                "triangles": NDArrayValidator(np.uint, shape=(None, 3))
```

hoomd/mesh.py
```diff
@@ -53,19 +54,25 @@ class Mesh(_HOOMDBaseObject):
...
+            }),
+                                 allow_none=True))
```
b-butler:
```suggestion
                                 postprocess=self._ensure_same_size,
                                 allow_none=True))
```

hoomd/mesh.py
```diff
@@ -86,46 +95,48 @@ def _attach_hook(self):
...
 
-                self._cpp_obj.setTriangleData(triag, self._type_ids)
+    def _set_triangulation(self, value):
+        if self._attached:
+            self._cpp_obj.setTriangleData(value["triangles"], value["type_ids"])
         else:
-            self.size = len(triag)
-        self._triangles = triag
+            self._param_dict._dict["triangulation"] = value
+            self.size = len(value["triangles"])
```
b-butler: The parameter dict should be capable of handling most of this directly (besides the size parameter) which could be a property.

hoomd/mesh.py
```diff
@@ -86,46 +95,48 @@ def _attach_hook(self):
...
+            return
+        super()._setattr_param(attr, value)
 
-    @triangles.setter
-    def triangles(self, triag):
-        if self._attached:
+    def _getattr_param(self, attr):
+        if attr == "triangulation":
+            return self._get_triangulation()
+        return super()._getattr_param(attr)
```
b-butler: I think these should be unnecessary.

hoomd/mesh.py
```diff
@@ -86,46 +95,48 @@ def _attach_hook(self):
...
 
-            if len(triag) != len(self._type_ids):
-                raise ValueError(
-                    "Number of type_ids do not match number of triangles.")
-            else:
-                self._cpp_obj.setTypes(list(self._param_dict['types']))
+    def _get_triangulation(self):
+        if self._attached:
+            return dict(type_ids=self.type_ids, triangles=self.triangles)
+        return self._param_dict._dict["triangulation"]
```
b-butler: Same as below. This should be able to be handled by the ParameterDict automatically.

hoomd/mesh.py
```diff
@@ -86,46 +95,48 @@ def _attach_hook(self):
...
+            self.size = len(value["triangles"])
 
-    @log(category='sequence')
+    @log(category='sequence', requires_run=True)
     def type_ids(self):
         """((*N*) `numpy.ndarray` of ``uint32``): Triangle type ids."""
-        if self._attached:
-            return self._cpp_obj.getTriangleData().typeid
-        return self._type_ids
+        return self._cpp_obj.getTriangleData().typeid
```
b-butler:
```suggestion
        return self.triangulation["typeid"]

File: hoomd/mesh.py [hunk 21]
```diff
@@ -86,46 +95,48 @@ def _attach_hook(self):
...
+    @log(category='sequence', requires_run=True)
+    def triangles(self):
+        """((*N*, 3) `numpy.ndarray` of ``uint32``): Mesh triangulation.
 
-            self._cpp_obj.setTriangleData(self._triagles, tid)
-        self._type_ids = tid
+        A list of triplets of particle tags which encodes the
+        triangulation of the mesh structure.
+        """
+        return self._cpp_obj.getTriangleData().group
```
b-butler:
```suggestion
        return self.triangulation["triangles"]
```

b-butler (changes requested):
The `ParameterDict` class should be able to handle most of your needs if you export a `triangulation` property from the C++ class. Including automatic getting and setting from the C++ side.

hoomd/MeshDefinition.cc
```diff
@@ -61,10 +61,44 @@ TriangleData::Snapshot MeshDefinition::getTriangleData()
...
+    {
+    pybind11::dict triangulation;
+
+    TriangleData::Snapshot triangle_data = getTriangleData();
+
+    unsigned int len_triang = triangle_data.getSize();
+
+    std::vector<size_t> dims(2);
+    dims[0] = len_triang;
+    dims[1] = 3;
```
b-butler:
```suggestion
    std::vector<size_t> dims{len_triang, 3};
```

hoomd/MeshDefinition.h
```diff
@@ -15,6 +15,7 @@
 
 #include <memory>
 #include <pybind11/pybind11.h>
+// #include <pybind11/numpy.h>
```
b-butler:
```suggestion
```

hoomd/MeshDefinition.h
```diff
@@ -109,7 +110,9 @@ class PYBIND11_EXPORT MeshDefinition
 
     TriangleData::Snapshot getTriangleData();
 
-    void setTriangleData(pybind11::array_t<int> triangles, pybind11::array_t<int> type_ids);
+    pybind11::object getTriangulationData(); // pybind11::object self);
```
b-butler:
```suggestion
    pybind11::object getTriangulationData();
```

hoomd/mesh.py
```diff
@@ -82,52 +82,31 @@ def _attach_hook(self):
...
         self._cpp_obj.setTypes(list(self._param_dict['types']))
-        if self._param_dict._dict["triangulation"] is not None:
-            self._set_triangulation(self._param_dict._dict["triangulation"])
 
         if hoomd.version.mpi_enabled:
             pdata = self._simulation.state._cpp_sys_def.getParticleData()
             decomposition = pdata.getDomainDecomposition()
             if decomposition is not None:
                 # create the c++ Communicator
+                self._cpp_obj.setTriangulation(
```
b-butler: Oh was this the problem with the communicator?

b-butler (commented):
A few comments for clean up but otherwise looks good.

hoomd/mesh.py
SchoeniPhlippsn: Yes, that was the issue!

SchoeniPhlippsn:
I fixed the MPI issue, I talked about earlier this week. This PR can be reviewed again!

hoomd/mesh.py
```diff
@@ -128,7 +123,7 @@ def bonds(self):
 
     @property
     def size(self):
-        """(`uint32``): Number of triangles in the mesh."""
+        """(``uint32``): Number of triangles in the mesh."""
```
b-butler:
```suggestion
        """(int): Number of triangles in the mesh."""
```
Unless this is a NumPy type the type is just int in Python.

b-butler (commented):
Looks good. Just change the typing before merging.

b-butler (approved):

joaander (approved):
Thanks! The new code is much simplified compared to the original while also adding functionality.

PR #1485, opened 2023-02-07
fix: Add deprecation warnings to deprecated features.

b-butler:
## Description
So `gh` removes all data when moving to the browser... This adds deprecation warnings.

## Motivation and context
Will warn those who test hoomd code.

## How has this been tested?
No need beyond existing tests.

## Change log
NA

## Checklist:

### DIFF
--- a/sphinx-doc/migrating.rst
+++ b/sphinx-doc/migrating.rst
@@ -3,6 +3,7 @@
 
 Note about Warnings
 ===================
+
 HOOMD developers may use ``FutureWarnings`` to provide warnings for breaking changes in the next
 major release. For users who wish to silence these warnings which may not be correctable until the
 next major release use Python's warnings module. Specifically, this filter will work,
--- a/sphinx-doc/migrating.rst
+++ b/sphinx-doc/migrating.rst
@@ -1,17 +1,6 @@
 .. Copyright (c) 2009-2023 The Regents of the University of Michigan.
 .. Part of HOOMD-blue, released under the BSD 3-Clause License.
 
-Note about Warnings
-===================
-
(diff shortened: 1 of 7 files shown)


joaander (changes requested):
Thanks! Shouldn't these all be `FutureWarning` as you have said previously?

joaander:
Also, some tests are failing: https://github.com/glotzerlab/hoomd-blue/actions/runs/4118267747/jobs/7110757262#step:6:5523

b-butler:
Depends. Do we want to warn users for something they can't fix? My opinion is to error on that side, but I know that other HOOMD developers such as yourself disagree, so I don't want to just force my opinion through.

joaander:
I have stated that I don't want to provide non-silencable warnings. You mentioned that the user can control the Python warning subsystem to silence `FutureWarning`. If the code to do so is not obvious in the documentation (https://docs.python.org/3/library/warnings.html#module-warnings) then we should provide a how to guide in HOOMD.

I agree that `DeprecationWarning` is useless as few, if any, HOOMD users will run their scripts with the proper environment variables, command line options, or warnings filter code to display those.

sphinx-doc/migrating.rst
```diff
@@ -1,6 +1,16 @@
 .. Copyright (c) 2009-2023 The Regents of the University of Michigan.
 .. Part of HOOMD-blue, released under the BSD 3-Clause License.
 
+Note about Warnings
+===================
```
b-butler:
```suggestion
===================

```

joaander (approved):
Thanks! 

I applied minor edits directly to the branch.

PR #1487, opened 2023-02-09
Virtual sites

### LINKED ISSUE #1486: Virtual Sites
### Description

Moving forward to martini v3 will require that I use virtual sites for my own simulations, which are currently not available in hoomd.

### Proposed solution

Defining a virtual site is very similar to a bonded type: a list of particles from which the site is derived and the class of site (e.g. 2 particles linear, etc.). Additionally, a particle type for non-bonded interactions with the virtual site has to be supplied. 

Implementation-wise, this can be done easily with an almost carbon-copy of the rigid bodies. Instead of decomposing multiple particles onto the center particle, we decompose the virtual site onto the non-virtual particles. Each "type" of virtual site can be created as a template. This will be supplied shortly. 

### Additional context

_No response_

MartinGirard:
## Description
Adds virtual site functionality, with implementation derived from rigid bodies. 

## Motivation and context

I need this for martini v3. Some users may enjoy it for similar reasons. 

Resolves #1486

## How has this been tested?
Currently untested, best test needs to be discussed and afterwards implemented. 

## Checklist:


joaander:
Thanks for working on this, many HOOMD users will find this useful.

I need to carefully evaluate how the rest of HOOMD's code uses the `body` tag and see how we can make the new sentinel value interoperate with that framework.

MartinGirard:
@joaander Alternatively, we can make a new particle property, but it'd be just a flag; mainly so that virtual particles are not integrated (as for constituant particles), and that virtual particles are not used as a basis for further virtual particles. There's less internal consistency that needs to be maintained here.

joaander:
Using the `body` tag is fine, there no need for the overhead of a new per-particle quantity for this. Either way, we need to account for the particles correctly in all appropriate conditions.

MartinGirard:
@joaander I've pushed a potential gpu implementation of the kernels (w/o the book keeping yet). Its overly simple and easy to extend,  but I'm not sure its the most efficient in terms of computation efficiency, in particular for the force decomposition. At the moment, a single thread computes all of the projections from the virtual particle to the real particles (from which the v-site is derived). Virtual sites are derived from 2 - 4 real particles. 

The alternative is to make each thread compute a single projection. For linear combinations, this is somewhat trivial to do without branching. For non-linear site combinations, each thread would have a different functional form for the projection, which would lead to branching. 

A potential solution to this issue is to batch virtual sites so that each warp computes a single projection type. Taking an example where the virtual particle is created from 3 real particles, and there is M such virtual sites. Warp0 would compute the projection of the first 32 virtual particles onto their first real particles. Warp 1 would compute the projection of the same 32 virtual particles onto their second real particles. There would be multiple memory loads from the same memory location across the warps, which I assume would then be in cache. Thoughts?

joaander:
In general, computation and branching is inexpensive while memory reads and writes are very expensive - especially atomic operations with many collisions and scattered writes. In specific cases, one can profile the kernel with Nsight compute and use the performance counters to evaluate whether the kernel is compute or memory limited and identify the hotspots with line level details. The ldg cache can help with reads, but understand that warps are asynchronous and with many warps in flight the cache turns over very quickly.

How long does your implementation take to run for a typical simulation condition? If it is only tens of microseconds per kernel call, then there is very little to gain from optimization.

joaander:
Currently, HOOMD has the concept of *central*, *free*, and *constituent* particles. (There is also *floppy*, but it is unused). This PR adds *site* particles. The new *site* particles are not **integrable**. This matters both in error checking the integration methods (ongoing bugfix for in #1503), and in counting the number of degrees of freedom for the kinetic temperature computation (recently fixed in #1488): https://github.com/glotzerlab/hoomd-blue/blob/fbe46e9b1a321940b9b37bb016063747cdffcd58/hoomd/md/IntegratorTwoStep.cc#L161-L196

I think the right solution is to:
1) Provide helper methods for testing whether a particle is *central*, *free*, *constituent*, *integrable*, and in this PR add *site*.
2) Use these helpers throughout the code.
3) In many cases where we currently check for the union of *central* and *free*, replace with an *integrable* check.

I will implement these in another pull request. 

With those changes in place, this PR then will need to update the existing helpers if needed, add the *site* helper and modify the `validateGroup` and `getTranslationalDOF` methods appropriately. Some care will be needed to ensure that `getTranslationalDOF` counts correctly in all combinations of rigid/no rigid and virtual/no virtual.

MartinGirard:
@joaander That seems reasonable, and will simplify any further implementation of methods with non-integrable particles. 
There's another side note; the initial implementation I had, based on MolecularForceCompute, does not work if multiple virtual particles are constructed using a given particle. It'll take me some time to redesign.

PR #1500, opened 2023-02-17
File object notice stream

### LINKED ISSUE #1449: File-like object for notice stream
### Description

Provide a Python file-like object that writes to the HOOMD notice message stream.

### Proposed solution

```
device = hoomd.device.CPU()
notice_file = device.make_notice_file(notice_level=1)
notice_file.write("Message\n")

table_writer = hoomd.write.Table(trigger=hoomd.trigger.Periodic(4000),
                                                         logger=logger,
                                                         output=notice_file)
```

### Additional context

This will allow users to write [`Table` output](https://hoomd-blue.readthedocs.io/en/v3.7.0/module-hoomd-write.html#hoomd.write.Table) to the same output file as [`notice`](https://hoomd-blue.readthedocs.io/en/v3.7.0/module-hoomd-device.html#hoomd.device.Device).

I would use this to include status updates with Table mixed with other status messages printed by `device.notice()` into a separate file for each partition in an aggregate run. `stdout` is mixed from all partitions in aggregate runs.

`notice` accepts one line per call, while `file.write()` accepts arbitrary input. To properly support the file-like protocol, this shim class will need to buffer input, split on lines and issue one `notice()` call per line. 

My main interest for this is adding `notice()` support to `writer.Table`, so a simpler implementation would be to instead make this an option in `Table`:
```
table_writer = hoomd.write.Table(trigger=hoomd.trigger.Periodic(4000),
                                                         logger=logger,
                                                         output='notice')
```
When `output == 'notice'`, `Table` would call `device.notice` instead of `output.write`.

iblanco11981870:
## Description

Provide a Python file-like object that writes to the HOOMD notice message stream
## Motivation and context
This will allow user to write Table output to the same output file as notice.

It allows the user with more flexibility to write on Table output

Resolves #1449 

## How has this been tested?
We implemented the appropriate tests in ``test_table.py`` 

We parameterized the relevant tests to also test the new implementation of notice. 

## Change log

```
-Added the functionality, tests and documentation needed to recognize device from the string 'notice'
```

## Checklist:

### DIFF
--- a/hoomd/write/table.py
+++ b/hoomd/write/table.py
@@ -176,24 +176,30 @@ class _TableInternal(_InternalAction):
 
     def __init__(self,
                  logger,
-                 output=stdout,
+                 output=stdout or str,
                  header_sep='.',
                  delimiter=' ',
                  pretty=True,
                  max_precision=10,
                  max_header_len=None):
 
-        param_dict = ParameterDict(header_sep=str,
-                                   delimiter=str,
-                                   min_column_width=int,
-                                   max_header_len=OnlyTypes(int,
-                                                            allow_none=True),
-                                   pretty=bool,
-                                   max_precision=int,
-                                   output=OnlyTypes(
-                                       _OutputWriter,
-                                       postprocess=_ensure_writable),
-                                   logger=Logger)
+        if output == 'notice':
+            output_type = str
+        else:
+            output_type = OnlyTypes(_OutputWriter, postprocess=_ensure_writable)
+
+        param_dict = ParameterDict(
+            header_sep=str,
+            delimiter=str,
+            min_column_width=int,
+            max_header_len=OnlyTypes(int, allow_none=True),
+            pretty=bool,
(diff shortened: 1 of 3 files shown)


hoomd/write/table.py
```diff
@@ -176,24 +176,30 @@ class _TableInternal(_InternalAction):
 
     def __init__(self,
                  logger,
-                 output=stdout,
+                 output=stdout or str,
```
b-butler, line 179: This is wrong. or is a Boolean operator and will make output either `True` or `False`.
```suggestion
                 output=stdout
```

b-butler (changes requested):
I would prefer if you created a file handle like object that a device could give rather than add this to the `Table` writer directly.

joaander (commented):
I agree with @b-butler. You should be able to use the updated test framework here with little modification, undo the changes to `write/table.py` and implement the `NoticeFile` object.

PR #1519, opened 2023-03-15
Added different error messages to table writer

### LINKED ISSUE #1510: Add different error messages in Table Writer
### Description

More descriptive error messages about why a logger is not excepted would be helpful.

### Proposed solution

Add messages that mention when an unsupported logger category is active.

### Additional context

_No response_

janbridley:
## Description

- Improved error logic for setting ``LoggerCategories`` in ``hoomd.write.Table``.
- Added new error message that details if invalid categories have been set

## Motivation and context

Prior to this change, users adding disallowed categories to ``hoomd.write.Table`` would receive an error intended for when ``LoggerCategories.scalar`` was not set. This fix updates the error logic to raise a more descriptive error detailing which disallowed categories have been set by the user.

Resolves #1510 

## How has this been tested?

Changes were tested with every permutation of LoggerCategories, which return the expected errors.

## Change log

```
Fixed:

* hoomd.write.Table now raises correct error if invalid LoggerCategories are set.
```

## Checklist:

### DIFF
--- a/hoomd/write/table.py
+++ b/hoomd/write/table.py
@@ -211,7 +211,6 @@ class _TableInternal(_InternalAction):
         _invalid_inputs = logger.categories & self._invalid_logger_categories
 
         # Ensure that only scalar and potentially string are set for the logger
-        # If logger.categories == none, pass automatically
         if logger.categories == LoggerCategories.NONE:
             pass
         elif LoggerCategories.any([
--- a/hoomd/write/table.py
+++ b/hoomd/write/table.py
@@ -208,13 +208,21 @@ class _TableInternal(_InternalAction):
 
         # internal variables that are not part of the state.
         # Ensure that only scalar and potentially string are set for the logger
-        if LoggerCategories.scalar & LoggerCategories.string not in logger.categories:
-            raise ValueError("Given Logger must have the scalar or string categories set.")
-        elif (
-            logger.categories & self._invalid_logger_categories != LoggerCategories.NONE
+        if (
+            LoggerCategories.scalar & LoggerCategories.string
+            not in logger.categories
         ):
-            # logger.categories != NONE passes if the logger is empty (no invalid categories)
-            # invalid_categories != NONE passes if the 
+            raise ValueError(
+                "Given Logger must have the scalar or string categories set."
+            )
+        elif (
+            logger.categories & self._invalid_logger_categories
+            != LoggerCategories.NONE
+        ):
+            # If logger.categories != NONE passes:
+            # -  Logger is empty (no invalid categories)
+            # If invalid_categories != NONE passes: 
+            # - No invalid categories are set
             raise ValueError(
                 "{} are incompatible with write.Table: use write.GSD instead.".format(
                     logger.categories & self._invalid_logger_categories
--- a/hoomd/write/table.py
+++ b/hoomd/write/table.py
@@ -205,29 +205,24 @@ class _TableInternal(_InternalAction):
                  output=output,
                  logger=logger))
         self._param_dict = param_dict
-
         # internal variables that are not part of the state.
+
+        # Generate list of current invalid categories
+        _invalid_inputs = logger.categories & self._invalid_logger_categories
+
         # Ensure that only scalar and potentially string are set for the logger
-        if (
-            LoggerCategories.scalar & LoggerCategories.string
-            not in logger.categories
-        ):
+        if (LoggerCategories.scalar & LoggerCategories.string
+                not in logger.categories):
             raise ValueError(
-                "Given Logger must have the scalar or string categories set."
-            )
-        elif (
-            logger.categories & self._invalid_logger_categories
-            != LoggerCategories.NONE
-        ):
+                "Given Logger must have the scalar or string categories set.")
+        elif _invalid_inputs != LoggerCategories.NONE:
             # If logger.categories != NONE passes:
-            # -  Logger is empty (no invalid categories)
-            # If invalid_categories != NONE passes: 
+            # - Logger is empty (no invalid categories)
+            # If invalid_categories != NONE passes:
             # - No invalid categories are set
             raise ValueError(
-                "{} are incompatible with write.Table: use write.GSD instead.".format(
-                    logger.categories & self._invalid_logger_categories
-                )
-            )
+                "{} are incompatible with write.Table: use write.GSD instead."
+                .format(_invalid_inputs))
 
         self._cur_headers_with_width = dict()
         self._fmt = _Formatter(pretty, max_precision)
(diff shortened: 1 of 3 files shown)


hoomd/write/table.py
```diff
@@ -208,12 +208,16 @@ def __init__(self,
...
         # Ensure that only scalar and potentially string are set for the logger
-        if (LoggerCategories.scalar not in logger.categories
-                or logger.categories & self._invalid_logger_categories
-                !=  # noqa: W504 (yapf formats this incorrectly
-                LoggerCategories.NONE):
+        if LoggerCategories.scalar not in logger.categories:
+            raise ValueError("Given Logger must have the scalar categories set.")
+        elif (
+            logger.categories & self._invalid_logger_categories != LoggerCategories.NONE
+        ):
```
tommy-waltmann: Is the boolean expression in the elif condition of the if-elif general enough to cover the boolean expression in the if condition?

tommy-waltmann (commented):
Great work @janbridley . I wanted to bring one point up for consideration which may make the if-elif logic simpler.

hoomd/write/table.py
```diff
@@ -208,12 +208,16 @@ def __init__(self,
 
         # internal variables that are not part of the state.
         # Ensure that only scalar and potentially string are set for the logger
-        if (LoggerCategories.scalar not in logger.categories
-                or logger.categories & self._invalid_logger_categories
-                !=  # noqa: W504 (yapf formats this incorrectly
-                LoggerCategories.NONE):
+        if LoggerCategories.scalar not in logger.categories:
```
joaander: Why not accept a logger that only takes `string` categories?

janbridley: I'm not sure - I retained the original behavior (that required ``scalar`` categories to be set) but I don't see why a ``string``-only logger wouldn't work.  @b-butler, do you have any insight into why it was initially set up this way? If you don't recall any issues, I will refine the logic to allow ``string``-only loggers.

b-butler: We should. I just didn't originally for who knows why. @janbridley please change this here to fix my past mistake.

hoomd/write/table.py
janbridley: Yes, as far as I am aware: the cases are as follows.

**If scalar category is not set:** the logger fails. This is in discussion [below](https://github.com/glotzerlab/hoomd-blue/pull/1519#discussion_r1138716821), but would work the same way if the case checked for scalar OR string

**If there are logger categories AND invalid logger categories:** the logger fails as intended.

**Other cases:**

- **There are no logger categories:** logger passes, but provides no output (expected)
- **There are logger categories but no invalid categories:** logger passes.
- **There are ONLY invalid categories:** this is the same as the first case, and will fail because there is no scalar category set.
- **There is only the string category:** this is also the same as the first case, and fails because scalar is not set

joaander: It would be awesome if you added a pytest unit test that checks these.

janbridley: Added - see [68df2a3](https://github.com/glotzerlab/hoomd-blue/pull/1519/commits/68df2a30053109ba1eef7edd49e10a6cfcc373a4)

hoomd/write/table.py
janbridley: Done! See [d4a292b](https://github.com/glotzerlab/hoomd-blue/pull/1519/commits/d4a292b17cfa118d77cffee53f558b0ffd7ea953)

hoomd/write/table.py
joaander: Thanks!

joaander (approved):
Thanks, this looks good to me!

tommy-waltmann (approved):

hoomd/write/table.py
```diff
@@ -205,15 +205,28 @@ def __init__(self,
...
+
+        # Generate list of current invalid categories
+        _invalid_inputs = logger.categories & self._invalid_logger_categories
+
         # Ensure that only scalar and potentially string are set for the logger
-        if (LoggerCategories.scalar not in logger.categories
-                or logger.categories & self._invalid_logger_categories
-                !=  # noqa: W504 (yapf formats this incorrectly
-                LoggerCategories.NONE):
+        # If logger.categories == none, pass automatically
```
b-butler: This comment just repeats the next line. Clarification comments like this are good when the next line(s) is/are difficult to parse or call some obscure function.
```suggestion
```

hoomd/write/table.py
```diff
@@ -205,15 +205,28 @@ def __init__(self,
...
+                "Table Logger may only have scalar or string categories set.")
+        elif _invalid_inputs != LoggerCategories.NONE:
+            # If logger.categories != NONE passes:
+            # - Logger is empty (no invalid categories)
+            # If invalid_categories != NONE passes:
+            # - No invalid categories are set
             raise ValueError(
-                "Given Logger must have the scalar categories set.")
+                "{} are incompatible with write.Table: use write.GSD instead."
+                .format(_invalid_inputs))
```
b-butler: I still don't see the need for two warnings at this point maybe something like `Table Logger may only have scalar or string categories set. Use hoomd.write.GSD for other categories.`. Also if a user only asks to log invalid categories, the error does not mention this but instead tells them to add valid categories.

b-butler (changes requested):

b-butler:
I requested changes to prevent a pre-mature merging as I wanted to get @joaander and @tommy-waltmann thoughts on my comment above.

joaander:
> I requested changes to prevent a pre-mature merging as I wanted to get @joaander and @tommy-waltmann thoughts on my comment above.

I don't have a preference. The docs and tutorials are very clear, but if a user finds more detailed error messages helpful at the cost of a few lines of Python code then I'm fine with that.

hoomd/write/table.py
janbridley: If we want to condense to a single error message, we could use the following code: 
 
```python
# Generate LoggerCategories for valid and invalid categories
_valid_categories = LoggerCategories.any(
    [LoggerCategories.scalar, LoggerCategories.string])
_invalid_inputs = logger.categories & self._invalid_logger_categories

if logger.categories == LoggerCategories.NONE:
    pass
elif (_valid_categories ^ LoggerCategories.ALL
      ) & logger.categories == LoggerCategories.NONE:
    pass
else:
    raise ValueError(
        "Table Logger may only have scalar or string categories set. \
            Use hoomd.write.GSD for {}.".format(_invalid_inputs))
```

~~This may be possible without converting the input categories to string...~~

Code has been updated to directly compare logger categories!

tommy-waltmann: I don't have a strong preference either way on this.

b-butler (approved):

PR #1529, opened 2023-03-30
Autotune hard-coded kernel block sizes

### LINKED ISSUE #1374: Use Autotuners instead of `m_block_size` in kernels.
## Description

Replace the uses of fixed `m_block_size` with an `Autotuner`.

On first check, this will require some additional work. For example, `TwoStepLangevinGPU` will also need to dynamically resize the corresponding scratch buffer and will need to call the tally kernel from a separate driver function so that it can be tuned independently.

## Motivation and context

Improve performance by tuning block sizes for these kernels.

## Classes to update:
Found by grep:

```
$ rg "unsigned int m_block_size"                                                                                                          [refactor-autotuner]
hoomd/md/ComputeThermoGPU.h
41:    unsigned int m_block_size; //!< Block size executed

hoomd/md/TwoStepLangevinGPU.h
40:    unsigned int m_block_size;       //!< block size for partial sum memory

hoomd/md/TwoStepRATTLELangevinGPU.h
55:    unsigned int m_block_size;       //!< block size for partial sum memory

hoomd/md/FIREEnergyMinimizerGPU.h
43:    unsigned int m_block_size; //!< block size for partial sum memory

hoomd/md/TwoStepRATTLEBDGPU.h
54:    unsigned int m_block_size; //!< block size

hoomd/md/TwoStepBerendsenGPU.h
43:    unsigned int m_block_size; //!< Block size to launch on the GPU

hoomd/md/PPPMForceComputeGPU.h
135:    unsigned int m_block_size;                //!< Block size for fourier mesh reduction

hoomd/md/ComputeThermoHMAGPU.h
39:    unsigned int m_block_size;       //!< Block size executed
```

tommy-waltmann:
## Description

This PR autotunes many GPU kernels for which the block size was hard-coded.

## Motivation and context

All kernel block sizes should be autotuned since we don't know a priori what the block size with the highest performance will be.

Resolves #1374 

## How has this been tested?

~~Existing tests should suffice. I can ensure locally that the block sizes change as the simulation starts running.~~

I have added tests to make sure the new autotuners work properly.

## Change log

```
Fixed:
  - All GPU kernels autotune their block sizes. 
```

## Checklist:

- [ ] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/CONTRIBUTING.rst).


tommy-waltmann:
still working on this

janbridley (approved):
Clear code - looks good to me!

tommy-waltmann:
It is worth mentioning here that I was not able to autotune kernels for `PPPMForceCompute` (because the calculation is complex) or `FIREEnergyMinimizer` (due to current design around updating the integration methods list).

hoomd/md/ComputeThermoGPU.cc
```diff
@@ -36,7 +36,10 @@ ComputeThermoGPU::ComputeThermoGPU(std::shared_ptr<SystemDefinition> sysdef,
         throw std::runtime_error("Error initializing ComputeThermoGPU");
         }
 
-    m_block_size = 256;
+    m_tuner.reset(new Autotuner<1>({AutotunerBase::makeBlockSizeRange(m_exec_conf)},
```
joaander: The partial sum kernel assumes that block sizes are a power of 2:
https://github.com/glotzerlab/hoomd-blue/blob/6c7eb5373665360d75abeef7d7d4ad10beb35ecb/hoomd/md/ComputeThermoGPU.cu#L98-L110

Also, the final sum kernel can handle any (power of 2) block size regardless of the number of partial sums:
https://github.com/glotzerlab/hoomd-blue/blob/6c7eb5373665360d75abeef7d7d4ad10beb35ecb/hoomd/md/ComputeThermoGPU.cu#L342-L384
This is needed in cases where the number of particles is very large and the number of partial sums is therefore larger than the maximum block size on the GPU.

However, the timing of the second kernel will depend on the block size for the first (as that sets the number of partial sums). Therefore, we should use a 2 dimensional autotuner where the parameters are 1) the block size of the partial sum kernel and 2) the block size of the final sum kernel. It should include all combinations of two values from valid power of 2 block sizes (from the warp size on up, use the device properties to get the warp size). The tuner `end` call should be after the final sum.

hoomd/md/ComputeThermoGPU.cu
```diff
@@ -676,11 +676,10 @@ hipError_t gpu_compute_thermo_final(Scalar* d_properties,
...
 
     // setup the grid to run the final kernel
-    int final_block_size = 256;
-    dim3 grid = dim3(1, 1, 1);
-    dim3 threads = dim3(final_block_size, 1, 1);
+    dim3 grid = dim3(args.n_blocks, 1, 1);
+    dim3 threads = dim3(args.block_size, 1, 1);
 
-    size_t shared_bytes = sizeof(Scalar4) * final_block_size;
+    size_t shared_bytes = sizeof(Scalar4) * args.block_size;
```
joaander: The block sizes chosen by the autotuner go up to the maximum possible on the GPU. This may be larger than the maximum possible for a specific kernel due to register and/or shared memory usage. Query the maximum and clamp the autotuned block size to this value. See `ConstantVolume` for an example.
https://github.com/glotzerlab/hoomd-blue/blob/6c7eb5373665360d75abeef7d7d4ad10beb35ecb/hoomd/md/TwoStepConstantVolumeGPU.cu#L116-L121

Need to apply this to both the partial and final sum kernels. Also note that because these kernels require a power of 2 block size, you need to clamp the maximum given to you by the driver to the nearest power of 2 lower than or equal to the given value.

hoomd/md/ComputeThermoHMAGPU.cc
joaander, line 1: Apply the same changes suggested to `ComputeThermoGPU`.

hoomd/md/TwoStepBerendsenGPU.cc
joaander, line 1: No need to change `Berendsen`, these files no longer exist in v4.

hoomd/md/TwoStepLangevinGPU.cu
joaander, line 1: Clamp Langevin kernel block sizes to the maximum. Also, when `tally` is true, block sizes for the step two must be chosen as a power of 2.

hoomd/md/TwoStepRATTLEBDGPU.cuh
joaander, line 1: Clamp RATTLE BD block sizes to the maximum.

hoomd/md/TwoStepRATTLELangevinGPU.cuh
joaander, line 1: Clamp the block sizes to their maximums. Also, when `tally` is true, block sizes for step two must be chosen as a power of 2.

joaander (changes requested):
Thanks for working on this! On review, I noticed some subtleties to these kernels that were not documented in comments. See the inline PR comments for details.

hoomd/md/ComputeThermoGPU.cu
tommy-waltmann: Will I need to do this separately for each kernel launch? For example, inside `gpu_compute_thermo_partial` there are many kernel calls; each one requests a different amount of shared memory, and that could result in a different value for `run_block_size` for each kernel launch.

joaander: Yes. Each kernel launch may have a different maximum block size.

hoomd/md/ComputeThermoGPU.cu
```diff
@@ -675,12 +720,11 @@ hipError_t gpu_compute_thermo_final(Scalar* d_properties,
...
 
-    // setup the grid to run the final kernel
-    int final_block_size = 256;
-    dim3 grid = dim3(1, 1, 1);
-    dim3 threads = dim3(final_block_size, 1, 1);
-
-    size_t shared_bytes = sizeof(Scalar4) * final_block_size;
+    // clamp the block size for this kernel
+    auto bs_thermo_final = clamp_block_size(args.block_size, (const void *) gpu_compute_thermo_final_sums);
+    dim3 grid = dim3(num_partial_sums / bs_thermo_final + 1, 1, 1);
```
joaander:
```suggestion
    dim3 grid = dim3(1, 1, 1);
```
The final kernel runs only 1 block.

hoomd/md/ComputeThermoGPU.cc
```diff
@@ -160,7 +165,7 @@ void ComputeThermoGPU::computeProperties()
 
         // build up args list
         kernel::compute_thermo_args args;
-        args.n_blocks = num_blocks;
+        args.n_blocks = num_blocks_partial;
```
joaander: `n_blocks` is unused in the new scheme. Remove it.

hoomd/md/ComputeThermoGPU.cc
```diff
@@ -59,16 +63,17 @@ void ComputeThermoGPU::computeProperties()
...
+    unsigned int num_blocks_partial
+        = m_group->getNumMembers() / block_size_partial + m_exec_conf->getNumActiveGPUs();
 
     // resize work space
     size_t old_size = m_scratch.size();
 
-    m_scratch.resize(num_blocks);
-    m_scratch_pressure_tensor.resize(num_blocks * 6);
-    m_scratch_rot.resize(num_blocks);
+    m_scratch.resize(num_blocks_partial);
```
joaander: These may resize arrays to an incorrect size. For example, with a block size of 1024, 4096 particles would allocate 4 partial sums. However if the kernel run clamps the block size to 512, 8 partial sums are needed. 

The scratch arrays need to be sized according to the *minimal* block size - the warp size in this case.

As `num_blocks_partial` is unused below, rename this to `max_num_blocks_partial`.

hoomd/md/ComputeThermoGPU.cc
```diff
@@ -182,6 +187,7 @@ void ComputeThermoGPU::computeProperties()
         args.external_energy = m_pdata->getExternalEnergy();
 
         // perform the computation on the GPU(s)
+        m_tuner->begin();
         gpu_compute_thermo_partial(d_properties.data,
```
joaander, line 188: `gpu_compute_thermo_partial` determines the number of partial sums it computes. In order to pass that information on to `gpu_compute_thermo_final`, add an output argument `unsigned int &num_partial_sums` to `gpu_compute_thermo_partial` and pass it to `gpu_compute_thermo_final`.

hoomd/md/ComputeThermoGPU.cc
```diff
@@ -200,7 +206,16 @@ void ComputeThermoGPU::computeProperties()
         // converge GPUs
         m_exec_conf->endMultiGPU();
 
+        // reset block size and number of blocks for the final reduction kernel
+        auto block_size_final = m_tuner->getParam()[1];
+        args.block_size = block_size_final;
+        args.n_blocks = num_blocks_partial / block_size_final + 1;
```
joaander:
```suggestion
```
Unused.

joaander (commented):
@tommy-waltmann Thanks for working on this! Your solution for `ComputeThermoGPU` is almost complete. I added inline suggestions to complete the implementation.

It is somewhat concerning that the unit tests pass even with this broken code. I think to fully test this we would need a) a large system size (enough to generate many partial sums even at the max block size) b) to run many compute calls (and check the output) in a loop to explore all the possible block size combinations and c) To have the partial sum kernel complex enough that the clamp actually does something (or force it to).

I'm not sure on c) - would need to add some `cout`s to check what the max block size of the kernel is on some of the GPUs we test on.

tommy-waltmann:
@joaander I have addressed your comments for `ComputeThermoGPU`, please have another look at it before I make the same changes to the other altered classes.

tommy-waltmann:
To summarize conversations I've had with @joaander offline: There are many corner cases involved in autotuning `ComputeThermoGPU` and `ComputeThermoGPUHMA`; we don't want to put the time and effort into autotuning for only miniscule performance benefits with the added potential bug risk.

There are other classes in this PR which would still benefit from autotuning, however, so we'll limit the scope of this PR to those classes.

Issue #1536, opened 2023-04-13
Release v3.11.0


joaander:
Release checklist:

- [x] Update tutorial submodule.
- [x] Update change log.
  - ``git log --format=oneline --first-parent `git log -n 1 --pretty=format:%H -- CHANGELOG.rst`...``
  - [milestone](https://github.com/glotzerlab/hoomd-blue/milestones)
- [x] Check readthedocs build, especially change log formatting.
  - [Build status](https://readthedocs.org/projects/hoomd-blue/builds/)
  - [Output](https://hoomd-blue.readthedocs.io/en/latest/)
- [x] Update actions versions.
  - See current actions usage with: `rg --no-filename --hidden uses: | awk '{$1=$1;print}' | sort | uniq`
  - Use global search and replace to update them to the latest tags
- [x] Check for new or duplicate contributors since the last release:
  `comm -13 <(git log LAST_TAG --format="%aN <%aE>" | sort | uniq) <(git log --format="%aN <%aE>" | sort | uniq)`.
  Add entries to `.mailmap` to remove duplicates.
- [x] Run [hoomd-benchmarks](https://github.com/glotzerlab/hoomd-benchmarks), check for performance
  regressions with the previous release, and post the tables in the release pull request.
- [x] Run *bumpversion*.
- [x] Tag and push.
- [x] Update conda-forge recipe.
- [x] Update glotzerlab-software.
- [x] Announce release on mailing list.

joaander:
Continuing contributors:
Corwin Kerr, Joshua A. Anderson, Melody Zhang, Tim Moore

New contributors:
Jen Bradley

joaander:
hoomd-benchmarks results using hoomd-benchmarks@ec5f12b010af1a81cd9d888eabd2215e54ed7053

V100 GPU:
```
                                     3.10.0        3.11.0         3.8.0         3.9.0
HPMCSphere                     1.043639e+03  1.058749e+03  1.063163e+03  1.053143e+03
MDPairLJ                       3.162095e+03  3.167564e+03  3.151214e+03  3.170358e+03
MDPairOPP                      1.633523e+03  1.630468e+03  1.630245e+03  1.620379e+03
MDPairTable                    1.966039e+03  1.977673e+03  1.975161e+03  1.992570e+03
MDPairWCA                      6.596048e+03  6.463683e+03  6.603566e+03  6.512428e+03
[...]

PR #1537, opened 2023-04-13
Release 3.11.0

joaander:
## Description

Changes needed to release HOOMD-blue 3.11.0.

## Motivation and context

Make the recent changes available to users.

See checklist in #1536.

## Checklist:


.github/workflows/stale.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Run stale action
-      uses: actions/stale@v7.0.0
```
b-butler, line 15: This changes the name and version number. Given we use `actions/checkout` elsewhere at a version `3.5.2` this seems wrong.

b-butler (changes requested):
There seems to be an inconsistency in the action update.

.github/workflows/stale.yml
joaander: Good catch. Fixed.

b-butler (approved):
:+1:

Issue #1266, opened 2022-03-04
Remove diameter.


joaander:
## Description

Remove the diameter shifting logic in all C++ neighbor list implementations.

## Motivation and context

With the removal of SLJ and related methods in v3, this code is no longer needed.

joaander:
Includes:
- Removing diameter shifting in the neighbor list.
- Removing diameter from snapshot data structures.
- Modifying the built in GSD writer and reader to ignore diameters.
- Possibly removing the diameter particle property (though this is a lot of work with little to gain).
- Find and remove references to diameter in the documentation.

Reason: Diameters cause more confusion that they help when working with visualization tools like OVITO. Many users also make assumptions that HOOMD uses the diameter in some way when it does not.

mphoward:
Strongly support, diameter shifting triggers some exotic and little used code paths in the neighbor list. I think there might also be some code floating around in the communicator related to this too, but I don't remember if/how @jglaser might have refactored some of that. Suggest to add:

* Check `Communicator` and `CommunicatorGPU` for diameter shifting in MPI and remove as needed

to your list! I think you might need to be careful doing this part, though, because there may be some stuff in the that looks like diameter shifting but is actually related to rigid body diameters rather than particle diameters.

joaander:
Will do.

MartinGirard:
I have a bit of a side comment to this. The presence of the diameter property was practical for custom evaluators. By abusing this property, one could write a potential that depended on a per-particle property. It's still possible to do without, but it requires a significant amount of code, for instance new force compute kernel, need to connect to particle sorting signal, etc. 

It might be useful to also provide a method to create custom evaluator that require custom per-particle properties when diameters are removed.

joaander:
> I have a bit of a side comment to this. The presence of the diameter property was practical for custom evaluators. By abusing this property, one could write a potential that depended on a per-particle property. It's still possible to do without, but it requires a significant amount of code, for instance new force compute kernel, need to connect to particle sorting signal, etc.
> 
> It might be useful to also provide a method to create custom evaluator that require custom per-particle properties when diameters are removed.

Do you need both custom per-particle quantities *and* charge? If not, you could abuse charge instead of diameter.

I agree that the ideal solution would be a system that supports dynamically added and arbitrarily named per-particle quantities. That is a very complex feature to introduce. I would hope to add it some day but I don't have a time frame.

MartinGirard:
[...]

PR #1538, opened 2023-04-14
Improve `write.GSD` and `get_snapshot` performance.

joaander:
## Description

* Rewrite GSD writer to directly use local particle properties in place of `takeSnapshot`.
* Rewrite serial `takeSnapshot` code path to improve performance.
* The MPI code path in `takeSnapshot` remains slow for now, but could later be rewritten with code similar to the new code in `GSDDumpWriter`.
* Topology (bond, angle, ...) snapshots also remain slow (most use cases only write this in the first frame).
* Remove `ExternalCallback` code, as it is unused and relies on the now removed `std::map` in the snapshot.
* Expand the `dynamic` flag to take in a list of individual particle fields. Keep the existing `'property'`, `'attribute'`, and `'momentum'` meta-fields for backward compatibility.

## Motivation and context

`write.GSD` is extremely slow in current releases. On profiling, I found the majority of the time was spent in creating and iterating through the `std::map` used by `takeSnapshot`. Next after that is std::vector allocations in the snapshot objects.

* The new GSD writer accesses only the particle data needed given the user parameters (e.g. in the default dynamic=['property'], only position and orientation are accessed). 
* GSD now only sorts the non-default values it will write. 
* GSD now holds on to the `std::vector` instances from one frame to the next to avoid reallocation costs.
* In MPI, GSD now uses an optimized k-way merge sort to order the particles in increasing tag order.

The optimizations to `takeSnapshot` are in this pull request because I took those steps first in an effort to boost GSD performance. There is more work to make `takeSnapshot` as fast in MPI as GSD is, but I leave that work for a future pull request.

In log only GSD files (`filter=hoomd.filter.Null()`), performance is now ~1000x faster (speedup depends on system size) as GSD no longer accesses any particle data in this case - before it always called `takeSnapshot`.

## How has this been tested?

Existing unit tests pass.

Will post benchmark results when ready to review.

## Change log

```
Changed:

* Increase performance of ``hoomd.write.GSD``.
* Increase performance of ``hoomd.State.get_snapshot`` in serial.
* `hoomd.write.GSD.dynamic` now allows fine grained control over individual particle fields.
```

## Checklist:


joaander:
Benchmark results on `cheme-hodges` (NVME SSD):
* 1 rank, N=64000
  * This branch: 241 MiB/s
  * `trunk-major`: 34.8 MiB/s
* 1 rank, N=1000000
  * This branch: 203 MiB/s
  * `trunk-major`: 33.2 MiB/s
* 6 ranks, N=64000:
  * This branch: 150 MiB/s
  * `trunk-major`: 13.1 MiB/s
* 6 ranks, N=1000000)
  * This branch: 191 MiB/s
  * `trunk-major`: 11.1 MiB/s

I will add this benchmark to `hoomd-benchmarks` soon. To test, I ran` python3 -m hoomd_benchmarks.write_gsd --device CPU -v --repeat=5 --warmup_steps=100 --benchmark_steps=500 -N <N> --bandwidth`.

joaander:
Benchmark results on `summit` (Alpine):
* 1 rank, N=64000
  * This branch: 12.2 MiB/s (16.5 frames/second)
  * `trunk-major`: 8.7 MiB/s (11.6 frames/second)
* 1 rank, N=1000000
  * This branch:  84.4 MiB/s (7.5 frames/second)
  * `trunk-major`: 14.0 MiB/s (1.28 frames/second)
* 6 ranks, N=64000:
  * This branch: 11.5 MiB/s
  * `trunk-major`:  4.79 MiB/s
* 6 ranks, N=1000000)
  * This branch: 81.6 MiB/s
  * `trunk-major`:  6.02 MiB/s

joaander:
With the new fine grained dynamic control, setting `dynamic=['particles/position']` offers a slight performance increase over the default `dynamic=['property']` (which includes orientation).

joaander:
@b-butler @tommy-waltmann I ask you both to review as this is a critical piece of code. Exiting unit test pass. Adding new unit tests that cover all 1024+ combinations of parameters is not feasible.

hoomd/GSDDumpWriter.cc
```diff
@@ -150,85 +299,61 @@ void GSDDumpWriter::analyze(uint64_t timestep)
...
+        writeFrameHeader(m_local_frame);
+        writeAttributes(m_local_frame);
+        writeProperties(m_local_frame);
+        writeMomenta(m_local_frame);
         }
 
     // topology is only meaningful if this is the all group
     if (m_group->getNumMembersGlobal() == m_pdata->getNGlobal()
-        && (m_write_topology || nframes == 0))
+        && (m_write_topology || m_nframes == 0))
```
b-butler, line 346: Sort of tangential, but if we are not going to follow a user's specification (even if it doesn't make sense) , shouldn't we error rather than silently not follow his/her specification?

hoomd/GSDDumpWriter.cc
```diff
@@ -1110,6 +1015,482 @@ void GSDDumpWriter::populateNonDefault()
...
+                {
+                image = h_image.data[index];
+                }
+
+            frame.global_box.wrap(position, image);
+
+            if (m_particle_dynamic[gsd_flag::position] || m_nframes == 0)
+                {
+                if (position.x != Scalar(0.0) || position.y != Scalar(0.0)
+                    || position.z != Scalar(0.0))
```
b-butler:
```suggestion
                if (position != vec3<Scalar>{0.0, 0.0, 0.0})
```
Is there a reason not to use the overloaded `==`?

hoomd/GSDDumpWriter.cc
```diff
@@ -1110,6 +1015,482 @@ void GSDDumpWriter::populateNonDefault()
...
+                    {
+                    all_default[gsd_flag::position] = false;
+                    }
+
+                frame.particle_data.pos.push_back(vec3<float>(position));
+                }
+
+            if (m_particle_dynamic[gsd_flag::image] || m_nframes == 0)
+                {
+                if (image.x != 0 || image.y != 0 || image.z != 0)
```
b-butler:
```suggestion
                if (image.x != int3{0.0, 0.0, 0.0})
```

hoomd/GSDDumpWriter.cc
```diff
@@ -1110,6 +1015,482 @@ void GSDDumpWriter::populateNonDefault()
...
+                    {
+                    all_default[gsd_flag::mass] = false;
+                    }
+
+                frame.particle_data.mass.push_back(mass);
+                }
+
+            if (m_particle_dynamic[gsd_flag::velocity] || m_nframes == 0)
+                {
+                if (velocity.x != 0.0f || velocity.y != 0.0f || velocity.z != 0.0f)
```
b-butler:
```suggestion
                if (velocity.x != vec3<float>{0.0, 0.0, 0.0f})
```

hoomd/GSDDumpWriter.cc
```diff
@@ -1110,6 +1015,482 @@ void GSDDumpWriter::populateNonDefault()
...
+                                       access_location::host,
+                                       access_mode::read);
+
+        frame.particle_data_present[gsd_flag::inertia] = true;
+
+        for (unsigned int index : m_index)
+            {
+            vec3<float> inertia = vec3<float>(h_inertia.data[index]);
+
+            if (inertia.x != 0.0f || inertia.y != 0.0f || inertia.z != 0.0f)
```
b-butler:
```suggestion
            if (inertia != vec3<float>{0.0, 0.0, 0.0})
```

hoomd/GSDDumpWriter.cc
```diff
@@ -1110,6 +1015,482 @@ void GSDDumpWriter::populateNonDefault()
     gsd_close(&m_handle);
     }
 
+void GSDDumpWriter::populateLocalFrame(GSDDumpWriter::GSDFrame& frame, uint64_t timestep)
```
b-butler, line 1018: Why the decision to iterate over index N attribute times rather than once with conditionals inside? Is it faster or just cleaner?

hoomd/md/pytest/test_gsd.py
```diff
@@ -237,7 +237,7 @@ def test_write_gsd_dynamic(simulation_factory, create_md_sim, tmp_path):
 
     sim = create_md_sim
 
-    # test default dynamic=['property']
+    # test default dynamic=['property', 'property']
```
b-butler:
```suggestion
    # test default dynamic=['property']
```

hoomd/md/pytest/test_gsd.py
```diff
@@ -107,7 +107,7 @@ def test_write_gsd_trigger(create_md_sim, tmp_path):
     gsd_writer = hoomd.write.GSD(filename=filename,
```
b-butler, line 107: While I agree that we should/can not test all combinations. It would be worth testing once using the individual attributes rather than the group ones.

hoomd/write/gsd.py
```diff
@@ -80,37 +80,37 @@ class GSD(Writer):
...
 
     * **momentum**
 
-        * particles/velocity
-        * particles/angmom (*only written when values are not the
-          default: [0,0,0,0]*)
-        * particles/image (*only written when values are not the
-          default: [0,0,0]*)
+      * particles/velocity
+      * particles/angmom (*only written when values are not the
```
b-butler: If we are going to document when we don't write them should we not do this for all of them?

hoomd/write/gsd.py
tommy-waltmann, line 1: We should add something to the docstring saying that individual particle quantities can be made dynamic a la carte. Also, line 76 seems to cut off in the middle of a sentence.

hoomd/write/gsd.py
```diff
@@ -80,37 +80,37 @@ class GSD(Writer):
```
tommy-waltmann, line 81: The line above this one says that *property* is always dynamic, but I don't think that is true anymore.

hoomd/md/pytest/test_gsd.py
tommy-waltmann, line 1: Let's add at least a few test cases where individual quantities are made dynamic. For example, just making `particles/velocity` dynamic instead of all of the momentum quantities, or just `particles/position` instead of all of the property quantities. I imagine these will be relatively frequent use cases.

hoomd/hpmc/ExternalCallback.h
tommy-waltmann, line 1: What does this have to do with improving `gsd` and `get_snapshot` performance?

tommy-waltmann (changes requested):
This is the start of my review, I'll go through the guts of the changes in the future.

b-butler (changes requested):
We need to document the setting of individual quantities in gsd.py.

Besides that and my other comments, the code looks good and well written.

hoomd/write/gsd.py
joaander: I updated the docstring `gsd.py`.

hoomd/write/gsd.py
joaander: Fixed in the docstring update.

hoomd/hpmc/ExternalCallback.h
joaander: As stated in the pull request description:
> Remove `ExternalCallback` code, as it is unused and relies on the now removed `std::map` in the snapshot.

My options were to either refactor the unused and untested code, or remove it.

hoomd/GSDDumpWriter.cc
joaander: It should not be an error to write the first frame of the file with a filter that selects a subset. Otherwise, users could never select a subset.

Do you think it would be useful for users to receive an exception on the 2nd frame written to the file if they both select a subset and include `'topology'` in the dynamic list?

hoomd/GSDDumpWriter.cc
joaander: I forget that these exist.

hoomd/hpmc/ExternalCallback.h
tommy-waltmann: Apologies for that, I should have read closer.

hoomd/GSDDumpWriter.cc
joaander: Instantiating all of the `ArrayHandle`s outside the scope of one loop would require that all particle data is copied from the GPU to the CPU. Instantiating the `ArrayHandle`s inside the conditional scope in an inner loop would add excessive overhead with all the calls to `GPUArray::acquire`. Multiple loops over `m_index` is also not ideal, but I thought it to be the best of the available options.

hoomd/write/gsd.py
joaander: I pushed an update to `gsd.py` just after this review that removed all these notes. Let me know if you prefer that we instead document all the defaults here instead.

hoomd/md/pytest/test_gsd.py
joaander: Implemented.

hoomd/md/pytest/test_gsd.py
joaander: Done.

hoomd/GSDDumpWriter.cc
b-butler: @joaander I was thinking of cases where users specify to write out `topology` that is explicitly in `dynamic`, but the filter excludes some particles. So only erroring when `m_write_topology == true` and `m_group->getNumMembersGlobal() != m_pdata->getNGlobal()`

The error doesn't need to be here, but could be put in `setDynamic`.

hoomd/write/gsd.py
b-butler: I am good with the removal.

hoomd/write/gsd.py
```diff
@@ -70,47 +70,44 @@ class GSD(Writer):
...
+    writes non-dynamic fields only the first frame. When reading a GSD file,
     the data in frame 0 is read when a quantity is missing in frame *i*,
-    supplying data that is static over the entire trajectory.  Set the *dynamic*
+    supplying data that is static over the entire trajectory.
 
-    Specify the one or more of the following strings in **dynamic** to make the
-    corresponding quantities dynamic (**property** is always dynamic):
+    Provide one or more of the following strings in the `dynamic` list to make
+    the corresponding field dynamic (or all fields in a given category)
+    **dynamic**:
```
tommy-waltmann: I think this is worded a bit awkwardly, the word dynamic appears twice in a row, ignoring parentheses.

hoomd/write/gsd.py
```diff
@@ -70,47 +70,44 @@ class GSD(Writer):
     To reduce file size, `GSD` does not write properties that are set to
     defaults. When masses, orientations, angular momenta, etc... are left
     default for all particles, these fields will not take up any space in the
-    file. Additionally, `GSD` only writes *dynamic* quantities to all frames. It
-    writes non-dynamic quantities only the first frame. When reading a GSD file,
+    file. Additionally, `GSD` only writes *dynamic* fields to all frames. It
+    writes non-dynamic fields only the first frame. When reading a GSD file,
```
tommy-waltmann: If I understand this correctly, when a user sets a given field to the default value in frame 0, it will not be written in later frames even if the user requests that it be dynamic? That seems counterintuitive to me.

joaander:
And on Frontier (Orion):
* 1 rank, N=64000
  * This branch: 11.64 MiB/s (15.5 frames/s)
* 1 rank, N=1000000
  * This branch:  99.8 MiB/s
* 6 ranks, N=64000:
  * This branch: 10.75 MiB/s (8.5 frames/s)
* 6 ranks, N=1000000)
  * This branch: 101 MiB/s

I have confirmed that the per-frame `fsync` for stability limits the performance on these massive scale systems to ~10-15 frames/s. When I remove the `fsync`, the benchmark achieves ~500-600 MiB/s regardless of system size.

hoomd/GSDDumpWriter.cc
joaander: I don't think the error can be checked in `setDynamic`. The user may provide group that changes membership from one frame to the next.

hoomd/GSDDumpWriter.h
```diff
@@ -153,49 +124,109 @@ class PYBIND11_EXPORT GSDDumpWriter : public Analyzer
...
-    bool m_write_momentum = false;  //!< True if momenta should be written
-    bool m_write_topology = false;  //!< True if topology should be written
-    bool m_write_diameter = false;  //!< True if the diameter attribute should be written
-    gsd_handle m_handle;            //!< Handle to the file
+    std::string m_fname;           //!< The file name we are writing to
+    std::string m_mode;            //!< The file open mode
+    bool m_truncate = false;       //!< True if we should truncate the file on every analyze()
+    bool m_is_initialized = false; //!< True if the file is open
+    bool m_write_topology = false; //!< True if topology should be written
+    bool m_write_diameter = false; //!< True if the diameter attribute should be written
```
tommy-waltmann, line 201: Why do we still carry around a flag specifically for particle diameter instead of checking `m_particle_dynamic[gsd_flag::diameter]`?

hoomd/GSDDumpWriter.cc
```diff
@@ -63,63 +64,211 @@ GSDDumpWriter::GSDDumpWriter(std::shared_ptr<SystemDefinition> sysdef,
...
+#ifdef ENABLE_MPI
+    if (m_sysdef->isDomainDecomposed())
+        {
+        m_gather_tag_order = GatherTagOrder(m_exec_conf->getMPICommunicator());
+        }
+#endif
+
+    m_particle_dynamic.reset();
+    m_particle_dynamic[gsd_flag::position] = true;
+    m_particle_dynamic[gsd_flag::orientation] = true;
```
tommy-waltmann, line 77: According to the migration guide, these are not dynamic by default any more.

tommy-waltmann (changes requested):
Here are a few more comments after I went through all the changes.

hoomd/GSDDumpWriter.cc
joaander: I tried to be clear in the migration guide:
> `hoomd.write.GSD` no longer enforces 'property' as an always dynamic quantity

I did not change the default (see `gsd.py` line 178) or the corresponding docstring.

The migration guide covers the case where a user provides a value for `dynamic` that doesn't include `'property'`.

For example, in v3:
```
>>> write_gsd = hoomd.write.GSD(..., dynamic = ['momentum'])
>>> print(write_gsd.dynamic)
['property', 'momentum']
```

In v4:
```
>>> write_gsd = hoomd.write.GSD(..., dynamic = ['momentum'])
>>> print(write_gsd.dynamic)
['particles/velocity', 'particles/angmom', 'particles/image']
```

If you have any suggestions on how to make this more clear in the migration guide, let me know.

hoomd/GSDDumpWriter.h
joaander: Because I wish the user to opt-in to writing diameters on both frame 0 and subsequent frames. So far, we only have 1 user who has expressed an interest in this: #1266

hoomd/write/gsd.py
joaander: When a field is non-default in frame 0, GSD will write it in subsequent frames when the user requests it to be dynamic.

joaander: I noted this corner case in the documentation.

hoomd/write/gsd.py
joaander: I revised this language.

b-butler (approved):

tommy-waltmann (approved):
I think the documentation is about as clear as its going to get for now. We can always go back and adjust wording based on confusion from users on the mailing list if needed.

PR #1541, opened 2023-04-28
Further improve GSD performance.

joaander:
## Description

Add write buffers to further improve GSD performance.

## Motivation and context

See https://github.com/glotzerlab/gsd/pull/237.

Most users will not need to change the write buffer size. I make the API available so that users wishing to push the limits can increase the buffer size.

## How has this been tested?

* Existing unit tests pass after adding needed calls to flush().
* Will report on performance tests on Frontier.

TODO:
* [x] Test and update tutorial notebooks as needed.

## Change log

```
Added:

* ``hoomd.write.GSD.maximum_write_buffer_size`` - Set the maximum size of the GSD write buffer.
* ``hoomd.write.GSD.flush`` - flush the write buffer of an open GSD file.
* On importing ``hoomd``, install a ``SIGTERM`` handler that calls ``sys.exit(1)``.
```

## Checklist:

### DIFF
--- a/hoomd/GSDDumpWriter.cc
+++ b/hoomd/GSDDumpWriter.cc
@@ -54,8 +54,7 @@ GSDDumpWriter::GSDDumpWriter(std::shared_ptr<SystemDefinition> sysdef,
                              std::shared_ptr<ParticleGroup> group,
                              std::string mode,
                              bool truncate)
-    : Analyzer(sysdef, trigger), m_fname(fname), m_mode(mode), m_truncate(truncate),
-      m_group(group)
+    : Analyzer(sysdef, trigger), m_fname(fname), m_mode(mode), m_truncate(truncate), m_group(group)
     {
     m_exec_conf->msg->notice(5) << "Constructing GSDDumpWriter: " << m_fname << " " << mode << " "
                                 << truncate << endl;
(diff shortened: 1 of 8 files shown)


joaander:
Frontier performance (Orion) with 1 rank writing frames with N positions:
* Default buffer size (64 MiB) (356 frames/second)
  * N=64000: 296 MiB/s 
  * N=1000000: 188 MiB/s
* 256 MiB write buffer:
  * N=64000:  344 MiB/s (480 frames/second)
  * N=1000000:  227 MiB/s
* 1 byte buffer for comparison to https://github.com/glotzerlab/hoomd-blue/pull/1538#issuecomment-1520312988
  * N=64000: 10.6 MiB/s
  * N=1000000: 98.2 MiB/s

Log-only files show a massive improvement:
* Default buffer size (64 MiB): 31134 frames/second
* 256 MiB write buffer: 32028 frames/second
* 1 byte buffer: 82 frames/second

The limitation here might be the index buffer limit. Perhaps we should re-implement the index buffer limit as a fraction of the write buffer so they scale together. Not that there is anything wrong with a 380x speedup, but there might be room for a little more.

b-butler (approved):
LGTM

PR #1543, opened 2023-05-02
Feat/gsd with queue

b-butler:
## Description

Add a `hoomd.write.Deque` class which stores a `std::deque` of logger and trajectory data up to a specified maximum size `N`. The deque pushes data to the front and pops old data off the back of the deque. The `analyze` method only collects and removes data as necessary. All writes are triggered by the `dump` method exposed in Python. The purpose of this class is to allow selective high frequency data storing, by allowing users to determine when to write data.

One foreseen pattern of use would be to create a custom action which composes a `Deque` object calling `dump` when the specified conditions have been met.

## Motivation and context

To provide a means for high-performance, high-frequency writes in HOOMD.

## How has this been tested?

Tests based on the GSD tests were added.

## Change log

```
Added: `hoomd.write.Deque` class for enabling selective high-frequency frame writing.
```

## Checklist:


hoomd/md/pytest/test_burst_writer.py
```diff
@@ -0,0 +1,202 @@
...
+except ImportError:
+    pytest.skip("gsd not available", allow_module_level=True)
+
+from hoomd.pytest.test_snapshot import assert_equivalent_snapshots
+
+N_RUN_STEPS = 3
+
+
+@pytest.fixture(scope='function')
+def hoomd_snapshot(lattice_snapshot_factory):
```
AlainKadar, line 20: Do we want to call this `frame` instead of `snapshot`? Or better to keep consistent with the name of the module? Are there plans to rename this module to `frame` in the future?

AlainKadar (approved):
I just had one suggestion

b-butler:
@joaander The test errored due to timeout which is currently set at 3000 second or 50 minutes but MPI tests finish in 30 seconds on hodges. Also, MPI tests for the neighbor list are erroring for so reason on hodges. Will look into that.

joaander:
> @joaander The test errored due to timeout which is currently set at 3000 second or 50 minutes but MPI tests finish in 30 seconds on hodges. Also, MPI tests for the neighbor list are erroring for so reason on hodges. Will look into that.

I can't look at it right now. Maybe later this week. If you can find a configuration that reproduces the deadlock, I usually debug them by 1) Running the tests with mpirun and wait for the deadlock. 2) Attach a debugger to the processes and see where each rank is waiting.

b-butler:
pre-commit.ci autofix

b-butler:
The deadlocking is solved. I forgot to not call gsd functions on all ranks.

hoomd/GSDDumpWriter.cc
```diff
@@ -342,56 +352,57 @@ void GSDDumpWriter::analyze(uint64_t timestep)
...
+                          frame.angle_data,
+                          frame.dihedral_data,
+                          frame.improper_data,
+                          frame.constraint_data,
+                          frame.pair_data);
             }
         }
 
     // emit on all ranks, the slot needs to handle the mpi logic.
     m_write_signal.emit(m_handle);
```
joaander: If this code was used, it would be a problem for the burst writer. However, this has now been replaced by `Logger`. I opened #1554 to remove these dead code paths. No need to make any changes in this pull request.

hoomd/md/pytest/test_burst_writer.py
joaander: No, these are `hoomd.Snapshot` objects. I renamed `gsd.Snapshot` to `gsd.Frame` to avoid confusion with `hoomd.Snapshot`.

hoomd/write/gsd_burst.py
```diff
@@ -0,0 +1,86 @@
...
+from hoomd import _hoomd
+from hoomd.filter import All
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.write.gsd import GSD
+
+
+class Burst(GSD):
+    r"""Write last :math:`N` stored frames at user trigger in the GSD format.
+
+    This class stores the last :math:`N` frames in an interal deque
```
joaander:
```suggestion
    This class stores up to the last :math:`N` frames in an interal deque
```

hoomd/write/gsd_burst.py
```diff
@@ -0,0 +1,86 @@
...
+        filter (hoomd.filter.filter_like): Select the particles to write.
+            Defaults to `hoomd.filter.All`.
+        mode (str): The file open mode. Defaults to ``'ab'``.
+        dynamic (list[str]): Field names and/or field categores to save in
+            all frames. Defaults to ``['property']``.
+        logger (hoomd.logging.Logger): Provide log quantities to write. Defaults
+            to `None`.
+        max_burst_frames (int): The maximum number of frames to store before
+            between writes. -1 represents no limit. Defaults to -1.
+
```
joaander, line 44: Add a note that when there are no frames in the GSD file, `Burst` always one frame the first time it is triggered to capture all the static quantities.

hoomd/write/gsd_burst.py
```diff
@@ -0,0 +1,86 @@
...
+
+class Burst(GSD):
+    r"""Write last :math:`N` stored frames at user trigger in the GSD format.
+
+    This class stores the last :math:`N` frames in an interal deque
+    which only writes the frames out when `dump` is called. When the writer is
+    triggered and the next frame would result in :math:`N + 1` frames being
+    stored, the oldest frame is popped off the deque and the new frame is
+    pushed to the front of the deque.
```
joaander:
```suggestion
    stored, the oldest frame is removed from the deque and the new frame is
    added.
```
Also, consider replacing `deque` (an implementation detail) with a more generic term such as `queue` or `buffer` to help users understand.

joaander (changes requested):
Thanks! The implementation is very clean. I have a few suggestions to improve the documentation.

hoomd/GSDDumpWriter.cc
joaander: By `this`, I refer to `m_write_signal.emit`.

b-butler:
@joaander I just thought of a potential mitigation of the initial write from the burst writer. We could either
* provide a method to write similar to `GSD.write`
* or show how to get this to run at the very beginning of a simulation with an `And` trigger.

What do you think?

joaander:
Yes, we could encourage users to use `GSD.write` to write the initial file and then open it in append mode. We could also enforce this behavior by making it an error for `BurstWriter` to create files or append to files with 0 frames.

hoomd/write/gsd_burst.py
```diff
@@ -0,0 +1,106 @@
...
+        filter (hoomd.filter.filter_like): Select the particles to write.
+            Defaults to `hoomd.filter.All`.
+        mode (str): The file open mode. Defaults to ``'ab'``.
+        dynamic (list[str]): Field names and/or field categores to save in
+            all frames. Defaults to ``['property']``.
+        logger (hoomd.logging.Logger): Provide log quantities to write. Defaults
+            to `None`.
+        max_burst_frames (int): The maximum number of frames to store before
+            between writes. -1 represents no limit. Defaults to -1.
+        write_at_start (bool): When ``True`` **and** the file does not exist or
```
joaander, line 33: `write_at_start` seems like the wrong name for this. Consider using `write_initial_frame`.

joaander (commented):
I pushed some documentation revisions directly to the branch. Please review and make any further edits as needed. It looks like you need to merge in conflicts from `trunk-major`.

I have one comment on parameter naming to think about. Other than that, this is good to merge.

hoomd/write/gsd_burst.py
b-butler: Decided offline that the current name is okay.

joaander (approved):
Thanks!

Issue #1457, opened 2023-01-04
Attaching a logger to a writer prevents GC from cleaning up memory


ianrgraham:
### Description

I was recently running a number of simulations in a loop, instantiating a new simulation object on each pass, and I noticed that upon each pass the GPU memory used by my process kept growing linearly. I tracked this down to the use of the standard logger. Removing the logger from the writer, or removing all analyzers from the logger eliminates the leak. This effects both CPU and GPU memory.

### Script

```python
import hoomd
import freud
import matplotlib.pyplot as plt

import gc
import os
import psutil

# memory leak occurs both on CPU and GPU
device = hoomd.device.GPU()
# device = hoomd.device.CPU()

def setup_sim(snap):
    sim = hoomd.Simulation(device, seed=0)
    sim.create_state_from_snapshot(snap)

    integrator = hoomd.md.Integrator(1e-3)

    nlist = hoomd.md.nlist.Tree(buffer=0.3)
    lj = hoomd.md.pair.LJ(nlist, default_r_cut=2.5)
    lj.params[('A', 'A')] = dict(sigma=1.0, epsilon=1.0)

    nve = hoomd.md.methods.NVT(hoomd.filter.All(), 1.0, 0.1)
    integrator.forces = [lj]
    integrator.methods = [nve]
    sim.operations.integrator = integrator

    return sim


def make_snapshot():
    snap = hoomd.Snapshot()

    box, pos = freud.data.UnitCell.fcc().generate_system(20)

    snap.particles.N = len(pos)
    snap.particles.position[:] = pos
    snap.particles.typeid[:] = 0
    snap.particles.types = ["A"]
    snap.configuration.box = box
    return snap

snap = make_snapshot()

mem_usage = []

for i in range(10):
    process = psutil.Process(os.getpid())
    mem = process.memory_info().rss//1024**2
    mem_usage.append(mem)
    print("Iter:", i, "Memory Usage (MB):", mem)

    sim = setup_sim(snap)

    log = hoomd.logging.Logger()
    analyzer = hoomd.md.compute.ThermodynamicQuantities(filter=hoomd.filter.All())

    sim.operations.computes.append(analyzer)

    log.add(sim, quantities=["timestep"])  # commenting out this line removes the leak

    write_trigger = hoomd.trigger.Periodic(1000, phase=sim.timestep)
    # removing the log parameter from the writer also removes the leak
[...]

PR #1546, opened 2023-05-09
Flush GSD buffers on exit.

joaander:
## Description

Maintain a list of active GSD writers. Use an `atexit` method to flush them on exit. Also flush the files in the exception handler to attempt to keep data before calling MPI_Abort.

## Motivation and context

Due to #1457 or a related issue, GSD files may not flush **even when the program cleanly exits**. Explicit is better than implicit, so forcibly flushing the GSD files at exit ensures as much data as possible is written. 

Note: We may be able to solve #1457 using some `weakref` objects to break cycles.

## How has this been tested?

Existing unit tests pass. Examined the length of `_open_gsd_writers` with and without the `finalizer` object and got the expected results.

## Change log

Not needed. Fixing a new feature in this release.

## Checklist:

### DIFF
--- a/hoomd/write/gsd.py
+++ b/hoomd/write/gsd.py
@@ -14,6 +14,22 @@ from hoomd.logging import Logger, LoggerCategories
 from hoomd.operation import Writer
 import numpy as np
 import json
+import atexit
+import weakref
+
+# Track open gsd writers to flush at exit.
+_open_gsd_writers = []
+
+
+def _flush_open_gsd_writers():
+    """Flush all open gsd writers at exit."""
+    for weak_writer in _open_gsd_writers:
+        writer = weak_writer()
+        if writer is not None:
+            writer.flush()
+
+
+atexit.register(_flush_open_gsd_writers)
 
 
 def _array_to_strings(value):
@@ -202,6 +218,12 @@ class GSD(Writer):
             self.truncate)
 
         self._cpp_obj.log_writer = self.logger
+
+        # Maintain a list of open gsd writers
+        weak_writer = weakref.ref(self)
+        _open_gsd_writers.append(weak_writer)
+        self._finalizer = weakref.finalize(self, _open_gsd_writers.remove,
+                                           weak_writer)
 
     @staticmethod
     def write(state, filename, filter=All(), mode='wb', logger=None):
--- a/hoomd/write/gsd.py
+++ b/hoomd/write/gsd.py
@@ -44,6 +44,12 @@ def _array_to_strings(value):
         return value
 
 
+def _finalize_gsd(weak_writer, cpp_obj):
+    """Finalize a GSD writer."""
+    _open_gsd_writers.remove(weak_writer)
+    cpp_obj.flush()
+
+
 class GSD(Writer):
     r"""Write simulation trajectories in the GSD format.
 
@@ -222,8 +228,8 @@ class GSD(Writer):
         # Maintain a list of open gsd writers
         weak_writer = weakref.ref(self)
         _open_gsd_writers.append(weak_writer)
-        self._finalizer = weakref.finalize(self, _open_gsd_writers.remove,
-                                           weak_writer)
+        self._finalizer = weakref.finalize(self, _finalize_gsd, weak_writer,
+                                           self._cpp_obj),
 
     @staticmethod
     def write(state, filename, filter=All(), mode='wb', logger=None):
--- a/hoomd/GSDDumpWriter.cc
+++ b/hoomd/GSDDumpWriter.cc
@@ -206,6 +206,7 @@ void GSDDumpWriter::flush()
     {
     if (m_exec_conf->isRoot())
         {
+        m_exec_conf->msg->notice(5) << "GSD: flush gsd file " << m_fname << endl;
         int retval = gsd_flush(&m_handle);
         GSDUtils::checkError(retval, m_fname);
         }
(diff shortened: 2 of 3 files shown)


b-butler:
@joaander, Yeah I think switching all our simulation references to weakrefs could work. Or we could remove the simulation storage in many of them and optionally store a weak reference through the use of `weakref` and `property`.

hoomd/write/gsd.py
```diff
@@ -203,6 +225,12 @@ def _attach_hook(self):
 
         self._cpp_obj.log_writer = self.logger
 
+        # Maintain a list of open gsd writers
+        weak_writer = weakref.ref(self)
+        _open_gsd_writers.append(weak_writer)
+        self._finalizer = weakref.finalize(self, _finalize_gsd, weak_writer,
+                                           self._cpp_obj),
```
tommy-waltmann, line 232: It is possible to do this in `__del__` instead of using `weakref.finalize`?

tommy-waltmann (commented):
There are a couple of other potential solutions I'd like to discuss. I mentioned one in a comment and I'll mention another below:

Can we just call `flush` from the destructor of the `GSDDumpWriter`? That destructor is guaranteed to run in every case.

b-butler:
@tommy-waltmann Actually, since the process is a Python interpreter, the destructor of any object is not guaranteed to be called if Python just decides to exit without cleaning up all its memory (which it does) this is why we cannot use `__del__` or destructors, because when the kernel cleans up Python's mess, it has no idea what byte represents what.

joaander:
The destructor of `GSDDumpWriter` implicitly calls flush via `gsd_close()`. However, this was not being called in `hoomd-validation`'s `lj-fluid` project, *even for clean exits*.

These are the five behaviors I have seen or been able to create and how they are handled by this pull request.
1. Clean exit / SIGTERM (or unhandled exception in serial): Python's garbage collector cleans up both the Python level `GSD` object and the C++ `GSDDumpWriter`
  *  The finalizer triggers, flushes the file and removes it from `_open_gsd_writers`.
  * `GSDDumpWriter` closes the file.
  * `_open_gsd_writers` is empty and the `atexit` routine does nothing.
2. Clean exit / SIGTERM (or unhandled exception in serial): Python's garbage collector cleans up the Python level `GSD` object *but not* the C++ `GSDDumpWriter`
  *  The finalizer triggers, flushes the file and removes it from `_open_gsd_writers`.
  * `_open_gsd_writers` is empty and the `atexit` routine does nothing.
  * The files are closed by the kernel during tear down.
3. Clean exit / SIGTERM (or unhandled exception in serial): Python's garbage collector fails to clean up both the Python level `GSD` object and the C++ `GSDDumpWriter`
  * `_open_gsd_writers` contains the writers and the `atexit` routine flushes the files.
  * The files are closed by the kernel during tear down.
4. Unhandled exception in MPI on the root rank.
  * `_hoomd_sys_excepthook` flushes the write buffers in all writers in `_open_gsd_writers`, then calls `MPI_Abort` which kills the other rank's processes.
  * The files are closed by the kernel during tear down.
5. Unhandled exception in MPI on a non-root rank.
  * `_hoomd_sys_excepthook` calls `MPI_Abort` which kills the other rank's processes including the root rank. This is `SIGKILL` which cannot be caught, so buffers are not flushed.
  * The files are closed by the kernel during tear down.
 
Unhandled exceptions thrown on all ranks can potentially follow either 4 or 5 depending on race conditions (I got 4 in a few quick tests). There is no way to avoid this without causing deadlock. I add the call to hopefully preserve intended GSD file contents as much as possible.

Behaviors 2 and 3 do not appear in minimal examples. #1457 shows one. The `lj-fluid.py` I am currently working on is another. That one gave behavior 3 reproducibly for most of the morning. I took a break and went back later in the day, then it gave behavior 2 reproducibly. You can test this by running with `notice_level=5`. If the garbage collector works correctly, you should see many object deconstruction notices and notices that the GSD file are closed.

Even if we could somehow work around the bugs in Python's garbage collector so that behaviors 2 and 3 never occur, we will need this PR to cover 4 even if it is not perfect.

hoomd/write/gsd.py
joaander: Yes. However the finalizer is more flexible for future development. For example if we were to add a `close` method to `GSD`, the finalizer is a cleaner solution. See this example for an explanation: https://docs.python.org/3/library/weakref.html#comparing-finalizers-with-del-methods

joaander:
Due to the way that `MPI_Abort` works, uncaught exceptions in one partition will send `SIGKILL` to the other partitions (at least in OpenMPI): https://www.open-mpi.org/doc/v4.1/man3/MPI_Abort.3.php
> The long-term goal of the Open MPI implementation is to terminate all processes in all tasks that contain a process in comm, and the error code is not returned to the invoking environment. At the moment, this isn’t fully implemented and MPI_Abort will terminate the entire job.

This means that users wishing to obtain partial data files from jobs that fail should either catch exceptions and/or explicitly call `flush` at specific points needed to ensure data integrity (i.e. flush trajectory files when writing a restart file). Perhaps we should demonstrate this in the workflow tutorial.

tommy-waltmann:
It looks like there are a lot of details related to cleaning up that I had not anticipated. It seems that this PR handles the possible outcomes as cleanly as it can (outcome 5 doesn't seem to be fixable).

tommy-waltmann (approved):

PR #1549, opened 2023-05-10
Refactor gamma distribution.

joaander:
## Description

* Replace floating point literals with explicit Real constants in random number generation.
* Test gamma distribution with a larger alpha.

## Motivation and context

After getting some surprising results from the Bussi thermostat (which turned out to be a poor seed choice), I reviewed the GammaDistribution code. I found no errors in the code, but there were floating point literals. I replaced these with Real casted literals to ensure there are no unintended conversions. This should be a no-op for double precision Real as the compiler would have upcasted implicitly. 

In Bussi, GammaDistribution is used with alpha=ndof/2, so I added a unit test to cover something more reasonable. I had to lower the tolerance because the `sample_x3` used in the skew computation goes to 1e10 and suffers from round off error in the accumulation. A better solution would be fix the skew computation with kahan summation or another method, but I do not have time for that right now.

## How has this been tested?

Existing tests pass. *hoomd-validation* simulations using Bussi produce the expected K distribution.

## Change log

Internal change, no change log necessary.

## Checklist:

### DIFF
--- a/hoomd/RandomNumbers.h
+++ b/hoomd/RandomNumbers.h
@@ -418,7 +418,7 @@ template<typename Real> class SpherePointGenerator
         do
             {
             u = UniformDistribution<Real>(Real(-1.0), Real(1.0))(rng);
-            one_minus_u2 = 1.0f - u * u;
+            one_minus_u2 = Real(1.0) - u * u;
             } while (one_minus_u2 < Real(0.0));
 
         // project onto the sphere surface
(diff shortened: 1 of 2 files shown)


tommy-waltmann (commented):
I have no issues with the code, but there appear to be build issues in github actions with no output to explain the failures.

joaander:
> I have no issues with the code, but there appear to be compilation issues.

GitHub Actions is not currently functional: https://www.githubstatus.com/

PR #1551, opened 2023-05-12
Fix Circular References

### LINKED ISSUE #1457: Attaching a logger to a writer prevents GC from cleaning up memory

b-butler:
## Description

This fixes memory leaks associated with using a logger and perhaps other parts of HOOMD's interface using `weakref`. Why this is necessary, I don't know. I think it may have something to do with a C++ object holding a reference to a Python object which holds a reference to the C++ object, but that is purely conjecture.

This moves simulation references to weak references generally and moves the logger to use weak references for objects expected to live beyond the call. Unfortunately simply always wrapping leads to cases like
```python
logger[("new", "loggable")] = (lambda : 42, "scalar")
```
not working since the only reference to the function is in the logger. Similarly and perhaps more damningly,

```python
def add_quantities(logger):
    log_quantities = LogValues()
    logger["A"] = (log_quantities, "A")
    logger["B"] = (log_quantities, "B")
    logger["C"] = (log_quantities, "C")
```

would not work

## Motivation and context

This is to fix multiple issues we have had with the garbage collector such as #1457.

Resolves #1457

## How has this been tested?

Existing tests pass which ensures the same behavior as before (well tests will pass when I make this not a draft).

## Change log

```
Fixed: Potential memory leaks.
```

## Checklist:

### DIFF
--- a/hoomd/data/syncedlist.py
+++ b/hoomd/data/syncedlist.py
@@ -215,7 +215,7 @@ class SyncedList(MutableSequence):
 
     @property
     def _simulation(self):
-        sim = self.__simulation
+        sim = self._simulation_
         if sim is not None:
             sim = sim()
             if sim is not None:
@@ -225,7 +225,7 @@ class SyncedList(MutableSequence):
     def _simulation(self, simulation):
         if simulation is not None:
             simulation = weakref.ref(simulation)
-        self.__simulation = simulation
+        self._simulation_ = simulation
 
     def _sync(self, simulation, synced_list):
         """Attach all list items and update for automatic attachment."""
@@ -260,7 +260,7 @@ class SyncedList(MutableSequence):
     def __getstate__(self):
         """Get state for pickling."""
         state = copy(self.__dict__)
-        state['_simulation'] = None
+        state['_simulation_'] = None
         state.pop('_synced_list', None)
         return state
 
--- a/hoomd/data/syncedlist.py
+++ b/hoomd/data/syncedlist.py
@@ -5,6 +5,7 @@
 
 from collections.abc import MutableSequence
 from copy import copy
+import weakref
 
 from hoomd.data.typeconverter import _BaseConverter, TypeConverter
 
@@ -211,6 +212,20 @@ class SyncedList(MutableSequence):
         except ValueError as verr:
             raise ValueError(f"Validation failed: {verr.args[0]}") from verr
         return processed_value
+
+    @property
+    def _simulation(self):
+        sim = self.__simulation
+        if sim is not None:
+            sim = sim()
+            if sim is not None:
+                return sim
+
+    @_simulation.setter
+    def _simulation(self, simulation):
+        if simulation is not None:
+            simulation = weakref.ref(simulation)
+        self.__simulation = simulation
 
     def _sync(self, simulation, synced_list):
         """Attach all list items and update for automatic attachment."""
--- a/hoomd/data/syncedlist.py
+++ b/hoomd/data/syncedlist.py
@@ -71,6 +71,7 @@ class SyncedList(MutableSequence):
                  callable_class=False,
                  attach_members=True):
         self._attach_members = attach_members
+        self._simulation_ = None
         if to_synced_list is None:
             to_synced_list = identity
 
--- a/hoomd/data/syncedlist.py
+++ b/hoomd/data/syncedlist.py
@@ -220,6 +220,8 @@ class SyncedList(MutableSequence):
             sim = sim()
             if sim is not None:
                 return sim
+            else:
+                self._unsync()
 
     @_simulation.setter
     def _simulation(self, simulation):
--- a/hoomd/operation.py
+++ b/hoomd/operation.py
@@ -52,8 +52,9 @@ class _HOOMDGetSetAttrBase:
 
     def __getattr__(self, attr):
         if attr in self._reserved_default_attrs:
-            default = self._reserved_default_attrs[attr]
-            value = default() if callable(default) else default
+            value = self._reserved_default_attrs[attr]
+            if callable(value):
+                value = value()
             object.__setattr__(self, attr, value)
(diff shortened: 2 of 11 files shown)


hoomd/data/syncedlist.py
```diff
@@ -212,6 +213,20 @@ def _validate_or_error(self, value):
             raise ValueError(f"Validation failed: {verr.args[0]}") from verr
         return processed_value
 
+    @property
+    def _simulation(self):
+        sim = self._simulation_
+        if sim is not None:
+            sim = sim()
+            if sim is not None:
+                return sim
```
tommy-waltmann, line 223: What should this property return in the case that `sim is None` after dereferencing the weakref to `sim`? I can't imagine code that uses the `_simulation` property will want to always have to double check to make sure `_simulation` actually returned something.

hoomd/operation.py
```diff
@@ -310,6 +312,21 @@ def _attach(self, simulation):
...
         return self._cpp_obj is not None
 
+    @property
+    def _simulation(self):
+        """Get/set reference to weakly referenced simulation."""
+        sim = self._simulation_
+        if sim is not None:
+            sim = sim()  # grab weakref
+            if sim is not None:
+                return sim
```
tommy-waltmann, line 329: Same as above comment.

tommy-waltmann (changes requested):
Have the changes in this PR been tested to make sure the memory leaks are removed?

joaander:
> Have the changes in this PR been tested to make sure the memory leaks are removed?

Yes. It solves the issue in both `hoomd-validation` and the case in #1457.

hoomd/data/syncedlist.py
joaander: As far as I understand it, that case should never happen. All objects holding this are held by a `Simulation`. `_simulation` is a back reference to that object. When `_simulation` returns None (the implied return value when there is no `return` statement), it is the same as if the object has not yet been added to a Simulation.

We could add this check for completeness. The only way to trigger it would be to directly access internal data structures and bypass HOOMD's attach/detach and/or add/remove APIs.
```python
            if sim is not None:
                return sim
            else:
                raise RuntimeError("Parent Simulation was garbage collected.")
```

b-butler: Yeah, we could error here or make sure we are detached (which would be my preference. This could happen in some unusual cases where the operation's lifespan is longer than the simulation's.

tommy-waltmann: In my original comment, I didn't think about the implied return value of `None` when there is no return statement, but I like the idea of fully checking for attachment when the weakref points to `None`

tommy-waltmann (approved):

PR #1552, opened 2023-05-15
More explicit attachment errors.

b-butler:
## Description
Add an error message when any object fails to attach to a simulation with the class name.

## Motivation and context
This will significantly aid debugging by providing necessary information at attachment about what is causing the problem. Despite the `ParameterDict` and `TypeparameterDict` infrastructure, there are times where pybind11 will raise an error without much context making debugging very difficult. This can be seen in multiple mailing list questions.

## How has this been tested?

We don't normally test error messages, and so I did not here. A test can be added if desired.

## Change log

```
Added:
+ More descriptive error messages when calling `Simulation.run`.
```

## Checklist:

### DIFF
--- a/hoomd/operation.py
+++ b/hoomd/operation.py
@@ -307,8 +307,8 @@ class _HOOMDBaseObject(_HOOMDGetSetAttrBase,
             self._apply_typeparam_dict(self._cpp_obj, self._simulation)
         except Exception as err:
             raise type(err)(
-                f"Object of type {type(self)} could not be setup for "
-                f"simulating.") from err
+                f"Error applying parameters for object of type {type(self)}."
+                ) from err
         self._post_attach_hook()
 
     @property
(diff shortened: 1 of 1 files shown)


hoomd/operation.py
```diff
@@ -302,8 +302,13 @@ def _attach(self, simulation):
...
             self._use_count -= 1
             raise err
-        self._apply_param_dict()
-        self._apply_typeparam_dict(self._cpp_obj, self._simulation)
+        try:
+            self._apply_param_dict()
+            self._apply_typeparam_dict(self._cpp_obj, self._simulation)
+        except Exception as err:
+            raise type(err)(
+                f"Object of type {type(self)} could not be setup for "
```
joaander: "setup for simulating" sounds awkward. 
How about: "Error when applying parameters for object of type {type(self)}"?

hoomd/operation.py
```diff
@@ -302,8 +302,13 @@ def _attach(self, simulation):
...
+                f"simulating.") from err
```
b-butler:
```suggestion
                f"Error applying parameters for object of type {type(self)}."
                ) from err
```

joaander:
pre-commit.ci autofix

joaander (approved):
Thanks!

PR #1556, opened 2023-05-19
File object notice stream

### LINKED ISSUE #1449: File-like object for notice stream

iblanco11981870:
## Description

Provide a Python file-like object that writes to the HOOMD notice message stream.
## Motivation and context
Write a notice message to a file using `hoomd.Device`.

This change is useful for tracking error outputs when you are running operations in parallel or with multiple partitions. This will allow the user to save errors or outputs to each partition into a separate file. 

Resolves #1449 
## How has this been tested?
We implemented the appropriate tests in ``test_device.py``

Tested outputs of different data types and notice levels.

## Change log

```
`hoomd.device.NoticeFile` provides a Python file-like object that can save notice message streams.
```

## Checklist:


hoomd/device.py
```diff
@@ -27,6 +27,42 @@
 from hoomd import _hoomd
 
 
+class NoticeFile:
+    """Write a notice message to a file using `hoomd.Device`.
```
joaander:
```suggestion
    """A file-like object that writes to a `hoomd.Device` notice stream.
```

hoomd/pytest/test_table.py
joaander, line 1: Yes, please restore `test_table.py` and `table.py` to their previous state. You can clean up the history with an interactive rebase, or make a new branch and cherry-pick the changes to keep: c84ba40 and 312aa9c (plus any commits you make applying suggestions).

hoomd/device.py
```diff
@@ -27,6 +27,42 @@
...
+        lines = self._buff.split("\n")
+        for line in lines:
+
+            if "\n" == line[-2:]:
+                pass
+            else:
+                line += "\n"
+
+            self._msg.notice(self._level, line)
+        self._buff = ""
```
joaander: This is not working as intended.
With this test script:
```
import hoomd
cpu = hoomd.device.CPU()
notice_file = hoomd.device.NoticeFile(cpu)

notice_file.write('t')
notice_file.write('e')
notice_file.write('s')
notice_file.write('t')
notice_file.write('1')
notice_file.write('\n')

notice_file.write('test2')
notice_file.write('\n')

notice_file.write('test3\ntest4\n')
```
I get the output:
```
t
e
s
t
1


test2


test3
test4

```
where I would expect:
```
test1
test2
test3
test4
```
(you can set `notice_file` to `sys.stdout` to confirm).

This change allows buffering of input missing `\n` over to the next call to `write` and avoids adding extra newlines to the output.
```suggestion
        lines = self._buff.split("\n")
        for line in lines[:-1]:
            self._msg.notice(self._level, line + "\n")

        if self._buff.endswith("\n"):
            self._msg_notice(self._level, lines[-1] + "\n")
            self._buff = ""
        else:
            self._buff = lines[-1]
```

hoomd/device.py
```diff
@@ -27,6 +27,42 @@
 from hoomd import _hoomd
 
 
+class NoticeFile:
+    """Write a notice message to a file using `hoomd.Device`.
+
+    Args:
+        device (`hoomd.Device`): The `Device` object.
+        level (int): Message notice level. Default value is 1
```
joaander:
```suggestion
        level (int): Message notice level. Default value is 1.
```

hoomd/device.py
```diff
@@ -27,6 +27,42 @@
...
+            if "\n" == line[-2:]:
+                pass
+            else:
+                line += "\n"
+
+            self._msg.notice(self._level, line)
+        self._buff = ""
+
+    def flush(self):
+        """Flush the device."""
```
joaander:
```suggestion
        """Flush the output."""
```

hoomd/device.py
```diff
@@ -27,6 +27,42 @@
...
+        level (int): Message notice level. Default value is 1
+    """
+
+    def __init__(self, device, level=1):
+        self._msg = device._cpp_msg
+        self._buff = ""
+        self._level = level
+
+    def write(self, message):
+        """Writes the notice message to the device.
```
joaander:
```suggestion
        """Writes data to the associated devices notice stream.
```

hoomd/device.py
```diff
@@ -27,6 +27,42 @@
 from hoomd import _hoomd
 
 
+class NoticeFile:
+    """Write a notice message to a file using `hoomd.Device`.
+
+    Args:
+        device (`hoomd.Device`): The `Device` object.
+        level (int): Message notice level. Default value is 1
+    """
```
joaander, line 39:
```suggestion
    Note:
        Use this in combination with `Device.message_filename` to combine notice
        messages with output from code that expects file-like objects (such as
        `hoomd.write.Table`.
    """
```

joaander (changes requested):
Thanks! Let's make a few changes, then this is ready to merge and include in v4.0.

hoomd/pytest/test_table.py
iblanco11981870: I just pushed a version that restores test_table.py and table.py to their previous state.

joaander:
@iblanco11981870 It appears that your force push reverted all of the suggested changes. Did you perhaps commit suggestions on GitHub, then force push from local?

iblanco11981870:
> @iblanco11981870 It appears that your force push reverted all of the suggested changes. Did you perhaps commit suggestions on GitHub, then force push from local?

That is exactly what I did. Should I now do the suggested changes manually and add them to my local environment?

hoomd/pytest/test_device.py
```diff
@@ -138,3 +138,36 @@ def test_device_notice(device, tmp_path):
     if device.communicator.rank == 0:
         with open(device.msg_file) as fh:
             assert fh.read() == ""
+
+
+def test_noticefile(device, tmp_path):
+
+    # Message file declared. Should output in specified file.
+    device.message_filename = str(tmp_path / "str_message")
+    msg = "This message should output."
```
joaander: These tests are failing with the changes I suggested to `write`. You need to include `'\n'` in the messages to write. Also, it would be helpful if each `assert` as in a separate `test_noticefile_*(device, tmp_path)` function.

joaander (changes requested):
Thanks! Now the tests need to be updated to work with the fixed `write` method.

hoomd/device.py
```diff
@@ -27,6 +27,50 @@
...
+    def write(self, message):
+        """Writes data to the associated devices notice stream.
+
+        Args:
+            message (str): Message to write.
+        """
+        if type(message) != str:
+            self._buff += str(message) + "\n"
+        else:
+            self._buff += message
```
joaander:
```suggestion
        self._buff += message
```

file-like objects expect `str` arguments. It is the responsibility of the caller to convert non-string arguments into strings with appropriate formatting.
```
>>> import sys
>>> sys.stdout.write(1234)
TypeError: write() argument must be str, not int
```

You can add an explicit type check and `raise TypeError`, or let the `+=` operator on `str` raise the error. Then you should remove the test that attempts to write an int.

hoomd/pytest/test_device.py
```diff
@@ -138,3 +138,37 @@ def test_device_notice(device, tmp_path):
...
+    test_msg = str(msg) + "\n"
+    device.message_filename = str(tmp_path / "int_message")
+    notice_file.write(msg)
+
+    if device.communicator.rank == 0:
+        with open(device.message_filename) as fh:
+            assert fh.read() == test_msg + "\n"
+
+    # Test the level argument
+    msg = "This message should not output./n"
```
joaander:
```suggestion
    msg = "This message should not output.\n"
```

joaander: '\n' is a newline character. '/n' is the characters '/' and 'n'.

hoomd/pytest/test_device.py
```diff
@@ -138,3 +138,37 @@ def test_device_notice(device, tmp_path):
...
+
+    # Test notice with a message that is not a string.
+    msg = 123456
+    test_msg = str(msg) + "\n"
+    device.message_filename = str(tmp_path / "int_message")
+    notice_file.write(msg)
+
+    if device.communicator.rank == 0:
+        with open(device.message_filename) as fh:
+            assert fh.read() == test_msg + "\n"
```
joaander:
```suggestion
    # Test notice with a message that is not a string.
    msg = 123456
    device.message_filename = str(tmp_path / "int_message")
    with pytest.raises(TypeError):
         notice_file.write(msg)
```

joaander (changes requested):
There was no need to merge `trunk-patch` into this branch. The feature you are developing does not rely on any of those bug fixes. GitHub is now confused about which changes were actually made on your branch and is showing changes from other pull requests. To attempt to resolve this, merge the latest `trunk-major` into your branch.

iblanco11981870:
> There was no need to merge `trunk-patch` into this branch. The feature you are developing does not rely on any of those bug fixes. GitHub is now confused about which changes were actually made on your branch and is showing changes from other pull requests. To attempt to resolve this, merge the latest `trunk-major` into your branch.

My apologies. I just did that. Please let me know if there is anything else I should do

joaander:
It appears that GitHub will remain hopelessly confused. Reviewers will need to ignore the changes GitHub is showing that come from #1551.

joaander:
pre-commit.ci autofix

joaander (approved):
Thanks for working on this! I made a few changes so that `NoticeFile` will appear in the documentation. With that, this is good to merge.

Issue #568, opened 2019-12-19
hoomd.update.box_resize shifting particle positions


patricklawton:
## Description
When running with mpi, a call to hoomd.update.box_resize causes the particle positions to shift in the dimensions resized, even if scale_particles=False in hoomd.update.box_resize. 

## Script

import hoomd
from hoomd import hpmc
import numpy as np
hoomd.context.initialize("")
system = hoomd.init.create_lattice(hoomd.lattice.sc(a=0.5), n=[10,10,10])
mc = hpmc.integrate.sphere(d=1.05, seed=1)
mc.shape_param.set('A', diameter=1.0)

print ('Simulation box before resizing:', system.box)
print ('min and max particle positions before resizing:{0}/{1}'.format(np.amin([p.position for p in system.particles], axis=0),np.amax([p.position for p in system.particles], axis=0)))

hoomd.run(1e2)
hoomd.update.box_resize(Lx=system.box.Lx,Ly=system.box.Ly*2,Lz=system.box.Lz*2, scale_particles=False, period=None)

print ('Simulation box after resizing:', system.box)
print ('min and max particle positions after resizing:{0}/{1}'.format(np.amin([p.position for p in system.particles], axis=0),np.amax([p.position for p in system.particles], axis=0)))

## Output
Simulation box before resizing: Box: Lx=5.0 Ly=5.0 Lz=5.0 xy=0.0 xz=0.0 yz=0.0 dimensions=3
Simulation box before resizing: Box: Lx=5.0 Ly=5.0 Lz=5.0 xy=0.0 xz=0.0 yz=0.0 dimensions=3
min and max particle positions before resizing:[-2.25 -2.25 -2.25]/[2.25 2.25 2.25]
min and max particle positions before resizing:[-2.25 -2.25 -2.25]/[2.25 2.25 2.25]
resize_test.py:012  |  hoomd.run(1e2)
** starting run **
Time 00:00:00 | Step 100 / 100 | TPS 258.419 | ETA 00:00:00
Average TPS: 258.393
---------
notice(2): -- HPMC stats:
notice(2): Average translate acceptance: 0
notice(2): Trial moves per second:        458708
notice(2): Overlap checks per second:     4.4919e+06
notice(2): Overlap checks per trial move: 9.79251
notice(2): Number of overlap errors:      0
** run complete **
[...]

PR #1564, opened 2023-05-25
Account for grid shift in HPMC external fields

### LINKED ISSUE #1562: Unexpected behavior with CPPExternalPotential
### Description

I am running an HPMC simulation with a custom external potential using `hpmc.external.user` and am getting unexpected results. The potential should bias particle positions to a specific z-value, but that is not happening. The particles do not appear to be influenced by the external potential.

@joaander found the problem: the functions in the `ExternalFieldJIT` C++ class do not account for the grid shift when computing energies. I will implement a bugfix very soon.

### Script

```python
import hoomd
```

### Input files

_No response_

### Output

```shell
No output.
```

### Expected output

_No response_

### Platform

CPU, Linux

### Installation method

Compiled from source

### HOOMD-blue version

3.11.0

### Python version

3.9.15

tcmoore3:
## Description

Account for grid shift in member functions of C++ class `ExternalFieldJIT`.

## Motivation and context

Fixes a bug.

Resolves #1562 

## How has this been tested?

I updated the `test_gravity` unit test to be more robust. I have also observed the correct behavior in the simulation mentioned in #1562. 

## Change log

```
- Fix bug in hpmc.external.user.CPPExternalPotential
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/ExternalFieldJIT.h
+++ b/hoomd/hpmc/ExternalFieldJIT.h
@@ -103,7 +103,9 @@ template<class Shape> class ExternalFieldJIT : public hpmc::ExternalFieldMono<Sh
             // read in the current position and orientation
             Scalar4 postype_i = h_postype.data[i];
             unsigned int typ_i = __scalar_as_int(postype_i.w);
-            vec3<Scalar> pos_i = vec3<Scalar>(postype_i);
+            vec3<Scalar> pos_i = vec3<Scalar>(postype_i) - vec3<Scalar>(this->m_pdata->getOrigin());
+            int3 image = make_int3(0, 0, 0);
+            box.wrap(pos_i, image);
 
             total_energy += energy(box,
                                    typ_i,
@@ -161,7 +163,12 @@ template<class Shape> class ExternalFieldJIT : public hpmc::ExternalFieldMono<Sh
             // read in the current position and orientation
             Scalar4 postype_i = h_postype.data[i];
             unsigned int typ_i = __scalar_as_int(postype_i.w);
-            vec3<Scalar> pos_i = vec3<Scalar>(postype_i);
+            int3 image = make_int3(0, 0, 0);
+            vec3<Scalar> pos_i = vec3<Scalar>(postype_i) - vec3<Scalar>(this->m_pdata->getOrigin());
+            box_new.wrap(pos_i, image);
+            image = make_int3(0, 0, 0);
+            vec3<Scalar> old_pos_i = vec3<Scalar>(*(position_old + i));
+            box_old.wrap(old_pos_i, image);
             dE += energy(box_new,
                          typ_i,
                          pos_i,
@@ -170,7 +177,7 @@ template<class Shape> class ExternalFieldJIT : public hpmc::ExternalFieldMono<Sh
                          h_charge.data[i]);
             dE -= energy(box_old,
                          typ_i,
-                         vec3<Scalar>(*(position_old + i)),
+                         old_pos_i,
                          quat<Scalar>(*(orientation_old + i)),
                          h_diameter.data[i],
                          h_charge.data[i]);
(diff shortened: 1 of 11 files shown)


hoomd/hpmc/pytest/test_external_user.py
joaander, line 1: We discussed improvements to the test offline.

hoomd/hpmc/ExternalFieldJIT.h
```diff
@@ -161,7 +163,12 @@ template<class Shape> class ExternalFieldJIT : public hpmc::ExternalFieldMono<Sh
             // read in the current position and orientation
             Scalar4 postype_i = h_postype.data[i];
             unsigned int typ_i = __scalar_as_int(postype_i.w);
-            vec3<Scalar> pos_i = vec3<Scalar>(postype_i);
+            int3 image = make_int3(0, 0, 0);
+            vec3<Scalar> pos_i = vec3<Scalar>(postype_i) - vec3<Scalar>(this->m_pdata->getOrigin());
+            box_new.wrap(pos_i, image);
+            image = make_int3(0, 0, 0);
+            vec3<Scalar> old_pos_i = vec3<Scalar>(*(position_old + i));
```
joaander: * `old_pos_i` must be shifted by the old origin, which the caller must provide.
* `IntegratorHPMC::attemptBoxResize` should scale the origin.

joaander (changes requested):
Thanks for the fix. I have suggestions for improvement.

hoomd/hpmc/IntegratorHPMC.cc
```diff
@@ -153,6 +153,12 @@ bool IntegratorHPMC::attemptBoxResize(uint64_t timestep, const BoxDim& new_box)
 
     m_pdata->setGlobalBox(new_box);
 
+    // scale the origin
+    Scalar3 old_origin = m_pdata->getOrigin();
```
joaander, line 157: Excellent. Please make similar changes to `BoxResizeUpdater.cc`.

hoomd/hpmc/UpdaterBoxMC.cc
```diff
@@ -305,6 +311,9 @@ inline bool UpdaterBoxMC::box_resize_trial(Scalar Lx,
 
         m_pdata->setGlobalBox(curBox);
 
+        // reset origin
+        m_pdata->translateOrigin(-origin_shift);
```
joaander, line 315: Excellent. Please make similar changes to `UpdaterQuickCompress.cc`.

joaander (changes requested):
Thanks! This looks to be exactly what we need to fix external fields in HPMC with box moves.

I would like to see a test for this. First let's look at the trajectory of a simulation run (not a unit test) and see if we can identify a way to write a unit test. You could use your existing harmonic test, though I think it the results might be more clear if we instead place disks outside of a cylindrical wall. 

Before this fix: when run in MPI, the origin will shift inconsistently with the particles, so at some point particles will either end up stuck inside the cylinder and/or the cylinder may appear to diffuse through the simulation box. After this fix, the trajectory should look normal with particles moving around a fixed cylinder.

tcmoore3:
> Thanks! This looks to be exactly what we need to fix external fields in HPMC with box moves.
> 
> I would like to see a test for this. First let's look at the trajectory of a simulation run (not a unit test) and see if we can identify a way to write a unit test. You could use your existing harmonic test, though I think it the results might be more clear if we instead place disks outside of a cylindrical wall.
> 
> Before this fix: when run in MPI, the origin will shift inconsistently with the particles, so at some point particles will either end up stuck inside the cylinder and/or the cylinder may appear to diffuse through the simulation box. After this fix, the trajectory should look normal with particles moving around a fixed cylinder.

@joaander and I worked out a test that highlights the unexpected behavior, which I will summarize here for future reference. The test system is a system of hard spheres initially at a low packing fraction with an HPMC integrator with move size 0 and a `hpmc.update.BoxMC` updater set to `betaP = 10`. As the simulation runs, the box updater causes the system density to increase. Before the fix (specifically, using the v3.10 release), the grid shift is apparent if you track the trajectories of individual particles as the system compresses; the trajectories appear as wavy lines. In contrast, the trajectories are straight lines with this fix. The attached figure illustrates these differences. Note that in this example, only the "dynamics" are affected and the thermodynamics are still correct. However, in cases where a potential energy function depends on spatial coordinates, as discussed in #1562, the thermodynamics of the system are also incorrect.
image

This fix may also resolve the behavior described in #568.

joaander (approved):
Thanks!

PR #1574, opened 2023-06-13
Setup sybil

joaander:
## Description

Add Sybil documentation example tests to `methods.py` and `thermostats.py`.

## Motivation and context

Users wish to copy and paste example code from the documentation. Sybil tests those snippets to ensure that they work now and in the future.

These tests currently run both in serial and MPI. That may be helpful. However future examples we write might only work in MPI and/or only in serial (or only on the GPU or only on the CPU). `pytest.skip` works within these blocks, however Sybil generates each code block as a separate test. Users will see the `pytest.skip` logic in tests that need to be skipped in some configurations. If we were to add `pytest.skip` in a hidden code block, it would only skip that block and not the next. The `skip: next if` syntax may be a solution to this, I just found it and have not yet tried it: https://sybil.readthedocs.io/en/latest/rest.html#skipping-examples

In my testing, Sybil only parses ``.. code-block::`` sphinx directives and not literals ``Example::``. This format makes the raw docstring more verbose (for users of ``help()``), but the HTML output is as expected:
```
.. rubric:: Example:

.. code-block:: python

    example code
```

Note that the package-level example code the users sees in `module-md-methods.html` comes from `md/methods/__init__.py`, but the example code that Sybil runs in the namespace of all the MD methods is in `md/methods/methods.py`. This is because `__init__.py` imports from `methods.py` to create a different canonical namespace for the methods while Sybil runs all examples in the context of a document (a single .py file).

This pull request adds documentation for previously undocumented methods in `hoomd.util`. Should these be documented, or should they remain hidden?
* `dict_map`
* `dict_fold`
* `dict_flatten`
* `dict_filter`
* `GPUNotAvailableError`

## How has this been tested?

* I have tested this locally and ensured that all code blocks are indeed being tested.
* I have examined the generated html docs and ensured that the formatting is consistent and readable.

## Change log

```
Added:

* Tested example code snippets in select modules.
* ``hoomd.util.make_example_simulation`` - create an example Simulation object.
```

## Checklist:

### DIFF
--- a/hoomd/device.py
+++ b/hoomd/device.py
@@ -33,7 +33,7 @@ See Also:
 
 .. invisible-code-block: python
 
-    # Rename tmp_path to path to avoid giving the users the wrong signal.
+    # Rename pytest's tmp_path fixture for clarity.
     path = tmp_path
 """
 
--- a/hoomd/device.py
+++ b/hoomd/device.py
@@ -16,13 +16,13 @@ with many `hoomd.Simulation` objects.
 
 .. code-block:: python
 
-    cpu = hoomd.device.CPU()
+    device = hoomd.device.CPU()
 
 .. skip: next if(gpu_not_available)
 
 .. code-block:: python
 
-    gpu = hoomd.device.GPU()
+    device = hoomd.device.GPU()
 
 Tip:
     Reuse `Device` objects when possible. There is a non-negligible overhead
@@ -53,7 +53,7 @@ class NoticeFile:
 
     .. code-block:: python
 
-        notice_file = hoomd.device.NoticeFile(device=cpu)
+        notice_file = hoomd.device.NoticeFile(device=device)
 
     Note:
         Use this in combination with `Device.message_filename` to combine notice
@@ -154,7 +154,7 @@ class Device:
 
         .. code-block:: python
 
-            cpu.notice_level = 4
+            device.notice_level = 4
         """
         return self._cpp_msg.getNoticeLevel()
 
@@ -180,11 +180,11 @@ class Device:
 
         .. code-block:: python
 
-            cpu.message_filename = str(path / 'messages.log')
+            device.message_filename = str(path / 'messages.log')
 
         .. code-block:: python
 
-            cpu.message_filename = None
+            device.message_filename = None
 
         Note:
             All MPI ranks within a given partition must open the same file.
@@ -192,7 +192,7 @@ class Device:
             other ranks. Different partitions may open separate files. For
             example:
 
-            .. skip: next if(cpu.communicator.num_ranks % 2 != 0)
+            .. skip: next if(device.communicator.num_ranks != 2)
 
             .. code-block:: python
 
@@ -248,7 +248,7 @@ class Device:
 
         .. code-block:: python
 
-            cpu.notice('Message')
+            device.notice('Message')
 
         Hint:
             Use `notice` instead of `print` to write status messages and your
--- a/hoomd/device.py
+++ b/hoomd/device.py
@@ -12,12 +12,29 @@ User scripts may instantiate multiple `Device` objects and use each with a
 different `hoomd.Simulation` object. One `Device` object may also be shared
 with many `hoomd.Simulation` objects.
 
+.. rubric:: Examples:
+
+.. code-block:: python
+
+    cpu = hoomd.device.CPU()
+
+.. skip: next if(gpu_not_available)
+
+.. code-block:: python
+
+    gpu = hoomd.device.GPU()
+
 Tip:
     Reuse `Device` objects when possible. There is a non-negligible overhead
     to creating each `Device`, especially on the GPU.
 
 See Also:
     `hoomd.Simulation`
+
+.. invisible-code-block: python
+
+    # Rename tmp_path to path to avoid giving the users the wrong signal.
+    path = tmp_path
 """
 
 import contextlib
@@ -31,6 +48,12 @@ class NoticeFile:
     Args:
         device (`Device`): The `Device` object.
         level (int): Message notice level. Default value is 1.
+
+    .. rubric:: Example:
+
+    .. code-block:: python
+
+        notice_file = hoomd.device.NoticeFile(device=cpu)
 
     Note:
         Use this in combination with `Device.message_filename` to combine notice
@@ -48,6 +71,12 @@ class NoticeFile:
 
         Args:
             message (str): Message to write.
+
+        .. rubric:: Example:
+
+        .. code-block:: python
+
+            notice_file.write('Message\\n')
         """
         self._buff += message
 
@@ -120,6 +149,12 @@ class Device:
         default level of 2 shows messages that the developers expect most users
         will want to see. Set the level lower to reduce verbosity or as high as
         10 to get extremely verbose debugging messages.
+
+        .. rubric:: Example:
+
+        .. code-block:: python
+
+            cpu.notice_level = 4
         """
         return self._cpp_msg.getNoticeLevel()
 
@@ -141,18 +176,30 @@ class Device:
         Set `message_filename` to `None` to use the system's ``stdout`` and
         ``stderr``.
 
+        .. rubric:: Examples:
+
+        .. code-block:: python
+
+            cpu.message_filename = str(path / 'messages.log')
+
+        .. code-block:: python
+
+            cpu.message_filename = None
+
         Note:
             All MPI ranks within a given partition must open the same file.
             To ensure this, the given file name on rank 0 is broadcast to the
             other ranks. Different partitions may open separate files. For
             example:
 
-            .. code::
+            .. skip: next if(cpu.communicator.num_ranks % 2 != 0)
+
+            .. code-block:: python
 
                 communicator = hoomd.communicator.Communicator(
                     ranks_per_partition=2)
                 filename = f'messages.{communicator.partition}'
-                device = hoomd.device.GPU(communicator=communicator,
+                device = hoomd.device.CPU(communicator=communicator,
                                           message_filename=filename)
         """
         return self._message_filename
@@ -196,6 +243,12 @@ class Device:
 
         Write the given message string to the output defined by
         `message_filename` on MPI rank 0 when `notice_level` >= ``level``.
+
+        .. rubric:: Example:
+
+        .. code-block:: python
+
+            cpu.notice('Message')
 
         Hint:
             Use `notice` instead of `print` to write status messages and your
@@ -276,6 +329,14 @@ class GPU(Device):
         that all GPUs support concurrent managed memory access and have high
         bandwidth interconnects.
 
+    .. rubric:: Example:
+
+    .. skip: next if(gpu_not_available)
+
+    .. code-block:: python
+
+        gpu = hoomd.device.GPU()
+
     """
 
     def __init__(
@@ -307,6 +368,14 @@ class GPU(Device):
         When `False` (the default), error messages from the GPU may not be
         noticed immediately. Set to `True` to increase the accuracy of the GPU
         error messages at the cost of significantly reduced performance.
+
+        .. rubric:: Example:
+
+        .. skip: next if(gpu_not_available)
+
+        .. code-block:: python
+
+            gpu.gpu_error_checking = True
         """
         return self._cpp_exec_conf.isCUDAErrorCheckingEnabled()
 
@@ -361,10 +430,15 @@ class GPU(Device):
         context manager and continue the simulation for a time. Profiling stops
         when the context manager closes.
 
-        Example::
+        .. rubric:: Example:
 
-            with device.enable_profiling():
-                sim.run(1000)
+        .. skip: next if(gpu_not_available)
+
+        .. code-block:: python
+
+            simulation = hoomd.util.make_example_simulation(device=gpu)
+            with gpu.enable_profiling():
+                simulation.run(1000)
         """
         try:
             self._cpp_exec_conf.hipProfileStart()
@@ -392,6 +466,12 @@ class CPU(Device):
     .. rubric:: MPI
 
     In MPI execution environments, create a `CPU` device on every rank.
+
+    .. rubric:: Example:
+
+    .. code-block:: python
+
+        cpu = hoomd.device.CPU()
     """
 
     def __init__(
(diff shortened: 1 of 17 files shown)


joaander:
@b-butler @DomFijan Take a look at the changes and rendered Sphinx docs and let me know what you think.

hoomd/md/methods/methods.py
```diff
@@ -327,29 +372,79 @@ class ConstantPressure(Thermostatted):
...
+
+            Save and restore the barostat degrees of freedom when continuing
+            simulations:
+
+            .. rubric:: Examples:
+
+            .. code-block:: python
+
+                saved_barostat_dof = npt.barostat_dof
+                # Save to a file.
```
DomFijan: Would it be too much to write the save function here as well? Is this a problem for sybil
```suggestion
                # Save to a file.
                import numpy
                numpy.savetxt('barostat_dof.txt', saved_barostat_dof)
```

hoomd/md/methods/methods.py
```diff
@@ -327,29 +372,79 @@ class ConstantPressure(Thermostatted):
...
+
+            .. code-block:: python
+
+                # Load from the file on next execution.
```
DomFijan: Again, same as above.
```suggestion
                # Load from the file on next execution.
                import numpy
                saved_barostat_dof = numpy.loadtxt('barostat_dof.txt')
```

hoomd/md/methods/methods.py
```diff
@@ -602,17 +704,47 @@ class Langevin(Method):
...
 
         gamma (TypeParameter[ ``particle type``, `float` ]): The drag
             coefficient for each particle type
             :math:`[\mathrm{mass} \cdot \mathrm{time}^{-1}]`.
 
+            .. rubric:: Example:
+
+            .. code-block:: python
+
+                langevin.gamma['A'] = 0.5
```
DomFijan, line 751: Add a comment to explain particle types? Is this too much?
```suggestion
                # set drag for particle type 'A'
                langevin.gamma['A'] = 0.5
```

hoomd/md/methods/methods.py
```diff
@@ -602,17 +704,47 @@ class Langevin(Method):
...
+
         gamma_r (TypeParameter[``particle type``,[`float`, `float` , `float`]]):
             The rotational drag coefficient tensor for each particle type
             :math:`[\mathrm{time}^{-1}]`.
+
+            .. rubric:: Example:
+
+            .. code-block:: python
+
+                langevin.gamma_r['A'] = [1.0, 2.0, 3.0]
```
DomFijan, line 761:
```suggestion
                # set drag for particle type 'A'
                langevin.gamma_r['A'] = [1.0, 2.0, 3.0]
```

hoomd/md/methods/methods.py
```diff
@@ -900,9 +1079,22 @@ class OverdampedViscous(Method):
             coefficient for each particle type
             :math:`[\mathrm{mass} \cdot \mathrm{time}^{-1}]`.
 
+            .. rubric:: Example:
+
+            .. code-block:: python
+
+                overdamped_viscous.gamma['A'] = 0.5
```
DomFijan, line 1099:
```suggestion
                # set drag for particle type 'A'
                overdamped_viscous.gamma['A'] = 0.5
```

hoomd/md/methods/methods.py
```diff
@@ -900,9 +1079,22 @@ class OverdampedViscous(Method):
...
+
         gamma_r (TypeParameter[``particle type``,[`float`, `float` , `float`]]):
             The rotational drag coefficient tensor for each particle type
             :math:`[\mathrm{time}^{-1}]`.
+
+            .. rubric:: Example:
+
+            .. code-block:: python
+
+                overdamped_viscous.gamma_r['A'] = [1.0, 2.0, 3.0]
```
DomFijan, line 1110:
```suggestion
                # set drag for particle type 'A'
                overdamped_viscous.gamma_r['A'] = [1.0, 2.0, 3.0]
```

hoomd/md/methods/methods.py
```diff
@@ -750,42 +892,70 @@ class Brownian(Method):
...
 
-        brownian = hoomd.md.methods.Brownian(filter=hoomd.filter.All(), kT=0.2)
-        brownian.gamma.default = 2.0
-        brownian.gamma_r.default = [1.0, 2.0, 3.0]
-        integrator = hoomd.md.Integrator(dt=0.001, methods=[brownian],
-        forces=[lj])
+    .. code-block:: python
+
+        brownian = hoomd.md.methods.Brownian(filter=hoomd.filter.All(), kT=1.5)
+        simulation.operations.integrator.dt = 0.0001
```
DomFijan: dt is defined already at the beggining? This shouldn't be needed?
```suggestion
```

hoomd/md/methods/thermostats.py
```diff
@@ -73,29 +100,77 @@ class MTTK(Thermostat):
...
 
+            Save and restore the thermostat degrees of freedom when continuing
+            simulations:
+
+            .. rubric:: Examples:
+
+            .. code-block:: python
+
+                saved_translational_dof = mttk.translational_dof
+                # Save to a file.
```
DomFijan:
```suggestion
                # Save to a file.
                import numpy
                numpy.savetxt('translational_dof.txt', saved_translational_dof)
```

hoomd/md/methods/thermostats.py
```diff
@@ -73,29 +100,77 @@ class MTTK(Thermostat):
...
+
+            .. code-block:: python
+
+                # Load from the file on next execution.
+                mttk.translational_dof = saved_translational_dof
```
DomFijan:
```suggestion
                import numpy
                saved_translational_dof = numpy.loadtxt('translational_dof.txt')
                mttk.translational_dof = saved_translational_dof
```

hoomd/md/methods/thermostats.py
```diff
@@ -73,29 +100,77 @@ class MTTK(Thermostat):
...
+
+            Save and restore the thermostat degrees of freedom when continuing
+            simulations:
+
+            .. rubric:: Examples:
+
+            .. code-block:: python
+
+                saved_rotational_dof = mttk.rotational_dof
+                # Save to a file.
```
DomFijan:
```suggestion
                # Save to a file.
                import numpy
                numpy.savetxt('rotational_dof.txt', saved_rotational_dof)
```

hoomd/md/methods/thermostats.py
```diff
@@ -73,29 +100,77 @@ class MTTK(Thermostat):
...
+
+            .. code-block:: python
+
+                # Load from the file on next execution.
```
DomFijan:
```suggestion
                # Load from the file on next execution.
                import numpy
                saved_rotational_dof = numpy.loadtxt('rotational_dof.txt')
```

DomFijan (changes requested):
I went through the documentation. This is looks really good and I think it will be very useful, especially for newish users.

I have some suggestions, they might be too much in some cases and if you feel like it is too much feel free to discard proposed changes.

hoomd/md/methods/methods.py
joaander: This is not a problem with Sybil. I'm concerned about dictating a particular solution for something that users may be opinionated on and there are so many potential solutions (numpy, json, bson, signac job doc, signac data store, h5py). I'm also concerned about providing a solution that will "just work" for a high fraction of users that copy and paste it (e.g. saving to 'file.ext' assumes that signac users are using the `with job` to change the current working directory)..

We could reword the comment to "# Add your own code to save barostat_dof to persistent storage." @b-butler what do you think?

hoomd/md/methods/methods.py
joaander: Yes, this is too much. Comments should not restate the line or line(s) of code that follow.

hoomd/md/methods/methods.py
joaander: This dt is 1/10 that of the previous dt. It goes along with the new warning about numerical stability. My hope is that users copy it with Brownian and I get future questions in the future about Brownian simulations blowing up :)

hoomd/md/methods/methods.py
joaander: I've thought about this more. We should either remove the example entirely or include an example of saving/restoring. `numpy` is the lowest common denominator for this. Users of workflows will need to adapt filenames as needed after pasting. I would also use `numpy.save` instead of `savetxt` though to encourage good practices. Agreed?

b-butler: @joaander I think keeping the storage and using NumPy (definitely `save` over `savetxt`) is fine. I also think JSON through `json.dump` is also fine and perhaps another low common denominator.

In general with questions like this, I think showing the saving of the barostat DOF is better than not because it moves users towards restartable jobs over not including it. Examples which have multiple solutions, but emphasize a general best practice are fine to include. Power users will still do what they need and new users wouldn't have even saved the barostat DOF to begin with.

joaander: It would be ideal to place these generated files in a temp directory (otherwise, they appear in the directory where pytest is executed). Would the examples look OK if they wrote to `path / 'file.extension'`? The definition of `path` would not be shown and up to the user - in pytest I would make it be `tmp_path` https://docs.pytest.org/en/7.3.x/how-to/tmp_path.html#tmp-path.

joaander:
Updated with `device.py` examples ensuring that the skip next syntax allows us to write GPU and MPI specific examples that are tested when possible and skipped when not. This is as far as I'm taking this specific pull request. Future pull requests will add examples in batches to other parts of the code.

hoomd/device.py
```diff
@@ -141,18 +176,30 @@ def message_filename(self):
...
+            cpu.message_filename = None
+
         Note:
             All MPI ranks within a given partition must open the same file.
             To ensure this, the given file name on rank 0 is broadcast to the
             other ranks. Different partitions may open separate files. For
             example:
 
-            .. code::
+            .. skip: next if(cpu.communicator.num_ranks % 2 != 0)
```
b-butler: Why not use an equal sign here `== 2`? We only run tests on 1 or two ranks.

hoomd/md/methods/__init__.py
```diff
@@ -15,6 +15,21 @@
...
+.. rubric:: Preparation
+
+Create a `hoomd.md.Integrator` to accept an integration method (or methods):
+
+.. code-block:: python
+
+    simulation = hoomd.util.make_example_simulation()
+    simulation.operations.integrator = hoomd.md.Integrator(dt=0.001)
+
+Create a `hoomd.logging.Logger` that will be used in the examples below:
```
b-butler: From a documentation user's perspective, there is no need to create the object beforehand. I would probably make this hidden given it is already quite separate from the examples using it.

hoomd/md/methods/__init__.py
```diff
@@ -15,6 +15,21 @@
 .. rubric:: Integration methods with constraints
 
 For methods that constrain motion to a manifold see `hoomd.md.methods.rattle`.
+
+.. rubric:: Preparation
```
b-butler: I am not sure that preparation makes sense as a title from a documentation reader's perspective. Maybe "Attaching an Integrator" could be the title.

hoomd/md/methods/methods.py
```diff
@@ -72,45 +82,61 @@ def _thermostat_setter(self, new_thermostat):
...
 
     Args:
         filter (hoomd.filter.filter_like): Subset of particles on which to
             apply this method.
 
-        thermostat (hoomd.md.methods.thermostats.Thermostat): thermostat use do
-            to control temperature. Setting this to ``None`` produces NVE
-            dynamics. Defaults to ``None``
+        thermostat (hoomd.md.methods.thermostats.Thermostat): Thermostat to
+            control temperature. Setting this to ``None`` yields constant energy
```
b-butler: Wouldn't use yield as it is a Python keyword.
```suggestion
            control temperature. Setting this to ``None`` generates constant energy
```

hoomd/md/methods/methods.py
```diff
@@ -154,45 +180,45 @@ class ConstantPressure(Thermostatted):
         filter (hoomd.filter.filter_like): Subset of particles on which to apply
             this method.
 
-        thermostat (hoomd.md.methods.thermostats.Thermostat): Thermostat used to
-            control temperature. Setting this to ``None`` yields NPH
-            integration.
+        thermostat (hoomd.md.methods.thermostats.Thermostat): Thermostat to
+            control temperature. Setting this to ``None`` yields constant
```
b-butler:
```suggestion
            control temperature. Setting this to ``None`` generates constant
```

hoomd/md/methods/thermostats.py
```diff
@@ -3,15 +3,47 @@
 
 """Provide classes for thermostatting simulations.
 
-Classes are for use with `hoomd.md.methods.ConstantVolume` and
+The thermostat classes are for use with `hoomd.md.methods.ConstantVolume` and
 `hoomd.md.methods.ConstantPressure`.
 
+.. rubric:: Preparation
```
b-butler: See comment above.

hoomd/md/methods/thermostats.py
```diff
@@ -3,15 +3,47 @@
...
+    constant_volume = hoomd.md.methods.ConstantVolume(filter=hoomd.filter.All())
+    simulation.operations.integrator = hoomd.md.Integrator(
+        dt=0.001,
+        methods=[constant_volume])
+
+Create a `hoomd.logging.Logger` that will be used in the examples below:
+
+.. code-block:: python
+
+    logger = hoomd.logging.Logger()
```
b-butler: See comment above.

b-butler (commented):
This is a great start and definitely going to aid users. I left some small comment to address.

hoomd/md/methods/__init__.py
DomFijan: Or maybe move logger creation to the two examples that actually use the logger?

joaander: We discussed this at the hoomd development meeting and decided to hide the setup code. The minimal examples should remain minimal. They are copy and pasteable to the extent that the user has common variable names in their scripts ("simulation", "logger", etc...). If they do not have these objects already, then can find examples with a search.

The examples would all be very long if they included all setup necessary, and would then not be copy and pasteable. Copying `logger = hoomd.logging.Logger()` into your script would overwrite any existing `logger` instance you might have.

b-butler (approved):
LGTM!

DomFijan (approved):

PR #1576, opened 2023-06-19
Prevent ValueError when importing HOOMD in a non-main thread.

joaander:
## Description

See https://github.com/glotzerlab/gsd/pull/258 for details.

## Change log

```
*Fixed:*

* Prevent ``ValueError: signal only works in main thread of the main interpreter`` when importing
  gsd in a non-main thread.
```

## Checklist:

### DIFF
--- a/hoomd/__init__.py
+++ b/hoomd/__init__.py
@@ -104,4 +104,7 @@ sys.excepthook = _hoomd_sys_excepthook
 
 # Install a SIGTERM handler that gracefully exits, allowing open files to flush
 # buffered writes and close.
-signal.signal(signal.SIGTERM, lambda n, f: sys.exit(1))
+try:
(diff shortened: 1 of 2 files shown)


hoomd/__init__.py
```diff
@@ -104,4 +104,7 @@ def _hoomd_sys_excepthook(type, value, traceback):
 
 # Install a SIGTERM handler that gracefully exits, allowing open files to flush
 # buffered writes and close.
```
b-butler: Probably worth mentioning why we are using the try block here as it is non-obvious.

b-butler (commented):
Looks fine, just think a comment would help future developers if there is ever need to touch this code.

b-butler (approved):

Issue #775, opened 2020-09-04
Migrate MPCD component to v4


mphoward:
## Description

The MPCD component needs to be migrated to the v4 API. This is more involved than some other components because MPCD has its own particle data structures. Tasks and related issues:

- [X] PR #743 removed all of the MPCD python unit tests. These files need to be restored and converted to pytest framework.
- [X] MPCD has many C++ unit tests. I think these mostly need to stay because they test features that are not exposed to python, but they should be checked carefully if there are any that can be consolidated or removed.
- [X] MPCD has some slow-running C++ unit tests that we previously marked as "validation" so that they ran infrequently. This choice should be revisited given the new test setup.
- [x] Issue #662 is a C++ refactoring of the geometries that should be documented and tests updated.
- [X] Issue #663 is a C++ and/or CMake refactoring that might make plugins work a little better and support HIP.
- [x] Issue #772 is a C++ refactoring of the the virtual particle filler API to be more flexible.
- [x] MPCD needs inline examples updated to the new API and possibly expanded as jupyter notebooks.
- [x] Make sure code formatting is enabled for MPCD.
- [x] Do a final review of the API and make any last tweaks.
- [ ] Run tests on Frontier (AMD GPUs).
- [x] Write release notes on MPCD. Consider a full migration guide.

joaander:
@mphoward While I still don't have a hard timeframe for a 3.0.0 final release, the todo list on the project board is getting smaller. I would guess that 3.0.0 final will be completed this summer. Do you plan to update MPCD for that release, or should we delay MPCD until a future 3.x feature release?

mphoward:
I think I can do the migration and breaking changes by this summer, as we would like to be able to use MPCD with other v3 features. Could we have a short discussion to walk through what needs to be done to migrate a component? Feel free to email me.

mphoward:
@joaander I am still interested to migrate this component, but I don't really know where to start given the large number of changes that have been made on the python side.

I am guessing this will not get done in time for a 3.0.0 release, and I would rather wait to implement all of the proposed changes above in one 3.x release.

joaander:
@mphoward understood. If you need guidance, feel free to set up a real time meeting with @b-butler and I and we can give you an overview of the new Python API internals and how the data model is laid out in v3. We have yet to write overview documentation for it (#979)

To time all this work in one 3.x release, we can target the independently reviewed PRs to a staging branch and then merge the staging branch to master when all work is complete.

[...]

PR #1580, opened 2023-06-21
Update mpcd initialization for v4

matthewddo:
## Description

Update snapshot and system data to new API, currently work in progress.

To be more consistent with the new initialization API, the MPCD snapshot data structure is integrated into the HOOMD snapshot data so that MPCD particle data is synced with the HOOMD system. We are going to move the MPCD `SystemData` into the HOOMD `SystemDefinition` so that the the core data structures are also more closely synced during the simulation. This will simplify processes like initializing, replicating, etc.

## Motivation and context

It didn't work anymore. :)

This is the first step in #775. Additional PRs will follow to migrate other features. These PRs will all be targeted onto `mpcd-v4` as a staging branch, which is based off `v4.0.0`. This branch can eventually be merged into `trunk-minor` in a feature release because v3 did not have MPCD as part of its API.

## How has this been tested?

Unit tests are being ported to pytest as features are moved.

## Change log

```
MPCD particle data is re-enabled and included in snapshots.
```

## Checklist:


joaander:
Please maintain equal performance for non-MPCD users and be aware of the `get_snapshot` performance improvements I made in #1538.

Write as little or as detailed information in the change log and migration guide as you prefer.

mphoward:
> Please maintain equal performance for non-MPCD users and be aware of the `get_snapshot` performance improvements I made in #1538.

Sounds good! We will check in #1538 and apply to MPCD since it was copied originally.

For non-MPCD users, I think these changes will fully preserve performance since there is very little overhead to the empty MPCD data structures, similarly to the topology data structures. We can be even more aggressive than we currently are, though, and not allocate *any* MPCD data if there are no particles present in the snapshot.

For MPCD users, I anticipate `get_snapshot` is going to get very slow if the user only wants access to the MD data, since it will pull all the MPCD data if present. This would not be desirable if someone wanted to do something with only the MD particles (common use case) in a snapshot. A related issue is possible in `set_snapshot`, where partial initialization may be desirable (e.g., only initialize from part of a snapshot). I propose adding boolean flags to the principle groups of data initialized with the following defaults:
```
state.get_snapshot(particles=True, topology=True, mpcd=None)
state.set_snapshot(particles=True, topology=True, mpcd=None)
```
(`None` converting to `True` in MPCD builds and `False` in non-MPCD builds). This would preserve the current defaults. I can open a separate issue after this PR is ready so we can discuss this separately.

FYI: this is still a WIP, but we are getting closer.

hoomd/mpcd/Communicator.cc
```diff
@@ -204,7 +196,7 @@ void mpcd::Communicator::communicate(uint64_t timestep)
 
     // force the cell list to adopt the correct dimensions before proceeding,
     // which will trigger any size change signals
-    m_mpcd_sys->getCellList()->computeDimensions();
+    m_cl->computeDimensions();
```
mphoward, line 204: Need to add a check that `m_cl` is not null

hoomd/mpcd/__init__.py
```diff
@@ -171,9 +171,9 @@ def __init__(self, dt, aniso=None):
             self.set_params(aniso=aniso)
 
     _aniso_modes = {
-        None: _md.IntegratorAnisotropicMode.Automatic,
-        True: _md.IntegratorAnisotropicMode.Anisotropic,
-        False: _md.IntegratorAnisotropicMode.Isotropic
+        #None: _md.IntegratorAnisotropicMode.Automatic,
```
mphoward: This was a quick hack to get this to run. Will make an empty dict before merging, then this will be rewritten later anyway.

hoomd/snapshot.py
```diff
@@ -288,6 +288,38 @@ def constraints(self):
...
+                Names of the particle types.
+
+            mpcd.typeid ((*N*, ) `numpy.ndarray` of ``uint32``):
+                Particle type id.
+
+            mpcd.mass (float): Particle mass.
+
+        Note:
+            Set ``N`` to change the size of the arrays.
+        """
```
mphoward, line 408: Maybe use `hoomd.version` to check for MPCD support and error out first, that will simplify what comes next.

hoomd/snapshot.py
```diff
@@ -288,6 +288,38 @@ def constraints(self):
...
+            mpcd.types (list[str]):
+                Names of the particle types.
+
+            mpcd.typeid ((*N*, ) `numpy.ndarray` of ``uint32``):
+                Particle type id.
+
+            mpcd.mass (float): Particle mass.
+
+        Note:
+            Set ``N`` to change the size of the arrays.
```
mphoward, line 403: We need to note here that this is only when built with MPCD.

hoomd/snapshot.py
mphoward:
```
if not hoomd.version.mpcd_built:
    raise RuntimeError("MPCD component not built")

if self.communicator.rank == 0:
    return self._cpp_obj.mpcd
else:
    raise RuntimeError("Snapshot data is only present on rank 0")
```

hoomd/mpcd/Communicator.cc
mphoward: Put this at top of function.
```
if (!m_cl)
    {
    throw std::runtime_error("Cell list must be set");
    }
```
See base `CollisionMethod`

cslominski:
The flags BUILD_MPCD and BUILD_VALIDATION (which appears to only be used in mpcd) are undocumented either by HOOMD or by the command `cmake -LAH`. I am turning off BUILD_MPCD and am not specifying BUILD_VALIDATION in my builds. Are these the correct settings, and could this information please be added to the documentation moving forward? Thanks.

joaander:
@cslominski `BUILD_VALIDATION` will be removed and `BUILD_MPCD` will be documented by this pull request or a future pull request in this series.

mphoward:
pre-commit.ci autofix

hoomd/CMakeLists.txt
```diff
@@ -214,6 +214,18 @@ set(_hoomd_cu_sources BondedGroupData.cu
                       ParticleGroup.cu
                       SFCPackTunerGPU.cu)
 
+# add the MPCD base parts that should go into _hoomd (i.e., core particle data)
+if (BUILD_MPCD)
```
mphoward: We need the same check as below on HIP here, or we will get compile errors.

hoomd/mpcd/data.py
mphoward, line 1: I think this entire file probably should be removed at this point.

mphoward (commented):
Just a few small things to cleanup and then this is ready for review.

I accidentally merged `trunk-major` in when I meant `trunk-minor`. I think this did not add anything, and I still just need to merge in `trunk-minor` and fix small merge conflicts.

mphoward:
> @cslominski `BUILD_VALIDATION` will be removed and `BUILD_MPCD` will be documented by this pull request or a future pull request in this series.

`BUILD_VALIDATION` is now removed. I will make a note to redocument `BUILD_MPCD`, since it will now affect compilation.

joaander:
> I accidentally merged `trunk-major` in when I meant `trunk-minor`. I think this did not add anything, and I still just need to merge in `trunk-minor` and fix small merge conflicts.

Correct. `trunk-major` is currently behind `trunk-minor` (I synchronize these on releases).

hoomd/mpcd/pytest/test_snapshot.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+@pytest.fixture
+def snap():
+    snap_ = hoomd.Snapshot()
+    snap_.configuration.box = [20, 20, 20, 0, 0, 0]
+    return snap_
+
+
+def test_make(snap):
```
tommy-waltmann:
```suggestion
def test_get_set_mpcd_data(snap):
```
I would like to see this test have a name more specific to what its doing.

hoomd/mpcd/pytest/test_snapshot.py
```diff
@@ -0,0 +1,261 @@
...
+        snap.mpcd.types = test_types
+        snap.mpcd.mass = test_mass
+        np.testing.assert_array_equal(snap.mpcd.position, test_positions)
+        np.testing.assert_array_equal(snap.mpcd.velocity, test_velocities)
+        np.testing.assert_array_equal(snap.mpcd.typeid, test_typeids)
+        np.testing.assert_array_equal(snap.mpcd.types, test_types)
+
+
+# test that snapshot can be resized without losing data
+def test_resize(snap):
```
tommy-waltmann:
```suggestion
def test_resize(snap):
    """Test that snapshot can be resized without losing data."""
```
Personally, I like to put these little descriptions of what the test is doing in a one line docstring.

hoomd/mpcd/pytest/test_snapshot.py
```diff
@@ -0,0 +1,261 @@
...
+
+        scale = nx * ny * nz
+        assert snap.mpcd.N == len(pos0) * scale
+        np.testing.assert_allclose(snap.mpcd.position, pos1)
+        np.testing.assert_allclose(snap.mpcd.velocity, vel1)
+        np.testing.assert_array_equal(snap.mpcd.typeid, typeid1)
+        np.testing.assert_equal(snap.mpcd.types, types0)
+
+
+def test_create_from_snap(snap, simulation_factory):
```
tommy-waltmann: This test assumes it gets launched with either 1 or 2 ranks, we should have a line which skips the test if this is not true.

hoomd/mpcd/pytest/test_snapshot.py
```diff
@@ -0,0 +1,261 @@
...
+            np.testing.assert_array_equal(dat[:, 7], targets["typeid"][0:2])
+    else:
+        np.testing.assert_array_equal(dat[:, 0], [0, 1, 2])
+        np.testing.assert_array_equal(dat[:, 1:4], targets["position"])
+        np.testing.assert_array_equal(dat[:, 4:7], targets["velocity"])
+        np.testing.assert_array_equal(dat[:, 7], targets["typeid"])
+
+
+# test for typeid out of range
+def test_bad_typeid(snap, simulation_factory):
```
tommy-waltmann:
```suggestion
def test_bad_typeid(snap, simulation_factory):
    """Test for typeid out of range."""
```

hoomd/mpcd/pytest/test_snapshot.py
```diff
@@ -0,0 +1,261 @@
...
+        snap.mpcd.N = test_positions.shape[0]
+        snap.mpcd.typeid[:] = test_typeids
+        snap.mpcd.types = ["A", "B"]
+    sim = simulation_factory()
+    with pytest.raises(RuntimeError):
+        sim.create_state_from_snapshot(snap)
+
+
+# check for position outside the box
+def test_bad_positions(snap, simulation_factory):
```
tommy-waltmann:
```suggestion
def test_bad_positions(snap, simulation_factory):
    """Check for position outside the box."""
```

hoomd/mpcd/pytest/test_snapshot.py
```diff
@@ -0,0 +1,261 @@
...
+            np.testing.assert_array_equal(dat[:, 4:7], test_velocities[0::2])
+            np.testing.assert_array_equal(dat[:, 7], test_typeids[0::2])
+    else:
+        np.testing.assert_array_equal(dat[:, 0], [0, 1, 2])
+        np.testing.assert_array_equal(dat[:, 1:4], test_positions)
+        np.testing.assert_array_equal(dat[:, 4:7], test_velocities)
+        np.testing.assert_array_equal(dat[:, 7], test_typeids)
+
+
+def test_take_restore(snap, simulation_factory):
```
tommy-waltmann: There is a lot of repeat code between this test and the previous one, and they do very similar things. Can the code be reorganized so there is no duplicated code between them?

tommy-waltmann (changes requested):
Great work so far on the C++ code, it looks like all the necessary changes have been made there!

Now, the python side of the code needs to be updated. Python classes need to be reorganized to use the new data structures for exposing properties in python. The names of the classes should also be capitalized and base classes should not have underscores before the class names.

After that, some basic tests can be written which make sure the python classes are functioning properly. I think some of the current C++ tests can be replaced by equivalent python tests, but that is an optional detail for later.

mphoward:
All suggestions have been applied! For your other comments:

> Now, the python side of the code needs to be updated. Python classes need to be reorganized to use the new data structures for exposing properties in python. The names of the classes should also be capitalized and base classes should not have underscores before the class names.

Yes, this will be done in many additional pull requests. This PR only implements the particle data / initialization interfaces, as the rest of the code cannot be tested without this change. The next PR should be the `Integrator`.

> After that, some basic tests can be written which make sure the python classes are functioning properly.

Yes, there is an entire suite of Python tests that was removed during the v3 port and needs to be copied over & updated:

https://github.com/glotzerlab/hoomd-blue/tree/v2.9.7/hoomd/mpcd/test-py

For example, `test_snapshot.py` is a reimplementation of `data_snapshot_test.py`.

> I think some of the current C++ tests can be replaced by equivalent python tests, but that is an optional detail for later.

I do not plan to replace the C++ tests because (1) they are already written and (2) test some things that are critical for the algorithm to work correctly but are not (and should not be) exposed at the Python level.

tommy-waltmann (commented):
Sounds good! FWIW, I think the rest of the python stuff can be done in a single PR and doesn't need to be split up into a long series.

mphoward:
> Sounds good! FWIW, I think the rest of the python stuff can be done in a single PR and doesn't need to be split up into a long series.

@joaander and I had discussed this a while ago, and we both think doing this in a few PRs will be easier to review. The MPCD python code & tests are probably a few thousand lines. Some of the MPCD objects also don't derive from any of the existing HOOMD classes, so new code will be needed there. I am of course planning to logically group changes into PRs when possible!

tommy-waltmann:
> > Sounds good! FWIW, I think the rest of the python stuff can be done in a single PR and doesn't need to be split up into a long series.
> 
> @joaander and I had discussed this a while ago, and we both think doing this in a few PRs will be easier to review. The MPCD python code & tests are probably a few thousand lines. Some of the MPCD objects also don't derive from any of the existing HOOMD classes, so new code will be needed there. I am of course planning to logically group changes into PRs when possible!

Ok, whatever works.

tommy-waltmann (approved):

mphoward:
There is still one open review request from @klywang, but I think this PR is otherwise ready to merge!

Does somebody need to manually trigger the validation tests? Maybe I can do that, but I'm not sure how.

hoomd/CMakeLists.txt
```diff
@@ -401,10 +413,9 @@ if (BUILD_METAL AND BUILD_MD)
     add_subdirectory(metal)
 endif()
 
-if (NOT ENABLE_HIP OR HIP_PLATFORM STREQUAL "nvcc")
-    if (BUILD_MPCD)
-        add_subdirectory(mpcd)
-    endif()
+if (BUILD_MPCD AND (NOT ENABLE_HIP OR HIP_PLATFORM STREQUAL "nvcc"))
+    target_compile_definitions(_hoomd PUBLIC ENABLE_MPCD)
```
joaander:
```suggestion
    target_compile_definitions(_hoomd PUBLIC BUILD_MPCD)
```
Name this definition consistent with the CMake variable that controls it. Also change all instances where this is used (global search and replace may be useful here).

joaander (changes requested):
Thanks! I have one change request.

hoomd/CMakeLists.txt
mphoward: Done!

joaander (approved):
Thanks!

PR #1586, opened 2023-07-14
Add sybil examples to more files

joaander:
## Description

Add sybil examples to all *.py files in the `hoomd` source directory. Also reword some documentation for clarity and reduce the toctree depth.

## Motivation and context

Provide users with code examples that are tested and known to execute without errors.

## How has this been tested?

I executed sybil tests locally (serial CPU build).

## Change log

No additional change log is needed over that in #1574.

## Checklist:


hoomd/write/gsd.py
```diff
@@ -55,7 +55,7 @@ class GSD(Writer):
 
     Args:
         trigger (hoomd.trigger.trigger_like): Select the timesteps to write.
-        filename (str): File name to write.
+        filename (`str`): File name to write.
```
b-butler: Given the explicit conversion to a string, we now would accept `PurePath` objects as well.

hoomd/simulation.py
```diff
@@ -187,6 +215,12 @@ def create_state_from_gsd(self,
             automatically selected direction. The default value of ``(None,
             None, None)`` will automatically select the number of domains in all
             directions.
+
+        .. rubric:: Example:
+
+        .. code-block:: python
+
+            simulation.create_state_from_gsd(filename = path / 'file.gsd')
```
b-butler: I think that adding a variable `filename` to the invisible code block would be more understandable for the new Python user as `Path` objects are not well known from my estimation.

hoomd/md/update.py
```diff
@@ -220,7 +220,7 @@ def _attach_hook(self):
 
     @log(category="scalar", requires_run=True)
     def summed_exchanged_momentum(self):
-        R"""Returned the summed up exchanged velocity of the full simulation."""
+        """Returned the summed up exchanged velocity of the full simulation."""
```
b-butler:
```suggestion
        """Returns the summed up exchanged velocity of the full simulation."""
```

hoomd/logging.py
```diff
@@ -657,14 +625,20 @@ class they come from. For instance, the `hoomd.md.pair.LJ` class has a
...
         only_default (`bool`, optional): Whether to log only quantities that are
             logged by default. Defaults to ``True``. Non-default quantities
             are typically measures of operation performance rather than
             information about simulation state.
+
+    .. rubric:: Example:
+
+    .. code-block:: python
+
+        logger = hoomd.logging.Logger()
```
b-butler, line 641: See my PR #1588 where I added example logger constructions. We should choose one to keep.

b-butler (commented):
The PR looks good overall I just have a few small questions/comments.

hoomd/write/gsd.py
joaander: Do most HOOMD users know what a `PurePath` object is? We also accept any time that will convert to a filename with `str()` - e.g. `filename=12345`. What is the universal way of documenting that we accept any type that will convert to a string?

hoomd/simulation.py
joaander: I'm not certain that `filename=filename` would be any less confusing to novice users.

hoomd/logging.py
joaander: If you think the additional examples are helpful, we can keep them. I have been avoiding code comments in examples. If additional explanation or a header is needed, put that outside the code block. For example: https://github.com/glotzerlab/hoomd-blue/blob/4e30e11957421a349da1371a6980c12497d058cf/hoomd/md/methods/methods.py#L107-L123

hoomd/box.py
```diff
@@ -176,15 +160,19 @@ class Box:
     .. rubric:: Factory Methods
 
     `Box` has factory methods to enable easier creation of boxes: `cube`,
-    `square`, `from_matrix`, and `from_box`. See the method documentation for
-    usage.
+    `square`, `from_matrix`, and `from_box`. See the each methods documentation
```
SchoeniPhlippsn:
```suggestion
    `square`, `from_matrix`, and `from_box`. See the methods documentation
```

hoomd/box.py
```diff
@@ -455,12 +546,17 @@ def to_matrix(self):
         """(3, 3) `numpy.ndarray` `float`: The upper triangular matrix that \
         defines the box.
 
-        .. code-block:: python
+        .. code-block::
```
SchoeniPhlippsn, line 554: In line 227 you removed `.. code-block::` completely. I would make it consistent throughout the whole file.

hoomd/logging.py
```diff
@@ -681,13 +655,13 @@ def categories(self):
 
     @property
     def string_categories(self):
-        """`list` of `str`: A list of the string names of the allowed \
+        """list[str]: A list of the string names of the allowed \
```
SchoeniPhlippsn, line 658:
```suggestion
        """`list` [`str`]: A list of the string names of the allowed \
```

hoomd/simulation.py
SchoeniPhlippsn: I agree that `filename = filename` might not be the best choice as well. Maybe we can rename `path` to something more specific such as `path_to_input_file`.

hoomd/state.py
```diff
@@ -348,32 +363,74 @@ def set_snapshot(self, snapshot):
 
     @property
     def particle_types(self):
-        """list[str]: List of all particle types in the simulation state."""
+        """list[str]: List of all particle types in the simulation state.
```
SchoeniPhlippsn, line 366:
```suggestion
        """`list` [`str`]: List of all particle types in the simulation state.
```
Same for all the following

SchoeniPhlippsn (commented):
PR looks good. Only some minor comments.

hoomd/simulation.py
b-butler: It isn't so much the name, but the `Path("/path/to/tmp/dir") / "file.gsd")` syntax that I thought might be advanced.

hoomd/logging.py
joaander: These are parsed correctly as is. The napoleon engine in sphinx understands these type annotations and expands `list` and `str` to links. We need explicit links only in cases where we add additional text to the type e.g. "`list` [`str`], optional".

hoomd/box.py
joaander: I made this change intentionally. This block is not a code example, it is a monospace font formatted matrix. I removed the sphinx `.. code-block::` directive and used the restructured text literal block instead: https://docutils.sourceforge.io/0.4/docs/ref/rst/restructuredtext.html#literal-blocks

All visible code examples that execute in sybil must be preceded by the sphinx block `.. code-block:: python`: https://sybil.readthedocs.io/en/latest/rest.html#python - other `.. code-block::` blocks are ignored by sybil.

hoomd/logging.py
joaander: `logger` is needed to run the later examples in the file. We will add more examples (if any) during merge conflict resolution.

hoomd/box.py
joaander: I added a math formatted matrix.

b-butler (approved):
LGTM.

SchoeniPhlippsn (approved):

PR #1588, opened 2023-07-17
HDF5 Logging 

b-butler:
## Description

Adds a HDF5 logger backend for use in HOOMD. We use HDF5 resizable datasets with specified chunk sizes for fast read performance for typical workflows.

## Motivation and context

This provides a small performant binary file object for logger backends.

Resolves #???

## How has this been tested?

Tests have been added, documentation written, and doc-tests made.

## Change log

```
Added: HDF5 logger backend (hoomd.write.HDF5Logger)
```

## Checklist:

### DIFF
--- a/hoomd/logging.py
+++ b/hoomd/logging.py
@@ -660,19 +660,42 @@ class Logger(_SafeNamespaceDict):
         occur for user specified names that are reused.
 
     Args:
-        categories (`list` of `str`, optional): A list of string categories
-            (list of categories can be found in `LoggerCategories`). These are
-            the only types of loggable quantities that can be logged by this
-            logger. Defaults to allowing every type.
+        categories (`list`[`str` ], `LoggerCategories`, optional): Either a
+            list of string categories (list of categories can be found in
+            `LoggerCategories`) or a `LoggerCategories` instance with the
+            desired flags set. These are the only types of loggable quantities
+            that can be logged by this logger. Defaults to allowing every type
+            ``LoggerCategories.ALL``.
         only_default (`bool`, optional): Whether to log only quantities that are
             logged by default. Defaults to ``True``. Non-default quantities
             are typically measures of operation performance rather than
             information about simulation state.
+
+    .. rubric:: Example
+
+    .. code-block:: python
+
+        import hoomd
+        from hoomd.logging import LoggerCategories
+
+        # Accept all loggables
+        logger = hoomd.logging.Logger()
+
+        # Accept only string(s)
+        logger = hoomd.logging.Logger(categories=["string", "strings"])
+
+        # Accept particle and bond categories
+        logger = hoomd.logging.Logger(
+            categories=LoggerCategories.particle | LoggerCategories.bond)
     """
 
     def __init__(self, categories=None, only_default=True):
-        self._categories = LoggerCategories.ALL if categories is None else \
-            LoggerCategories.any(categories)
+        if categories is None:
+            self._categories = LoggerCategories.ALL
+        if isinstance(categories, LoggerCategories):
+            self._categories = categories
+        else:
+            self._categories = LoggerCategories.any(categories)
         self._only_default = only_default
         super().__init__()
 
--- a/hoomd/logging.py
+++ b/hoomd/logging.py
@@ -642,26 +642,17 @@ class Logger(_SafeNamespaceDict):
     .. rubric:: Examples
 
     There are various ways to create a logger with different available
-    loggables. To allow all loggable categories just create a logger.
+    loggables. Create a `Logger` with no options to allow all categories.
 
     .. code-block:: python
 
-        from hoomd.logging import LoggerCategories
-
         logger = hoomd.logging.Logger()
 
-    When restricting to a subset of categories you can use a list of strings.
+    Use a list of strings to log a subset of categories:
 
     .. code-block:: python
 
         logger = hoomd.logging.Logger(categories=["string", "strings"])
-
-    The `LoggerCategories` enum values can also be used directly.
-
-    .. code-block:: python
-
-        logger = hoomd.logging.Logger(
-            categories=LoggerCategories.particle | LoggerCategories.bond)
     """
 
     def __init__(self, categories=None, only_default=True):
--- a/hoomd/logging.py
+++ b/hoomd/logging.py
@@ -653,11 +653,13 @@ class Logger(_SafeNamespaceDict):
     When restricting to a subset of categories you can use a list of strings.
 
     .. code-block:: python
+
         logger = hoomd.logging.Logger(categories=["string", "strings"])
 
     The `LoggerCategories` enum values can also be used directly.
 
     .. code-block:: python
+
         logger = hoomd.logging.Logger(
             categories=LoggerCategories.particle | LoggerCategories.bond)
     """
--- a/hoomd/logging.py
+++ b/hoomd/logging.py
@@ -812,6 +812,10 @@ class Logger(_SafeNamespaceDict):
 
         .. rubric:: Example:
 
+        .. invisible-code-block: python
+
+            logger = hoomd.logging.Logger()
+
         .. code-block:: python
 
             logger[('custom', 'name')] = (lambda: 42, 'scalar')
--- a/hoomd/logging.py
+++ b/hoomd/logging.py
@@ -660,7 +660,7 @@ class Logger(_SafeNamespaceDict):
         occur for user specified names that are reused.
 
     Args:
-        categories (`list`[`str` ], `LoggerCategories`, optional): Either a
+        categories (`list` [`str` ], `LoggerCategories`, optional): Either a
             list of string categories (list of categories can be found in
             `LoggerCategories`) or a `LoggerCategories` instance with the
             desired flags set. These are the only types of loggable quantities
--- a/hoomd/logging.py
+++ b/hoomd/logging.py
@@ -12,6 +12,11 @@ See Also:
     Tutorial: :doc:`tutorial/02-Logging/00-index`
 
     Tutorial: :doc:`tutorial/04-Custom-Actions-In-Python/00-index`
+
+.. invisible-code-block: python
+
+    if hoomd.version.md_built:
+        lj = hoomd.md.pair.LJ(nlist=hoomd.md.nlist.Cell(buffer=0.4))
 """
 
 import copy
@@ -29,11 +34,8 @@ from collections.abc import Sequence
 class LoggerCategories(Flag):
     """Enum that marks all accepted logger types.
 
-    This class does not need to be used by users directly. We directly convert
-    from strings to the enum wherever necessary in the API. This class is
-    documented to show users what types of quantities can be logged, and what
-    categories to use for limiting what data is logged, user specified logged
-    quantities, and custom actions (`hoomd.custom.Action`).
+    The attribute names of `LoggerCategories` are valid strings for the
+    category argument of `Logger` constructor and the `log` method.
 
     Attributes:
         scalar: `float` or `int` object.
@@ -416,12 +418,7 @@ def log(func=None,
             loggable quantity, defaults to 'scalar'. See `LoggerCategories` for
             available types. Keyword only argument.
         default (`bool`, optional): Whether the quantity should be logged
-            by default. Defaults to True. This is orthogonal to the loggable
-            quantity's type. Many users may not want to log such quantities even
-            when logging other quantities of that type. An example would be
-            performance orientated loggable quantities, like the number of
-            neighborlist builds. The `default` argument allows for these to be
-            skipped by `Logger` objects by default. Keyword only argument.
+            by default. Defaults to True. Keyword only argument.
         requires_run (`bool`, optional): Whether this property requires the
             simulation to run before being accessible.
 
@@ -435,19 +432,18 @@ def log(func=None,
         helps to prevent naming conflicts and automates the logging
         specification for developers and users.
 
-    Example::
+    .. rubric:: Example:
 
-        # Metaclass specification is not necessary for
-        # subclasses of HOOMD classes as they already use this
-        # metaclass.
-        class LogExample(metaclass=hoomd.logging.Loggable)
-            @log(category="string")
+    .. code-block:: python
+
+        class LogExample(metaclass=hoomd.logging.Loggable):
+            @log()
             def loggable(self):
-                return "log_me"
+                return 1.5
 
-            @log(is_property=False, default=False)
-            def not_property(self, a=4):
-                return 2 ** a
+    Note:
+        The metaclass specification is not necessary for subclasses of HOOMD
+        classes as they already use this metaclass.
 
     See Also:
         Tutorial: :doc:`tutorial/04-Custom-Actions-In-Python/00-index`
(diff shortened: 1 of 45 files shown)


hoomd/write/hdf5.py
```diff
@@ -0,0 +1,225 @@
...
+    Warning:
+        This class cannot handle string loggables.
+
+    Args:
+        trigger (hoomd.trigger.trigger_like): The trigger to determine when to
+            run the HDF5 backend.
+        filename (str): The filename of the HDF5 file to write to.
+        logger (hoomd.logging.Logger): The logger instance to use for querying
+            log data.
+        mode (`str`, optional): The mode to open the file in. Defaults to "a".
```
tommy-waltmann: Let's add more documentation on supported modes. If they mirror h5 file modes exactly, let's add a link to h5's documentation on that.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,225 @@
...
+
+    Attributes:
+        cls.accepted_categories (hoomd.logging.LoggerCategories): The enum value
+            for all accepted categories for `HDF5Logger` instances.
```
tommy-waltmann: Can we list out the accepted categories somewhere, or would that list be cumbersomely long?

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def flush(self):
+        """Write out all data currently buffered in memory.
+
+        Without calling this, data may be stored in the h5py.File object without
+        being written to the on disk file yet.
+        """
+        self._fh.flush()
+
+    def _create_database(self, key: str, shape, dtype, chunk_size):
```
tommy-waltmann:
```suggestion
    def _create_dataset(self, key: str, shape, dtype, chunk_size):
```
Does this seem like a more appropriate name, given the name of the h5 call in its implementation and context it gets used in?

hoomd/write/__init__.py
tommy-waltmann, line 1: Should we add some description of the HDF5 writer to the docstring at the top of this file? Upon further inspection, it seems we forgot to do this with the burst writer in a previous PR.

hoomd/md/pytest/test_hdf5.py
```diff
@@ -0,0 +1,128 @@
...
+    sim = simulation_factory(two_particle_snapshot_factory())
+    sim.operations.integrator = lj_integrator()
+
+    return sim
+
+
+@pytest.fixture
+def logger():
+    logger = hoomd.logging.Logger(categories=['scalar'])
+    return logger
```
tommy-waltmann: Does a very simple one-liner like this warrant a fixture?

tommy-waltmann (changes requested):
Great work on this so far, I just have a few comments related to code style and documentation.

hoomd/write/hdf5.py
b-butler: That would be quite long (there are only three disallowed categories), but we can mention that.

hoomd/write/hdf5.py
b-butler: Good point.

hoomd/md/pytest/test_hdf5.py
b-butler: Not really, just a stylistic preference, but I will add the line to the tests.

b-butler:
@joaander we in gloterlab/ci-containers

https://github.com/glotzerlab/ci-containers/blob/5795b26faaab6b64d48abae78dff4b629097069e/ubuntu22.04/py310/Dockerfile#L28

exclude h5py from installation in the docker file. Can we undo that, or is there a reason for the commenting out?

hoomd/write/__init__.py
```diff
@@ -30,3 +33,7 @@
 from hoomd.write.gsd_burst import Burst
 from hoomd.write.dcd import DCD
 from hoomd.write.table import Table
+try:
```
joaander: Move the try/except around `import h5py` in `write/hdf5.py`. Will also need to add an error check (e.g. if `h5py is None`) check in `HDF5Logger`. This way, HOOMD developers that modify `write/hd5.py` will see the cause of any `ImportError` related to `hdf5.py` and the `HDF5Logger` class is available (e.g. for sphinx doc builds) when `h5py` is not installed.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Tests were done on 1,000 frame files for writes and reads and tested for
+        writing and reading speed.
+        """
+        for key, (value, category) in log_dict.items():
+            chunk_size = None
+            if category == "scalar":
+                data_shape = (1,)
+                dtype = "f8" if isinstance(value, float) else "i8"
+                chunk_size = (500,)
```
joaander:
```suggestion
                chunk_size = (512,)
```
Not really necessary, but this will make the chunks nicely map neatly into power of 2 filesystem block sizes.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,272 @@
...
+            else:
+                if not isinstance(value, np.ndarray):
+                    value = np.asarray(value)
+                data_shape = (1,) + value.shape
+                dtype = value.dtype
+                chunk_size = [dim if dim <= 10 else dim for dim in data_shape]
+                chunk_size[0] = 1
```
joaander:
```suggestion
                chunk_size = data_shape
```
Unless this code is an error and you intend to perform some modification to the chunk size in the list comprehension.

Also, I would assume that `chunk_size = data_shape` works very well for typical per-particle, per-bond, and other large arrays. It might not work so well for fixed length 2 arrays. For parity with the scalar chunk size above, try:
```
chunk_size[0] = max(4096 // value.nbytes, 1)
```
If this works well, remove magic numbers by making 4096 a named constant and using it in both places we compute a chunk size in dimensions.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,272 @@
...
+
+    Args:
+        trigger (hoomd.trigger.trigger_like): The trigger to determine when to
+            run the HDF5 backend.
+        filename (str): The filename of the HDF5 file to write to.
+        logger (hoomd.logging.Logger): The logger instance to use for querying
+            log data.
+        mode (`str`, optional): The mode to open the file in. Available values
+            are "w", "x" and "w-", "a", and "r+". Defaults to "a". See the
+            h5py_ documentation for more).
```
joaander:
```suggestion
            h5py_ documentation for more details).
```

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,272 @@
...
+    .. invisible-code-block: python
+
+        import hoomd
+
+        simulation = hoomd.util.make_example_simulation()
+
+    .. code-block:: python
+
+        logger = hoomd.logging.Logger(
+            hoomd.write.HDF5Logger.accepted_categories)
```
joaander: Nice!

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,272 @@
...
+        h5_writer = hoomd.write.HDF5Logger(10_000, "simulation-log.h5", logger)
```
joaander:
```suggestion
        h5_writer = hoomd.write.HDF5Logger(
            trigger=10_000,
            filename=h5_filename,
            logger=logger)
```
I like to be explicit with keyword arguments in examples. It makes the code more readable.

Also, per our discussions on #1586, create a `h5_filename` variable in the invisible codeblock to use here:
```
h5_filename = tmp_path / 'file.h5'
```
Without the tmp_path fixture, the file will be placed in the current working directory when executing `pytest`.

sphinx-doc/requirements.txt
```diff
@@ -1,3 +1,4 @@
+h5py
```
joaander:
```suggestion
```
With my suggested change to the `try / except ImportError` above, this is no longer needed.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,272 @@
...
+        simulation.operations += h5_writer
+    """
+    _internal_class = _HDF5LoggerInternal
+
+    def write(self):
+        """Write out data to the HDF5 file.
+
+        Writes out a frame from the composed logger.
+        """
+        self._action.act()
```
joaander: How does this work? `act` takes a `timestep` argument. Also, should we be providing a public facing `write` API at all?

sphinx-doc/module-hoomd-write.rst
```diff
@@ -24,5 +25,8 @@ hoomd.write
     :members: Burst, DCD, CustomWriter, GSD
     :show-inheritance:
 
+    .. autoclass:: HDF5Logger(trigger, filename, logger, mode="a")
+        :members:
```
joaander, line 29:
```suggestion
        :members: flush write
```
`flush` currently doesn't appear in the documentation. If I recall correctly, the internal action's methods are somehow automatically made available to the parent - right? Does this work to make flush appear in the sphinx docs? Or do we need to define a placeholder in `HDF5Logger`?

joaander (changes requested):
Thanks! This will be very helpful for space-efficient log-only files. I have some suggestions for improvement.

h5py for py310 is available in the 2023.07.20 ci container build. At the time I initially created the py310 container, wheels were not available for h5py. This is no longer the case.

The failing test demonstrates that h5py is not actually an optional test dependency as documented. We should fix that.

hoomd/write/hdf5.py
b-butler: I originally did, but it did not change performance. Good catch.

hoomd/write/hdf5.py
b-butler: The timestep is optional in many custom writers. Also, we provide a `write` method in `_InternalCustomWriter`. We could remove that (can't remember why I put it there), but it would technically be a breaking change.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+        The HDF5 file can be used for other data storage; however, the
+        "hoomd-data" key is reserved for use by this class. An exception will be
+        thrown if this requirement is not met.
+
+    Warning:
+        This class cannot handle string, strings, or object loggables.
+
+    Args:
+        trigger (hoomd.trigger.trigger_like): The trigger to determine when to
+            run the HDF5 backend.
```
tommy-waltmann:
```suggestion
        trigger (hoomd.trigger.trigger_like): The trigger to determine when to
            write to the HDF5 file.
```

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+        state = copy.copy(self.__dict__)
+        del state["_fh"]
+        state["_attached_"] = False
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__ = state
+
+
+class HDF5Logger(_InternalCustomWriter):
```
tommy-waltmann: Would this class more appropriately be named simply `HDF5` instead of `HDF5Logger`? It only writes log data (i.e. not snapshots), but its still a writer and hoomd's other writers are named according to the type of file they write to.

If we change this class's name, we may also want to change the name of the internal class.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+    """An HDF5 logger backend.
```
tommy-waltmann:
```suggestion
    """Write loggable simulation data to HDF5 files.
```

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+
+    This class handles scalar and array data storing them in HDF5 resizable
+    datasets.
```
tommy-waltmann:
```suggestion
    This class stores resizable scalar and array data in the HDF5 file format.
```

tommy-waltmann (changes requested):
This looks to be in really good shape, I just have some naming and re-wording suggestions to clarify the documentation.

hoomd/write/hdf5.py
joaander: `HDF5Logger` is awkward because it is _not_ a `Logger`. At the same time putting `HDF5` alongside `GSD` and `DCD` gives some hint that it might also write the system state when it does not.

What do you think of `HDF5Log` as a compromise solution?

tommy-waltmann: I would be okay with that. My idea to fix the potential user confusion was to clarify repeatedly in the class documentation that it writes log data only, which I've tried to do to some extent in the other suggestions in my PR review.

hoomd/logging.py
```diff
@@ -625,25 +628,49 @@ class they come from. For instance, the `hoomd.md.pair.LJ` class has a
...
         only_default (`bool`, optional): Whether to log only quantities that are
             logged by default. Defaults to ``True``. Non-default quantities
             are typically measures of operation performance rather than
             information about simulation state.
 
-    .. rubric:: Example:
+    .. rubric:: Examples
+
+    There are various ways to create a logger with different available
+    loggables. To allow all loggable categories just create a logger.
```
joaander:
```suggestion
    loggables. Create a `Logger` with no options to allow all categories.
```
"just" belittles the reader.

hoomd/logging.py
```diff
@@ -625,25 +628,49 @@ class they come from. For instance, the `hoomd.md.pair.LJ` class has a
...
 
     .. code-block:: python
 
+        from hoomd.logging import LoggerCategories
+
         logger = hoomd.logging.Logger()
```
joaander:
```suggestion
        logger = hoomd.logging.Logger()
```
Unnecessary import.

hoomd/logging.py
```diff
@@ -625,25 +628,49 @@ class they come from. For instance, the `hoomd.md.pair.LJ` class has a
...
+    .. code-block:: python
+
+        logger = hoomd.logging.Logger(categories=["string", "strings"])
+
+    The `LoggerCategories` enum values can also be used directly.
+
+    .. code-block:: python
+
+        logger = hoomd.logging.Logger(
+            categories=LoggerCategories.particle | LoggerCategories.bond)
```
joaander:
```suggestion
    Use a list of strings to log a subset of categories:

    .. code-block:: python

        logger = hoomd.logging.Logger(categories=["string", "strings"])
```
I think using the enum directly is an advanced topic and therefore doesn't need an example. Users that need it can infer the appropriate syntax from the reference documentation.

hoomd/md/pytest/test_hdf5.py
```diff
@@ -0,0 +1,128 @@
...
+
+from hoomd.conftest import operation_pickling_check
+import hoomd
+import hoomd.write
+
+try:
+    import h5py
+except ImportError:
+    pytestmark = pytest.skip("h5py required to test this feature.",
+                             allow_module_level=True)
```
joaander:
```suggestion
pytest.importorskip("h5py")
```

hoomd/write/hdf5.py
joaander: Also fix the description in `Attributes:` below.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+        h5_filename = tmp_path / "simulation_log.h5"
+
+    .. skip: start if(h5py_not_available)
+
+    .. code-block:: python
+
+        logger = hoomd.logging.Logger(
+            hoomd.write.HDF5Logger.accepted_categories)
+        h5_writer = hoomd.write.HDF5Logger(
+            trigger=10_000,
```
joaander:
```suggestion
            trigger=hoomd.trigger.Periodic(10_000),
```
Explicit is better than implicit.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+        self._frame += 1
+        self._fh["hoomd-data"].attrs["frames"] = self._frame
+
+    @_skip_fh
+    def flush(self):
+        """Write out all data currently buffered in memory.
+
+        Without calling this, data may be stored in the h5py.File object without
+        being written to disk yet.
+        """
```
joaander, line 183:
```suggestion
        .. rubric:: Examples
        
        Flush one writer:
        
        .. skip: if(h5py_not_available)
        
        .. code-block:: python
        
            hdf5_log_writer.flush()
            
        Flush all write buffers:
        
        .. skip: if(h5py_not_available)
        
        .. code-block:: python
        
            for writer in simulation.operations.writers:
                if hasattr(writer, 'flush'):
                    writer.flush()
        """
```
Will need to move the invisible setup code up higher in the document for this to pass tests.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+
+        try:
+            import h5py
+            h5py_not_available = False
+        except ImportError:
+            h5py_not_available = True
+
+
+        simulation = hoomd.util.make_example_simulation()
+        h5_filename = tmp_path / "simulation_log.h5"
```
joaander:
```suggestion
        hdf5_filename = tmp_path / "simulation_log.h5"
```
Use the same name throughout. We don't abbreviate with h5 anywhere else. Alternately, use H5 everywhere: e.g. name the class `H5Log` to be consistent with the file extension.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+
+    .. skip: start if(h5py_not_available)
+
+    .. code-block:: python
+
+        logger = hoomd.logging.Logger(
+            hoomd.write.HDF5Logger.accepted_categories)
+        h5_writer = hoomd.write.HDF5Logger(
```
joaander:
```suggestion
        hdf5_log_writer = hoomd.write.HDF5Logger(
```
Match the (expected new) class name.

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+            trigger=10_000,
+            filename=h5_filename,
+            logger=logger)
+        simulation.operations += h5_writer
```
joaander:
```suggestion
            filename=hdf5_filename,
            logger=logger)
        simulation.operations += hdf5_writer
```

hoomd/write/hdf5.py
```diff
@@ -0,0 +1,295 @@
...
+    """
+    _internal_class = _HDF5LoggerInternal
+    _wrap_methods = ("flush",)
+
+    def write(self, timestep=None):
+        """Write out data to the HDF5 file.
+
+        Writes out a frame from the composed logger.
+        """
```
joaander, line 322: If I understand correctly, users can call `write` to add frames to the log that are not triggered by `trigger`.
We need to:
1) Add an example snippet.
2) Test `write` in the unit test.
and 3) Explain clearly that calling `write` will add an additional frame even if not triggered.

One issue with `write` is that the fields requiring any of the the flags `ROTATIONAL_KINETIC_ENERGY`, `PRESSURE_TENSOR`, or `EXTERNAL_FIELD_VIRIAL` may be NaN. Those quantities may or may not be available depending on whether the last timestep executed had those flags set by some other means.

joaander (changes requested):
Thanks! I noted some more consistency and documentation updates to improve this further.

Looking at the implementation of `write` (which accepts `timestep=None`), I realize that this writer doesn't output the timestep.  As a user, I would want to know what timestep my log frames are on, especially in cases with non-periodic triggers.

`GSD` writes the timestep by default. Should we do that here? Or nudge users into using the `Simulation.timestep` loggable by including that in the example?

joaander (approved):
Thanks!

tommy-waltmann (approved):
LGTM

PR #1594, opened 2023-07-25
Fix plugin bond

### LINKED ISSUE #1591: bonds don't work with plugins
### Description

The [bond python class](https://github.com/glotzerlab/hoomd-blue/blob/trunk-patch/hoomd/md/bond.py) isn't working with  plugins because it is missing the string that let's you redefine where the module is implemented, see [pair python class](https://github.com/glotzerlab/hoomd-blue/blob/trunk-patch/hoomd/md/pair/pair.py) for reference. The fix should be fairly simple, see attached python snippet with proposed changes for the bond class. 

I am happy to supply a bug fix. 

### Script
```python
class Bond(Force):
    """Base class bond force.

    `Bond` is the base class for all bond forces.

    Warning:
        This class should not be instantiated by users. The class can be used
        for `isinstance` or `issubclass` checks.
    """

    def __init__(self):
        super().__init__()

    def _attach_hook(self):
        """Create the c++ mirror class."""
        if isinstance(self._simulation.device, hoomd.device.CPU):
            cpp_cls = getattr(_md, self._cpp_class_name)
        else:
            cpp_cls = getattr(_md, self._cpp_class_name + "GPU")

        self._cpp_obj = cpp_cls(self._simulation.state._cpp_sys_def)
 

changed to:

class Bond(Force):
    """Base class bond force.

    `Bond` is the base class for all bond forces.

    Warning:
        This class should not be instantiated by users. The class can be used
        for `isinstance` or `issubclass` checks.
    """

    # Module where the C++ class is defined. Reassign this when developing an
    # external plugin.
    _ext_module = _md

    def __init__(self):
        super().__init__()

    def _attach_hook(self):
        """Create the c++ mirror class."""
        if isinstance(self._simulation.device, hoomd.device.CPU):
            cpp_cls = getattr(self._ext_module, self._cpp_class_name)
        else:
            cpp_cls = getattr(self._ext_module, self._cpp_class_name + "GPU")

        self._cpp_obj = cpp_cls(self._simulation.state._cpp_sys_def)
```

### Input files

_No response_

### Output

When developing a plugin with a new bond potential, it will compile but then not correctly find it, i.e when wanting to use or import "plugin_name.bond.BondPotentialName" it will instead look for "hoomd.bond.BondPotentialName". The simple inclusion of "_ext_module = _md" that can be overwritten by the plugin fixes this issue.

### Expected output

Plugins should be able to develop bond potentials the same way as pair potentials. 

### Platform

CPU, GPU, Linux, macOS

### Installation method

Compiled from source

### HOOMD-blue version

4.0.1

### Python version

3.10

astatt:
## Description

A simple bug fix that allows bond potentials to be developed in external plugins in the same way pair potentials are. 

## Motivation and context

Without this fix, an external plugin with a new bond potential will compile but fail to find the bond when run, i.e. the link will still be to `hoomd.md.bond` instead of `external_plugin_name.bond`. The simple fix is to add the same "_ext_module = _md" logic as is present in the pair potential class, see [here](https://github.com/glotzerlab/hoomd-blue/blob/efbeae37947a739146f8e7fc1da151db34366db5/hoomd/md/pair/pair.py#L65). 

Resolves #1591

## How has this been tested?

Tested via small "example_plugin" with a test bond implementation. No major changes to the code, so all other tests and functionalities are unaffected.  No changes to documentation aside from adding contributor name. 

## Change log

```
Fix: Bond potentials can now be implemented via external plugins 
```

## Checklist:

### DIFF
--- a/hoomd/md/bond.py
+++ b/hoomd/md/bond.py
@@ -52,15 +52,19 @@ class Bond(Force):
         for `isinstance` or `issubclass` checks.
     """
 
+    # Module where the C++ class is defined. Reassign this when developing an
+    # external plugin.
+    _ext_module = _md
+
     def __init__(self):
         super().__init__()
 
     def _attach_hook(self):
         """Create the c++ mirror class."""
         if isinstance(self._simulation.device, hoomd.device.CPU):
-            cpp_cls = getattr(_md, self._cpp_class_name)
+            cpp_cls = getattr(self._ext_module, self._cpp_class_name)
         else:
-            cpp_cls = getattr(_md, self._cpp_class_name + "GPU")
+            cpp_cls = getattr(self._ext_module, self._cpp_class_name + "GPU")
 
         self._cpp_obj = cpp_cls(self._simulation.state._cpp_sys_def)
 
--- a/hoomd/md/bond.py
+++ b/hoomd/md/bond.py
@@ -247,6 +247,10 @@ class Table(Bond):
         width (int): Number of points in the table.
     """
 
+    # Module where the C++ class is defined. Reassign this when developing an
+    # external plugin.
+    _ext_module = _md
+
     def __init__(self, width):
         super().__init__()
         param_dict = hoomd.data.parameterdicts.ParameterDict(width=int)
@@ -266,9 +270,9 @@ class Table(Bond):
     def _attach_hook(self):
         """Create the c++ mirror class."""
         if isinstance(self._simulation.device, hoomd.device.CPU):
-            cpp_cls = _md.BondTablePotential
+            cpp_cls = self._ext_module.BondTablePotential
         else:
-            cpp_cls = _md.BondTablePotentialGPU
+            cpp_cls = self._ext_module.BondTablePotentialGPU
 
         self._cpp_obj = cpp_cls(self._simulation.state._cpp_sys_def, self.width)
 
--- a/hoomd/md/bond.py
+++ b/hoomd/md/bond.py
@@ -247,10 +247,6 @@ class Table(Bond):
         width (int): Number of points in the table.
     """
 
-    # Module where the C++ class is defined. Reassign this when developing an
-    # external plugin.
-    _ext_module = _md
-
     def __init__(self, width):
         super().__init__()
         param_dict = hoomd.data.parameterdicts.ParameterDict(width=int)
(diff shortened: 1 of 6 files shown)


sphinx-doc/credits.rst
```diff
@@ -17,6 +17,7 @@ The following people have contributed to the to HOOMD-blue:
 * Andrew Schultz, University at Buffalo
 * Andrew Mark, Max Planck Institute
 * Andrey Kazennov, Joint Institute for High Temperatures of RAS
+* Antonia Statt, Universit of Illinois
```
tcmoore3:
```suggestion
* Antonia Statt, University of Illinois
```

tcmoore3 (approved):
This looks good, thanks for the fix. Echoing [Josh's comment on the issue](https://github.com/glotzerlab/hoomd-blue/issues/1591#issuecomment-1651801758), doing the same for angle, dihedrals, impropers, and special pairs would be good for completeness.

hoomd/md/bond.py
```diff
@@ -243,6 +247,10 @@ class Table(Bond):
         width (int): Number of points in the table.
     """
 
+    # Module where the C++ class is defined. Reassign this when developing an
+    # external plugin.
+    _ext_module = _md
+
```
b-butler: Unnecessary. Just leave in base class.
```suggestion
```

b-butler (changes requested):
Thank for the fix. Just one change, and I think this is good to go.

astatt:
As @joaander suggested, the same fix should be done for angles, dihedrals, impropers, and special pairs. Would you prefer that I do this in this fix/pull request, your should we do the other fixes seperately? Even if it's less clean, this is such a small fix, it might be easiest to simply leave this pull request open and do all of the fixes at the same time. @b-butler and @joaander, any preference?

joaander:
You are welcome to apply the other fixes in this pull request. They are closely related and therefore fine to combine together.

joaander:
pre-commit.ci autofix

hoomd/md/bond.py
```diff
@@ -262,9 +266,9 @@ def __init__(self, width):
     def _attach_hook(self):
         """Create the c++ mirror class."""
         if isinstance(self._simulation.device, hoomd.device.CPU):
-            cpp_cls = _md.BondTablePotential
+            cpp_cls = self._ext_module.BondTablePotential
         else:
-            cpp_cls = _md.BondTablePotentialGPU
+            cpp_cls = self._ext_module.BondTablePotentialGPU
 
         self._cpp_obj = cpp_cls(self._simulation.state._cpp_sys_def, self.width)
```
b-butler, line 273: This seems to be an unnecessary function (not relevant to this PR though).

b-butler (approved):
Thanks @astatt for the PR.

PR #1609, opened 2023-08-17
Fix GSD file exists error message.

joaander:
## Description

Provide "RuntimeError: GSD: File exists" error message instead of "Bad file descriptor." when using the 'xb' file mode and the file exists.

## Motivation and context

Users expect to see a "File exists" error when the file exists. See https://github.com/glotzerlab/gsd/issues/270.

## How has this been tested?

Added a unit tests that verifies the correct message. The test fails before the fix and passes after.

## Change log

```
Fixed:

* ``hoomd.write.GSD`` reports "File exists" in the exception description when using the ``'xb'`` mode and the file exists.
```

## Checklist:


AlainKadar (approved):

tommy-waltmann (approved):
LGTM

PR #1619, opened 2023-09-15
Changes to Bussi thermostat

### LINKED ISSUE #1618: Add option of tau in Bussi thermostat
### Description

In the original formulation of Bussi thermostat, a time constant tau is included, but in current hoomd version is not included.

### Proposed solution

Add an option for input tau

### Additional context

_No response_

shihkual:
## Description

Correct the behavior of Bussi thermostat, which currently has two problems.
This PR resolves #1618

## Motivation and context

1. The time constant tau only default to 0. Add an option for users to set finite tau larger than 0.
2. Potential problem that the current implementation is problematic, which relates to an uncorrected typo in the original velocity rescaling thermostat paper.
3. Need to update current documentation based on the changes.

## How has this been tested?

1. Need to check the kinetic energy distribution and the kinetic energy autocorrelation sampled from the corrected Bussi thermostat are correct and reasonable.

## Change log

```
Pass `tau` into BussiThermostat class. Defaults to 0.
Calculate the rescaling factor for both zero and positive non-zero tau.
Calculate the rescaling factor so it is always 1 when the timestep is 0.
Modify the hoomd/md/pytest/test_methods.py according to the changes above.
Update the documentation of the Python code according to the changes.
```

## Checklist:

### DIFF
--- a/hoomd/md/Thermostat.h
+++ b/hoomd/md/Thermostat.h
@@ -425,7 +425,7 @@ class BussiThermostat : public Thermostat
         else if (nrdof_left == 1)
             {
             Scalar r_random_temp = normal_rotation(rng);
-            r_random_left = rr_temp * rr_temp;
+            r_random_left = r_random_temp * r_random_temp;
             }
         else if (nrdof_left_even)
             {
--- a/hoomd/md/Thermostat.h
+++ b/hoomd/md/Thermostat.h
@@ -358,9 +358,6 @@ class BussiThermostat : public Thermostat
         {
         m_thermo->compute(timestep);
 
-        Scalar time_decay_factor;
-        Scalar t_random_left;
-        Scalar r_random_left;
         // Integration methods will always operate on subsets with integer degrees of freedom.
         // Partial degrees of freedom are removed by getTranslationalDOF() when the subset of
         // particles is not the same subet selected for integration.
@@ -382,6 +379,7 @@ class BussiThermostat : public Thermostat
 
         const auto set_T = m_T->operator()(timestep);
 
+        Scalar time_decay_factor;
         if (m_tau != 0.0)
             {
             time_decay_factor = exp(-deltaT / m_tau);
@@ -396,51 +394,8 @@ class BussiThermostat : public Thermostat
         Scalar t_random_one = normal_translation(rng);
         Scalar r_random_one = normal_rotation(rng);
 
-        const int ntdof_left = ntdof - 1;
-        const int nrdof_left = nrdof - 1;
-        const bool ntdof_left_even = ntdof_left % 2 == 0;
-        const bool nrdof_left_even = nrdof_left % 2 == 0;
-        if (ntdof_left == 0)
-            {
-            t_random_left = 0.;
-            }
-        else if (ntdof_left == 1)
-            {
-            Scalar t_random_temp = normal_translation(rng);
-            t_random_left = t_random_temp * t_random_temp;
-            }
-        else if (ntdof_left_even)
-            {
-            GammaDistribution<double> gamma_translation(ntdof_left / 2.0, 1.0);
-            t_random_left = 2.0 * gamma_translation(rng);
-            }
-        else
-            {
-            GammaDistribution<double> gamma_translation((ntdof_left - 1) / 2.0, 1.0);
-            Scalar t_random_temp = normal_translation(rng);
-            t_random_left = 2.0 * gamma_translation(rng) + t_random_temp * t_random_temp;
-            }
-
-        if (nrdof_left == 0)
-            {
-            r_random_left = 0.;
-            }
-        else if (nrdof_left == 1)
-            {
-            Scalar r_random_temp = normal_rotation(rng);
-            r_random_left = r_random_temp * r_random_temp;
-            }
-        else if (nrdof_left_even)
-            {
-            GammaDistribution<double> gamma_rotation(nrdof_left / 2.0, 1.0);
-            r_random_left = 2.0 * gamma_rotation(rng);
-            }
-        else
-            {
-            GammaDistribution<double> gamma_rotation((nrdof_left - 1) / 2.0, 1.0);
-            Scalar r_random_temp = normal_rotation(rng);
-            r_random_left = 2.0 * gamma_rotation(rng) + r_random_temp * r_random_temp;
-            }
+        Scalar t_random_left = sample_sumnormal(ntdof - 1, rng);
+        Scalar r_random_left = sample_sumnormal(nrdof - 1, rng);
 
         Scalar t_rescale
             = sqrt(time_decay_factor
@@ -462,6 +417,37 @@ class BussiThermostat : public Thermostat
 
     protected:
     Scalar m_tau;
+
+    Scalar sample_sumnormal(int dof_left, RandomGenerator rng)
+        {
+        Scalar sum_normal_square;
+        const bool dof_left_even = ((dof_left & 1) == 0);
+
+        if (dof_left == 0)
+            {
+            sum_normal_square = 0.;
+            }
+        else if (dof_left == 1)
+            {
+            NormalDistribution<double> normal(1.0);
+            Scalar normal_random = normal(rng);
+            sum_normal_square = normal_random * normal_random;
+            }
+        else if (dof_left_even)
+            {
+            GammaDistribution<double> gamma(dof_left / 2.0, 1.0);
+            sum_normal_square = 2.0 * gamma(rng);
+            }
+        else
+            {
+            NormalDistribution<double> normal(1.0);
+            GammaDistribution<double> gamma((dof_left - 1) / 2.0, 1.0);
+            Scalar normal_random = normal(rng);
+            sum_normal_square = 2.0 * gamma(rng) + normal_random * normal_random;
+            }
+
+        return sum_normal_square;
+        }
     };
 
 /// Implement the Berendsen velocity recalcing thermostat.
--- a/hoomd/md/Thermostat.h
+++ b/hoomd/md/Thermostat.h
@@ -432,16 +432,15 @@ class BussiThermostat : public Thermostat
         Scalar r_normal_one = normal(rng);
 
         GammaDistribution<double> gamma((degrees_of_freedom - 1.0) / Scalar(2.0), Scalar(1.0));
-        double r_gamma_n_minus_one = 0.0;
+        double r_gamma = 0.0;
         if (degrees_of_freedom > 1.0)
             {
-            r_gamma_n_minus_one = 2.0 * gamma(rng);
+            r_gamma = 2.0 * gamma(rng);
             }
 
-        double term1 = set_T / 2.0 / K * (1.0 - time_decay_factor)
-                       * (r_gamma_n_minus_one + r_normal_one * r_normal_one);
-        double term2 = 2.0 * r_normal_one
-                       * sqrt(set_T / 2.0 / K * (1.0 - time_decay_factor) * time_decay_factor);
+        double v = set_T / 2.0 / K;
+        double term1 = v * (1.0 - time_decay_factor) * (r_gamma + r_normal_one * r_normal_one);
+        double term2 = 2.0 * r_normal_one * sqrt(v * (1.0 - time_decay_factor) * time_decay_factor);
 
         return Scalar(sqrt(time_decay_factor + term1 + term2));
         }
--- a/hoomd/md/Thermostat.h
+++ b/hoomd/md/Thermostat.h
@@ -365,8 +365,7 @@ class BussiThermostat : public Thermostat
         const int nrdof = static_cast<int>(m_thermo->getRotationalDOF());
         const auto ket_int = m_thermo->getTranslationalKineticEnergy();
         const auto ker_int = m_thermo->getRotationalKineticEnergy();
-        if ((ntdof != 0 && m_thermo->getTranslationalKineticEnergy() == 0)
-            || (nrdof != 0 && m_thermo->getRotationalKineticEnergy() == 0))
+        if ((ntdof != 0 && ket_int == 0) || (nrdof != 0 && ker_int == 0))
             {
             throw std::runtime_error("Bussi thermostat requires non-zero initial temperatures");
             }
--- a/hoomd/md/Thermostat.h
+++ b/hoomd/md/Thermostat.h
@@ -398,8 +398,8 @@ class BussiThermostat : public Thermostat
         Scalar t_random_first = normal_translation(rng);
         Scalar r_random_first = normal_rotation(rng);
 
-        Scalar t_random_left = sample_sumnormalsq(ntdof - 1, rng);
-        Scalar r_random_left = sample_sumnormalsq(nrdof - 1, rng);
+        Scalar t_random_left = sample_sumnormal_squared(ntdof - 1, rng);
+        Scalar r_random_left = sample_sumnormal_squared(nrdof - 1, rng);
 
         Scalar t_rescale
             = sqrt(time_decay_factor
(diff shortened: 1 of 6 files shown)


shihkual:
pre-commit.ci autofix

hoomd/md/Thermostat.h
```diff
@@ -337,38 +337,128 @@ class MTTKThermostat : public Thermostat
...
 
-        const auto ntdof = m_thermo->getTranslationalDOF();
-        const auto nrdof = m_thermo->getRotationalDOF();
+        Scalar e_factor;
+        Scalar rt_left;
+        Scalar rr_left;
+        const int ntdof = static_cast<int>(m_thermo->getTranslationalDOF());
+        const int nrdof = static_cast<int>(m_thermo->getRotationalDOF());
+        const auto ket_int = m_thermo->getTranslationalKineticEnergy();
+        const auto ker_int = m_thermo->getRotationalKineticEnergy();
```
b-butler: These shouldn't be ints right? The int in the name is confusing especially with the `auto` keyword.

hoomd/md/Thermostat.h
```diff
@@ -337,38 +337,128 @@ class MTTKThermostat : public Thermostat
...
+            e_factor = 1.0;
+            }
+        else if (m_tau / deltaT > 0.1)
+            {
+            e_factor = exp(-deltaT / m_tau);
+            }
+        else // the limit case when tau is near 0
+            {
+            e_factor = 0.0;
+            }
```
b-butler: Are the first two branches necessary? and for the second couldn't we do a check to see if `tau` is zero or very close to zero?

hoomd/md/methods/thermostats.py
```diff
@@ -287,13 +287,17 @@ class Bussi(Thermostat):
                                               t_ramp=1_000_000)
     """
 
-    def __init__(self, kT):
+    def __init__(self, kT, tau=0.):
```
b-butler:
```suggestion
    def __init__(self, kT, tau=0.0):
```

hoomd/md/Thermostat.h
```diff
@@ -337,38 +337,128 @@ class MTTKThermostat : public Thermostat
...
-        const auto nrdof = m_thermo->getRotationalDOF();
+        Scalar e_factor;
+        Scalar rt_left;
+        Scalar rr_left;
+        const int ntdof = static_cast<int>(m_thermo->getTranslationalDOF());
+        const int nrdof = static_cast<int>(m_thermo->getRotationalDOF());
+        const auto ket_int = m_thermo->getTranslationalKineticEnergy();
+        const auto ker_int = m_thermo->getRotationalKineticEnergy();
         if ((ntdof != 0 && m_thermo->getTranslationalKineticEnergy() == 0)
             || (nrdof != 0 && m_thermo->getRotationalKineticEnergy() == 0))
```
b-butler: Why are you not using `ker_int` and `ket_int`?

hoomd/md/Thermostat.h
```diff
@@ -337,38 +337,128 @@ class MTTKThermostat : public Thermostat
...
 
     std::array<Scalar, 2> getRescalingFactorsOne(uint64_t timestep, Scalar deltaT) override
         {
         m_thermo->compute(timestep);
 
-        const auto ntdof = m_thermo->getTranslationalDOF();
-        const auto nrdof = m_thermo->getRotationalDOF();
+        Scalar e_factor;
+        Scalar rt_left;
+        Scalar rr_left;
```
b-butler: Define these closer to where they are needed.

joaander: These variables need more descriptive names.

hoomd/md/Thermostat.h
joaander: Simplify and remove magic numbers:
```suggestion
        if (m_tau != 0.0)
            {
            e_factor = exp(-deltaT / m_tau);
            }
        else
            {
            e_factor = 0.0;
            }
```

hoomd/md/Thermostat.h
```diff
@@ -337,38 +337,128 @@ class MTTKThermostat : public Thermostat
...
+        Scalar t_rescale = sqrt(
+            e_factor
+            + set_T / Scalar(2.0) / ket_int * (Scalar(1.0) - e_factor) * (rt_left + rt * rt)
+            + Scalar(2.0) * rt
+                  * sqrt(set_T / Scalar(2.0) / ket_int * (Scalar(1.0) - e_factor) * e_factor));
+        Scalar r_rescale = sqrt(
+            e_factor
+            + set_T / Scalar(2.0) / ker_int * (Scalar(1.0) - e_factor) * (rr_left + rr * rr)
+            + Scalar(2.0) * rr
+                  * sqrt(set_T / Scalar(2.0) / ker_int * (Scalar(1.0) - e_factor) * e_factor));
```
joaander: Refactor into one function that you call twice to get translation and rotation.

hoomd/md/Thermostat.h
```diff
@@ -337,38 +337,128 @@ class MTTKThermostat : public Thermostat
...
+            }
+
+        NormalDistribution<double> normal_translation(1.0);
+        NormalDistribution<double> normal_rotation(1.0);
+        Scalar rt = normal_translation(rng);
+        Scalar rr = normal_rotation(rng);
+
+        const int ntdof_left = ntdof - 1;
+        const int nrdof_left = nrdof - 1;
+        const bool ntdof_left_even = ntdof_left % 2 == 0;
```
joaander:
```suggestion
        const bool ntdof_left_even = (ntdof_left & 1 == 0);
```
Improve readability and use bit-wise operations when possible.

hoomd/md/Thermostat.h
```diff
@@ -337,38 +337,128 @@ class MTTKThermostat : public Thermostat
...
+            {
+            GammaDistribution<double> gamma_translation(ntdof_left / 2.0, 1.0);
+            rt_left = 2.0 * gamma_translation(rng);
+            }
+        else
+            {
+            GammaDistribution<double> gamma_translation((ntdof_left - 1) / 2.0, 1.0);
+            Scalar rt_temp = normal_translation(rng);
+            rt_left = 2.0 * gamma_translation(rng) + rt_temp * rt_temp;
+            }
```
joaander: It is not clear what this sequence of conditions achieves. You mentioned it came from a correction to the original article. Is that correction published? We need to cite an explanation for this logic for future reference.

hoomd/md/methods/thermostats.py
joaander, line 1: Update the documentation to correctly and concisely describe the current behavior of the code.

hoomd/md/Thermostat.h
```diff
@@ -337,38 +337,128 @@ class MTTKThermostat : public Thermostat
...
         {
         m_thermo->compute(timestep);
 
-        const auto ntdof = m_thermo->getTranslationalDOF();
-        const auto nrdof = m_thermo->getRotationalDOF();
+        Scalar e_factor;
+        Scalar rt_left;
+        Scalar rr_left;
+        const int ntdof = static_cast<int>(m_thermo->getTranslationalDOF());
+        const int nrdof = static_cast<int>(m_thermo->getRotationalDOF());
```
joaander:
```suggestion
        // Integration methods will always operate on subsets with integer degrees of freedom.
        // Partial degrees of freedom are removed by getTranslationalDOF() when the subset of
        // particles is not the same subet selected for integration.
        const int ntdof = static_cast<int>(m_thermo->getTranslationalDOF());
        const int nrdof = static_cast<int>(m_thermo->getRotationalDOF());
```
It isn't immediately obvious that the number of degrees of freedom will always be an integer here. Add a comment to clarify.
https://github.com/glotzerlab/hoomd-blue/blob/4184da26c976722d9852100c14941b62a018ca05/hoomd/md/IntegratorTwoStep.cc#L181-L186

joaander (changes requested):
Thanks! I have made suggestions to improve the readability of the code. Also, the documentation must be updated.

hoomd/md/Thermostat.h
shihkual: @joaander The correct implementation written by Bussi himself is here:
https://github.com/GiovanniBussi/StochasticVelocityRescaling/blob/master/resamplekin.cpp

Specifically, the `resamplekin_sumnoises` function.

hoomd/md/Thermostat.h
```diff
@@ -337,38 +337,131 @@ class MTTKThermostat : public Thermostat
...
-        return {std::sqrt(gamma_translation(rng) / m_thermo->getTranslationalKineticEnergy()),
-                std::sqrt(gamma_rotation(rng) / m_thermo->getRotationalKineticEnergy())};
+        if (m_tau != 0.0)
+            {
+            time_decay_factor = exp(-deltaT / m_tau);
+            }
+        else
+            {
+            time_decay_factor = 0.0;
+            }
```
shihkual: @joaander It seems to me that it will be problematic when both deltaT and m_tau are zero when calculating exp(-deltaT / m_tau)

joaander: You could add:
```
if (deltaT == 0.0)
    {
    return {1.0, 1.0};
    }
```
as a clear special case at the top of the function.

hoomd/md/Thermostat.h
MartinGirard: There is a major implementation difference between the gamma distribution of Bussi, and hoomd's that we should consider here. The gamma parameter needed is a half or full integer vale. 

Bussi's implementation of the gamma distribution requires an integer ndof. So when he needs a half integer value, he computes one less DOF, then adds that from a Gaussian distribution (see relation between sums squares of gaussian and gamma). 

The implementation we have in RandomNumbers.h is based on a FP number. It therefore does not matter for us whether this is integer or not. We shouldn't do this split between odd/even DOFs as we dont have any gain from it.

joaander: Thanks for confirming this, Martin. I suspected as such. Alex, please simplify the code accordingly.

shihkual: @joaander @MartinGirard Thanks for checking that for me! I will adjust the code later.

b-butler (changes requested):
I am leaving some code clean up suggestions.

hoomd/md/Thermostat.h
shihkual: The `ket_int` and `ker_int` here refer to instantaneous translational and rotational kinetic energy here, but I see why the naming is confusing. Maybe `ket_i` or `ket_ins` will be better?

hoomd/md/Thermostat.h
shihkual: Thanks for spotting this! I think I missed this part.

hoomd/md/pytest/test_methods.py
```diff
@@ -163,7 +165,9 @@ def test_attributes_attached(self, simulation_factory,
         check_instance_attrs(thermostat, constructor_args)
 
         change_attrs = thermostat_definition.generate_all_attr_change()
-        if isinstance(thermostat, hoomd.md.methods.thermostats.Berendsen):
+        if isinstance(thermostat,
+                      hoomd.md.methods.thermostats.Berendsen) or isinstance(
+                          thermostat, hoomd.md.methods.thermostats.Bussi):
```
joaander: Is there a fundamental reason we do not wish users to change the value of tau for an existing thermostat?
```
bussi.tau = 5
```
If not, remove this isinstance check, add `.def_property("tau", &BussiThermostat::getTau, &BussiThermostat::setTau)` and the corresponding `getTau` and `setTau` methods.

hoomd/md/Thermostat.h
joaander: I see, the intention is to perhaps make the long math equation easier to read with shorter variable names. I will push a commit showing another way to handle such long equations in code.

hoomd/md/Thermostat.h
```diff
@@ -337,37 +337,113 @@ class MTTKThermostat : public Thermostat
...
+                         * sqrt(set_T / Scalar(2.0) / ker_int * (Scalar(1.0) - time_decay_factor)
+                                * time_decay_factor));
+
+        return {t_rescale, r_rescale};
+        }
+
+    protected:
+    Scalar m_tau;
+
+    Scalar sample_sumnormalsq(int dof_left, RandomGenerator rng)
```
MartinGirard: Shouldn't the `RandomGenerator` be a reference here? Otherwise we will create correlations between the different scaling factors. 

The two branches at L434 & L440 (case ndof==1) can be merged. The distribution Gamma(ndof / 2, 2.0) is a chi-squared distribution with ndof degrees of freedom, which encompasses both branches. Is the cost of generating from a Gaussian distribution compared to a Gamma lower than the branching cost here?

hoomd/md/Thermostat.h
MartinGirard: Does it even matter if ndof is an integer or not? I see an edge case if 0 < ndof < 1 and m_tau != 0, but otherwise I think it works out fine; mathematically speaking I mean, I'm not sure why and how someone would end up with non-integer dofs.

MartinGirard (changes requested):

hoomd/md/Thermostat.h
joaander: Good catch, the `RandomGenerator` must be passed by reference.

I would like to improve code readability by using one branch here. However, the Gamma distribution code currently only implements code for alpha > 1. The special case at line 434 can be removed after we add the alpha < 1 case to `Gamma` - which is a straightforward transformation: https://www.hongliangjie.com/2012/12/19/how-to-generate-gamma-random-variables/

Note that `ndof_left==1` indicates the user is running something like 2 particles in 2D - or a system containing only 2 rotational degrees of freedom.

joaander (commented):
Thanks for working on this. I have one question below we should discuss.

I will push commits fixing the docs and cleaning up the code further, then we can make one final review pass.

hoomd/md/Thermostat.h
shihkual: Thanks. That will be very helpful!

hoomd/md/Thermostat.h
shihkual: Yes, I agree. `ndof_left==1` is not a usual case we will see in simulations but has to be taken care of, so merging the two branches should improve the overall performance.

hoomd/md/pytest/test_methods.py
shihkual: Good point. I could not come up with a reason why it would be problematic for users to change the tau, so yes, I will add the suggested methods.

joaander:
I have implemented the code style changes I suggested. @shihkual, please run double check that my changes are correct and run latest commit in your test notebook. Reviewers, this is ready for a final review.

b-butler (approved):
The code looks good. I did not check the math for correctness.

shihkual (commented):
@joaander Thanks! This is my testing results of the new Bussi thermostat. The testing system contains 125 ALJ hard cubes. I simulated 100,000 steps, and the timestep is 0.0005. The resulting autocorrelation functions match the expected decaying behavior, and the sampled kinetic energy distributions are nearly identical to the target gamma distribution.
K_test

hoomd/md/methods/thermostats.py
```diff
@@ -248,16 +248,38 @@ class Bussi(Thermostat):
...
+        f_N(g) = \frac{1}{\Gamma(N)} g^{N-1} e^{-g}.
 
-    where :math:`N_f` is the number of degrees of freedom thermalized.
+    When `tau` is non-zero, the kinetic energies decay to equilibrium with the
+    given characteristic time constant and :math:`\alpha` is given by:
+
+    .. math::
+        \alpha = \sqrt{e^{\delta t / \tau}
+                 + (1 - e^{\delta t / \tau}) \frac{(2 g_{N-1} + n^2) kT}{2 K}
+                 + 2 n \sqrt{e^{\delta t / \tau} (1-e{\delta t / \tau})
```
shihkual: I believe there is a typo
```suggestion
                 + 2 n \sqrt{e^{\delta t / \tau} (1-e^{\delta t / \tau})
```

joaander (approved):
Thanks!

MartinGirard (approved):

PR #1621, opened 2023-09-19
Fix/hdf5 scalar detection

### LINKED ISSUE #1620: HDF5 logger does not check data type correctly
### Description

Currently `hoomd.write.HDF5Log` checks the data type of the values with `dtype = "f8" if isinstance(value, float) else "i8"`, which does not account for data types such as `numpy.float64`. We should either include these types because users might define custom quantities which are calculated from NumPy etc, or provide an option for the user to specify the data type? In the following example, originally the acceptance ratio is an integer, but after several trial moves are made it will be float. We will probably also need to account for this when checking the data types, or document this behavior.

### Script

```python
@log(category="scalar", requires_run=True)
def acceptance_ratio(self):
    total_moves = sum(self._counter)
    accept_moves = self._counter[0]
    if total_moves == 0:
        return 0
    else:
        return accept_moves / total_moves

# Create a logger to log the acceptance ratio
def create_writer(self, log_fn, mode, dump_period):
    logger = hoomd.logging.Logger(categories=["scalar"])
    logger.add(self, quantities=[
                    "acceptance_ratio",
                ])
     return hoomd.write.HDF5Log(
            trigger=hoomd.trigger.Periodic(dump_period, 1),
            filename=log_fn,
            logger=logger,
            mode=mode
        )
```

### Input files

_No response_

### Output

```shell
[0, 0, ...., 1, 1, ..., 1]
```

### Expected output

_No response_

### Platform

Linux

### Installation method

glotzerlab-software container

### HOOMD-blue version

4.1.0

### Python version

3.11.3

b-butler:
## Description
Improves and corrects `hoomd.write.HDF5Log`'s detection of scalar value types and converts to the appropriate NumPy dtype.

## Motivation and context

Resolves #1620

## How has this been tested?
An appropriate test was added to validate the typing behavior.

## Change log

```
Fixed: ``hoomd.write.HDF5Log's scalar loggable typing.``
```

## Checklist:

### DIFF
--- a/hoomd/write/hdf5.py
+++ b/hoomd/write/hdf5.py
@@ -209,7 +209,7 @@ class _HDF5LogInternal(custom._InternalAction):
             chunk_size = None
             if category == "scalar":
                 data_shape = (1,)
-                if isinstance(value, (np.integer, np.floating, np.bool_)):
+                if isinstance(value, (np.number, np.bool_)):
                     dtype = value.dtype
                 elif isinstance(value, int):
                     dtype = np.dtype(bool) if isinstance(value, bool) else "i8"
--- a/hoomd/write/hdf5.py
+++ b/hoomd/write/hdf5.py
@@ -209,7 +209,12 @@ class _HDF5LogInternal(custom._InternalAction):
             chunk_size = None
             if category == "scalar":
                 data_shape = (1,)
-                dtype = "f8" if isinstance(value, float) else "i8"
+                if isinstance(value, (np.integer, np.floating, np.bool_)):
+                    dtype = value.dtype
+                elif isinstance(value, int):
+                    dtype = np.dtype(bool) if isinstance(value, bool) else "i8"
+                else:
+                    dtype = "f8"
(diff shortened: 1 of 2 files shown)


rayasare:
will have to take a look at this after Friday

hoomd/write/hdf5.py
```diff
@@ -209,7 +209,12 @@ def _initialize_datasets(self, log_dict):
             chunk_size = None
             if category == "scalar":
                 data_shape = (1,)
-                dtype = "f8" if isinstance(value, float) else "i8"
+                if isinstance(value, (np.integer, np.floating, np.bool_)):
```
joaander:
```suggestion
                if isinstance(value, (np.number, np.bool_)):
```
This covers integer, floating, and complex types which h5py supports: https://docs.h5py.org/en/stable/faq.html#what-datatypes-are-supported

tcmoore3 (approved):

rayasare (approved):

PR #1628, opened 2023-09-26
Add Sybil examples to most classes in the root hoomd namespace.

joaander:
## Description

Add Sybil examples and revise documentation for all classes in the `hoomd` namespace (excluding `hoomd.md` and `hoomd.hpmc` which will be added in a future pull request).

## Motivation and context

Many users prefer to copy and paste these minimal examples. Sybil tests the examples to ensure they operate.

## How has this been tested?

I rendered the documentation locally. Pull request reviewers can click the "Details" link next to readthedocs.io in the checks below to see the new documentation.

## Change log

```
Changed:

* Revise class documentation.
* Add more code snippets to the class documentation.
```

## Checklist:

### DIFF
--- a/hoomd/data/typeparam.py
+++ b/hoomd/data/typeparam.py
@@ -27,12 +27,10 @@ class TypeParameter(MutableMapping):
     and multiple parameters on one line.
 
     Important:
-
         Parameters for all types (or unordered pairs of types) in the simulation
         state must be defined prior to calling `Simulation.run()`.
 
     Note:
-
         `TypeParameter` removes types (or type pairs) not present in the
         simulation state *after* its operation is added to the simulation and
         the simulation has been run for 0 or more steps.
@@ -66,7 +64,7 @@ class TypeParameter(MutableMapping):
     def __setitem__(self, key, value):
         """Set parameters for a given type (or type pair).
 
-        .. rubric:: Examples
+        .. rubric:: Examples:
 
         Index types by name:
 
@@ -105,7 +103,6 @@ class TypeParameter(MutableMapping):
             lj.params[(['B', 'C'], ['B', 'C'])] = dict(epsilon=1, sigma=1)
 
         Note:
-
             Setting the value for *(a,b)* automatically sets the symmetric
             *(b,a)* parameter to the same value.
         """
@@ -192,7 +189,7 @@ class TypeParameter(MutableMapping):
         the simulation state that is not explicitly set by `__setitem__`
         or `setdefault`.
 
-        .. rubric:: Examples
+        .. rubric:: Examples:
 
         Set a default value:
 
--- a/hoomd/data/typeparam.py
+++ b/hoomd/data/typeparam.py
@@ -11,87 +11,40 @@ from collections.abc import MutableMapping
 # documentation. For documentation on the mechanics and internal structure go to
 # the documentation for hoomd.data.parameterdicts.TypeParameterDict.
 class TypeParameter(MutableMapping):
-    """Implement a type based mutable mapping.
+    """Store parameters by type or type pair.
 
-    *Implements the* `collections.abc.MutableMapping` *interface
-    (* ``__delitem__`` *is disallowed).*
+    *Implements the* `collections.abc.MutableMapping` *interface* (excluding
+    ``__delitem__``).
 
-    `TypeParameter` instances extend the base Python mapping interface with
-    smart defaults, value/key validation/processing, and advanced indexing. The
-    class's intended purpose is to store data per type or per unique
-    combinations of type (such as type pairs for `hoomd.md.pair` potentials) of
-    a prescribed length.
+    Many operations in HOOMD-blue utilize parameters that depend on the type or
+    pairs of types. For example, the Langevin drag coefficient
+    (`hoomd.md.methods.Langevin.gamma`) are set by **particle type** and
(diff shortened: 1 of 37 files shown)


sphinx-doc/module-hoomd-write.rst
```diff
@@ -22,11 +22,27 @@ hoomd.write
...
+
+    .. autoclass:: CustomWriter
+        :show-inheritance:
+        :members:
+
+    .. autoclass:: DCD(trigger, filename, filter=hoomd.filter.All(), overwrite=False, unwrap_full=False, unwrap_rigid=False, angle_z=False)
+        :show-inheritance:
+        :members:
+
+    .. autoclass:: GSD(trigger, filename, filter=hoomd.filter.All(), mode='ab', truncate=False, dynamic=None, logger=None)
```
b-butler, line 38: Are these explicit signatures needed. I thought only the `CustomAction` writers needed this?

hoomd/data/typeparam.py
```diff
@@ -11,87 +18,39 @@
...
+    Lennard-Jones pair potential parameters (`hoomd.md.pair.LJ.params`) are set
+    by **pairs** of particle types.
 
-    For getting and setting values, multiple index formats are supported. The
-    base case is either a string representing the appropriate type or a tuple of
-    such strings if multiple types are required per key. This is the exact same
-    indexing behavior expect from a Python `dict`, and all functions (barring
-    those that delete keys) should function as expected for a Python
-    `collections.defaultdict`.
+    `TypeParameter` holds the values of these type (or type pair) depdendent
```
b-butler:
```suggestion
    `TypeParameter` holds the values of these type (or type pair) dependent
```

hoomd/data/typeparam.py
```diff
@@ -110,33 +69,93 @@ def __getattr__(self, attr):
...
+
+            lj.params[('A', ['B', 'C'])] = dict(epsilon=0, sigma=0)
+
+        Set parameters for multiple pairs (e.g. ('B', 'B'), ('B', 'C'), ('C',
+        'B'), and ('C', 'C')):
+
+        .. skip: next if(not hoomd.version.md_built)
+        .. code-block:: python
+
+            lj.params[(['B', 'C'], ['B', 'C'])] = dict(epsilon=1, sigma=1)
```
b-butler, line 111: I get the reason to include `('C', 'B')`, but that won't be set actually. Just wanted to mention this.

hoomd/data/typeparam.py
```diff
@@ -110,33 +69,93 @@ def __getattr__(self, attr):
...
+
+        Note:
+            Setting the value for *(a,b)* automatically sets the symmetric
+            *(b,a)* parameter to the same value.
+        """
         self.param_dict[key] = value
 
+    def __getitem__(self, key):
```
b-butler, line 119: Maybe we should mention that `__getitem__` supports the same indexing logic shown above?

b-butler (commented):
This was/is a lot of work! Overall looks great and should be helpful to users.

sphinx-doc/module-hoomd-write.rst
joaander: I added them to make the default for `filter` clear. Prior to this change, it renders as `filter=hoomd._hoomd.ParticleFilter`.

hoomd/data/typeparam.py
joaander: Yes, that note is preserved on the next line. I listed both pairs here for completeness.

hoomd/data/typeparam.py
joaander: Isn't this a weird behavior? Do other mapping types in the Python ecosystem follow this behavior?
```
>>> lj.params[('A', 'A')] = dict(epsilon=1, sigma=2)
>>> lj.params[('A', 'B')] = dict(epsilon=2, sigma=3)
>>> lj.params[('B', 'B')] = dict(epsilon=3, sigma=4)
>>> lj.params[('A', 'A')]
_HOOMDDict{'epsilon': 1.0, 'sigma': 2.0}
>>> lj.params[('A', ['A', 'B'])]
{('A', 'A'): _HOOMDDict{'epsilon': 1.0, 'sigma': 2.0}, ('A', 'B'): _HOOMDDict{'epsilon': 2.0, 'sigma': 3.0}}
```
I see that with a single pair, it returns a single dict. With smart indexing, it returns a dict of dicts. What use-case does this enable for users and how would I demonstrate it in an example? 

This seems very awkard:
```
>>> lj.params[('A', ['A', 'B'])][('A', 'A')]
_HOOMDDict{'epsilon': 1.0, 'sigma': 2.0}
```

b-butler: Most mapping types I am aware of do not support setting or getting in this abbreviated fashion. We only support it because of the setting which I thought was useful and somewhat intuitive when shown. I figured it would be confusing to only support the smart indexing for setting, so instead it is supported on any operation that takes in keys.

joaander: I see. I find it confusing that setting with multiple keys broadcasts a single dictionary to those values where getting with multiple keys returns a dictionary with multiple results. I cannot think of any use-case to demonstrate for users, so I added documentation that states multiple keys are allowed in `__getitem__` and show an equivalent notation to explain the output:
```
gammas = langevin.gamma[['A', 'B']]
gammas = {key: langevin.gamma[key] for key in ['A', 'B']}
```

b-butler (approved):

PR #1644, opened 2023-10-27
added alcheimcal param for opp

Lina492375qW1188:
## Description

Adding alchemical parameters and corresponding derivatives to OPP.

## Motivation and context

The OPP is important for testing new alchemostat.

Resolves #???

## How has this been tested?

It is tested on placing two particles in the simulation box.

## Change log

```
Added: alchemical parameters for OPP
```

## Checklist:


j-proc:
Just wanted to bring up ahead of time that with our implementation of alpha being multiplied (to keep it dimensionless) the parameters entered cannot be zero. A zero parameter would have been nonsensical for LJG, but for OPP you have $\phi$. I don't think it makes sense to change the implementation, but adding a note for the user would be appropriate here I think. Perhaps specific to OPP or general to alchemical MD. Others will likely have stronger thoughts on where is best.

Lina492375qW1188:
pre-commit.ci autofix

Lina492375qW1188:
pre-commit.ci autofix

hoomd/md/alchemy/pair.py
```diff
@@ -9,6 +9,7 @@
 from hoomd.operation import _HOOMDBaseObject
 from hoomd.data.parameterdicts import ParameterDict
 from hoomd.md.pair import LJGauss as BaseLJGauss
+from hoomd.md.pair import OPP as BASEOPP
```
tcmoore3:
```suggestion
from hoomd.md.pair import OPP as BaseOPP
```

hoomd/md/alchemy/pair.py
```diff
@@ -364,3 +365,116 @@ def __init__(self,
                  mode='none'):
         _AlchemicalPairForce.__init__(self)
         super().__init__(nlist, default_r_cut, default_r_on, mode)
+
+
+@_modify_pair_cls_to_alchemical
+class OPP(BASEOPP, _AlchemicalPairForce):
```
tcmoore3:
```suggestion
class OPP(BaseOPP, _AlchemicalPairForce):
```

tcmoore3 (commented):
Looks good, but needs tests. At a minimum, you should test that you can attach the item to the simulation, and you should also test that the alchemical parameters evolve when they should.

hoomd/md/alchemy/pair.py
joaander: Better yet:
```class OPP(hoomd.md.pair.OPP, ...)```
Explicit is better than implicit.

joaander: So using `BaseOPP` violates pep8-naming rule N811 because it thinks `OPP` is the name of a constant: https://peps.python.org/pep-0008/#constants

I think the simplest solution here is to be explicit with `class OPP(hoomd.md.pair.OPP, ...)` and then the imported alias is not necessary.

Lina492375qW1188: Yeah that's why I replaced the name by Opp. I'll make it explicit. Thanks!

PR #1655, opened 2023-11-16
Port MPCD integrator, streaming methods, and collision methods

mphoward:
## Description

This PR ports core MPCD integration capabilities to v4. Specifically, it reimplements the `Integrator` that extends `md.Integrator` to include a `StreamingMethod` and `CollisionMethod`. All of the existing implementations of these two classes have also been reimplemented.

~~I targeted this PR onto `trunk-minor` rather than a staging branch to try to make it easier to keep things synchronized. If we do this, I propose that we do not document the inclusion of these MPCD features in the public documentation or changelog yet. I am concerned that some aspects of the API may need to change as I add back in some of the other capabilities, and I don't want to worry about making breaking changes. I think it would be good to get all necessary features added back, do some private beta testing with it, and then complete & link in documentation.~~

Per below, this will go onto the integration branch `mpcd-v4`, so this PR still needs documentation to be linked in and built, and a changelog entry. Will take care of this after initial review is done.

## Motivation and context

This is the next big change needed to make MPCD available again in HOOMD.

## How has this been tested?

The previous unit tests have been reimplemented as best I could with `pytest`. It would be great to get some feedback on whether I've followed the level of coverage and patterns that you are aiming for before I work on additional classes & tests in subsequent PRs.

## Change log

~~I propose not to add a change at this point. I can write a set of release notes when everything is ready.~~
Will add this before merging.

## Checklist:


mphoward:
@joaander @b-butler if you have time, I would appreciate it if one of you could also have a look at whether I've used developer features of the new Python interface as intended before I repeat my mistakes in other classes. =) I think I figured it out in the end (& got a lot less code as a result!), but it was difficult to follow the documentation & code for it in a couple places.

joaander:
Thanks! I'll review this soon.

Unless I am missing something, the only additional maintenance burden would involve periodically merging `trunk-minor` into the integration branch. With an integration branch, you could enable and test the `mpcd` package import and sphinx documentation at every step along the way. 

I would not want to show documentation for an API in a 4.x release that is going to drastically change in 4.x+n. I am also reluctant to import `mpcd` in any release until it is documented.

mphoward:
> Unless I am missing something, the only additional maintenance burden would involve periodically merging `trunk-minor` into the integration branch.

Nope, that should be it. We had talked about potentially avoiding that to avoid long-lived branches parallel to the trunk ones, but I am fine doing it this way. I'll retarget this PR onto `mpcd-v4`, I had created that branch for this purpose!

joaander:
> Nope, that should be it. We had talked about potentially avoiding that to avoid long-lived branches parallel to the trunk ones, but I am fine doing it this way. I'll retarget this PR onto `mpcd-v4`, I had created that branch for this purpose!

Thanks. Hopefully the branch will not be that long lived :)

hoomd/md/integrate.py
```diff
@@ -278,38 +279,42 @@ class Integrator(_DynamicIntegrator):
...
         self.half_step_hook = half_step_hook
 
     def _attach_hook(self):
         # initialize the reflected c++ class
-        self._cpp_obj = _md.IntegratorTwoStep(
-            self._simulation.state._cpp_sys_def, self.dt)
+        # checking for None allows deriving classes to construct this first
+        if self._cpp_obj is None:
+            self._cpp_obj = _md.IntegratorTwoStep(
+                self._simulation.state._cpp_sys_def, self.dt)
```
mphoward: Just wanted to add a note because this was a spot I wanted someone to have a look at.

Without this change, I think the `_cpp_obj` created in `mpcd.Integrator` would get overwritten by this one?

My alternative was to leave this alone & call the grandparent method from `mpcd.Integrator`. This would work currently because this method doesn't do anything other than initialize `_cpp_obj` then call the parent method, but I was worried that I would need to duplicate anything that got added here in future.

b-butler: It depends on how you called it for instance,
```python
super()._attach_hook()
# mpcd logic
```
would overwrite the MD integrator.

However, I don't feel too strongly about this, but I probably would call the grandparent method.

hoomd/md/integrate.py
```diff
@@ -28,15 +28,16 @@ def __init__(self, forces, constraints, methods, rigid):
         constraints = [] if constraints is None else constraints
         methods = [] if methods is None else methods
         self._forces = syncedlist.SyncedList(
-            Force, syncedlist._PartialGetAttr('_cpp_obj'), iterable=forces)
+            Force, syncedlist._PartialGetAttr("_cpp_obj"), iterable=forces)
```
b-butler: Is there a reason for all this unrelated changes?

hoomd/mpcd/CollisionMethod.cc
```diff
@@ -176,12 +136,15 @@ void mpcd::detail::export_CollisionMethod(pybind11::module& m)
...
+                                   auto group = method->getEmbeddedGroup();
+                                   return (group) ? group->getFilter()
+                                                  : std::shared_ptr<hoomd::ParticleFilter>();
+                               })
         .def("setEmbeddedGroup", &mpcd::CollisionMethod::setEmbeddedGroup)
-        .def("setPeriod", &mpcd::CollisionMethod::setPeriod)
-        .def_property("instance",
-                      &mpcd::CollisionMethod::getInstance,
-                      &mpcd::CollisionMethod::setInstance);
+        .def_property_readonly("every", &mpcd::CollisionMethod::getPeriod);
```
b-butler: I would call this something more descriptive than `every` `period` seems fine.

hoomd/mpcd/CollisionMethod.cc
```diff
@@ -176,12 +136,15 @@ void mpcd::detail::export_CollisionMethod(pybind11::module& m)
         m,
         "CollisionMethod")
         .def(pybind11::init<std::shared_ptr<SystemDefinition>, uint64_t, uint64_t, int>())
-        .def("enableGridShifting", &mpcd::CollisionMethod::enableGridShifting)
+        .def_property_readonly("embed",
```
b-butler: Perhaps, `embeded_group` would be better?

hoomd/mpcd/StreamingMethod.cc
```diff
@@ -120,9 +120,8 @@ void mpcd::detail::export_StreamingMethod(pybind11::module& m)
         m,
         "StreamingMethod")
         .def(pybind11::init<std::shared_ptr<SystemDefinition>, unsigned int, unsigned int, int>())
-        .def("setPeriod", &mpcd::StreamingMethod::setPeriod)
-        .def("setField", &mpcd::StreamingMethod::setField)
-        .def("removeField", &mpcd::StreamingMethod::removeField);
+        .def_property_readonly("every", &mpcd::StreamingMethod::getPeriod)
```
b-butler: Similar comment to above.

hoomd/mpcd/Integrator.cc
```diff
@@ -42,20 +64,22 @@ mpcd::Integrator::~Integrator()
...
+    if (!m_gave_warning && m_methods.size() == 0)
         {
-        m_exec_conf->msg->warning()
-            << "mpcd.integrate: No integration methods are set." << std::endl;
+        if (!m_stream)
+            {
+            m_exec_conf->msg->warning()
+                << "mpcd.integrate: No integration methods are set." << std::endl;
+            }
+        // setting this to true will shut up subsequent warnings if there is only a streaming method
```
b-butler: Why not add `!m_stream` to the first conditional?

hoomd/mpcd/pytest/test_integrator.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) 2009-2023 The Regents of the University of Michigan.
```
b-butler, line 1: I would add tests for setting and getting stream and collision methods,  detaching, and pickling.

hoomd/mpcd/collide.py
```diff
@@ -1,367 +1,225 @@
...
-        """ Disable the collision method
 
-        Examples::
+class CollisionMethod(AutotunedObject):
+    """Base collision method.
 
-            method.disable()
+    Args:
+        every (int): Number of integration steps between collisions.
+        embed (hoomd.filter.filter_like): HOOMD particles to include in collision.
```
b-butler: I think period and `filter` make the most sense. Here, `filter` at least is the common term used when we have a feature like this.

hoomd/mpcd/collide.py
```diff
@@ -1,367 +1,225 @@
...
-        collide.srd(seed=77, period=50, angle=130., group=hoomd.group.all())
-        collide.srd(seed=1991, period=10, angle=90., kT=1.5)
+    particles exchange momentum to reach an equilibrium temperature.) A
+    thermostat can be applied in conjunction with the SRD method through the
+    `kT` parameter. SRD employs a `Maxwell-Boltzmann thermostat
+    <https://doi.org/10.1016/j.jcp.2009.09.024>`_ on the cell level, which
+    generates the (correct) isothermal ensemble. The temperature is defined
+    relative to the cell-average velocity, and so can be used to dissipate heat
+    in nonequilibrium simulations. Under this thermostat, the SRD algorithm
+    still conserves linear momentum, but kinetic energy is of course no longer
```
b-butler:
```suggestion
    still conserves linear momentum, but kinetic energy is no longer
```
Try to avoid language like obviously, clearly, of course, etc.

hoomd/mpcd/pytest/test_collide.py
```diff
@@ -0,0 +1,101 @@
...
+        assert ig.collision_method is cm
+        assert cm.every == 5
+
+        ig.collision_method = None
+        sim.run(0)
+        assert ig.collision_method is None
+
+        ig.collision_method = cm
+        sim.run(0)
+        assert ig.collision_method is cm
```
b-butler: This is testing the integrator more than the collision method outside line 50, and should be in the integrator tests.

hoomd/mpcd/pytest/test_collide.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def test_embed(self, small_snap, simulation_factory, cls, init_args):
+        sim = simulation_factory(small_snap)
+        cm = cls(**init_args, embed=hoomd.filter.All())
+        sim.operations.integrator = hoomd.mpcd.Integrator(dt=0.02,
+                                                          collision_method=cm)
+
+        sim.run(0)
+        assert isinstance(cm.embed, hoomd.filter.All)
```
b-butler: Generally if we are going to test a parameter like this, we would test before and after run is called.

hoomd/mpcd/pytest/test_collide.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def test_temperature(self, small_snap, simulation_factory, cls, init_args):
```
b-butler, line 82: This entire test has no asserts. There is something to a test just to check an exceptions isn't raised, but should we check that the values are set correctly?

hoomd/mpcd/pytest/test_collide.py
```diff
@@ -0,0 +1,101 @@
...
+        sim.operations.integrator = hoomd.mpcd.Integrator(dt=0.02,
+                                                          collision_method=cm)
+
+        sim.run(1)
+
+        if "kT" not in init_args:
+            init_args["kT"] = 1.0
+        sim.operations.integrator.collision_method = cls(
+            **init_args, embed=hoomd.filter.All())
+        sim.run(1)
```
b-butler: Once again this is testing the integrator more than the collision method.

hoomd/mpcd/pytest/test_stream.py
```diff
@@ -0,0 +1,439 @@
...
+    ],
+    ids=["Bulk", "ParallelPlates", "PlanarPore"],
+)
+def test_create(simulation_factory, snap, cls, init_args):
+    sim = simulation_factory(snap)
+    sm = cls(every=5, **init_args)
+    ig = hoomd.mpcd.Integrator(dt=0.02, streaming_method=sm)
+    sim.operations.integrator = ig
+
+    sim.run(0)
```
b-butler: We typically test before and after calling run.

hoomd/mpcd/pytest/test_stream.py
```diff
@@ -0,0 +1,439 @@
...
+        sim.run(1)
+        snap = sim.state.get_snapshot()
+        if snap.communicator.rank == 0:
+            np.testing.assert_array_almost_equal(
+                snap.mpcd.position, [[1.3, -4.85, 3.3], [-3.3, 4.95, -1.3]])
+
+        # change streaming method to use a different period, and change integrator step
+        # running again should not move the particles since we haven't hit next period
+        ig.dt = 0.05
+        ig.streaming_method = hoomd.mpcd.stream.Bulk(every=4)
```
b-butler: Couldn't we test changing `every` here.

hoomd/mpcd/pytest/test_stream.py
```diff
@@ -0,0 +1,439 @@
...
+        # take another step where one particle will now hit the wall
+        sim.run(1)
+        snap = sim.state.get_snapshot()
+        if snap.communicator.rank == 0:
+            np.testing.assert_array_almost_equal(
+                snap.mpcd.position, [[-4.95, 4.95, 3.95], [-0.2, -0.2, -4.0]])
+            np.testing.assert_array_almost_equal(
+                snap.mpcd.velocity, [[-1.0, 1.0, -1.0], [-1.0, -1.0, -1.0]])
+
+        # take another step, wrapping the second particle through the boundary
```
b-butler: It seems the second particle is being reflected not passing through?

hoomd/mpcd/pytest/test_stream.py
```diff
@@ -0,0 +1,439 @@
...
+        # take another step where one particle will now hit the wall
+        sim.run(1)
+        snap = sim.state.get_snapshot()
+        if snap.communicator.rank == 0:
+            np.testing.assert_array_almost_equal(
+                snap.mpcd.position, [[-4.85, 4.85, 3.95], [-0.2, -0.2, -4.0]])
+            np.testing.assert_array_almost_equal(
+                snap.mpcd.velocity, [[1.0, -1.0, -1.0], [-1.0, -1.0, -1.0]])
+
+        # take another step, wrapping the second particle through the boundary
```
b-butler: Same as previous question.

hoomd/mpcd/pytest/test_stream.py
```diff
@@ -0,0 +1,439 @@
...
+                snap.mpcd.position, [[-4.75, 4.75, 3.85], [-0.3, -0.3, -3.9]])
+            np.testing.assert_array_almost_equal(
+                snap.mpcd.velocity, [[1.0, -1.0, -1.0], [-1.0, -1.0, 1.0]])
+
+    def test_step_moving_wall(self, simulation_factory, snap):
+        """Test step with moving wall.
+
+        The first particle is matched exactly to the wall speed, and so it will
+        translate at same velocity along +x for 3 steps. It will bounce back in
+        y and z to where it started. (vx stays the same, and vy and vz flip.)
```
b-butler: While, this may be true (didn't check), we only run for 1 step, so how is this relevant?

hoomd/mpcd/pytest/test_stream.py
```diff
@@ -0,0 +1,439 @@
...
+        ig = hoomd.mpcd.Integrator(dt=0.1, streaming_method=sm)
+        sim.operations.integrator = ig
+
+        with pytest.raises(RuntimeError):
+            sim.run(1)
+
+    def test_test_of_bounds(self, simulation_factory, snap):
+        """Test box validation raises an error on run."""
+        if snap.communicator.rank == 0:
+            snap.mpcd.position[:] = [[4.95, -4.95, 3.85]]
```
b-butler:
```suggestion
            snap.mpcd.position[0] = [[4.95, -4.95, 3.85]]
```
Only one particle anyways.

b-butler (changes requested):
Tried to go through all the code, and leave comments.

hoomd/md/integrate.py
mphoward: No, a VSCode format-on-save did this. I can revert the changes if you want.

hoomd/mpcd/CollisionMethod.cc
mphoward: I switched to `every` because `period` wasn't used elsewhere anymore, but this is not a general `trigger` either. Typically when people write up these methods they say something like "a collision was performed every 5 steps" so I thought that fit. I can switch back to `period` if you prefer though!

hoomd/mpcd/CollisionMethod.cc
mphoward: Does HOOMD use the concept of a "group" elsewhere now that there are "filters" though? Maybe `embedded_particles`?

hoomd/md/integrate.py
joaander: Please do. Numerous unrelated changes caused by slightly differently configured auto-formatters makes code review challenging. yapf doesn't currently allow us to auto-format a specific quote style.

If we switch to ruff in the future, then we can enforce string quote styles automatically via pre-commit.

mphoward: Gotcha, this was probably done by black being enabled at a user level, and I think I thought pre-commit would catch it and put things back. I'll revert.

hoomd/mpcd/CollisionMethod.cc
joaander: I would also prefer `period`. I find class attributes that are not nouns awkward in general.

hoomd/mpcd/Integrator.cc
mphoward: You are right that it should only be a warning if there are no integration methods and no streaming method.

In MPCD, it is fine to run with only a streaming method (pure solvent), but a warning will be issued by the parent integrator when it update gets called below because there are no integration methods. So, we enter this branch if there are no integration methods and pretend to issue the warning (`m_gave_warning = true`) to avoid that case in the parent.

This would be better solved by making the decision to issue a warning a virtual method. I wasn't sure if that was overkill though.

hoomd/mpcd/pytest/test_integrator.py
mphoward: Is there a file with good examples I can model these after?

hoomd/mpcd/CollisionMethod.cc
joaander: Groups are still used internally. The user-facing docs allude to this in the description: https://hoomd-blue.readthedocs.io/en/v4.3.0/module-hoomd-filter.html . Namely, the filter is the object that selects particles and the group is the (not user facing, not accessible) object that represents the particles selected the last time the filter was evaluated. This is also discussed in the `FilterUpdater` docs: https://hoomd-blue.readthedocs.io/en/v4.3.0/module-hoomd-update.html#hoomd.update.FilterUpdater

hoomd/mpcd/collide.py
mphoward: I considered `filter` but decided against it because I thought it sounded like the method would only be applied to the particles in `filter`, and not also the MPCD particles. How about `embedded_particles` in keeping with Josh's preference for nouns?

hoomd/mpcd/collide.py
mphoward: That is a copy–paste from the original documentation. All of docstrings need review and revision, and will keep an eye out for this.

hoomd/mpcd/pytest/test_collide.py
mphoward: I think I wanted to test the attach method her and this can only be done with an integrator, but lines 52–58 are not needed for this.

hoomd/mpcd/pytest/test_collide.py
mphoward: Yes, this is a good idea.

hoomd/mpcd/pytest/test_collide.py
mphoward: The intent of this test in the old code was to trigger all of the internals of the collision method to make sure they worked without error. We can consolidate this with the streaming method in a run test of the integrator though.

hoomd/mpcd/pytest/test_stream.py
mphoward: Yes, we can. I think this might have been cumbersome with the old API.

hoomd/mpcd/pytest/test_stream.py
mphoward: Yes good catch, the is a copy–paste error that was probably in the old code too.

hoomd/mpcd/pytest/test_stream.py
mphoward: `every` is 3, so the particles advance 3 steps when the first run step triggers (the MPCD particles do not stay in sync with the timestep counter). I will update the comment below to make this clear.

hoomd/md/integrate.py
joaander: I would also call the grandparent method. I don't anticipate adding more logic to this integrator attach hook.

hoomd/mpcd/Integrator.cc
joaander: I have been tempted to remove this warning altogether from `IntegratorTwoStep`. If a user choose to add no methods, that is their choice and hoomd's behavior is perfectly well defined in this case. Many users often use this for testing the value of forces/energies.

mphoward: I'm fine with that, and I would then remove this warning here too.

joaander (commented):
Thanks for working on this. I don't have any detailed comments outside what @b-butler has raised.

I am confused how the unit tests are working. The
```
# if version.mpcd_built:
#     from hoomd import mpcd
```
is commented out in `hoomd/__init__.py` - yet the tests somehow successfully use `hoomd.mpcd`. How does this work?

Also, to enable pre-commit formatting and style checks, remove the `mpcd` exclusion lines from `setup.cfg` and `.pre-commit-config.yaml`. Depending on your progress you may or may not want to take these steps in this pull request or defer to the next.

Similarly, you will want to test any code snippets in the documentation (again, probably in a future PR that revises the docstrings). See https://github.com/glotzerlab/hoomd-blue/pulls?q=is%3Apr+is%3Aclosed+sybil for examples of pull requests that use sybil to test `.. code-block:: python` blocks.

mphoward:
> I am confused how the unit tests are working. The
> 
> ```
> # if version.mpcd_built:
> #     from hoomd import mpcd
> ```
> 
> is commented out in `hoomd/__init__.py` - yet the tests somehow successfully use `hoomd.mpcd`. How does this work?

That is a great question! I didn't realize this import was commented out since the tests ran without error...

> Also, to enable pre-commit formatting and style checks, remove the `mpcd` exclusion lines from `setup.cfg` and `.pre-commit-config.yaml`. Depending on your progress you may or may not want to take these steps in this pull request or defer to the next.

I will revert the format change we noticed above, then enable this in the next PR.

> Similarly, you will want to test any code snippets in the documentation (again, probably in a future PR that revises the docstrings). See https://github.com/glotzerlab/hoomd-blue/pulls?q=is%3Apr+is%3Aclosed+sybil for examples of pull requests that use sybil to test `.. code-block:: python` blocks.

I will defer this to a later PR that improves the docstrings.

hoomd/mpcd/pytest/test_collide.py
mphoward: I think we should leave this test in place since it checks that a run with only a collision method can execute a real step. This only flags if something catastrophic were to happen, and more detailed validation of the physics is currently done in C++.

The streaming methods have their own tests that run, and particle level details are checked.

I think it's still valuable to do these checks separately from each other, and that this check is primarily of the collision method (and not integrator). We should add some integration / validation type checks later that do run a full simulation with both attached to an integrator, and I'll make a note to do so.

hoomd/mpcd/pytest/test_stream.py
mphoward: I was mistaken about this. `period` is currently readonly.

hoomd/mpcd/pytest/test_stream.py
mphoward: I changed the period to 1, and increased the timestep to 0.3 so this was less confusing.

hoomd/mpcd/__init__.py
```diff
@@ -30,57 +30,36 @@
...
-    mpcd.collide.srd(seed=42, period=1, angle=130., kT=1.0)
+`hoomd.md`. Getting started can look like:
 
-    hoomd.run(2000)
+    1. Initialize the solvent through `Snapshot.mpcd`. You can include any solute
+       particles in the snapshot as well.
+    2. Create the MPCD `Integrator`.
+    3. Choose the streaming method from :mod:`.mpcd.stream`.
+    4. Choose the collision rule from :mod:`.mpcd.collide`.
+    5. Setup solute particle integration methods and interactions as you normally
```
b-butler: Is is possible/practical to enumerate these restrictions in the documentation (sorry if you do this elsewhere and I missed it).

b-butler (approved):
I think the code is in a good place. The documentation will need a thorough going through before the final merge, but that can take place later when the code examples and such are in place.

hoomd/mpcd/__init__.py
mphoward: The main restriction is that the embedded solute particles cannot be in a rigid body. (We are going to remove this restriction as part of a funded project on this code.) I think this was explicit in the old version of the docs but got dropped out in revision. I'll add it back in `mpcd.collide.CollisionMethod`, and see if this statement can also be tightened up since it sounds more restrictive than it actually is!

mphoward:
> The documentation will need a thorough going through before the final merge, but that can take place later when the code examples and such are in place.

Thanks for the feedback and reviews! Yes, I plan to go through all of the documentation and add examples as the final PR to this staging branch.

We also plan to add some how-tos / examples since those are sorely missing, but that can take place after the staging branch is merged in.

I will implement the doc fix above and also merge up `trunk-minor`, then this should be ready.

hoomd/mpcd/__init__.py
joaander: This block (and a few others) are not intended the correct number of spaces. Sphinx is rendering them as an item in a definition list: https://hoomd-blue--1655.org.readthedocs.build/en/1655/package-mpcd.html

See https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#bullet-lists.

joaander:
Let me know when this is ready to merge.

mphoward:
This should be ready once all the tests pass!

joaander (approved):
Thanks!

PR #1656, opened 2023-11-17
Deprecate TBB.

joaander:
## Description

Deprecate the use of TBB. Warn the user when `num_cpu_threads > 1` and an operation would use more than one thread.

## Motivation and context

See  #1650.

## How has this been tested?

I verified that the `pytest` scripts generated the new warnings in the expected configurations.

## Change log

```
Deprecated:

* ``num_cpu_threads > 1``. Use ``num_cpu_threads = 1``.
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/CONTRIBUTING.rst).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/ContributorAgreement.md).
- [x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

Issue #1651, opened 2023-11-08
Remove implicit depletant code paths.


joaander:
### Description

In a branch off `trunk-minor`:
* [x] Document that implicit depletants are deprecated.
* [x] Issue deprecation warnings when implicit depletants are used.

In a branch off `trunk-major` (later, close to a v5.0 release):
* [x] Remove all implicit depletant code paths.

### Motivation and context

The implicit depletants code paths are broken (#1448). Current developers are not interested in fixing and maintaining the code.

PR #1657, opened 2023-11-20
Deprecate depletant fugacity.

joaander:
## Description

Document that `depletant_fugacity > 0` is deprecated.

Also adjust the deprecation notice grammar and stacklevels implemented in #1656. The stacklevel argument did not behave as I expected. See the commit message for further details.

## Motivation and context

See #1651.

## How has this been tested?

`pytest hoomd/hpmc -W all -v -x -ra` issues the expected warnings.

## Change log

```
Deprecated:

* ``HPMCIntegrator.depletant_fugacity > 0``.
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/integrate.py
+++ b/hoomd/hpmc/integrate.py
@@ -408,10 +408,9 @@ class HPMCIntegrator(Integrator):
         HPMC uses RNGs. Warn the user if they did not set the seed.
         """
         if any([f != 0 for f in self.depletant_fugacity.values()]):
-            warnings.warn(
-                    "depletant_fugacity > 0 is deprecated since 4.4.0.",
-                    FutureWarning,
-                    stacklevel=1)
+            warnings.warn("depletant_fugacity > 0 is deprecated since 4.4.0.",
+                          FutureWarning,
+                          stacklevel=1)
 
             if (isinstance(self._simulation.device, hoomd.device.CPU)
                     and self._simulation.device.num_cpu_threads > 1):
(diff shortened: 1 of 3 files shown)


cbkerr (approved):
I don't see anything obviously wrong but ony checked diffs

hoomd/hpmc/integrate.py
```diff
@@ -399,14 +407,18 @@ def _attach_hook(self):
...
+                          FutureWarning,
+                          stacklevel=1)
+
+            if (isinstance(self._simulation.device, hoomd.device.CPU)
+                    and self._simulation.device.num_cpu_threads > 1):
+                warnings.warn(
+                    "num_cpu_threads > 1 is deprecated since 4.4.0. "
+                    "Use num_cpu_threads=1.",
+                    FutureWarning,
+                    stacklevel=1)
```
SchoeniPhlippsn, line 421: Shouldn't this warning be thrown also when depletant_fugacity is set to 0? Otherwise, I don't understand why we throw this warning at all if depletant_fugacity > 0 is deprecated anyway.

SchoeniPhlippsn (approved):
I only have one comment. Otherwise, it looks straightforward.

hoomd/hpmc/integrate.py
joaander: In HPMC, only the implicit depletant code paths use TBB. The default behavior of TBB is set `num_cpu_threads` to the number of CPU cores on the system. By placing the `num_cpu_threads` warning inside the `depletant_fugacity > 0` check, I am preventing the `num_cpu_threads > 1` warning from appearing for all users of HPMC by default.

And yes, these warnings are redundant in this case. However, a user can act on the `num_cpu_threads` warning to future proof their code while still choosing to use the currently available, but deprecated, implicit depletant code path.

PR #1659, opened 2023-11-20
Remove MD integrator warning when no methods are set

### LINKED ISSUE #1658: Remove warning from MD integrator
### Description

The MD integrator warns when it runs with no integration methods attach. This warning should be removed because it is not necessary.

### Motivation and context

HOOMD will run correctly if the MD integrator has no methods attached to it, and this is a valid configuration that users may want to use (e.g., to test pair potentials). A warning should not be generated in such cases.

mphoward:
## Description

Remove the warning issued by `IntegratorTwoStep` when no integration methods are set.

## Motivation and context

This warning doesn't need to be there, and it can create inconveniences if a class derives from the MD integrator.

Resolves #1658 

## How has this been tested?

The unit tests used to raise lots of warnings about this. I verified they are not there after the change.

## Change log

```
* Remove warning when ``hoomd.md.Integrator`` is used without an integration method.
```

## Checklist:

### DIFF
--- a/hoomd/md/IntegratorTwoStep.cc
+++ b/hoomd/md/IntegratorTwoStep.cc
@@ -17,7 +17,7 @@ namespace hoomd
 namespace md
     {
 IntegratorTwoStep::IntegratorTwoStep(std::shared_ptr<SystemDefinition> sysdef, Scalar deltaT)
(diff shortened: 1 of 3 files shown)


joaander (approved):
Thanks!

PR #1662, opened 2023-11-26
Periodic improper support

kierannp:
## Description

Add support for a periodic improper force for 4 atoms bonded as an improper [https://manual.gromacs.org/current/reference-manual/functions/bonded-interactions.html#proper-dihedrals-periodic-type](url)

## Motivation and context

This is a common functional form in CHARMM-like forcefield and is an easy extension of the periodic dihedral already supported.

## How has this been tested?

Created python functions for calculating the force and energy resulting from this functional formal, see hoomd.md.pytest.test_improper.py for detailed test cases and functions to recreate test results. 

## Change log

```
Added test and support for periodic impropers

```

## Checklist:


CalCraven:
One thing to note about this PR is that the support for periodic impropers is important for running GAFF forcefields in HOOMD, which is recently supported in other engines via MoSDeF. One thing to note while @kierannp and I were working on this is the naming in HOOMD, which can be a bit confusing.
HOOMD has a periodic dihedral support already, which is called via hoomd.md.dihedral.Periodic. But the cpp code that is references is name [HarmonicDihedralForceCompute.cc(https://github.com/glotzerlab/hoomd-blue/blob/trunk-patch/hoomd/md/HarmonicDihedralForceCompute.cc). But again, it is a periodic dihedral force.

There already is a HarmonicImproperForceCompute.cc that computes a true a hoomd.md.improper.Harmonic functional form, and we are adding a hoomd.md.improper.Periodic. So we will call the cpp files PeriodicImproperForceCompute.cc. We may benefit at this time to rename the HarmonicDihedralForceCompute.cc (and like files) to PeriodicDihedralForceCompute.cc to reduce future confusion and potentially open the door for a true HarmonicDihedral force, which should look like any other harmonic force with an (a-a0)**2 type syntax, despite how [LAMMPS](https://docs.lammps.org/dihedral_harmonic.html) notates it with a cosine function, and uses the term `quadratic` to denote [harmonic type dihedrals](https://docs.lammps.org/dihedral_quadratic.html) to confuse all of us.

hoomd/md/PeriodicImproperForceCompute.h
```diff
@@ -0,0 +1,108 @@
...
+        flags |= ForceCompute::getRequestedCommFlags(timestep);
+        return flags;
+        }
+#endif
+
+    protected:
+    Scalar* m_K;     //!< K parameter for multiple improper tyes
+    Scalar* m_sign;  //!< sign parameter for multiple improper types
+    int* m_multi;    //!< multiplicity parameter for multiple improper types
+    Scalar* m_chi_0; //!< chi_0 parameter for multiple improper types
```
joaander: Replace this with a GPUArray of `periodic_improper_params`, similar to bond parameters in https://github.com/glotzerlab/hoomd-blue/blob/0048a57dec73de737f0f53a4a480dd31ef19659e/hoomd/md/PotentialBond.h#L64

Doing this will reduce duplicate code in the later GPU implementation.

hoomd/md/improper.py
```diff
@@ -110,3 +110,46 @@ def __init__(self):
...
+              :math:`\\phi_0` :math:`[\\mathrm{radian}]`.
+            * ``n`` (`float`, **required**), periodic number
+              :math:`\\n` :math:`[\\mathrm{dimensionless}]`.
+            * ``d`` (`float`, **required**), sign factor
+              :math:`\\n` :math:`[\\mathrm{dimensionless}]`.
+
+    Example::
+
+        harmonic = hoomd.md.improper.Periodic()
+        harmonic.params['A-B-C-D'] = dict(k=1.0, n = 1, phi0=0)
```
joaander: Make these changes to test the example code snippet using [sybil](https://sybil.readthedocs.io).

```suggestion
    .. rubric:: Example:

    .. code-block:: python

        periodic = hoomd.md.improper.Periodic()
        periodic.params['A-B-C-D'] = dict(k=1.0, n = 1, phi0=0)
        
        simulation.integrator.forces = [periodic]
```
and set the parameters appropriately.

Also, add the following to the top of the file in the module level docstring:
```
.. skip: next if(not hoomd.version.md_built)

.. invisible-code-block: python

    if hoomd.version.md_built:
        simulation = hoomd.util.make_example_simulation()
        simulation.operations.integrator = hoomd.md.Integrator(dt=0.001)
```

hoomd/md/PeriodicImproperForceCompute.cc
```diff
@@ -0,0 +1,373 @@
...
+        {
+        // lookup the tag of each of the particles participating in the improper
+        const ImproperData::members_t& improper = m_improper_data->getMembersByIndex(i);
+        assert(improper.tag[0] <= m_pdata->getMaximumTag());
+        assert(improper.tag[1] <= m_pdata->getMaximumTag());
+        assert(improper.tag[2] <= m_pdata->getMaximumTag());
+        assert(improper.tag[3] <= m_pdata->getMaximumTag());
+
+        // transform a, b, and c into indices into the particle data arrays
+        // MEM TRANSFER: 6 ints
```
joaander:
```suggestion
```

hoomd/md/improper.py
```diff
@@ -110,3 +110,46 @@ def __init__(self):
                 chi0=hoomd.data.typeconverter.nonnegative_real,
                 len_keys=1))
         self._add_typeparam(params)
+
+class Periodic(Improper):
```
joaander, line 123: Add this class name to the list in `module-md-improper.rst` so the documentation appears.

hoomd/md/pytest/test_improper.py
```diff
@@ -7,8 +7,56 @@
                             autotuned_kernel_parameter_check)
 import pytest
 import numpy
+import math
 
 import itertools
+
+# these series of functions are used to calculate the values for the test cases used below
```
joaander: These don't seem to be used below. I would suggest either using them in this method or preparing a separate `generate_periodic_improper_test_cases.py` file that uses them to generate the test cases used below.

joaander (commented):
This looks like a good start! Here are some suggestions for improvement.

joaander:
> There already is a HarmonicImproperForceCompute.cc that computes a true a hoomd.md.improper.Harmonic functional form, and we are adding a hoomd.md.improper.Periodic. So we will call the cpp files PeriodicImproperForceCompute.cc. We may benefit at this time to rename the HarmonicDihedralForceCompute.cc (and like files) to PeriodicDihedralForceCompute.cc to reduce future confusion and potentially open the door for a true HarmonicDihedral force, which should look like any other harmonic force with an (a-a0)**2 type syntax, despite how [LAMMPS](https://docs.lammps.org/dihedral_harmonic.html) notates it with a cosine function, and uses the term `quadratic` to denote [harmonic type dihedrals](https://docs.lammps.org/dihedral_quadratic.html) to confuse all of us.

Yes, feel free to rename `HarmonicDihedralForceCompute*` to `PeriodicDihedralForceCompute*` at the C++ level, both with the file and class names. Please make this change in a separate pull request.

hoomd/md/pytest/test_improper.py
CalCraven: @kierannp I think putting these into a seperate generate_periodic_improper_test_cases.py make sense, and then we can import them to generate the force and energy terms used in the test cases. We could then easily add a test case for a few different positions, as well as future users would be able to use the dchi_dr functions to test forces for any future functional forms using chi. Also tagging @calvinpozderac who actually got these derivatives using mathematica.

CalCraven (changes requested):
Looking good so far.

hoomd/md/pytest/test_improper.py
joaander: Now that they are called in the test, I'm fine with leaving them here. I would suggest wrapping them as static methods in a class to put them in a namespace that won't be confused with other tests:

```
class PeriodicImproperMath:
   @staticmethod
   def dchi_dr1(...)
```

kierannp: I added them to a separate file as suggested. I think this format is more clean.

hoomd/md/improper.py
kierannp: > Make these changes to test the example code snippet using [sybil](https://sybil.readthedocs.io).
> 
> and set the parameters appropriately.
> 
> Also, add the following to the top of the file in the module level docstring:
> 
> ```
> .. skip: next if(not hoomd.version.md_built)
> 
> .. invisible-code-block: python
> 
>     if hoomd.version.md_built:
>         simulation = hoomd.util.make_example_simulation()
>         simulation.operations.integrator = hoomd.md.Integrator(dt=0.001)
> ```

This example code produces the error:

```AttributeError: 'Simulation' object has no attribute 'integrator' ```

Do you have a suggestion for a different example case?

File: hoomd/md/PeriodicImproperForceCompute.cc (hunk 3 above)
Comment (CalCraven) [reply hoomd/md/PeriodicImproperForceCompute.cc]: Reminder to remove this line.

**Review by CalCraven — CHANGES_REQUESTED** [review]
Looking good, I think there are two more of Josh's edits still to get at. Also add your name to [this](https://github.com/glotzerlab/hoomd-blue/blob/0048a57dec73de737f0f53a4a480dd31ef19659e/sphinx-doc/credits.rst) file as well, should we add a note to the changelog?

**Comment by joaander** [comment]
> should we add a note to the changelog?

Please don't. HOOMD has so many PRs that all the change log entries would conflict. I build the change log at release time from the drafts in each of the PR descriptions.

File: hoomd/md/improper.py (hunk 2 above)
Comment (joaander) [reply hoomd/md/improper.py]: You can find many examples scattered throughout the HOOMD documentation, search for `.. invisible-code-block: python`. I am certain this code works. Here is a test:
```
>>> simulation = hoomd.util.make_example_simulation()
>>> simulation.operations.integrator = hoomd.md.Integrator(dt=0.001)
>>>
```

File: hoomd/md/PeriodicImproperForceGPU.cu [hunk 6]
```diff
@@ -0,0 +1,380 @@
...
+        Scalar dtfy = gaa * aay;
+        Scalar dtfz = gaa * aaz;
+        Scalar dtgx = fga * aax - hgb * bbx;
+        Scalar dtgy = fga * aay - hgb * bby;
+        Scalar dtgz = fga * aaz - hgb * bbz;
+        Scalar dthx = gbb * bbx;
+        Scalar dthy = gbb * bby;
+        Scalar dthz = gbb * bbz;
+
+        // Scalar df = -K * dfab;
```
tcmoore3:
```suggestion
```

tcmoore3 (approved):

joaander (approved):
Thanks for the complete PR and extensive tests.

I pushed a commit that moves the parameters into a GPUArray to reduce code duplication and the chances for errors in that duplication. I also made a few other changes to bring error messages up to modern HOOMD coding standards.

Could you double check my changes and let me know if they are correct? I got a little lost with the parameters. It seems that "d" and "sign" are the same parameter. It also seems that "n", "multi", and "m" are also all the same. You might consider modifying the code to keep the same names throughout to avoid confusion.

kierannp:
> Thanks for the complete PR and extensive tests.
> 
> I pushed a commit that moves the parameters into a GPUArray to reduce code duplication and the chances for errors in that duplication. I also made a few other changes to bring error messages up to modern HOOMD coding standards.
> 
> Could you double check my changes and let me know if they are correct? I got a little lost with the parameters. It seems that "d" and "sign" are the same parameter. It also seems that "n", "multi", and "m" are also all the same. You might consider modifying the code to keep the same names throughout to avoid confusion.

The changes you made appear correct. I tried to use the same naming conventions as in similar scripts in HOOMD for calculating forces. I will look into making these variables names more consistent and explicit.

CalCraven:
> > It seems that "d" and "sign" are the same parameter. It also seems that "n", "multi", and "m" are also all the same. 
> 
> I tried to use the same naming conventions as in similar scripts in HOOMD for calculating forces. 

To clarify a bit more, d is named sign because d is always +-1. n is the multiplicity of the cos function. I would second the fact that we should stick to one value within hoomd. I believe these were derived from [this line](https://github.com/glotzerlab/hoomd-blue/blob/3179f17d71b85aad8024a66f6cd55853feb86fa2/hoomd/md/HarmonicDihedralForceCompute.cc#L252C45-L252C45). When I pull a PR to rename the HarmonicDihedralForce to PeriodicDihedralForce, we can make the same consistent name changes to those files as well.

Issue #1652, opened 2023-11-08
Replace runtime CPPPotential potentials with compile time provided ones.


joaander:
### Description

On `trunk-minor`:
- [x] Note that CPP potential classes are deprecated.
- Add pair potentials:
  - [x] Lennard Jones
  - [x] Step function
  - [x] Union
  - [x] Angular step
- Add external potentials
  -  [x] Linear.
- [x] Add example plugin.
- [x] Update tutorials.
- [x] Update hoomd-validation to use new potentials.

On `trunk-major` (later, close to the release of v5.0):
- [x] Remove CPP potential classes.
- [x] Port walls and harmonic external potentials to new framework.
- [x] Remove JIT framework.
- [x] Remove LLVM from CMake.
- [x] Remove all mentions of run time compile potentials from the documentation.
- [x] Move all external hpmc potentials to `hoomd.hpmc.external`
- [x] Port muVT and SDF square well tests.

### Motivation and context

The LLVM run time compilation code path is very costly to maintain. Using `CPPPotential` requires that users write C++ code regardless, so asking them to compile it is not a significant additional burden.

PR #1676, opened 2023-12-12
Implement new HPMC pair potential framework

joaander:
## Description

* Implement a new framework for multiple HPMC pair potentials: `HPMCIntegrator.pair_potentials` on the CPU.
* Provide `LennardJones` as the first example using the new framework.

The new framework allows:
1) Multiple pair potentials in a list.
2) Potentials have proper type parameters.
3) Uses `ShortReal` appropriately for mixed precision builds.
4) Allows users to write custom potentials that are built at compile time.
5) Names the class `PairPotential` to match the `pair_potentials` name used in the documentation.

## Motivation and context

This is the first step toward implementing #1652. Future pull requests will add multi-site pair potentials (based on the `CPPPotentialUnion` code), anisotropic potentials (e.g. Kern-Frenkel) and provide an example component.

## How has this been tested?

TODO:

- [x] Add unit tests.
- [x] Fix ComputeSDF.
- [x] Fix UpdaterClusters.
- [x] Fix UpdaterMuVT.
- [x] Apply deprecations.
- [x] Validate `LennardJones` pair potential in `hoomd-validation`.
- [x] Validate `ComputeSDF` in `hoomd-validation`.

## Change log

```
Added:

* HPMC compile time pair potential framework (CPU only). Allows potential energy in HPMC simulations without `CPPPotential`.
  * ``hoomd.hpmc.pair.LennardJones`` - Evaluate Lennard Jones energy between particles
    (`#1676 <https://github.com/glotzerlab/hoomd-blue/pull/1676>`__).
  * ``HPMCIntegrator.pair_potentials`` - Set a list of pair potentials to evaluate
    (`#1676 <https://github.com/glotzerlab/hoomd-blue/pull/1676>`__). 
  * ``HPMCIntegrator.pair_energy`` (loggable) - Total pair energy from all pair potentials.
    (`#1676 <https://github.com/glotzerlab/hoomd-blue/pull/1676>`__).

Deprecated:

* ``hoomd.hpmc.pair.user.CPPPotentialBase`` 
  (`#1676 <https://github.com/glotzerlab/hoomd-blue/pull/1676>`__).
* ``hoomd.hpmc.pair.user.CPPPotential`` - Use a built-in potential or compile your code in a component 
  (`#1676 <https://github.com/glotzerlab/hoomd-blue/pull/1676>`__).
* ``hoomd.hpmc.pair.user.CPPPotentialUnion`` - Use a built-in potential or compile your code in a component
  (`#1676 <https://github.com/glotzerlab/hoomd-blue/pull/1676>`__).
* ``HPMCIntegrator.pair_potential`` - Use compiled potentials with ``pair_potentials`` 
  (`#1676 <https://github.com/glotzerlab/hoomd-blue/pull/1676>`__).

Changed:

* Refactored the C++ API for ``PatchEnergy`` potentials.
```

## Checklist:


joaander:
The validation tests pass:
lj_fluid_compare_kT1 0_density0 92_r_cut1 12_N1728
lj_fluid_compare_kT1 5_density0 63_r_cut4 0_N4096
lj_fluid_distribution_analyze_kT1 0_density0 92_r_cut1 12_N1728
lj_fluid_distribution_analyze_kT1 5_density0 63_r_cut4 0_N4096

The code for the tests is in: https://github.com/glotzerlab/hoomd-validation/pull/78

hoomd/hpmc/pair/__init__.py
```diff
@@ -3,9 +3,161 @@
...
+.. rubric:: Example:
+
+.. invisible-code-block: python
+
+    simulation = hoomd.util.make_example_simulation()
+    sphere = hoomd.hpmc.integrate.Sphere()
+    sphere.shape['A'] = dict(diameter=0.0)
+    simulation.operations.integrator = sphere
+
+    pair =  hoomd.hpmc.pair.LennardJones()
```
b-butler, line 20: I know we are moving towards more explicit naming, but we should be consistent on potentials across MD and HPMC.

hoomd/hpmc/pair/__init__.py
```diff
@@ -3,9 +3,161 @@
...
+            default_r_cut = float(default_r_cut)
+
+        params = hoomd.data.typeparam.TypeParameter(
+            'params', 'particle_types',
+            hoomd.data.parameterdicts.TypeParameterDict(
+                epsilon=float,
+                sigma=float,
+                r_cut=default_r_cut,
+                r_on=float(default_r_on),
+                mode=str(default_mode),
```
b-butler: We are making mode a per-pair type parameter and merging `r_cut` and `r_on` into `params`? Why the deviation from MD?

hoomd/hpmc/pytest/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ set(files __init__.py
           test_shape_updater.py
           test_shape_utils.py
           test_move_size_tuner.py
+          test_pair_lennard_jones.py
```
b-butler, line 16: Do we plan on adding other potentials to this file? Currently, `hoomd.hpmc.pair` contains `Pair` and `LennardJones` which suggests that we will keep one pair file like in MD. If I am right, we should rename this test file to just a generic `test_pair.py`.

hoomd/hpmc/UpdaterMuVT.h
```diff
@@ -1695,14 +1686,13 @@ bool UpdaterMuVT<Shape>::tryRemoveParticle(uint64_t timestep, unsigned int tag,
                                              access_mode::read);
 
                 // Check particle against AABB tree for neighbors
-                Scalar r_cut_patch = patch->getRCut() + 0.5 * patch->getAdditiveCutoff(type);
+                Scalar r_cut_patch = m_mc->getMaxPairEnergyRCutNonAdditive()
+                                     + 0.5 * m_mc->getMaxPairInteractionAdditiveRCut(type);
```
b-butler: Should this be
```suggestion
                                     + LongReal(0.5) * m_mc->getMaxPairInteractionAdditiveRCut(type);
```

hoomd/hpmc/pytest/test_pair_lennard_jones.py
```diff
@@ -0,0 +1,199 @@
...
+    dict(default_r_on=2.0),
+    dict(default_mode='shift'),
+]
+
+
+@pytest.mark.parametrize("constructor_args", valid_constructor_args)
+def test_valid_construction(device, constructor_args):
+    """Test that LennardJones can be constructed with valid arguments."""
+    lennard_jones = hoomd.hpmc.pair.LennardJones(**constructor_args)
+    assert lennard_jones is not None
```
b-butler: This isn't necessary. No assert is needed for this test.
```suggestion
```

hoomd/hpmc/pair/__init__.py
```diff
@@ -3,9 +3,161 @@
...
+
+    simulation.operations.integrator.pair_potentials = [pair]
 """
 
 from . import user
+
+import hoomd
+
+
+class Pair(hoomd.operation._HOOMDBaseObject):
```
b-butler: I generally prefer an emptier `__init__.py` with the logic moved to submodules.

hoomd/hpmc/pytest/test_pair_lennard_jones.py
```diff
@@ -0,0 +1,199 @@
...
+    dict(epsilon=1.0, sigma=1.0, r_cut='invalid'),
+    dict(epsilon=1.0, sigma=1.0, r_cut=2.5, r_on='invalid'),
+    dict(epsilon=1.0, sigma=1.0, r_cut=2.5, r_on=2.0, mode='invalid'),
+    dict(epsilon=1.0, sigma=1.0, r_cut=2.5, r_on=2.0, mode='shift', invalid=10),
+]
+
+
+@pytest.mark.parametrize("parameters", invalid_parameters)
+@pytest.mark.cpu
+def test_invalid_params(mc_simulation_factory, parameters):
```
b-butler:
```suggestion
def test_invalid_params_on_attach(mc_simulation_factory, parameters):
```

hoomd/hpmc/IntegratorHPMC.cc
```diff
@@ -6,6 +6,9 @@
 #include "hoomd/VectorMath.h"
 #include <sstream>
 
+#include <pybind11/stl_bind.h>
+PYBIND11_MAKE_OPAQUE(std::vector<std::shared_ptr<hoomd::hpmc::PairPotential>>);
```
b-butler, line 10: Do you need to define this both in `PairPotential.cc` and here?

hoomd/hpmc/PairPotentialLennardJones.cc
```diff
@@ -0,0 +1,93 @@
...
+                                           const LongReal charge_i,
+                                           const unsigned int type_j,
+                                           const quat<LongReal>& q_j,
+                                           const LongReal charge_j) const
+    {
+    unsigned int param_index = m_type_param_index(type_i, type_j);
+    const auto& param = m_params[param_index];
+
+    LongReal lj1 = param.epsilon_x_4 * param.sigma_6 * param.sigma_6;
+    LongReal lj2 = param.epsilon_x_4 * param.sigma_6;
```
b-butler: The compiler is likely smart enough for this, but just in case.
```suggestion
    LongReal lj2 = param.epsilon_x_4 * param.sigma_6;
    LongReal lj1 = lj2 * param.sigma_6;
```

hoomd/hpmc/PairPotential.h
```diff
@@ -0,0 +1,148 @@
...
+
+    /// Pre computed total r_cut squared (indexed by m_type_param_index).
+    std::vector<LongReal> m_r_cut_squared_total;
+
+    /// Update the pre-computed r_cut_total value.
+    void updateRCutTotal(unsigned int type_i, unsigned int type_j)
+        {
+        unsigned int type_pair = m_type_param_index(type_i, type_j);
+        LongReal r_cut = m_r_cut_non_additive[type_pair] + LongReal(0.5) * m_r_cut_additive[type_i]
+                         + m_r_cut_additive[type_j];
```
b-butler: Is this meant to be?
```suggestion
        LongReal r_cut = m_r_cut_non_additive[type_pair] + LongReal(0.5) * (m_r_cut_additive[type_i]
                         + m_r_cut_additive[type_j]);
```

hoomd/hpmc/IntegratorHPMC.h
```diff
@@ -437,6 +438,100 @@ class PYBIND11_EXPORT IntegratorHPMC : public Integrator
...
+            {
+            r_cut = m_patch->getRCut();
+            }
+        for (const auto& pair : m_pair_potentials)
+            {
+            for (unsigned int type_i = 0; type_i < m_pdata->getNTypes(); type_i++)
+                {
+                for (unsigned int type_j = 0; type_j < m_pdata->getNTypes(); type_j++)
+                    {
+                    r_cut = std::max(r_cut, pair->getRCutNonAdditive(type_i, type_j));
```
b-butler: Maybe it makes sense to have a `PairPotential::getMaxRCutNonAdditive` method to just do a linear search over half of the `std::vector`?

b-butler (changes requested):
The PR looks good. I only have a few comments. Also, if possible we should document the expected changes between potentials in MC and MD.

hoomd/hpmc/pair/__init__.py
joaander: With HPMC potentials I am aiming for the cleanest API, not necessarily one that is duck-type compatible with MD.

In MD, `r_cut` and `r_on` are explicitly managed by the template class `PotentialPair` and are stored separately from the potential specific parameters out of necessity. In HPMC, not all potentials will have `r_on` and some will not even have `r_cut` (e.g. angular patches). I thought the most natural expression was to combine it in params as it is a parameter of the function definition.

Similarly, not all HPMC potentials (e.g. square well, patch, ..) will have a mode. I could move `mode` out to be a member of `PairPotentialLennardJones` if you think strongly that the mode should be identical for all parameters in the matrix.

hoomd/hpmc/pytest/CMakeLists.txt
joaander: There will be a number of potentials in the near future. They will all have unique testing requirements (orientation, particle separation, etc...). I think they would be easiest to navigate in separate files. We can merge all tests into one massive test file if you prefer.

hoomd/hpmc/IntegratorHPMC.cc
joaander: I am fairly sure it does. From https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html (in an example on `std::vector<int>`):

> This macro must be specified at the top level (and outside of any namespaces), since it adds a template instantiation of type_caster. If your binding code consists of multiple compilation units, **it must be present in every file** (typically via a common header) preceding any usage of std::vector<int>.

hoomd/hpmc/PairPotential.h
joaander: Yes, good catch!

hoomd/hpmc/IntegratorHPMC.h
joaander: Yes. And/or have PairPotential cache the maximimum or move all calls to `getMaxRCutNonAdditive` outside of loops so it is called once/step instead of once/particle.

hoomd/hpmc/pytest/test_pair_lennard_jones.py
joaander: We either need this, or a `#noqa` to quiet flake8 warnings regarding the unused variable `lennard_jones`.

hoomd/hpmc/pair/__init__.py
joaander: Yes, this is consistent with a future version of HOOMD that names the MD potential `LennardJones`: https://github.com/glotzerlab/hoomd-blue/issues/1698

hoomd/hpmc/IntegratorHPMCMono.h
```diff
@@ -1191,21 +1222,20 @@ float IntegratorHPMCMono<Shape>::computePatchEnergy(uint64_t timestep)
...
 
         // the cut-off
-        ShortReal r_cut = ShortReal(m_patch->getRCut() + 0.5*m_patch->getAdditiveCutoff(typ_i));
+        LongReal r_cut = getMaxPairEnergyRCutNonAdditive() + static_cast<LongReal>(0.5) *
+                    getMaxPairInteractionAdditiveRCut(typ_i);
 
         // subtract minimum AABB extent from search radius
-        ShortReal R_query = std::max(shape_i.getCircumsphereDiameter()/ShortReal(2.0),
-            r_cut-getMinCoreDiameter()/(ShortReal)2.0);
+        LongReal R_query = r_cut-getMinCoreDiameter()/LongReal(2.0);
```
joaander: TODO: Restore std::max check.

joaander: Wait, nevermind. This is in `computeEnergy` which does not evaluate overlaps.

hoomd/hpmc/pytest/test_pair_lennard_jones.py
b-butler: You can just use either `_ = hoomd.hpmc.pair.LennardJones(...)` or `hoomd.hpmc.pair.LennardJones(...)` neither of which should warn.

hoomd/hpmc/IntegratorHPMC.cc
b-butler: Ah thanks for the info. Didn't know that.

b-butler (approved):
This PR looks ready to me.

joaander:
Here is a performance test of `hoomd-benchmarks.hpmc_pair_lj` with 1000 particles on cheme-hodges:
* 4.4.1 - LLVM: 117 sweeps per second
* `hpmc-pair-lj` - LLVM: 88.6 sweeps per second
* `hpmc-pair-lj` - compiled potential: 141 sweeps per second

There is some small overhead introduced to the LLVM code path from the generalization to add multiple pair potentials together (`computeOnePairEnergy`). However, the compiled code path is even faster than the LLVM code path, and should be even faster after we remove LLVM. Much of the LLVM path's overhead is in the virtual function calls to `getRCut` and `getAdditiveCutoff`. The new compiled code path makes these inlined non-virtual functions to avoid this overhead.

hoomd/hpmc/IntegratorHPMC.h
```diff
@@ -437,6 +438,111 @@ class PYBIND11_EXPORT IntegratorHPMC : public Integrator
...
+            r_cut = std::max(r_cut, pair->getRCutAdditive(type));
+            }
+
+        return r_cut;
+        }
+
+    __attribute__((always_inline)) inline LongReal computeOnePairEnergy(const LongReal r_squared,
+                                                                        const vec3<LongReal>& r_ij,
+
+                                                                        unsigned int type_i,
```
tcmoore3:
```suggestion
                                                                        const vec3<LongReal>& r_ij,
                                                                        unsigned int type_i,
```

hoomd/hpmc/IntegratorHPMC.h
```diff
@@ -437,6 +438,111 @@ class PYBIND11_EXPORT IntegratorHPMC : public Integrator
...
+            }
+        for (const auto& pair : m_pair_potentials)
+            {
+            r_cut = std::max(r_cut, pair->getRCutAdditive(type));
+            }
+
+        return r_cut;
+        }
+
+    __attribute__((always_inline)) inline LongReal computeOnePairEnergy(const LongReal r_squared,
```
tcmoore3, line 479: Why add `r_squared` to the argument list? Is the idea that you have generally already calculated `dot(r_ij, r_ij)` to check the pair distance against the cutoff and it's likely that you'll need it again when evaluating the pair energy, so you make it available to save a dot product calculation?

hoomd/hpmc/PairPotential.h
```diff
@@ -0,0 +1,175 @@
...
+
+namespace hoomd
+    {
+namespace hpmc
+    {
+/*** Functor that computes pair interactions between particles
+
+    PairPotential allows cutoff energetic interactions to be included in an HPMC simulation. This
+    abstract base class defines the API for the pair energy object, consisting of cutoff radius
+    and the pair energy evaluation fuction.
```
tcmoore3:
```suggestion
    PairPotential allows energetic interactions to be included in an HPMC simulation. This
    abstract base class defines the API for the pair energy object, consisting of a cutoff radius
    and the pair energy evaluation fuction.
```

hoomd/hpmc/pair/lennard_jones.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) 2009-2023 The Regents of the University of Michigan.
+# Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+"""Lennard Jones pair potential.
```
tcmoore3:
```suggestion
"""Lennard-Jones pair potential.
```

hoomd/hpmc/pair/lennard_jones.py
```diff
@@ -0,0 +1,96 @@
...
+        simulation.operations.integrator.pair_potentials = [lennard_jones]
+
+    .. py:attribute:: params
+
+        The potential parameters. The dictionary has the following keys:
+
+        * ``epsilon`` (`float`, **required**) -
+          Energy well depth :math:`\\varepsilon` :math:`[\\mathrm{energy}]`.
+        * ``sigma`` (`float`, **required**) -
+          Particle size :math:`\\sigma` :math:`[\\mathrm{length}]`.
```
tcmoore3:
```suggestion
          Characteristic length scale :math:`\\sigma` :math:`[\\mathrm{length}]`.
```
Should we be more exact and say that sigma isn't strictly a particle size?

tcmoore3: Maybe say "effective particle size"?

tcmoore3 (approved):
This is great. Just a few clarification questions, but otherwise seems ready to go.

hoomd/hpmc/IntegratorHPMC.h
joaander: Yes, this is a performance micro-optimization. I measured a noticeable increase in performance with this change. It isn't the best API design as `r_squared` must be commensurate with `r_ij` for correct results, but this API is only used internally.

hoomd/hpmc/pair/lennard_jones.py
joaander: "Characteristic length scale" is the most accurate name. We should update the MD documentation to match.

PR #1678, opened 2023-12-14
Add QuickCompress.allow_unsafe_resize

### LINKED ISSUE #1677: Allow 0 trial move size with `QuickCompress`.
### Description

`QuickCompress` computes the maximum safe box size change based on the translation move. When that translation move is 0, the box cannot be scaled at all.

### Proposed solution

Implement failsafe mode in `QuickCompress`. When the trial move size is 0, choose a random box move up to the user-provided `min_scale`. Reject box moves that introduce overlaps.

### Additional context

This will allow compressing simulation boxes while keeping particles fixed on a lattice. It is also required to implement fixed seeds as proposed in #1643. Slow compression (#1641) may also be useful when the min trial move size is 0 - but each of these features can be implemented separately.

janbridley:
## Description

Added check for minimum move size in `QuickCompress` that sets the box scale if min_move_size is 0.

## Motivation and context

Several researchers have attempted to use `QuickCompress` with a translational move size of zero to compress a crystal with fixed particle positions. While the code could be trivially updated to allow this, the default settings are likely to result in unresolvable overlaps. The `allow_unsafe_resize`, which defaults to False, provides a solution to this: users who set it to True will allow compression regardless of translational move sizes. In most cases, however, users should also set `max_overlaps_per_particle` to zero to avoid these unresolvable cases.

Resolves #1677

## How has this been tested?

TODO: tests will be added

## Change log

```
Added `allow_unsafe_resize` flag to `QuickCompress`
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/update.py
+++ b/hoomd/hpmc/update.py
@@ -852,10 +852,9 @@ class QuickCompress(Updater):
 
     Warning:
         When the smallest MC translational move size is 0, `allow_unsafe_resize`
-        must be set to `True` to progress toward the target box. A decrease in
-        `max_overlaps_per_particle` is recommended when using this setting. In
-        all other cases, using the default `allow_unsafe_resize` = `False` is
-        preferred, as it proposes better box moves.
+        must be set to `True` to progress toward the target box. Decrease
+        `max_overlaps_per_particle` when using this setting to prevent
+        unresolvable overlaps.
 
     Warning:
         Use `QuickCompress` *OR* `BoxMC`. Do not use both at the same time.
--- a/hoomd/hpmc/update.py
+++ b/hoomd/hpmc/update.py
@@ -879,16 +879,19 @@ class QuickCompress(Updater):
                  trigger,
                  target_box,
                  max_overlaps_per_particle=0.25,
-                 min_scale=0.99):
+                 min_scale=0.99,
+                 allow_unsafe_resize=False):
         super().__init__(trigger)
 
         param_dict = ParameterDict(max_overlaps_per_particle=float,
                                    min_scale=float,
                                    target_box=hoomd.Box,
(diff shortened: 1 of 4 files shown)


janbridley:
pre-commit.ci autofix

janbridley:
pre-commit.ci autofix

janbridley:
After some real world testing it appears that setting `max_overlaps_per_particle` to 0 is useful for configurations that are likely to cause overlaps (for example, axis-oriented octahedra in simple cubic). For easier cases though (cubes in generally cubic structures, or space filling shapes in their correct orientations) the "quick" part of `QuickCompress` works well.

hoomd/hpmc/update.py
```diff
@@ -838,16 +841,21 @@ class QuickCompress(Updater):
...
         acceptance ratio as the density of the system increases.
 
     Warning:
-        When the smallest MC translational move size is 0, `QuickCompress`
-        will scale the box by 1.0 and not progress toward the target box.
+        When the smallest MC translational move size is 0, `allow_unsafe_resize`
+        must be set to `True` to progress toward the target box. A decrease in
+        `max_overlaps_per_particle` is recommended when using this setting. In
+        all other cases, using the default `allow_unsafe_resize` = `False` is
+        preferred, as it proposes better box moves.
```
joaander:
```suggestion
        When the smallest MC translational move size is 0, `allow_unsafe_resize`
        must be set to `True` to progress toward the target box. Decrease
        `max_overlaps_per_particle` when using this setting to prevent
        unresolvable overlaps.
```
Users decide what they prefer.

joaander (approved):
Thanks for the clean PR! I have one tweak to the docs which I will apply and then merge.

PR #1687, opened 2023-12-19
Fix AABB tree performance.

joaander:
## Description

Use the fast AABB-AABB overlap check when traversing AABB trees.

I implemented this by moving the `overlap(AABB, AABB)` function to a member function of AABB: `aabb.overlaps(other_aabb)`. Explicitly calling the intended overlap test ensures that the correct function is called.

## Motivation and context

Increase performance.

For reasons I don't fully understand, the compiler chose to automatically convert AABBs to OBBs, then call `overlap(OBB, OBB)` - which is many times slower than `overlap(AABB, AABB)`.

## How has this been tested?

This pull request increases performance by ~40-50% in HPMC benchmarks (https://github.com/glotzerlab/hoomd-benchmarks/pull/68). It should also improve MD performance with `nlist.Tree`, though I haven't tested that.

## Change log

```
Fixed:

* Increase HPMC performance a further ~50% on the CPU.
```

## Checklist:


hoomd/hpmc/OBB.h
```diff
@@ -115,7 +115,7 @@ struct OBB
         is_sphere = 1;
         }
 
-    DEVICE OBB(const hoomd::detail::AABB& aabb)
+    DEVICE explicit OBB(const hoomd::detail::AABB& aabb)
```
joaander, line 118: This change ensures that no future automatic conversions from AABB to OBB are allowed.

Callers must opt into the conversion explicitly: `OBB obb(aabb)` or `OBB(aabb)`.

tommy-waltmann (commented):
Have you tested to see if it was necessary to move the function call to a class member of AABB, or if just adding `explicit` to the OBB constructor was enough to do the trick?

tommy-waltmann (approved):
Approving because I don't want to hold up the next release.

tcmoore3 (approved):

joaander:
> Have you tested to see if it was necessary to move the function call to a class member of AABB, or if just adding `explicit` to the OBB constructor was enough to do the trick?

When I only added explicit, I got an error compile error that there was no overlap(AABB, AABB) to call. I don't know why because it is clearly in the header file - so I moved it to a member instead.

tommy-waltmann:
> When I only added explicit, I got an error compile error that there was no overlap(AABB, AABB) to call. I don't know why because it is clearly in the header file - so I moved it to a member instead.

That seems like the reason the compiler was automatically converting the AABBs to OBBs, it couldn't find the overlap(AABB, AABB) declaration.

PR #1690, opened 2024-01-04
Reimplement MPCD virtual particle fillers

mphoward:
## Description

This PR reimplements the Python API for adding "virtual" solvent particles during the collision step.

~This code is branched off `mpcd-v4-geometries` and so should be reviewed after PR #1689 is merged.~ This one is ready to merge now!

## Motivation and context

This feature is needed to get the correct solvent properties near boundaries defined in `mpcd.geometry`.

## How has this been tested?

C++ tests already existed for the virtual particle fillers, and Python tests were rewritten.

## Change log

N/A

## Checklist:

- [X] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/CONTRIBUTING.rst).
- [X] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/ContributorAgreement.md).
- [X] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

PR #1691, opened 2024-01-06
Reimplement MPCD bounce-back integration method

mphoward:
## Description

This PR reimplements the bounce-back integrator for MD particles that is a companion to the bounce-back streaming method for the MPCD particles. It is added in `mpcd.methods` and uses `mpcd.geometry.Geometry` to define the bounce back geometry.

There are some documentation improvements to `mpcd.stream` included in this PR for stuff that I noticed when I was also writing the `mpcd.methods` documentation. Since everything is going into a staging branch, I just included them on this PR.

~This PR should be reviewed after #1690.~

## Motivation and context

This feature is helpful for coupling solute particles to the same geometry as the MPCD solvent, especially if it isn't easy to define an external potential to keep the solute particles out of some regions.

## How has this been tested?

Python unit tests from 2.9.7 have been reimplemented.

## Change log

N/A

## Checklist:

### DIFF
--- a/hoomd/mpcd/methods.py
+++ b/hoomd/mpcd/methods.py
@@ -1,0 +1,107 @@
+# Copyright (c) 2009-2023 The Regents of the University of Michigan.
+# Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+r""" MPCD integration methods
+
+Defines extra integration methods useful for solutes (MD particles) embedded in
+an MPCD solvent. However, these methods are not restricted to MPCD
+simulations: they can be used as methods of `hoomd.md.Integrator`. For example,
+`BounceBack` might be used to run DPD simulations with surfaces.
+
+"""
+
+import hoomd
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.filter import ParticleFilter
+from hoomd.md.methods import Method
+from hoomd.mpcd import _mpcd
+from hoomd.mpcd.geometry import Geometry
+
+
+class BounceBack(Method):
+    r"""Velocity Verlet integration method with bounce-back rule for surfaces.
+
+    Args:
+        filter (hoomd.filter.filter_like): Subset of particles on which to
+            apply this method.
+        geometry (hoomd.mpcd.geometry.Geometry): Surface to bounce back from.
+
+    A bounce-back method for integrating solutes (MD particles) embedded in
+    an MPCD solvent. The integration scheme is velocity Verlet with bounce-back
+    performed at the solid boundaries defined by a geometry, as in
+    `hoomd.mpcd.stream.BounceBack`. This gives a simple approximation of the
+    interactions required to keep a solute bounded in a geometry, and more complex
+    interactions can be specified, for example, by writing custom external fields.
+
+    Similar caveats apply to these methods as for `hoomd.mpcd.stream.BounceBack`.
+    In particular:
+
+    1. The simulation box is periodic, but the `geometry` may impose non-periodic
+       boundary conditions. You must ensure that the box is sufficiently large to
+       enclose the `geometry` and that all particles lie inside it, or an error will
+       be raised at runtime.
+    2. You must also ensure that particles do not self-interact through the periodic
+       boundaries. This is usually achieved for simple pair potentials by padding
+       the box size by the largest cutoff radius. Failure to do so may result in
+       unphysical interactions.
+    3. Bounce-back rules do not always enforce no-slip conditions at surfaces
+       properly. It may still be necessary to add additional "ghost" MD particles in
+       the surface to achieve the right boundary conditions and reduce density
+       fluctuations.
+
+    Warning:
+
+        This method does not support anisotropic integration because
+        torques are not computed for collisions with the boundary.
+        Rigid bodies will also not be treated correctly because the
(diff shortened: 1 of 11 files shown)


mphoward:
Not stale

mphoward:
This one should be good to go now! It is pretty small and mimics the one for `mpcd.stream`.

hoomd/mpcd/methods.py
```diff
@@ -0,0 +1,107 @@
...
+        group = sim.state._get_group(self.filter)
+        self._cpp_obj = class_(sim.state._cpp_sys_def, group,
+                               self.geometry._cpp_obj)
+        super()._attach_hook()
+
+    def _detach_hook(self):
+        self.geometry._detach()
+        super()._detach_hook()
+
+    _class_map = {}
```
b-butler: Bring to top of class definition. Class variables at top make it easier to read the code

hoomd/mpcd/methods.py
```diff
@@ -0,0 +1,107 @@
...
+        geom_type = type(self.geometry)
+        try:
+            class_info = self._class_map[geom_type]
+        except KeyError:
+            class_info = (_mpcd, "BounceBackNVE" + geom_type.__name__)
+        class_info = list(class_info)
+        if isinstance(sim.device, hoomd.device.GPU):
+            class_info[1] += "GPU"
+        class_ = getattr(*class_info, None)
+        assert class_ is not None, "Bounce back method for geometry not found"
```
b-butler, line 94: Shouldn't we raise the error every time this occurs, or is there a reason to limit it to running Python with asserts on?

hoomd/mpcd/methods.py
```diff
@@ -0,0 +1,107 @@
...
+        super()._attach_hook()
+
+    def _detach_hook(self):
+        self.geometry._detach()
+        super()._detach_hook()
+
+    _class_map = {}
+
+    @classmethod
+    def _register_geometry(cls, geometry, module, class_name):
```
b-butler: Don't see need to store `class_name` as this can be and is in some cases easily grabbed in `_attach_hook`.

hoomd/mpcd/methods.py
```diff
@@ -0,0 +1,107 @@
...
+        sim = self._simulation
+
+        self.geometry._attach(sim)
+
+        # try to find class in map, otherwise default to internal MPCD module
+        geom_type = type(self.geometry)
+        try:
+            class_info = self._class_map[geom_type]
+        except KeyError:
+            class_info = (_mpcd, "BounceBackNVE" + geom_type.__name__)
```
b-butler, line 89: Do we anticipate many cases like this? What is the argument not to require registration?

b-butler (commented):
Generally looks good. I just have a few comments.

Please add a test that checks that `BounceBack` can be used in a standard MD simulation as documented.

hoomd/mpcd/methods.py
mphoward: My thinking was that this error should only be raised for developers because they would need to implement and register the class. But, I can switch it to a regular exception if you think that's better.

hoomd/mpcd/methods.py
mphoward: I think my reasoning here was that `class_name` refers to the C++ class name, which may not necessarily have the same name as the Python one. I renamed this to `_register_cpp_class` on a subsequent branch to make that intent clearer.

If I remember right, I was also thinking that for this to work with plugins and not be strict about how things are imported, it was necessary to have the module and string name of the C++ class for use in `getattr`.

hoomd/mpcd/methods.py
mphoward: This is the default naming scheme, so I would expect it for most of the things implemented in HOOMD rather than a plugin. I used a similar pattern as this in `mpcd.stream`, and it is really helpful on PR #1712 because that adds a force class that the C++ code is also templated on, so we would have to explicitly register all combinations of geometries and forces.

hoomd/mpcd/methods.py
joaander: We could rename to `cpp_class_name` to make this clear.

joaander (approved):
Thanks!

Feel free to implement any of @b-butler's suggestions for improvement, then we can merge.

Looking at the limited cases where `BounceBack` works, I say we leave it in MPCD for now. Users can always apply it in MD only simulations if desired. We can consider moving it to `md` if/when `BounceBack` is expanded with significant new features.

hoomd/mpcd/methods.py
mphoward: I am going to leave this as an assert because it is a developer error, and I couldn't pick a more meaningful built-in exception to raise instead.

hoomd/mpcd/methods.py
mphoward: I'm going to leave this as is because it is the default naming scheme.

mphoward:
> Feel free to implement any of @b-butler's suggestions for improvement, then we can merge.
> 
> Looking at the limited cases where `BounceBack` works, I say we leave it in MPCD for now. Users can always apply it in MD only simulations if desired. We can consider moving it to `md` if/when `BounceBack` is expanded with significant new features.

Sounds good! I implemented the suggestions, so this is ready to merge.

PR #1696, opened 2024-01-15
Fix shapespace retreat

janbridley:
## Description

Properly call `retreat` for shape moves executed with the `hoomd.hpmc.updater.Shape` updater.

## Motivation and context

The `hoomd.hpmc.updater.Shape` updater was not rejecting shape alchemy over properly, resulting in `hoomd.hpmc.shape_move` methods would return particle geometries that did not correctly map back to the current value of the shape alchemy parameter(s). This means that the values would fluctuate into a region where every move was rejected, preventing progress from being made.

The tests added by this PR caught a new error only present when attempting to run shape updaters on GPU systems. For updaters that can change the memory footprint of the shape (e.g. updating the number of vertices with a ShapeSpace updater), the `managed` flag must be passed.

## How has this been tested?

The previous `test_python_callback_shape_move` method has been expanded on (and renamed to `test_python_callback_shape_move_ellipsoid`), and an additional `test_python_callback_shape_move_pyramid` has been added to test the feature with faceted geometries. Previous pytests have been updated to properly function with MPI parallelization.

More complex tests (running ~50_000 shape move trials) yield geometries that can be recreated from the logged alchemy parameter. Pytests run on GPU systems.

## Change log

```
- `hoomd.hpmc.updater.Shape` now properly restores shape alchemy parameters on rejected trial moves.
- `hoomd.hpmc.updater.Shape` now functions with `hoomd.device.GPU`
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/pytest/test_shape_updater.py
+++ b/hoomd/hpmc/pytest/test_shape_updater.py
@@ -222,7 +222,7 @@ def test_python_callback_shape_move_ellipsoid(simulation_factory,
 
     # run with 0 probability of performing a move:
     #  - shape and params should remain unchanged
-    #  - all moves accepted
+    #  - no shape moves proposed
     move.param_move_probability = 0
     sim.run(10)
     assert np.allclose(mc.shape["A"]["a"], ellipsoid["a"])
@@ -234,8 +234,8 @@ def test_python_callback_shape_move_ellipsoid(simulation_factory,
     #  - shape and params should change
     #  - volume should remain unchanged
     move.param_move_probability = 1
-    sim.run(20)
-    assert np.sum(updater.shape_moves) == 40
+    sim.run(50)
+    assert np.sum(updater.shape_moves) == 100
 
     # Check that the shape parameters have changed
     assert not np.allclose(mc.shape["A"]["a"], ellipsoid["a"])
@@ -321,8 +321,8 @@ def test_python_callback_shape_move_pyramid(simulation_factory,
     #  - shape and params should change
     #  - volume should remain unchanged
     move.param_move_probability = 1
-    sim.run(20)
-    assert np.sum(updater.shape_moves) == 40
+    sim.run(50)
+    assert np.sum(updater.shape_moves) == 100
 
     # Check that the shape parameters have changed
     current_h = move.params["A"][0]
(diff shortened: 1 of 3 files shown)


hoomd/hpmc/pytest/test_shape_updater.py
```diff
@@ -234,15 +234,109 @@ def __call__(self, type_id, param_list):
...
+    assert np.allclose(move.params["A"], [
+        mc.shape["A"]["a"] / mc.shape["A"]["b"],
+        mc.shape["A"]["a"] / mc.shape["A"]["c"]
+    ])
+
+    # Check that the callback is conserving volume properly
     assert np.allclose(updater.particle_volumes, 4 * np.pi / 3)
 
 
+def test_python_callback_shape_move_pyramid(simulation_factory,
```
tommy-waltmann, line 256: To make sure this actually tests the new line of code that was added, we need to make sure the parameter values aren't updated when the trial shape move creates overlaps between particles. I don't believe this test does that right now.

tommy-waltmann (changes requested):
Awesome work on this so far! I'm reviewing this even though I wasn't requested because this big fix will help code for one of my projects too.

hoomd/hpmc/pytest/test_shape_updater.py
janbridley: The test now fails if the overlap retreat is not implemented. Let me know if you can reproduce these results, and see [3aa5f41](https://github.com/glotzerlab/hoomd-blue/pull/1696/commits/3aa5f41818e84826d75f5f7feadd4eff7b35ef42) for the fix

tommy-waltmann: I am not able to reproduce that behavior with a distance of 3.0. I can reproduce it at a distance of 2.5 though.

janbridley: On reverification I have the same behavior - this is fixed in [70f0472](https://github.com/glotzerlab/hoomd-blue/pull/1696/commits/70f0472ad7840b4e28acc8e3859bbaf25b88640f)

tommy-waltmann (approved):
Looks good, looking forward to having this fixed so alchemy will work right again!

joaander (approved):
Thanks!

PR #1709, opened 2024-01-29
Fixed QuickCompress logic when changing box tilts

### LINKED ISSUE #1700: Negative tilt factors yield incorrect QuickCompress results
### Description

Running QuickCompress on a simulation state where any one (or multiple) tilt factors `xy`, `yz`, `xz` are less than 0 causes the tilt factors to change incorrectly. 

- Specifying a target box with the same (negative) tilt factor changes the current tilt factor by some amount `dx`
- Specifying a target box with any other (negative, positive, or zero) tilt factor changes the tilt factor by the same about `dx` (all else being constant)

Compression *should* work regardless of the sign of the tilt factors, but it seems like the negative sign causes an error: `UpdaterQuickCompress.cc::scaleValue()` calls `std::min(target, current/proposed_move_size)`. If current is less than 1, the scaled value will be the minimum and the incorrect box move will be accepted. 

I will implement a fix for this!

### Script

```python
import hoomd
import gsd.hoomd

sim = hoomd.Simulation(device=hoomd.device.CPU(), seed=1234)
sim.operations.integrator = hoomd.hpmc.integrate.Sphere()
sim.operations.integrator.shape["A"] = dict(diameter=1.0)

# Set up a simulation with a noncubic box and a negative tilt factor
frame = gsd.hoomd.Frame()
frame.particles.N = 1
frame.particles.position = [[0,0,0]]
frame.particles.types = ["A"]
frame.configuration.box = [3, 3, 3, -0.5, 0, 0]

sim.create_state_from_snapshot(frame)

final_box = hoomd.Box.from_box(sim.state.box)
final_box.volume = 0.5
assert(sim.state.box.xy==final_box.xy) # Shape should not change

qc = hoomd.hpmc.update.QuickCompress(trigger=1,target_box=final_box)
sim.operations.updaters.append(qc)

sim.run(1)
print(sim.state.box.xy,final_box.xy)


# Alternatively, trying to compress the box to its positive variant also fails
## final_box = hoomd.Box.from_matrix([[3,1.5,0],[0,3,0],[0,0,3]])
## final_box.volume = 0.5
## print(sim.state.box.xy, final_box.xy) # Shape should change in this example

## qc = hoomd.hpmc.update.QuickCompress(trigger=1,target_box=final_box)
## sim.operations.updaters.append(qc)

## sim.run(1)
## print(sim.state.box.xy,final_box.xy)
```

### Output

```shell
`>> -0.5010258004792931 -0.5`
```

### Expected output

`>> -0.5 -0.5` # The tilt factors should match

### Platform

CPU, macOS

### Installation method

Compiled from source

### HOOMD-blue version

4.4.1

### Python version

3.11.5

janbridley:
## Description

`QuickCompress` sometimes fails when reshaping a box, especially when tilt factors are less than or equal to zero. The additional logic provided in this PR addresses this issue.

## Motivation and context

Prior to this change, attempting to use `QuickCompress` in a few uncommon scenarios would result in incorrect behavior. The following box tilt factor changes now function properly:

`[0,0,0]` → `[xy,xz,yz]!=0`
`[xy,xz,yz]!=0` → `[0,0,0]`

`[xy,xz,yz] < 0` → `[xy,xz,yz]`
`[xy,xz,yz]`  → `[xy,xz,yz] < 0`

Resolves #1700 

## How has this been tested?

Additional testing has been added through in `test_quick_compress::test_quick_compression_triclinic`. This is a validate flagged test, as it tests many cases.

## Change log

```
`QuickCompress` can now reshape boxes with tilt factors <= 0.
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/pytest/test_quick_compress.py
+++ b/hoomd/hpmc/pytest/test_quick_compress.py
@@ -118,7 +118,7 @@ def test_valid_setattr_attached(attr, value, simulation_factory,
 def test_sphere_compression_triclinic(xy, xz, yz, phi, simulation_factory,
                                       lattice_snapshot_factory):
     """Test that QuickCompress can resize and reshape triclinic boxes."""
-    n = 3
+    n = (3, 3, 4)
     snap = lattice_snapshot_factory(n=n, a=1.1)
     snap.configuration.box = hoomd.Box.from_box([10, 9, 8, xy, xz, yz])
 
@@ -127,7 +127,7 @@ def test_sphere_compression_triclinic(xy, xz, yz, phi, simulation_factory,
     target_box = hoomd.Box.from_box([0.95, 1.05, 1, *tilts])
 
     v_particle = 4 / 3 * math.pi * (0.5)**3
-    target_box.volume = n**3 * v_particle / phi
+    target_box.volume = np.prod(n) * v_particle / phi
 
     qc = hoomd.hpmc.update.QuickCompress(trigger=hoomd.trigger.Periodic(25),
                                          target_box=target_box)
--- a/hoomd/hpmc/pytest/test_quick_compress.py
+++ b/hoomd/hpmc/pytest/test_quick_compress.py
@@ -116,9 +116,10 @@ def test_valid_setattr_attached(attr, value, simulation_factory,
 @pytest.mark.parametrize("phi", [0.01, 0.4, 0.6])
 @pytest.mark.validate
 def test_sphere_compression_triclinic(xy, xz, yz, phi, simulation_factory,
-                                      lattice_snapshot_factory):
+                                      lattice_snapshot_factory, device):
     """Test that QuickCompress can resize and reshape triclinic boxes."""
-    n = (5, 5, 5)
+    n = 6 if device.communicator.num_ranks > 1 else 3
+
     snap = lattice_snapshot_factory(n=n, a=1.1)
     snap.configuration.box = hoomd.Box.from_box([10, 9, 8, xy, xz, yz])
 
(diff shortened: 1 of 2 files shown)


janbridley:
~It seems like the new test is failing on MPI.~

I have added additional checks to the new test that increase the system size for tests on MPI or with GPU's enabled. The GPU test takes approximately 4 minutes on an NVIDIA A5000, and the MPI test is also on the order of a few minutes. The serial CPU test is significantly smaller, so it runs in under a minute.

hoomd/hpmc/pytest/test_quick_compress.py
```diff
@@ -105,6 +110,51 @@ def test_valid_setattr_attached(attr, value, simulation_factory,
     assert getattr(qc, attr) == value
 
 
+@pytest.mark.parametrize("xy", [-0.4, 0, 0.2])
+@pytest.mark.parametrize("xz", [-0.3, 0, 0.3])
+@pytest.mark.parametrize("yz", [-0.2, 0, 0.4])
+@pytest.mark.parametrize("phi", [0.01, 0.4, 0.6])
```
joaander, line 116:
```suggestion
@pytest.mark.parametrize("phi", [0.01, 0.4, 0.6])
pytest.mark.cpu
```
I know you got this running on the GPU - but it significantly increases the test run time. As the modified code runs on the CPU, I don't think we learn much by running the test on the GPU.

joaander (approved):
Thanks for condensing this down to the minimum changes needed. I have one change suggestion which I will apply and then this is good to merge.

PR #1711, opened 2024-01-30
Consider Walls when inserting or removing particles for the MuVT updater

Tobias-Dwyer:
## Description

Changes to the MuVT updater to calculate the energy of a particle insertion or removal for both hard walls and CPP potential field. 

Required implementation of an energy function for hard walls that checks if an inserted particle overlaps with the walls, and returns infinity. 

Required adding energy contributed by the wall to change in energy from inserting or removing a particle from the system. 

## Motivation and context

This solves the issue that hoomd does not consider walls when adding or removing particles in MuVT. Before this bug fix this means that when walls are added in a MuVT system, particles can be inserted intside the walls of the system where insertions should not be allowed or be otherwise highly energetically unfavoable in the case of jit. 

## How has this been tested?

I indepentently tested each hard wall geometry available in hoomd (sphere, cylinder, and plane) by starting with a single sphere inside the wall and testing whether particles would only be inserted inside the cylinder wall. I also test whether insertion and removal moves are performed. Finally, I test at a high fugacity meaning that more that just the starting particle should be inserted within the wall. I checked that these tests passed for the bug fix and didn't pass without them. 

To test the jit funtionality of the code I built the code with LLVM enabled on cheme-hodges. I then build a box with an energy constraint that any particle in the bottom half of the box would return an energy of Infinity. I then started the simulation with a single particle in the lower half of the box and a high fugacity. If hoomd considers the jit walls, the particle in the bottom part of the box will be removed and particles will only be inserted into the top of the box, which is what I tested for: At least one successful accepted removal move, no particles in the bottom half of the box, and particles inserted into the top half of the box. 

## Change log

```
MuVT ensembles considers walls when inserting or removing particles. 
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/ExternalField.h
+++ b/hoomd/hpmc/ExternalField.h
@@ -39,6 +39,25 @@ class ExternalField : public Compute
         return 0;
         }
 
+    //! Evaluate the energy of the force.
+    /*! \param box The system box.
+        \param type Particle type.
+        \param r_i Particle position
+        \param q_i Particle orientation.
+        \param diameter Particle diameter.
+        \param charge Particle charge.
+        \returns Energy due to the force
+    */
+    virtual float energy(const BoxDim& box,
+                         unsigned int type,
+                         const vec3<Scalar>& r_i,
+                         const quat<Scalar>& q_i,
+                         Scalar diameter,
+                         Scalar charge)
+        {
+        return 0;
+        }
+
     virtual bool hasVolume()
         {
         return false;
@@ -88,6 +107,7 @@ template<class Shape> void export_ExternalFieldInterface(pybind11::module& m, st
         .def(pybind11::init<std::shared_ptr<SystemDefinition>>())
         .def("compute", &ExternalFieldMono<Shape>::compute)
         .def("energydiff", &ExternalFieldMono<Shape>::energydiff)
+        .def("energy", &ExternalFieldMono<Shape>::energy)
         .def("calculateDeltaE", &ExternalFieldMono<Shape>::calculateDeltaE);
     }
 
--- a/hoomd/hpmc/ExternalFieldWall.h
+++ b/hoomd/hpmc/ExternalFieldWall.h
@@ -629,6 +629,56 @@ template<class Shape> class ExternalFieldWall : public ExternalFieldMono<Shape>
         return double(0.0);
         }
 
+    //! Evaluate the energy of the force.
+    /*! \param box The system box.
+        \param type Particle type.
+        \param r_i Particle position
+        \param q_i Particle orientation.
+        \param diameter Particle diameter.
+        \param charge Particle charge.
+        \returns Energy due to the force
+    */
+    virtual float energy(const BoxDim& box,
+                         unsigned int type,
+                         const vec3<Scalar>& r_i,
+                         const quat<Scalar>& q_i,
+                         Scalar diameter,
+                         Scalar charge)
+        {
+        const std::vector<typename Shape::param_type,
+                          hoomd::detail::managed_allocator<typename Shape::param_type>>& params
+            = m_mc->getParams();
+        Shape shape(q_i, params[type]);
+        vec3<Scalar> origin(m_pdata->getOrigin());
+
+        for (size_t i = 0; i < m_Spheres.size(); i++)
+            {
+            if (!test_confined(m_Spheres[i], shape, r_i, origin, box))
+                {
+                return INFINITY;
+                }
+            }
+
+        for (size_t i = 0; i < m_Cylinders.size(); i++)
+            {
+            set_cylinder_wall_verts(m_Cylinders[i], shape);
+            if (!test_confined(m_Cylinders[i], shape, r_i, origin, box))
+                {
+                return INFINITY;
+                }
+            }
+
+        for (size_t i = 0; i < m_Planes.size(); i++)
+            {
+            if (!test_confined(m_Planes[i], shape, r_i, origin, box))
+                {
+                return INFINITY;
+                }
+            }
+
+        return double(0.0);
+        }
+
     double calculateDeltaE(uint64_t timestep,
                            const Scalar4* const position_old,
                            const Scalar4* const orientation_old,
--- a/hoomd/hpmc/PatchEnergyJIT.cc
+++ b/hoomd/hpmc/PatchEnergyJIT.cc
@@ -69,7 +69,6 @@ void export_PatchEnergyJIT(pybind11::module& m)
                             Scalar,
                             pybind11::array_t<float>>())
         .def_property("r_cut", &PatchEnergyJIT::getRCut, &PatchEnergyJIT::setRCut)
-        .def("energy", &PatchEnergyJIT::energy)
         .def_property_readonly("param_array", &PatchEnergyJIT::getParamArray);
     }
 
--- a/hoomd/hpmc/PatchEnergyJIT.cc
+++ b/hoomd/hpmc/PatchEnergyJIT.cc
@@ -69,6 +69,7 @@ void export_PatchEnergyJIT(pybind11::module& m)
                             Scalar,
                             pybind11::array_t<float>>())
         .def_property("r_cut", &PatchEnergyJIT::getRCut, &PatchEnergyJIT::setRCut)
+        .def("energy", &PatchEnergyJIT::energy)
         .def_property_readonly("param_array", &PatchEnergyJIT::getParamArray);
     }
 
--- a/hoomd/hpmc/UpdaterMuVT.h
+++ b/hoomd/hpmc/UpdaterMuVT.h
@@ -1650,27 +1650,31 @@ bool UpdaterMuVT<Shape>::tryRemoveParticle(uint64_t timestep, unsigned int tag, 
         // do we have to compute energetic contribution?
         auto patch = m_mc->getPatchEnergy();
 
-     	// do we have to compute a wall contribution?
-    	auto field = m_mc->getExternalField();
+        // do we have to compute a wall contribution?
+        auto field = m_mc->getExternalField();
+        unsigned int p = m_exec_conf->getPartition() % m_npartition;
 
-     	if (field)
-     	    {
-     	   // getPosition() takes into account grid shift, correct for that
+        if (field && (!m_gibbs || p == 0))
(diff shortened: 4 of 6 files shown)


hoomd/hpmc/ExternalFieldWall.h
```diff
@@ -629,6 +629,56 @@ template<class Shape> class ExternalFieldWall : public ExternalFieldMono<Shape>
...
+    //! Evaluate the energy of the force.
+    /*! \param box The system box.
+        \param type Particle type.
+        \param r_i Particle position
+        \param q_i Particle orientation.
+        \param diameter Particle diameter.
+        \param charge Particle charge.
+        \returns Energy due to the force
+    */
+    virtual float energy(const BoxDim& box,
```
tommy-waltmann, line 641:
```suggestion
    virtual Scalar energy(const BoxDim& box,
```

hoomd/hpmc/ExternalField.h
```diff
@@ -39,6 +39,25 @@ class ExternalField : public Compute
...
+    //! Evaluate the energy of the force.
+    /*! \param box The system box.
+        \param type Particle type.
+        \param r_i Particle position
+        \param q_i Particle orientation.
+        \param diameter Particle diameter.
+        \param charge Particle charge.
+        \returns Energy due to the force
+    */
+    virtual float energy(const BoxDim& box,
```
tommy-waltmann, line 51:
```suggestion
    virtual Scalar energy(const BoxDim& box,
```

hoomd/hpmc/ExternalFieldWall.h
```diff
@@ -629,6 +629,56 @@ template<class Shape> class ExternalFieldWall : public ExternalFieldMono<Shape>
...
+                          hoomd::detail::managed_allocator<typename Shape::param_type>>& params
+            = m_mc->getParams();
+        Shape shape(q_i, params[type]);
+        vec3<Scalar> origin(m_pdata->getOrigin());
+
+        for (size_t i = 0; i < m_Spheres.size(); i++)
+            {
+            if (!test_confined(m_Spheres[i], shape, r_i, origin, box))
+                {
+                return INFINITY;
```
tommy-waltmann, line 658:
```suggestion
                return std:numeric_limits<Scalar>::max();
```
This is a better way to make sure we're returning the max value for the right type.

hoomd/hpmc/ExternalFieldWall.h
```diff
@@ -629,6 +629,56 @@ template<class Shape> class ExternalFieldWall : public ExternalFieldMono<Shape>
...
+
+        for (size_t i = 0; i < m_Planes.size(); i++)
+            {
+            if (!test_confined(m_Planes[i], shape, r_i, origin, box))
+                {
+                return INFINITY;
+                }
+            }
+
+        return double(0.0);
```
tommy-waltmann, line 679:
```suggestion
        return Scalar(0.0);
```

hoomd/hpmc/PatchEnergyJIT.cc
```diff
@@ -69,7 +69,6 @@ void export_PatchEnergyJIT(pybind11::module& m)
                             Scalar,
                             pybind11::array_t<float>>())
         .def_property("r_cut", &PatchEnergyJIT::getRCut, &PatchEnergyJIT::setRCut)
-        .def("energy", &PatchEnergyJIT::energy)
```
tommy-waltmann: Why remove this?

hoomd/hpmc/UpdaterMuVT.h
```diff
@@ -1650,6 +1650,28 @@ bool UpdaterMuVT<Shape>::tryRemoveParticle(uint64_t timestep, unsigned int tag,
...
+     	    vec3<Scalar> pos(p);
+     	    const BoxDim box = this->m_pdata->getGlobalBox();
+     	    unsigned int type = this->m_pdata->getType(tag);
+     	    quat<Scalar> orientation(m_pdata->getOrientation(tag));
+     	    lnboltzmann += field->energy(box,
+     	                                 type,
+     	                                 pos,
+     	                                 quat<float>(orientation),
+     	                                 1.0, // diameter i
+     	                                 0.0 // charge i
```
tommy-waltmann: We should get the diameter and charge from the particle data rather than assuming their values.

hoomd/hpmc/UpdaterMuVT.h
```diff
@@ -1945,6 +1971,18 @@ bool UpdaterMuVT<Shape>::tryInsertParticle(uint64_t timestep,
...
 
+        if (field)
+            {
+            const BoxDim box = this->m_pdata->getGlobalBox();
+            lnboltzmann -= field->energy(box,
+                                         type,
+                                         pos,
+                                         quat<float>(orientation),
+                                         1.0, // diameter i
+                                         0.0 // charge i
```
tommy-waltmann: I am in general not a fan of hard-coding diameter and charge like this, however, all the rest of the code seems to follow this pattern so I guess we can leave it as is for now.

hoomd/hpmc/pytest/test_muvt.py
```diff
@@ -159,3 +159,144 @@ def test_insertion_removal(device, simulation_factory,
...
+                                    )
+
+    sphere_radius = 0.6 
+    mc = hoomd.hpmc.integrate.Sphere(default_d=0.0, default_a=0.0)
+    mc.shape['A'] = dict(diameter=2*sphere_radius)
+
+    # code returns infinity if the particle center goes past the center of the box
+
+    center_wall = '''
+        if (r_i.z < 0. || r_i.z > box.getL().z/2)
```
tommy-waltmann: Is the second part of this boolean expression necessary? Any z-position greater than L/2 should be outside the box.

hoomd/hpmc/pytest/test_muvt.py
```diff
@@ -159,3 +159,144 @@ def test_insertion_removal(device, simulation_factory,
...
+    # We should have successfully removed the high energy particle
+    assert muvt.remove_moves[0] > 0
+
+    # We should have successfully inserted only to the top of the box and removed the high energy particle
+    assert  numpy.min(pos[:,2]) >= 0. 
+
+    # We should have added more than one particle to the box
+    assert  len(pos) > 1 
+
+def test_plane_wall_insertion(device, simulation_factory,
```
tommy-waltmann, line 233: I like this test, however I would like to see this test and the following two combined into a single test which is parametrized on the type of the wall.

tommy-waltmann (commented):
Great start on this! I just have a few small comments and a quick question.

Has this code be verified to work on a "real" system?

hoomd/hpmc/ExternalFieldWall.h
joaander: We need the infinity value here, not the maximum representable Scalar value:
You can use std::numeric_limits to get infinity on the CPU: https://en.cppreference.com/w/cpp/types/numeric_limits/infinity

hoomd/hpmc/UpdaterMuVT.h
joaander: The inserted particle does not exist in the particle data. If you have a need to insert particles with non-default charges, feel free to make a PR that adds the inserted particle charge as a parameter. `diameter` is unused.

hoomd/hpmc/pytest/test_muvt.py
joaander: A trial move temporarily generates positions outside the box. For technical reaons, HPMC does not wrap the trial move until long after it is accepted. Tobias and I discussed this and I will consider wrapping positions before evaluating wall potentials when I refactor this code. For now, this check is needed.

hoomd/hpmc/UpdaterMuVT.h
tommy-waltmann: This is inside `tryRemoveParticle` so the diameter and charge do exist in the particle data. We can just leave the diameter as 1 if its unused, but I think it more rigorous to at least get the charge from the particle data.

joaander: Yes, `tryRemoveParticle` should pass the particle's diameter and charge.

hoomd/hpmc/ExternalFieldWall.h
Tobias-Dwyer: not sure what the benefit is of having this as a float or scalar but happy to commit if it makes sense / would appreciate short explan.

hoomd/hpmc/pytest/test_muvt.py
Tobias-Dwyer: Chatted with @joaander about this and we think we should keep the tests this way for ease of readability, time writing the test, and simplicity.

hoomd/hpmc/UpdaterMuVT.h
Tobias-Dwyer: Neither am I. However, for now, I think this should be updated when this ends up getting rewritten.

hoomd/hpmc/UpdaterMuVT.h
Tobias-Dwyer: agreed and updated

hoomd/hpmc/PatchEnergyJIT.cc
Tobias-Dwyer: not sure. added it back in and tests pass fine and it builds.

Tobias-Dwyer:
> Great start on this! I just have a few small comments and a quick question.
> 
> Has this code be verified to work on a "real" system?

Thanks for the compliments! :D This code has been verified on a real system: here's a snapshot of a system with wall in middle and on top. Previously particles would be inserted in the bottom of the box but now insertion moves only happen on the top! So looks like everything's working great. 
wall_trajectory

hoomd/hpmc/ExternalFieldWall.h
Tobias-Dwyer: so when I implemented the changes from float back to Scalar, I started getting a bunch of errors while building. I think that replacing float with scalar might be more complicated than at first glance. Since this is a bug fix with code that's going to be rewritten soon anyway I think it's fine to leave as float for now.

tommy-waltmann (approved):
Looks good now, thanks for making those changes and showing the real system snapshot!

PR #1712, opened 2024-01-31
Reimplement MPCD solvent forces

### LINKED ISSUE #663: MPCD and GPU polymorphism
## Description

The MPCD forces are currently implemented with runtime polymorphism. This works well on the CPU but is complicated on the GPU because it requires a special wrapper to construct/destroy the force object and also needs separable device compilation. @jglaser has pointed that this polymorphism does not work on AMD GPUs. Another difficulty is how plugins can be built against hoomd.mpcd with either new streaming geometries or new forces, since they no longer have access to the unlinked device objects. We should try to figure out a better solution to this problem.

## Proposed solution

Two possibilities:

1. Supply the MPCD CUDA objects unlinked as a static library (in addition to the MPCD shared library), and plugins can then complete the device link themselves. I tried unsuccessfully to get this to work in older CMake, but it may be easier with modern CMake. This doesn't fix the AMD issue, though.

2. Scrap the runtime polymorphism and template the streaming methods on the force in addition to the geometry. This will increase compile time and forces developers to template all geometries when they add a new force, but will certainly work.

There may also be changes required for how integrators, streaming methods, and forces work for the v3 API, so I'm not sure if one of these might 

## Developer

Yes, I can look into this when I refactor for v3.

mphoward:
## Description

This PR adds back the ability to apply a body force on the solvent. The force is a class that is attached to a streaming method.

This PR also removes the complicated machinery that was in place to do GPU polymorphism, which barely worked with CUDA and did not work with ROCm. Instead, classes are now templated.

This PR should be merged after #1691.

## Motivation and context

Solvent forces are an important feature for people using MPCD. This also fixes a long-time issue with polymorphism that prevented hoomd.mpcd from compiling for AMD GPUs.

Resolves #663

## How has this been tested?

The force API and streaming method are tested at the Python level. The forces themselves are tested by a reimplemented C++ test.

## Change log

N/A, but it could be noted that `GPUPolymorph` was removed. As far as I know, it has only been used in one spot in the MPCD code, and it has some pretty severe limitations with compilation.

## Checklist:


mphoward:
Not stale

hoomd/mpcd/ConstantForce.cc
```diff
@@ -0,0 +1,39 @@
...
+            {
+                const auto F = force.getForce();
+                return pybind11::make_tuple(F.x, F.y, F.z);
+            },
+            [](ConstantForce& force, const pybind11::tuple& F)
+            {
+                force.setForce(make_scalar3(pybind11::cast<Scalar>(F[0]),
+                                            pybind11::cast<Scalar>(F[1]),
+                                            pybind11::cast<Scalar>(F[2])));
+            });
```
tommy-waltmann, line 34: I would prefer to see these defined in the `ConstantForce` class than in the export function if possible.

hoomd/mpcd/pytest/test_stream.py
```diff
@@ -65,6 +65,58 @@ def test_pickling(self, simulation_factory, snap, cls, init_args):
         sim.run(0)
         pickling_check(sm)
 
+    @pytest.mark.parametrize(
+        "force",
+        [
+            hoomd.mpcd.force.BlockForce(
+                force=2.0, half_separation=3.0, half_width=0.5),
+            hoomd.mpcd.force.ConstantForce(force=(1, -2, 3)),
+            hoomd.mpcd.force.SineForce(amplitude=2.0, wavenumber=1),
```
tommy-waltmann, line 75: Should a case where no force is added to the streaming method be included for completeness?

tommy-waltmann (commented):
Overall, looks to be in good shape. I just have two small things I want to mention.

hoomd/mpcd/CMakeLists.txt
```diff
@@ -51,94 +48,143 @@ set(_mpcd_headers
...
+    ParallelPlateGeometry
+    PlanarPoreGeometry
+    )
+foreach(_force ${_forces})
+    # append force itself first
+    list(APPEND _mpcd_cc_sources ${_force}.cc)
+    list(APPEND _mpcd_headers ${_force}.h)
+
+    # bulk geometry is special
+    configure_file(
```
joaander, line 120: Thanks for following this pattern! It really helps with compile times.

joaander (approved):
Thanks! @tommy-waltmann has a few suggestions for improvement if you would like to implement before merging.

Have you tested mpcd on an AMD GPU?

rayasare (approved):

hoomd/mpcd/CMakeLists.txt
mphoward: Yes definitely, and it is also way more convenient for generating the boilerplate code for the different templates!

hoomd/mpcd/ConstantForce.cc
mphoward: Since these are short functions that are only used for binding and would need to be conditionally compiled in the class, I prefer to keep them here.

hoomd/mpcd/pytest/test_stream.py
mphoward: I think this case is covered by `test_pickling` because `solvent_force=None` is the default, but it doesn't hurt to explicitly test with `None` here as well. Added!

mphoward:
I applied Tommy's suggestion on the unit tests and replied on the other one, so this should be ready to merge!
 
> Have you tested mpcd on an AMD GPU?

I unfortunately don't have access to an AMD GPU. I removed Jens' comment about MPCD not working with AMD GPUs because I remember that issue was related to device polymorphism, which is removed in this PR. (Jens had added CMake code to stop compilation of MPCD for AMD with HIP.)

tommy-waltmann (approved):

mphoward:
@joaander unless you want to hold off due to the comment about AMD above, could you merge this when you get a chance? Thanks!

joaander:
I will build the mpcd-v4 branch and run tests on Frontier when I get a chance. I would like to do that before merging mpcd-v4 into a trunk branch.

mphoward:
That would be great, thanks! If possible, let's do that as the final step before merging as there may be one or two more C++ refactorings if I have time.

I'll add this to the checklist in #775 .

Issue #772, opened 2020-09-04
Relax MPCD virtual particle filler design


mphoward:
## Description

The virtual particle filler specializations currently need to compute the total number of particles to add so that the particle arrays are sized by the base class. It would be helpful for geometries where this number is not easily calculated (e.g., in MPI with domains slicing through boundaries) to allow the deriving filler to have more control over how it determines the number of virtual particles (e.g., random sampling with rejection). This should be a small change at the C++ level but tweaks the API of the fillers a little bit. I think this should go into v3 breaking changes for MPCD.

I will work on this.

mphoward:
This is being worked on by someone in a fork:

https://github.com/YMWani/hoomd-blue/tree/feature/mpcd-sphere

That branch will need updates to v3 when it is ready, but I will handle that as part of the other v3 migration tasks.

mphoward:
We will need to address this as part of our new work.

mphoward:
Resolved in #1737

PR #1714, opened 2024-02-01
Rotate MPCD geometry and force coordinates

### LINKED ISSUE #662: Rotate MPCD coordinate system
## Description

For legacy reasons from another implementation, many of the geometries and features in MPCD are defined in the _xz_ plane, including `mpcd.stream.slit`, `mpcd.stream.slit_pore`, `mpcd.integrate.slit`, `mpcd.integrate.slit_pore`, and everything in `mpcd.force`. As pointed out by @j-proc in PR #443, it would be smarter to rotate everything into the _xy_ plane so that these features still work in 2d simulations, if such a collision rule is later implemented.

This is a breaking change that I didn't want to tackle before, but I think it would be smart to adjust in v3.0.

## Proposed solution

Swap out z coordinates for y coordinates everywhere in those classes. Alot of it should be localized to the geometry files.

## Developer

Yes, I can take care of it when I go through this code for v3.

mphoward:
## Description

This PR rotates the coordinate systems of some geometries and forces to be in the (x,y) plane rather than (x,z) plane. It mostly involved replacing "z" with "y" in a bunch of places.

While doing this refactoring, I removed the `const` ness of the `Geometry` and the internal validation of the geometry. There were a few reasons for this: `const` was clunky and couldn't really be enforce at the Python level, I have some future use cases where the `Geometry` needs to be able to change but it shouldn't be updated every time this happens, and actually validating the box for complicated geometries can be difficult. It should be up to the user to make sure they have chosen a `Geometry` that makes sense. However, I did expose a public method to Python that the user can call if they want to check that their particles are in the right spot.

## Motivation and context

As explained in #662, the use of (x,z) was a legacy choice from a previous code, but it prevented possible reuse of code for 2d MPCD simulations in HOOMD, which are in the (x,y) plane, if we ever wanted to be able to do this.

Resolves #662 

## How has this been tested?

Existing tests have been updated by exchanging y and z coordinates as needed.

## Change log

N/A

## Checklist:


mphoward:
Not stale

mphoward:
This PR is ready for review. It is the second to last C++ refactoring (one more will be completed related to #772), so getting it merged quickly would be great so that we can focus on finalizing the Python API where most of the user changes are happening.

joaander (approved):
Thanks! This looks good to me.

PR #1412, opened 2022-10-04
Add patchy pair potential

cbkerr:
## Description

Clean and publish code to simulate patchy particles, including janus particles, based on that used in [Phase behavior of Janus colloids determined by sedimentation equilibrium.
](https://pubs.rsc.org/en/content/articlelanding/2014/sm/c3sm53136h)

Using templates, the code can "modulate" any isotropic pair potential by an "envelope" that is created based on an angle dependence factor.

## TODOS remaining

* [x] Add support for patches not aligned with particle orientation.
* [x] Add tests in a new file
* [x] Check casting, line 112 of `GeneralEnvelope.h`
* [x] Confirm a negative sign relating to dot products, line 78 of `JanusFactor.h`
* [x] Confirm math in `PairModulator.h`, line 218
* [x] Add `type_shapes` in `aniso.py`.

## How has this been tested?

I wrote expressions in Mathematica and had it do the math to check the values used in the potential. That notebook will be attached to this PR.

<details>
<summary>Old nomenclature lookup table</summary>

Use this to translate when comparing old and new versions of the code.

<table id="orge7ef0a8" border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">

<colgroup>
<col  class="org-left" />

<col  class="org-left" />

<col  class="org-left" />
</colgroup>
<thead>
<tr>
<th scope="col" class="org-left">Kind of Entity</th>
<th scope="col" class="org-left">in Old Plugin</th>
<th scope="col" class="org-left">in New Code</th>
</tr>
</thead>

<tbody>
<tr>
<td class="org-left">thing created from template</td>
<td class="org-left">JanusDirectionalEvaluator</td>
<td class="org-left"><del>JanusEnvelope</del></td>
</tr>

<tr>
<td class="org-left"><b>Header file</b></td>
<td class="org-left">EvaluatorPairIsoModulated.h</td>
<td class="org-left">PairModulator.h</td>
</tr>

<tr>
<td class="org-left">template typename</td>
<td class="org-left">isoEval</td>
<td class="org-left">pairEvaluator</td>
</tr>

<tr>
<td class="org-left">variable</td>
<td class="org-left">iParam (&ldquo;isotropic&rdquo;)</td>
<td class="org-left">pairParam</td>
</tr>

<tr>
<td class="org-left">variable</td>
<td class="org-left">iP (for &ldquo;isotropic parameter&rdquo;), _iP</td>
<td class="org-left">pairP, _pairP</td>
</tr>

<tr>
[...]

PR #1728, opened 2024-02-26
Pair hpmc angular step potential 

### LINKED ISSUE #1652: Replace runtime CPPPotential potentials with compile time provided ones.

melodyyzh:
## Description
This PR adds angular step potential to the hpmc pair potential class. 

## Motivation and context
CPPotential will be replaced by the new class of pair potentials in hoomd-blue in the version 4.6 release with the removal of llvm's JIT engine. AngularStep potential is a type of pair potential that includes an isotropic component (e.g., LJ) and an angular step function that describes the patch interaction. 

Partially Resolves #1652 

## How has this been tested?
In progress of writing validation tests. 

## Change log

```
Added:
* `pair.AngularStep` to replace `CPPPotential` for patch interactions.
```

## Checklist:


joaander:
Validation tests run on 853ba7c merged with the step potential branch: https://github.com/glotzerlab/hoomd-validation/pull/89

patchy_particle_pressure_compare_density0 95_temperature1 0_pressure10 209_chi0 7_4096particles_epsilonrepulsive1 0

hoomd/hpmc/pair/angular_step.py
```diff
@@ -0,0 +1,134 @@
...
+import hoomd
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeparam import TypeParameter
+from hoomd.data.typeconverter import OnlyIf, to_type_converter
+
+from .pair import Pair
+
+
+@hoomd.logging.modify_namespace(('hpmc', 'pair', 'AngularStep'))
+class AngularStep(Pair):
```
tommy-waltmann, line 26: General question on the docs: Do we want to use the language of patchy particles throughout the documentation here, or do we want to describe this potential as one where the radial component is only defined for certain particle orientations?

In the union potentials, we stayed away from "patchy particle" language, instead choosing to describe the potential as a union of many potentials rather than a patchy particle.

hoomd/hpmc/PairPotentialAngularStep.cc
```diff
@@ -0,0 +1,154 @@
...
+        {
+        pybind11::tuple director_python = directors[i];
+        if (pybind11::len(director_python) != 3)
+            {
+            throw std::length_error("director must be a list of 3-tuples.");
+            }
+        m_directors[particle_type_id][i] = vec3<LongReal>(director_python[0].cast<LongReal>(),
+                                                          director_python[1].cast<LongReal>(),
+                                                          director_python[2].cast<LongReal>());
+        // TO DO: normalize m_director to be unit length
```
tommy-waltmann: Is this still needed?

hoomd/hpmc/pytest/test_pair_angular_step.py
```diff
@@ -0,0 +1,256 @@
...
+        hpmc.pair.AngularStep(LJ())
+
+
+@pytest.fixture(scope='function')
+def angular_step_potential(pair_potential):
+    return hpmc.pair.AngularStep(pair_potential)
+
+
+def _valid_particle_dicts():
+    valid_dicts = [
```
tommy-waltmann, line 42: I think at least one more case can be added to this list. For example, what if a user passes in arrays and not tuples?

tommy-waltmann (changes requested):

hoomd/hpmc/pair/angular_step.py
joaander: In this case, The multiplier is either 1 or 0. The "patches" do not sum so constituent is not the correct word. Do you have a suggestion for a better word in this sentence?

> The orientation masking function is 1 when any patch on particle i aligns with any patch on particle j.

hoomd/hpmc/PairPotentialAngularStep.h
```diff
@@ -0,0 +1,79 @@
...
+        @param r_ij Vector pointing from particle i to j.
+        @param type_i Integer type index of particle i.
+        @param charge_i Charge of particle i.
+        @param q_i Orientation quaternion of particle i.
+        @param type_j Integer type index of particle j.
+        @param q_j Orientation quaternion of particle j.
+        @param charge_j Charge of particle j.
+        @returns Energy of the pair interaction.
+    */
+
```
joaander:
```suggestion
```

hoomd/hpmc/PairPotentialAngularStep.h
```diff
@@ -0,0 +1,79 @@
...
+For use with HPMC simulations.
+*/
+class PairPotentialAngularStep : public hpmc::PairPotential
+    {
+    public:
+    PairPotentialAngularStep(std::shared_ptr<SystemDefinition> sysdef,
+                             std::shared_ptr<PairPotential> isotropic_potential);
+    virtual ~PairPotentialAngularStep() { }
+
+    void setPatch(std::string particle_type, pybind11::object v);
```
joaander:
```suggestion
    /// Set the patch parameters for a given particle type.
    void setPatch(std::string particle_type, pybind11::object v);
```

hoomd/hpmc/PairPotentialAngularStep.h
```diff
@@ -0,0 +1,79 @@
...
+
+    pybind11::object getPatch(std::string particle_type);
```
joaander:
```suggestion
    /// Get the patch parameters for a given particle type.
    pybind11::object getPatch(std::string particle_type);
```

hoomd/hpmc/pair/angular_step.py
tommy-waltmann: We could use the word "director"

joaander: If we use "director" for the object that has a direction and a width, then we need a new word for that direction (we are currently using director).
```
angular_step.directors['A'] = {'directors': [(1,0,0), (0,0,1)], deltas=[0.1, 0.2]}
```

hoomd/hpmc/PairPotentialAngularStep.h
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+    virtual LongReal computeRCutAdditive(unsigned int type) const
+        {
+        // Returns the additive part of the cutoff distance for a given type.
+        return m_isotropic_potential->computeRCutAdditive(type);
+        }
+
+    protected:
+    bool maskingFunction(LongReal r_squared,
```
joaander, line 63: `maskingFunction` should be defined in the `.h` file so the compiler can inline it and improve performance.

hoomd/hpmc/pytest/test_pair_angular_step.py
```diff
@@ -0,0 +1,256 @@
...
+    ),
+    (
+        dict(directors=[(1.0, 0, 0)], deltas=[0.1]),
+        0,
+        np.pi - 0.099,
+        2.0,  # < rcut
+        lj(2.0, 4.0, 1, 1),
+    ),
+    (
+        dict(directors=[(1.0, 0, 0)], deltas=[0.1]),
```
joaander, line 250: Please add some tests with two particle types that have different numbers of patches.
For example: 
```
patches['A'] = dict(directors=[(1.0, 0, 0)], deltas=[0.1])
patches['B'] = dict(directors=[(1.0, 0, 0), (0.0, 1.0, 0.0], deltas=[0.1, 0.2])
```
Then rotate the second particle through several angles and ensure that the energy is as expected.

hoomd/hpmc/PairPotentialAngularStep.cc
```diff
@@ -0,0 +1,154 @@
...
+                                               const quat<LongReal>& q_i,
+                                               const unsigned int type_j,
+                                               const quat<LongReal>& q_j) const
+    {
+    vec3<LongReal> rhat_ij = r_ij / fast::sqrt(r_squared);
+
+    for (int m = 0; m < m_directors[type_i].size(); m++)
+        {
+        LongReal cos_delta_m = cos(m_deltas[type_i][m]);
+        vec3<LongReal> ehat_m = rotate(q_i, m_directors[type_i][m]);
```
joaander: Pre-compute `cos_delta` in `setPatch`. The calls to `cos` in the inner loop are slowing performance.

hoomd/hpmc/PairPotentialAngularStep.cc
joaander: Yes, the patch director must be normalized.

hoomd/hpmc/pair/angular_step.py
melodyyzh: We used the word director to represent the vector pointing to the center of the patch, but we still need to consider how wide the patch is (i.e., half opening angle). Maybe we can use something like "angular range"?

joaander (changes requested):
Thanks!

I revised the documentation directly. My other suggestions are below.

I also benchmarked the code to confirm that `cos` and a non-inlined `maskingFunction` are indeed performance penalties:
```
  76.71%  python3  _hpmc.cpython-311-x86_64-linux-gnu.so              [.] hoomd::hpmc::IntegratorHPMCMono<hoomd::hpmc::ShapeSphere>::update(unsigne◆
   5.64%  python3  _hpmc.cpython-311-x86_64-linux-gnu.so              [.] hoomd::hpmc::PairPotentialAngularStep::maskingFunction(double, hoomd::vec▒
   2.86%  python3  libm.so.6                                          [.] __cos_fma                                                                ▒
   1.26%  python3  _hpmc.cpython-311-x86_64-linux-gnu.so              [.] void hoomd::hpmc::move_rotate<3u, hoomd::RandomGenerator>(hoomd::quat<dou▒
   1.08%  python3  _hoomd.cpython-311-x86_64-linux-gnu.so             [.] hoomd::SFCPackTuner::generateTraversalOrder(int, int, int, int, int, unsi▒
   0.97%  python3  _hpmc.cpython-311-x86_64-linux-gnu.so              [.] hoomd::hpmc::PairPotentialAngularStep::energy(double, hoomd::vec3<double>▒
   0.60%  python3  [unknown]                                         
```
`perf` reports these two methods as the top two after the main MC update method.

hoomd/hpmc/pair/angular_step.py
joaander: If it is not clear, I am happy with the word "patch". A given patch has a director and a half-width \delta.

I am open to suggestions if you can think of a better word that describes an object that decorates a particle and has both a direction and a width.

hoomd/hpmc/PairPotentialAngularStep.cc
melodyyzh: Nice catch, working on it

hoomd/hpmc/PairPotentialAngularStep.cc
```diff
@@ -0,0 +1,154 @@
...
+bool PairPotentialAngularStep::maskingFunction(LongReal r_squared,
+                                               const vec3<LongReal>& r_ij,
+                                               const unsigned int type_i,
+                                               const quat<LongReal>& q_i,
+                                               const unsigned int type_j,
+                                               const quat<LongReal>& q_j) const
+    {
+    vec3<LongReal> rhat_ij = r_ij / fast::sqrt(r_squared);
+
+    for (int m = 0; m < m_directors[type_i].size(); m++)
```
joaander:
```suggestion
    for (size_t m = 0; m < m_directors[type_i].size(); m++)
```

hoomd/hpmc/PairPotentialAngularStep.cc
```diff
@@ -0,0 +1,154 @@
...
+        {
+        LongReal cos_delta_m = cos(m_deltas[type_i][m]);
+        vec3<LongReal> ehat_m = rotate(q_i, m_directors[type_i][m]);
+
+        for (int n = 0; n < m_directors[type_j].size(); n++)
```
joaander:
```suggestion
        for (size_t n = 0; n < m_directors[type_j].size(); n++)
```

joaander (changes requested):

hoomd/hpmc/pair/angular_step.py
tommy-waltmann: How about envelope(s)? An opening envelope has a width and direction and would be consistent with the naming in #1412  .

I don't dislike "patch" as a name for the attribute, my original comment was more so related to how we word some parts of the documentation.

joaander: Envelope is an improvement. Although, let's avoid confusion with @cbkerr's MD patch potential where "envelope" is the word used to describe the shape of the mask as a function of the relative orientation.

@melodyyzh actually gives us the right word in the docs and in the C++ code: "mask".

Each _patch_ has a _director_ and _half-width_ (\delta). The _mask_ describes the distribution of patches on the particle's surface and the _masking function_ determines the interaction scale factor as a function of two interacting particle's _masks_.
```
angular_step.mask['A'] = {'directors': [(0,1,0), (1,0,0)], 'deltas': [0.1, 0.2]}
```

This is consistent with the rigid body / union potentials naming where each _constituent_ has a _position_, _orientation_, .... The _body_ describes the placement of the _constituent_ particles.
* _constituent_ <-> patch
* _body_ <-> _mask_

melodyyzh: Thanks, I made changes based on your suggestions.

hoomd/hpmc/PairPotentialAngularStep.cc
melodyyzh: Nice catch, I implemented the math to convert to unit vector

hoomd/hpmc/pytest/test_pair_angular_step.py
melodyyzh: I thought about adding a check for arrays/list originally, but in the `setMask` function in  `PairPotentialAngularStep.cc`, we pre-processed the director to be a tuple. Correct me if I am wrong here @joaander 
```
m_directors[particle_type_id][i] = vec3<LongReal>(director_python[0].cast<LongReal>(),
                                                          director_python[1].cast<LongReal>(),
                                                          director_python[2].cast<LongReal>());
```

hoomd/hpmc/pytest/test_pair_angular_step.py
melodyyzh: I am not sure if my changes match with what you referred to here. I created 2 test sets and 2 separated energy tests for the A-A only and A-B systems. let me know your suggestions.

hoomd/hpmc/PairPotentialAngularStep.h
```diff
@@ -0,0 +1,101 @@
...
+                    return true;
+                    }
+                }
+            }
+        return false;
+        }
+
+    /// Create a pointer to get the isotropic pair potential
+    std::shared_ptr<PairPotential> m_isotropic_potential;
+    // should this point to PairPotential or the LJ potential specifically?
```
tommy-waltmann:
```suggestion
```

hoomd/hpmc/pytest/test_pair_angular_step.py
tommy-waltmann: I disagree with the idea that arrays and lists should be considered invalid inputs, that's oftentimes the format users will want to give the data in. If they are considered invalid, then there should be cases in `test_invalid_particle_params` which make sure an exception is raised in those cases.

tommy-waltmann (changes requested):

hoomd/hpmc/pytest/test_pair_angular_step.py
melodyyzh: Sounds good, I added list and array format to `valid_params_dict`

joaander (approved):
Thanks!

melodyyzh:
pre-commit.ci autofix

tommy-waltmann (approved):

PR #1730, opened 2024-02-29
Fix cmake warnings.

joaander:
## Description

* Use `execute_process`.
* Use the new `FindPython` machinery when it is available.

Adjust CMake settings so that the new `FindPython` behaves similarly to the old. Especially in the case where Mac users expect their virtual environment Python to found before a system or homebrew installed one.

## Motivation and context

Address the following CMake warnings:
```
CMake Warning (dev) at /Users/joaander/miniforge3/share/cmake/pybind11/FindPythonLibsNew.cmake:98 (find_package):
  Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
  are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
  the cmake_policy command to set the policy and suppress this warning.

Call Stack (most recent call first):
  /Users/joaander/miniforge3/share/cmake/pybind11/pybind11Tools.cmake:50 (find_package)
  /Users/joaander/miniforge3/share/cmake/pybind11/pybind11Common.cmake:188 (include)
  /Users/joaander/miniforge3/share/cmake/pybind11/pybind11Config.cmake:250 (include)
  CMake/hoomd/HOOMDPythonSetup.cmake:2 (find_package)
  CMakeLists.txt:120 (include)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

```
CMake Warning (dev) at CMakeLists.txt:161 (exec_program):
  Policy CMP0153 is not set: The exec_program command should not be called.
  Run "cmake --help-policy CMP0153" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  Use execute_process() instead.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

## How has this been tested?

I tested these changes locally on my Mac. CI will check Linux builds.

## Change log

```
Changed:

* Use ``FindPython`` on modern CMake installations. You may need to adjust build scripts
  in cases where the new behavior does not exactly match the old (i.e. use 
  `-DPython_EXECUTABLE` in place of `-DPYTHON_EXECUTABLE`)
  (`#1730 <https://github.com/glotzerlab/hoomd-blue/pull/1730>`__).
* External components must switch from ``pybind11_add_module`` to ``hoomd_add_module``
  (`#1730 <https://github.com/glotzerlab/hoomd-blue/pull/1730>`__).
```

## Checklist:

## Todo:

- [ ] After release: update the CMakeLists.txt in `hoomd-component-template` to match.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,7 +158,7 @@ endif()
 
 #######################
 ## Get the compile date
-exec_program("date +%Y-%m-%d" OUTPUT_VARIABLE COMPILE_DATE)
+execute_process(COMMAND "date" "+%Y-%m-%d" OUTPUT_VARIABLE COMPILE_DATE)
 
 ################################
 # set up unit tests
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,12 +116,13 @@ include (HOOMDHIPSetup)
 
 # Find CUDA and set it up
 include (HOOMDCUDASetup)
-include (hoomd-macros)
 
 # setup MPI support
 include (HOOMDMPISetup)
 # find the python libraries to link to
 include(HOOMDPythonSetup)
+
+include (hoomd-macros)
 
 if (NOT ENABLE_HIP OR HIP_PLATFORM STREQUAL "nvcc")
     option(BUILD_MPCD "Build the mpcd package" on)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,10 @@
-CMAKE_MINIMUM_REQUIRED(VERSION 3.9 FATAL_ERROR)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.9...3.27 FATAL_ERROR)
 # >= 3.8 is required for CUDA language support
 # >= 3.9 is required for MPI::MPI_CXX target
+
+# Search for Python and other libraries in unix-like locations first and frameworks last.
+# This allows FindPython to find virtual environment Pythons before a homebrew or system Pythons.
+set(CMAKE_FIND_FRAMEWORK LAST)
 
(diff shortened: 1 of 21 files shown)


CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
-CMAKE_MINIMUM_REQUIRED(VERSION 3.9 FATAL_ERROR)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.9...3.27 FATAL_ERROR)
```
b-butler: Perhaps a comment on the version cap is warranted.

b-butler (approved):
Looks good. The version cap may be worth explaining in a code comment but otherwise :+1:

CMakeLists.txt
joaander: This is not a cap (even though it looks like one). This is stating that we support CMake 3.9 and accept the NEW policies introduced in all versions between 3.9 and 3.27. HOOMD will still configure with CMake >= 3.28, but will default to the OLD policies introduced in versions after 3.27.

Thew NEW policies in some newer versions make `FindPython` behave more like the old code. For example with the old policy CMP0094, CMake would prefer a homebrew installed Python 3.12 over a miniforge installed 3.11 - even when miniforge appears first on the PATH.

joaander:
This was more complex than I thought. The only reasonable way to solve the issue was to bump the minimum CMake to 3.15. This is fine following our stated policy as Ubuntu 20.04 provides CMake 3.16 via `apt-get`.

The new `pybind11_add_module` produces broken shared libraries for hoomd's non-standard usage. The only way to solve this problem was implement a `hoomd_add_module` function that replicates the old `pybind11_add_module` behavior with regards to NOT linking `libpython` to SHARED objects.

PR #1732, opened 2024-03-01
Step potential [HPMC].

joaander:
## Description

Add the `Step` potential.

## Motivation and context

Allow users to employ square well, square well with shoulder, and other similar potentials built on step functions.

Progress toward #1652.

## How has this been tested?

- [x] Unit tests
- [x] Validation tests

## Change log

```
Added:

* ``hoomd.hpmc.pair.Step`` - step function potential.
```

## Checklist:


hoomd/hpmc/pytest/test_pair_lennard_jones.py
```diff
@@ -76,7 +76,11 @@ def test_invalid_params_on_attach(mc_simulation_factory, parameters):
     simulation.operations.integrator.pair_potentials = [lennard_jones]
     simulation.run(0)
 
-    with pytest.raises(Exception):
+    with pytest.raises((
+            RuntimeError,
+            hoomd.error.TypeConversionError,
+            KeyError,
+    )):
```
tommy-waltmann, line 83: Was this test failling before this change?

hoomd/hpmc/PairPotential.h
```diff
@@ -134,7 +134,7 @@ class PairPotential
     /// Returns the additive part of the cutoff distance for a given type.
     virtual LongReal computeRCutAdditive(unsigned int type) const
         {
-        return m_r_cut_additive[type];
+        return 0;
```
tommy-waltmann, line 137: Why this change? Does it not impact other pair potentials as well?

hoomd/hpmc/pytest/test_pair_step.py
```diff
@@ -0,0 +1,173 @@
...
+
+valid_constructor_args = [
+    {},
+]
+
+
+@pytest.mark.parametrize("constructor_args", valid_constructor_args)
+def test_valid_construction(device, constructor_args):
+    """Test that Step can be constructed with valid arguments."""
+    hoomd.hpmc.pair.Step(**constructor_args)
```
tommy-waltmann: Is there a plan to add other constructor args at some point? If not, I don't think the fixture is necessary.

hoomd/hpmc/pytest/test_pair_step.py
```diff
@@ -0,0 +1,173 @@
...
+    (
+        dict(epsilon=[1, 2, 3], r=[0.5, 1.5, 2.5]),
+        1.6,
+        3,
+    ),
+    (
+        dict(epsilon=[1, 2, 3], r=[0.5, 1.5, 2.5]),
+        1.5,
+        2,
+    ),
```
tommy-waltmann: According to the formulas in the documentation, the energy should be 3 in this case. when r=r_k, the step function should return \epsilon_{k+1}.

hoomd/hpmc/pytest/test_pair_step.py
```diff
@@ -0,0 +1,173 @@
...
+    (
+        dict(epsilon=[1, 2, 3], r=[0.5, 1.5, 2.5]),
+        0.6,
+        2,
+    ),
+    (
+        dict(epsilon=[1, 2, 3], r=[0.5, 1.5, 2.5]),
+        0.5,
+        1,
+    ),
```
tommy-waltmann: Same as above comment; the step function should return 2 in this case.

hoomd/hpmc/PairPotentialStep.cc
```diff
@@ -0,0 +1,161 @@
...
+        {
+        return slow::sqrt(m_params[param_index].m_r_squared[n - 1]);
+        }
+    else
+        {
+        return 0;
+        }
+    }
+
+LongReal PairPotentialStep::energy(const LongReal r_squared,
```
tommy-waltmann, line 30: There may be some tricks we can play in this function to reduce average (not asymptotic) runtime.

For example, we could start by checking if the pair distance is larger or equal to the last value in `m_r_squared` before entering the while loop. I suspect this case will come up often, but it would take many iterations of the binary search to finally realize the distance value is outside the distance range.

We may also be able to lower average runtime by checking if the pair distance is less than the first value in `m_r_squared`, or just returning 0 immediately if `N == 0`.

hoomd/hpmc/pair/step.py
```diff
@@ -0,0 +1,72 @@
...
+"""
+
+import hoomd
+
+from .pair import Pair
+from hoomd.data.typeconverter import OnlyIf, to_type_converter
+
+
+@hoomd.logging.modify_namespace(('hpmc', 'pair', 'Step'))
+class Step(Pair):
```
tommy-waltmann, line 21: It may be helpful to add a note in the docs saying the r values need to be sorted by distance or else this will not work right.

tommy-waltmann (changes requested):
This looks mostly good, but I still have a few comments.

hoomd/hpmc/pytest/test_pair_lennard_jones.py
joaander: No, but I recognized that this needed to be changed after copying it.
`with pytest.raises(Exception):`
will *pass* for **ANY** exception (e.g. a divide by 0, passing the wrong number of arguments, etc..) and can give a false sense of the test passing when in reality the intended code is not even running.

hoomd/hpmc/PairPotential.h
joaander: This has no impact. It is a minor change reflecting that this method is "pure virtual" without going through the steps of making it actually pure virtual. The `m_r_cut_additive[type]` was a copy and paste error when this method was copied from `getRCutAdditive`.

hoomd/hpmc/PairPotentialStep.cc
joaander: The *caller* checks whether r is past the cutoff. This method should not be called with values larger than the cutoff. I include the 0 return to avoid out of bounds memory accesses in the rare case that the caller mistakenly does not check r_cut.

I will check on whether your other suggestions provide a measurable performance gain.

joaander: The early exit checks do provide a small but measurable benefit in the case where there is only 1 point in the potential. In the general case, the additional checks will slow the code down somewhat.

hoomd/hpmc/pair/step.py
joaander: We also need to validate this condition to prevent errors.

hoomd/hpmc/PairPotentialStep.cc
tommy-waltmann: This potential will frequently be used for square wells, so I think it will be worth it.

tommy-waltmann (approved):
The code looks good! Are there any validation tests planned for this before merging?

joaander:
Yes, I will execute validation tests before merging.

hoomd/hpmc/PairPotentialStep.cc
SchoeniPhlippsn: One more potential speed-up is to use an interpolating version of the binary search algorithm. In that case, you determine `m` according to the relation between `r_squared` and `param.m_r_squared[N-1]`: `size_t m = L*(1-r_squared/param.m_r_squared[N-1]) + R*r_squared/param.m_r_squared[N-1]`.
However, that might only help for large 'N' and when the distances are rather uniformly distributed.

SchoeniPhlippsn (commented):
Looks good, only have one suggestion for a potential with many steps.

hoomd/hpmc/PairPotentialStep.cc
joaander: I don't think the typical use-case of this will be a large number of points spaced linearly in `r**2`.

joaander:
Validation tests pass:
patchy_particle_pressure_compare_density0 6_temperature3 0_pressure2 125_chi1 0_4096particles_epsilonrepulsive-1 0
Tested in https://github.com/glotzerlab/hoomd-validation/pull/89

PR #1736, opened 2024-03-05
Enable slow compression with QuickCompress

### LINKED ISSUE #1641: Slow compression in HPMC simulations.
### Description

Add a method that slowly compresses in HPMC simulations.

### Proposed solution

Either implement `SlowCompress` or add optional arguments to `QuickCompress`. In either case, apply the `QuickCompress` algorithm to a variable box size.

### Additional context

`QuickCompress` works extremely well, but always quenches to the chosen box size. Some simulation workflows require slow annealing. MD simulations can use `BoxResize`. HPMC simulations need an alternate method that accounts for particle overlaps.

tcmoore3:
## Description

This PR changes the `target_box` member of `UpdaterQuickCompress` from a box object to a `VectorVariantBox` to enable a time-varying target box, which enables slow compression of the box in HPMC.

## Motivation and context

Resolves #1641 

## How has this been tested?

Added a validation test that uses QuickCompress with a `BoxVariant` and ensures that the box compression occurs no faster than what the time-varying target box dictates.

## Change log

```
`hoomd.hpmc.update.QuickCompress` now accepts a `hoomd.variant.box.BoxVariant` object for `target_box`, enabling slow compression of simulation boxes in HPMC simulations.
```

## Checklist:

### DIFF
--- a/hoomd/BoxDim.h
+++ b/hoomd/BoxDim.h
@@ -601,22 +601,7 @@ struct
         return L1 == L2 && xy1 == xy2 && xz1 == xz2 && yz1 == yz2;
         }
 
-    HOSTDEVICE bool operator==(const std::array<Scalar, 6>& other) const
-        {
-        Scalar3 L = getL();
-        Scalar xy = getTiltFactorXY();
-        Scalar xz = getTiltFactorXZ();
-        Scalar yz = getTiltFactorYZ();
-        std::array<Scalar, 6> box = {L.x, L.y, L.z, xy, xz, yz};
-        return box == other;
-        }
-
     HOSTDEVICE bool operator!=(const BoxDim& other) const
-        {
-        return !((*this) == other);
-        }
-
-    HOSTDEVICE bool operator!=(const std::array<Scalar, 6>& other) const
         {
         return !((*this) == other);
         }
--- a/hoomd/BoxDim.h
+++ b/hoomd/BoxDim.h
@@ -11,7 +11,9 @@
 #include "HOOMDMath.h"
 #include "VectorMath.h"
 
+#if !defined(HOOMD_LLVMJIT_BUILD)
 #include <array>
+#endif
 
 // Don't include MPI when compiling with __HIPCC__ or an LLVM JIT build
 #if defined(ENABLE_MPI) && !defined(__HIPCC__) && !defined(HOOMD_LLVMJIT_BUILD)
@@ -143,6 +145,7 @@ struct
         m_xz = m_xy = m_yz = Scalar(0.0);
         }
 
+#if !defined(HOOMD_LLVMJIT_BUILD)
     /// Constructs a box from a std::array<Scalar, 6>
     /** @param array Box parameters
      */
@@ -152,6 +155,7 @@ struct
         setTiltFactors(array[3], array[4], array[5]);
         m_periodic = make_uchar3(1, 1, 1);
         }
+#endif
 
     //! Get the periodic flags
     /*! \return Periodic flags
--- a/hoomd/BoxDim.h
+++ b/hoomd/BoxDim.h
@@ -141,6 +141,16 @@ struct
         m_xz = m_xy = m_yz = Scalar(0.0);
         }
 
+    /// Constructs a box from a std::array<Scalar, 6>
+    /** @param array Box parameters
+    */
+    HOSTDEVICE explicit BoxDim(const std::array<Scalar, 6>& array)
+        {
+        setL(make_scalar3(array[0], array[1], array[2]));
+        setTiltFactors(array[3], array[4], array[5]);
+        m_periodic = make_uchar3(1, 1, 1);
+        }
+
     //! Get the periodic flags
     /*! \return Periodic flags
      */
--- a/hoomd/BoxDim.h
+++ b/hoomd/BoxDim.h
@@ -11,6 +11,8 @@
 #include "HOOMDMath.h"
 #include "VectorMath.h"
 
+#include <array>
+
 // Don't include MPI when compiling with __HIPCC__ or an LLVM JIT build
 #if defined(ENABLE_MPI) && !defined(__HIPCC__) && !defined(HOOMD_LLVMJIT_BUILD)
 #include "HOOMDMPI.h"
--- a/hoomd/BoxDim.h
+++ b/hoomd/BoxDim.h
@@ -601,7 +601,22 @@ struct
         return L1 == L2 && xy1 == xy2 && xz1 == xz2 && yz1 == yz2;
         }
 
+    HOSTDEVICE bool operator==(const std::array<Scalar, 6>& other) const
+        {
+        Scalar3 L = getL();
+        Scalar xy = getTiltFactorXY();
+        Scalar xz = getTiltFactorXZ();
+        Scalar yz = getTiltFactorYZ();
+        std::array<Scalar, 6> box = {L.x, L.y, L.z, xy, xz, yz};
+        return box == other;
+        }
+
     HOSTDEVICE bool operator!=(const BoxDim& other) const
+        {
+        return !((*this) == other);
+        }
+
+    HOSTDEVICE bool operator!=(const std::array<Scalar, 6>& other) const
         {
         return !((*this) == other);
         }
--- a/hoomd/BoxDim.h
+++ b/hoomd/BoxDim.h
@@ -143,7 +143,7 @@ struct
 
     /// Constructs a box from a std::array<Scalar, 6>
     /** @param array Box parameters
-    */
+     */
     HOSTDEVICE explicit BoxDim(const std::array<Scalar, 6>& array)
         {
         setL(make_scalar3(array[0], array[1], array[2]));
--- a/hoomd/hpmc/UpdaterQuickCompress.cc
+++ b/hoomd/hpmc/UpdaterQuickCompress.cc
@@ -47,7 +47,6 @@ void UpdaterQuickCompress::update(uint64_t timestep)
     // count the number of overlaps in the current configuration
     auto n_overlaps = m_mc->countOverlaps(false);
     BoxDim current_box = m_pdata->getGlobalBox();
-
     BoxDim target_box = BoxDim((*m_target_box)(timestep));
     if (n_overlaps == 0 && current_box != target_box)
         {
@@ -61,9 +60,9 @@ void UpdaterQuickCompress::update(uint64_t timestep)
         m_is_complete = false;
     }
 
-void UpdaterQuickCompress::performBoxScale(uint64_t timestep)
+void UpdaterQuickCompress::performBoxScale(uint64_t timestep, const BoxDim& target_box)
     {
-    auto new_box = getNewBox(timestep);
+    auto new_box = getNewBox(timestep, target_box);
     auto old_box = m_pdata->getGlobalBox();
 
     Scalar3 old_origin = m_pdata->getOrigin();
@@ -140,7 +139,7 @@ static inline double scaleTilt(double current, double target, double s)
         }
     }
 
-BoxDim UpdaterQuickCompress::getNewBox(uint64_t timestep)
+BoxDim UpdaterQuickCompress::getNewBox(uint64_t timestep, const BoxDim& target_box)
     {
     // compute the current MC translate acceptance ratio
     auto current_counters = m_mc->getCounters();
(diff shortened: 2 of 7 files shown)


hoomd/hpmc/update.py
```diff
@@ -894,9 +897,11 @@ def __init__(self,
 
         param_dict = ParameterDict(max_overlaps_per_particle=float,
                                    min_scale=float,
-                                   target_box=hoomd.Box,
+                                   target_box=hoomd.variant.box.BoxVariant,
```
tommy-waltmann, line 917: If this attribute is settable, we need to coerce a regular box into a box variant whenever the attribute is set, not just in the constructor.

hoomd/hpmc/pytest/test_quick_compress.py
```diff
@@ -13,20 +13,35 @@
 # here that require preprocessing
 valid_constructor_args = [
     dict(trigger=hoomd.trigger.Periodic(10),
-         target_box=hoomd.Box.from_box([10, 10, 10])),
+         target_box=hoomd.variant.box.Constant(hoomd.Box.from_box([10, 10,
+                                                                   10]))),
```
tommy-waltmann, line 18: Let's keep at least one case where the constructor arg is still a regular box since it should be able to accept both.

hoomd/hpmc/pytest/test_quick_compress.py
```diff
@@ -35,8 +50,17 @@
     ('trigger', hoomd.trigger.Periodic(10000)),
     ('trigger', hoomd.trigger.After(100)),
     ('trigger', hoomd.trigger.Before(12345)),
-    ('target_box', hoomd.Box.from_box([10, 20, 30])),
-    ('target_box', hoomd.Box.from_box([50, 50])),
+    ('target_box', hoomd.variant.box.Constant(hoomd.Box.from_box([10, 20,
```
tommy-waltmann, line 59: Should I be able to set the target_box attribute to a regular box too?

tommy-waltmann (changes requested):
This looks pretty good, I just have some small changes to consider.

hoomd/BoxDim.h
```diff
@@ -601,10 +601,25 @@ struct
...
+    HOSTDEVICE bool operator==(const std::array<Scalar, 6>& other) const
+        {
+        Scalar3 L = getL();
+        Scalar xy = getTiltFactorXY();
+        Scalar xz = getTiltFactorXZ();
+        Scalar yz = getTiltFactorYZ();
+        std::array<Scalar, 6> box = {L.x, L.y, L.z, xy, xz, yz};
+        return box == other;
+        }
+
```
joaander: Let's not introduce unneeded == operators.
```suggestion
```

hoomd/BoxDim.h
```diff
@@ -601,10 +601,25 @@ struct
...
     HOSTDEVICE bool operator!=(const BoxDim& other) const
         {
         return !((*this) == other);
         }
+
+    HOSTDEVICE bool operator!=(const std::array<Scalar, 6>& other) const
+        {
+        return !((*this) == other);
+        }
```
joaander:
```suggestion
```

hoomd/hpmc/UpdaterQuickCompress.cc
```diff
@@ -48,13 +48,13 @@ void UpdaterQuickCompress::update(uint64_t timestep)
     auto n_overlaps = m_mc->countOverlaps(false);
     BoxDim current_box = m_pdata->getGlobalBox();
 
-    if (n_overlaps == 0 && current_box != *m_target_box)
+    if (n_overlaps == 0 && current_box != (*m_target_box)(timestep))
```
joaander:
```suggestion
    BoxDim target_box = BoxDim((*m_target_box)(timestep));
    if (n_overlaps == 0 && current_box != target_box)
```
Use the existing `BoxDim` `!=` check.

I will add a `BoxDim` constructor so this will work.

hoomd/hpmc/UpdaterQuickCompress.cc
```diff
@@ -48,13 +48,13 @@ void UpdaterQuickCompress::update(uint64_t timestep)
...
         {
         performBoxScale(timestep);
```
joaander:
```suggestion
        performBoxScale(timestep, target_box);
```
Add `const BoxDim& target_box` to the arguments of `performBoxScale`.

hoomd/hpmc/UpdaterQuickCompress.cc
```diff
@@ -48,13 +48,13 @@ void UpdaterQuickCompress::update(uint64_t timestep)
...
         }
 
     // The compression is complete when we have reached the target box and there are no overlaps.
-    if (n_overlaps == 0 && current_box == *m_target_box)
+    if (n_overlaps == 0 && current_box == (*m_target_box)(timestep))
```
joaander:
```suggestion
    if (n_overlaps == 0 && current_box == target_box)
```

hoomd/hpmc/UpdaterQuickCompress.cc
```diff
@@ -183,31 +183,37 @@ BoxDim UpdaterQuickCompress::getNewBox(uint64_t timestep)
     hoomd::UniformDistribution<double> uniform(min_scale, 1.0);
     double scale = uniform(rng);
 
-    const auto& target_box = *m_target_box;
+    const auto& target_box_dims = (*m_target_box)(timestep);
```
joaander:
```suggestion
```
Add `const BoxDim& target_box` to the arguments of `getNewBox`. Also, undo all the changes to `getNewBox` below that are no longer necessary as `target_box` is now set.

hoomd/hpmc/pytest/test_quick_compress.py
joaander: Yes, you need to be able to set a regular box both for backwards compatibility and convenience.

hoomd/hpmc/update.py
```diff
@@ -736,7 +736,8 @@ class QuickCompress(Updater):
         trigger (hoomd.trigger.trigger_like): Update the box dimensions on
             triggered time steps.
 
-        target_box (hoomd.box.box_like): Dimensions of the target box.
+        target_box (hoomd.box.box_like or hoomd.variant.box.BoxVariant):
```
joaander:
```suggestion
        target_box (`hoomd.box.box_like` or `hoomd.variant.box.BoxVariant`):
```
I think this is necessary for the links to work correctly.

hoomd/hpmc/update.py
joaander: Yes, use a converter method here.

hoomd/hpmc/pytest/test_quick_compress.py
```diff
@@ -235,6 +259,44 @@ def test_disk_compression(phi, simulation_factory, lattice_snapshot_factory):
...
+    mc.shape['A'] = dict(diameter=1)
+    sim.operations.integrator = mc
+    scale = hoomd.variant.Ramp(0, 1, 0, t_ramp)
+
+    while sim.timestep < t_ramp:
+        sim.run(10)
+        target_density = scale(sim.timestep) / final_volume + (
+            1 - scale(sim.timestep)) / initial_volume
+        target_volume = 1 / target_density
+        assert sim.state.box.volume >= target_volume
```
joaander: Does this test fail if you set `target_box` to the final box?

hoomd/hpmc/UpdaterQuickCompress.cc
```diff
@@ -183,31 +183,37 @@ BoxDim UpdaterQuickCompress::getNewBox(uint64_t timestep)
...
+    Scalar target_Lx = target_box_dims[0];
+    Scalar target_Ly = target_box_dims[1];
+    Scalar target_Lz = target_box_dims[2];
+    Scalar target_xy = target_box_dims[3];
+    Scalar target_xz = target_box_dims[4];
+    Scalar target_yz = target_box_dims[5];
```
joaander:
```suggestion
```

joaander (changes requested):
Thanks! I have many suggestions to reduce the number of changes needed and make this a clean change set.

hoomd/hpmc/update.py
tcmoore3: I think I implemented the preprocessing correctly, but I'm running into the same import issues that I was in the vector variant PR. I will need help fixing that.

hoomd/hpmc/pytest/test_quick_compress.py
tcmoore3: Agreed with Josh.

hoomd/hpmc/update.py
joaander: Yes, this is a circular import:
`hoomd/variant/box.py`:
```
from hoomd.data.typeconverter import box_preprocessing, variant_preprocessing
```
`hoomd/data/typeconvertor.py`:
```
from hoomd.variant.box import Constant as ConstantBox
```

The solution is to not use `from ... import ...`. I pushed a commit that fixes this.

hoomd/hpmc/pytest/test_quick_compress.py
tcmoore3: No. It would if it was `sim.state.box.volume > target_volume` but the `>=` handles the case where `target_box` is the final box. I updated the test to check that case.

joaander: Let me clarify: `t_ramp=100` is pretty short.
If you set the target box to the final, this test might still pass (where target_volume) is still determined from the ramp. We need `t_ramp` set large enough so that this test is validly checking that the target box variant is applied.

hoomd/hpmc/pytest/test_quick_compress.py
tommy-waltmann: A test case still needs to be added for this.

tommy-waltmann (changes requested):
I think there is just one more comment of mine from last time to address.

tcmoore3:
I only realized this while addressing [this comment](https://github.com/glotzerlab/hoomd-blue/pull/1736#discussion_r1514948806), but the notion of "complete" for a quick compress updater is somewhat ambiguous now that you can pass in a moving target box. The C++ updater sets `m_is_complete` if the box is equal to the target box (evaluated at the current timestep) and there are no overlaps. This means `quick_compress.complete` returns `True` if the box can easily compress to the setpoint for the specific step without overlaps. This situation occured when I was updating that test. The test contained:
```
# qc.t_ramp = 10000
while not qc.complete:
        sim.run(10)
        simulation_density = 1 / sim.state.box.volume
        target_density = 1 / hoomd.Box(*qc.target_box(sim.timestep)).volume
        assert simulation_density < target_density or (
            simulation_density == pytest.approx(target_density))
assert sim.timestep >= qc.target_box.t_ramp
```
After 10 steps, `qc.complete` returned `True` since the box reached the target for that timestep without overlaps, despite the step being far lower than the ramp length. 

Should we do anything to handle this scenario? It's easy enough to compute final target box for the `BoxVariant`s we implemented so far, but not all variants necessarily have a final box, e.g., I can imagine a `variant.box.Cycle`.  I imagine the typical use case is currently something along the lines of:
```
while not qc.complete:
    sim.run(N)
```
So do we now just let the user handle this, e.g., changing the above to something like
```
while sim.timestep < t_ramp and (not qc.complete):
    sim.run(N)
```
?

hoomd/hpmc/pytest/test_quick_compress.py
tcmoore3: You're right. Added.

tommy-waltmann (approved):

joaander:
> So do we now just let the user handle this, e.g., changing the above to something like
> 
> ```
> while sim.timestep < t_ramp and (not qc.complete):
>     sim.run(N)
> ```

Yes, we should make sure the behavior is documented clearly. I don't want to require that all box variants (including custom ones) provide a valid `complete` attribute. Users should instead adjust their `while` conditions appropriately for the variant of their choice.

hoomd/hpmc/update.py
joaander, line 1: This comment is not related to this pull request, but I noticed the docs are out of date. @janbridley can you please make a small PR to update the docs below. I think the `.. math::` block no longer accurately describes how the box is scaled after your fixes.

hoomd/hpmc/update.py
```diff
@@ -845,11 +848,25 @@ class QuickCompress(Updater):
...
+    evaluated at the current timestep and there are no overlaps in the system.
+    To ensure the updater has compressed the system to the final target box, use
+    a condition that checks both the `complete` attribute of the updater and the
+    simulation timestep. For example::
+
+        target_box = hoomd.variant.box.InverseVolumeRamp(
+            sim.state.box, sim.state.box.volume / 2, 0, 10000)
+        qc = hoomd.hpmc.update.QuickCompress(10, target_box)
+        while (not qc.complete) and sim.timestep < qc.target_box.t_ramp:
+            sim.run(100)
```
joaander:
```suggestion
        while sim.timestep < target_box.t_ramp + target_box.t_start or (not qc.complete):
            sim.run(100)
```
Use `or` to continue running past the end of `t_start + t_ramp` until the compression is complete. Also change the order of the lazy evaluation to make it more obvious this while loop is until the end of the ramp.

hoomd/hpmc/update.py
```diff
@@ -845,11 +848,25 @@ class QuickCompress(Updater):
...
 
+    When using a `BoxVariant` for `target_box`, `complete` returns `True` if the
+    current simulation box is equal to the box corresponding to `target_box`
+    evaluated at the current timestep and there are no overlaps in the system.
+    To ensure the updater has compressed the system to the final target box, use
+    a condition that checks both the `complete` attribute of the updater and the
+    simulation timestep. For example::
+
+        target_box = hoomd.variant.box.InverseVolumeRamp(
+            sim.state.box, sim.state.box.volume / 2, 0, 10000)
```
joaander:
```suggestion
        target_box = hoomd.variant.box.InverseVolumeRamp(
            sim.state.box, sim.state.box.volume / 2, 0, 1_000)
```
We should put this in Sybil, but I see that there are no other Sybil examples in this file. I'll make a separate PR for a large batch of Sybil updates.

For now, let's shorten the steps so the doc example will not run for long when it is added to Sybil.

joaander (approved):
Thanks! I have two suggestions to improve the docs. After those changes, this looks great.

joaander:
One more fix is needed. `test_inverse_volume_slow_compress[CPU-2]` uses too small of a box for domain decomposition in MPI: https://github.com/glotzerlab/hoomd-blue/actions/runs/8242679743/job/22542233097?pr=1736#step:8:2266

hoomd/hpmc/update.py
janbridley: Good catch! I just pushed `ac17bf8dd` which fixes this. The block is a little harder to read due to the more complicated formulas, but should still do the trick.

tcmoore3:
Resolves #1641

PR #1737, opened 2024-03-08
Refactor virtual particle interface

### LINKED ISSUE #772: Relax MPCD virtual particle filler design

mphoward:
## Description

This PR refactors the `VirtualParticleFiller` base class to be more general. It now has a pure virtual `fill` method that deriving classes must implement. The old `fill` method that calls virtual methods that compute the number of particles & draw them using known geometric properties has been moved to a `ManualVirtualParticleFiller` derived class. Implementations for the two geometries we have now derive from `ManualVirtualParticleFiller`. These are C++-only changes that could break plugins (like mine... *sigh*).

This PR also fixes a minor bug we identified where attaching multiple `VirtualParticleFiller`s would lead to the same sequence of drawn particles, due to having a common see and class identifier. I now assign a unique identifier for the class when it is created, and I use this as a counter in the PRNG.

## Motivation and context

The old base class required developers to implement methods to compute the number of particles to fill and to randomly draw the particles. This seemed like a flexible design at the time, but there are actually quite a few simple geometries where this no longer works, mainly when run under MPI. For example, it's easy to draw particles in a shell around a sphere using spherical coordinates on 1 processor, but this is hard to do with MPI because the local box of each rank is orthorhombic. You would need to know the volume defined by the intersection of the shell and the box in order to implement both required methods, and a formula for this is not known.

Rejection sampling is a solution to this problem, but then the number of particles is not deterministic, so the method that computes the number cannot be implemented.

Resolves #772 

## How has this been tested?

Everything still compiles and existing tests still pass.

## Change log

N/A

@kkritikak should be credited for helping with this implementation in a fork!

## Checklist:

### DIFF
--- a/hoomd/mpcd/CMakeLists.txt
+++ b/hoomd/mpcd/CMakeLists.txt
@@ -12,6 +12,7 @@ set(_mpcd_cc_sources
     CollisionMethod.cc
     Communicator.cc
     Integrator.cc
+    ManualVirtualParticleFiller.cc
     ParallelPlateGeometryFiller.cc
     PlanarPoreGeometryFiller.cc
     Sorter.cc
@@ -34,6 +35,7 @@ set(_mpcd_headers
     Communicator.h
     CommunicatorUtilities.h
     Integrator.h
+    ManualVirtualParticleFiller.h
     ParticleData.h
     ParticleDataSnapshot.h
     ParticleDataUtilities.h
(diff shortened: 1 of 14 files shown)


mphoward:
This one is ready for review now!

tommy-waltmann (approved):

AlainKadar (approved):

mphoward:
@joaander I think this one can be merged when you're ready.

mphoward:
@joaander would it be possible to get this merged today or tomorrow? I am starting to run some research-relevant regression tests and moving toward a final review of the new API so that we can get this porting wrapped up soon!

PR #1740, opened 2024-03-12
Deprecate the box1 box2 arguments to BoxResize.

### LINKED ISSUE #1640: A more flexible box resize updater
### Description

From a close read of the source for the BoxResizeUpdater, it seems to me that it resets the dimensions of the entire box every time it updates the box, even for dimensions that are not different between the two boxes. This is fine in an "NVT" ensemble calculation, but does not seem compatible with a mixed ensemble calculation such as allowing two dimensions of the box to be controlled by a barostat. With the current framework, is there a way to adjust a single dimension manually while allowing other dimensions to be managed by a barostat?

Additionally, as far as I can tell, there is no current support for adjusting box dimensions independently with a more complicated function than a linear interpolation between two endpoints. For example, uniaxial elongation at constant volume isn't supported, except for by stopping and starting the simulation. This could also likely be accomplished with a custom updater. However, in both cases, given the need to rescale every particle's coordinates, leveraging GPU resources would be ideal, and I don't think you can do that with a custom updater (I could be wrong).

### Proposed solution

I am almost finished writing a solution for this through modification to the C++ layer of HOOMD-blue, and would be excited to discuss whether including this in HOOMD-blue would be appropriate, and how to tweak the design to be in accordance with HOOMD-blue's design philosophy. 

The current form of my solution works as follows: the Python layer takes as input up to 6 variants, and dimensions to which these variants correspond. Any dimension not specified is untouched by the box updater (rather than maintained at its t=0 value). Hence, if something else (say a barostat) touches the box dimension, this isn't undone by the box updater. When triggered, the box updater sets each paramater value i to box0[i]*variant_i(timestep)/variant_i(0). The scale factor is normalized to the variant's value at time zero such that the box starts at its original size.

This supports varying any set of box dimensions with complete independence, but changes the philosophy from "interpolating between an initial and final box" to "scaling the original box". 

To not break backwards compatibility, this could be implemented in a few different ways -- 1) as its own separate Python class with its own separate C++ class, 2) as a switch within the current "hoomd.update.BoxResize" Python class that defaults to the original behavior and accepts the original arguments, but drops into different C++ classes,  3) as two separate Python classes that use the same C++ class (the new one) in different ways (since the original behavior is a special case of the general behavior I am suggesting), or 4) as a single Python class and single C++ class, with the same switch as mentioned in (2).

I am not sure of a clean way to accomplish the "switch" between behavior that I am suggesting, but it seems that it would absolutely need to be there if a single Python class is to be used, to support backwards compatibility.

Ultimately, if we decide to proceed, I have a lot of the code written for this and it could be reorganized accordingly. Looking forward to any thoughts or feedback! 

### Additional context

_No response_

joaander:
## Description

Add the `box` argument to `BoxResize`. Deprecate `box1`, `box2`, and `variant`.

## Motivation and context

Box variants provide a more flexible solution.

Resolves #1640

## How has this been tested?

Added unit tests.

## Change log

```
Added:

* ``box`` argument to ``hoomd.update.BoxResize`` that accepts a ``BoxVariant``.

Deprecated:

* ``box1``, ``box2``, and ``variant`` arguments to ``hoomd.update.BoxResize``.
```

## Checklist:

### DIFF
--- a/hoomd/BoxResizeUpdater.cc
+++ b/hoomd/BoxResizeUpdater.cc
@@ -15,21 +15,16 @@ using namespace std;
 
 namespace hoomd
     {
-/*! \param sysdef System definition containing the particle data to set the box size on
-    \param Lx length of the x dimension over time
-    \param Ly length of the y dimension over time
-    \param Lz length of the z dimension over time
-
-    The default setting is to scale particle positions along with the box.
+/** @param sysdef System definition containing the particle data to set the box size on
+    @param trigger Steps on which to execute.
+    @param box Box as a function of time.
+    @param group Particles to scale.
 */
-
 BoxResizeUpdater::BoxResizeUpdater(std::shared_ptr<SystemDefinition> sysdef,
                                    std::shared_ptr<Trigger> trigger,
-                                   std::shared_ptr<BoxDim> box1,
-                                   std::shared_ptr<BoxDim> box2,
-                                   std::shared_ptr<Variant> variant,
+                                   std::shared_ptr<VectorVariantBox> box,
                                    std::shared_ptr<ParticleGroup> group)
-    : Updater(sysdef, trigger), m_box1(box1), m_box2(box2), m_variant(variant), m_group(group)
+    : Updater(sysdef, trigger), m_box(box), m_group(group)
     {
     assert(m_pdata);
     assert(m_variant);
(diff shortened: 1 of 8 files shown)


tcmoore3 (approved):
LGTM!

tcmoore3:
pytest is failing because `variant.box.Interpolate` is not pickleable because I didn't add a `__reduce__` method in #1736 like I did for `variant.box.Constant`. I will add one shortly if it's not there before I can get to it.

PR #1748, opened 2024-03-22
Fix segfault when using `md.methods.rattle.Brownian` with domain decomposition.

joaander:
## Description

Access local group indices, not global group tags.

## Motivation and context

Prevent a seg fault when using domain decomposition.

Resolves #1742

## How has this been tested?

The test case in #1742 passes.

## Change log

```
* `md.methods.rattle.Brownian` executes without causing a segmentation fault on the CPU with domain decomposition.
```

## Checklist:

- [] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/ContributorAgreement.md).


tommy-waltmann (commented):
It appears there was no test to catch this segfault previously. Should a test be added to make sure this code works right now?

PR #1763, opened 2024-04-09
Tweaks to MPCD v4 API

mphoward:
## Description

This is a final set of API changes for MPCD in v4 resulting from using the code in some tests. The main changes are to clarify variable names in the force and geometry modules to make them more descriptive.

The pre-commit check failing on this PR due to a recent merge up of `trunk-minor` into `mpcd-v4` bumping the license year, so I ran pre-commit to update those.

## Motivation and context

This makes the API more readable.

## How has this been tested?

Tests have been updated to match the new API.

## Change log

N/A

## Checklist:

### DIFF
--- a/hoomd/mpcd/force.py
+++ b/hoomd/mpcd/force.py
@@ -1,17 +1,17 @@
 # Copyright (c) 2009-2024 The Regents of the University of Michigan.
 # Part of HOOMD-blue, released under the BSD 3-Clause License.
 
-r"""MPCD solvent forces.
+r"""MPCD particle body forces.
 
 MPCD can apply a body force to each MPCD particle as a function of position.
 The external force should be compatible with the chosen
 :class:`~hoomd.mpcd.geometry.Geometry`. Global momentum conservation can be
-broken by adding a solvent force, so care should be chosen that the entire model
+broken by adding a body force, so care should be chosen that the entire model
 is designed so that the system does not have net acceleration. For example,
 solid boundaries can be used to dissipate momentum, or a balancing force can be
-applied to particles that are embedded in the solvent through the collision
-step. Additionally, a thermostat will likely be required to maintain temperature
-control in the driven system.
+applied to particles that are embedded in the MPCD particles through the
+collision step. Additionally, a thermostat will likely be required to maintain
+temperature control in the driven system.
 
 .. invisible-code-block: python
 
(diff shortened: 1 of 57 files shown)


hoomd/mpcd/force.py
```diff
@@ -93,50 +90,49 @@ class BlockForce(SolventForce):
 
                 force.force = 1.0
 
-        half_separation (float): Half the distance between the centers of the
-            blocks.
+        separation (float): Ddistance between the centers of the blocks.
```
tommy-waltmann:
```suggestion
        separation (float): Distance between the centers of the blocks.
```

hoomd/mpcd/force.py
```diff
@@ -26,30 +26,29 @@
 from hoomd.operation import _HOOMDBaseObject
 
 
-class SolventForce(_HOOMDBaseObject):
-    """Solvent force.
+class BodyForce(_HOOMDBaseObject):
```
tommy-waltmann, line 29: Would it be more descriptive to name this class `MPCDParticleForce`? I don't understand why they are called body forces if they are just forces on the MPCD particles.

tommy-waltmann (changes requested):
I have one small correction and a comment about naming the forces on MPCD particles.

I am not the most familiar with the MPCD simulation mode, so feel free to disagree with my suggestion.

hoomd/mpcd/force.py
tommy-waltmann: The term "body force" is used throughout the documentation also.

mphoward: Thanks for the suggestion! I wanted to rename from `SolventForce`, and I switched to `BodyForce` because MPCD particles can only be acted on by (one) body forces:

https://github.com/glotzerlab/hoomd-blue/blob/4306d7a6be469148c6be2618b4a94c0cab386f59/hoomd/mpcd/BounceBackStreamingMethod.h#L136

This is one of the things that makes MPCD particles different from normal HOOMD particles.

However, `MPCDParticleForce` is also descriptive. Let me think about it, and I will comment on the PR later whether I will push this change.

mphoward: I decided I would like to stick with `BodyForce`. Its closest analog is `hoomd.md.external.field.Field`, but these also aren't exactly fields because they specify a force rather than a potential

mphoward:
@joaander I've implemented @tommy-waltmann's edit so I think this is ready to merge!

tommy-waltmann (approved):

PR #1785, opened 2024-05-16
Add error when ellipsoid axis is zero

### LINKED ISSUE #1762: Error when any semi-major axis is less than zero in `Ellipsoid`.
### Description

Add checks that `a`, `b`, and `c` are all greater than 0 in `hpmc.integrate.Ellipsoid`.

### Motivation and context

In #1759 a user reports that ellipsoids with `c=0` report no overlaps.

janbridley:
## Description

A check was added to verify values are within the expected range. ~I set the minimum allowed value to `ELLIPSOID_OVERLAP_PRECISION` (1e-6), which may or may not be the correct choice. We could do FLT_EPSILON from <cfloat> as well, but that may not be consistent if `HOOMD_SHORTREAL_SIZE==64`.~ I am open to suggestions about the correct value to assign.

Errors occur then any semimajor axis is less than `0.0f`.

## Motivation and context

Overlap checks do not work for ellipsoids with any semimajor axis less than zero. This error (on the c++ level) prevents users from inadvertently encountering this issue.

Resolves #1762

## How has this been tested?

A new test has been added that attempts to set invalid parameters. This tests values at zero, slightly above zero, and well below it.

## Change log

```
- Provide an error message for invalid Ellipsoid shape params.
```

## Checklist:

### DIFF
--- a/hoomd/hpmc/ShapeEllipsoid.h
+++ b/hoomd/hpmc/ShapeEllipsoid.h
@@ -67,8 +67,7 @@ struct EllipsoidParams : ShapeParams
         y = v["b"].cast<ShortReal>();
         z = v["c"].cast<ShortReal>();
 
-        if (x <= ELLIPSOID_OVERLAP_PRECISION || y <= ELLIPSOID_OVERLAP_PRECISION
-            || z <= ELLIPSOID_OVERLAP_PRECISION)
+        if (x <= 0.0f || y <= 0.0f || z <= 0.0f)
             {
             throw std::domain_error("All semimajor axes must be nonzero!");
             }
--- a/hoomd/hpmc/ShapeEllipsoid.h
+++ b/hoomd/hpmc/ShapeEllipsoid.h
@@ -66,6 +66,12 @@ struct EllipsoidParams : ShapeParams
         x = v["a"].cast<ShortReal>();
         y = v["b"].cast<ShortReal>();
         z = v["c"].cast<ShortReal>();
+
+        if (x <= ELLIPSOID_OVERLAP_PRECISION || y <= ELLIPSOID_OVERLAP_PRECISION
+            || z <= ELLIPSOID_OVERLAP_PRECISION)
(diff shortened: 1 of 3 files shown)


joaander:
I think a check against 0 is correct here. `ELLIPSOID_OVERLAP_PRECISION` is a relative tolerance. If users would like to work in a system of units where typical particle diameters are 1e-50, they we should allow that.

hoomd/hpmc/ShapeEllipsoid.h
```diff
@@ -66,6 +66,12 @@ struct EllipsoidParams : ShapeParams
         x = v["a"].cast<ShortReal>();
         y = v["b"].cast<ShortReal>();
         z = v["c"].cast<ShortReal>();
+
+        if (x <= ELLIPSOID_OVERLAP_PRECISION || y <= ELLIPSOID_OVERLAP_PRECISION
+            || z <= ELLIPSOID_OVERLAP_PRECISION)
```
janbridley:
```suggestion

        if (x <= 0.0f || y <= 0.0f || z <= 0.0f)
```

hoomd/hpmc/pytest/test_shape.py
```diff
@@ -80,6 +80,16 @@ def test_invalid_shape_params(invalid_args):
         mc.shape["A"] = args
 
 
+@pytest.mark.parametrize("c", [0.0, 1e-8, -0.5])
```
janbridley:
```suggestion
@pytest.mark.parametrize("c", [0.0, -0.5])
```

SchoeniPhlippsn (commented):
The pull request is straightforward, but we should raise a similar error message for `FacetedEllipsoid`.

tcmoore3 (approved):

janbridley:
pre-commit.ci autofix

joaander (approved):
Thanks!

PR #1812, opened 2024-06-12
Autogenerate `mpcd::BounceBackNVE` templates

mphoward:
## Description

This PR uses CMake to automatically generate the templates of `mpcd::BounceBackNVE`. I've had two developers get linker errors because this wasn't done automatically when they added new geometries. I've also refactored `module.cc` to use forward declaration like `module-md.cc` to speed up build times and avoid a recompile when a header changes.

## Motivation and context

This makes code development easier.

## How has this been tested?

Code compiles.

## Change log

No change needed, this is internal only.

## Checklist:


hoomd/mpcd/module.cc
```diff
@@ -86,7 +20,82 @@ namespace mpcd
  */
 namespace detail
     {
-    };
+void export_ATCollisionMethod(pybind11::module&);
```
joaander, line 23: Nice! This type of change can significantly reduce re-compilation time when developing.

joaander (approved):
Thanks!

Issue #773, opened 2020-09-04
Relax MPCD restrictions on boxes


mphoward:
## Description

MPCD currently requires that the box size be (1) orthorhombic, (2) constant, and (3) a multiple of the MPCD cell size. Only requirement (3) is strictly necessary for the algorithm to work right, so I will check over the code to see if it makes sense to support other (e.g., deformed) boxes. This is change is not necessarily breaking but may not be the behavior expected by the user, so I will look into it for v3.

mphoward:
We will need to address this as part of our new work.

mphoward:
Not stale

mphoward:
@joaander I am going to do my best to get this branch to a point that we can open a PR onto `trunk-major` by Monday. It has a bunch of changes that will potentially cause merge conflicts when you finalize 5.0, so I want to get it in as soon as I can.

joaander:
What planned changes to do you expect to conflict?

mphoward:
This change will only touch MPCD code, but it touches *alot* of it so I think the main conflicts will come from any global search & replace incidental to other changes. And then of course, application of code formatting.

[feature/mpcd-triclinic-box](https://github.com/glotzerlab/hoomd-blue/tree/feature/mpcd-triclinic-box)

joaander:
It looks like the majority of those changes are in C++ files, so the mostly likely cause of conflicts is #1938 which is already merged. The switch to ruff may cause conflicts in the `.py` files. I plan to switch to ruff as the last step before a 5.0 release. 

I will be sure to review your PR soon after you post it.

mphoward:
> It looks like the majority of those changes are in C++ files

Yes, that's right—this is a big generalization of some of the underlying C++ code. Sounds like a plan, thanks!

mphoward:
Done in #1950

PR #1813, opened 2024-06-12
Restrict `mpcd.collide.CellList` cell size

mphoward:
## Description

We are actively working to relax the requirement that the MPCD cells be cubic (#773). Allowing the user to specify, change, or access the scalar `cell_size` is problematic. To avoid introducing API breaking changes after MPCD is rereleased, I would prefer to specify the cell size is fixed at 1.0, which is the standard choice for most MPCD models anyway. We will then extend the `CellList` API to allow specifying the number of cells along each lattice vector in future, which will allow the cell size to change.

## Motivation and context

This will help smoothen the release of MPCD.

## How has this been tested?

Existing tests pass.

## Change log

No change needed.

## Checklist:

### DIFF
--- a/hoomd/mpcd/collide.py
+++ b/hoomd/mpcd/collide.py
@@ -29,12 +29,10 @@ class CellList(Compute):
     """Collision cell list.
 
     Args:
-        cell_size (float): Size of a collision cell.
         shift (bool): When True, randomly shift underlying collision cells.
 
-    The MPCD `CellList` bins particles into cubic cells of edge length
-    `cell_size`. Currently, the simulation box must be orthorhombic, and its
-    edges must be a multiple of `cell_size`.
+    The MPCD `CellList` bins particles into cubic cells of edge length 1.0.
+    The simulation box must be orthorhombic, and its edges must be an integer.
 
     When the total mean-free path of the MPCD particles is small, the cells
     should be randomly shifted in order to ensure Galilean invariance of the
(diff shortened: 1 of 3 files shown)


tommy-waltmann (commented):
Why not just allow the user to specify how many cells there should be in each direction right now, instead of waiting for a future PR?

mphoward:
Not a great answer, but because this was a quick fix and I’m leaving for a week after today 😅 I have the code to do that on:

https://github.com/glotzerlab/hoomd-blue/tree/feature/mpcd-triclinic-box

I could try to move only the minimal required parts here and leave a comment on its status at the end of the day.

There would still need to be an error check in place that after the user specifies the number of cells (as a tuple) that the same cell size is calculated for all dimensions.

There would also no longer be a default cell list because we can’t guess a default number of cells the way we guess the default cell size. All the examples and tests would need to be updated to reflect that.

FWIW, users rarely change the cell size from 1. I’ve seen it done only once with HOOMD.

tommy-waltmann:
Okay, this seems like a special situation. It may be okay long as the changes on this PR are going to be overwritten quickly by a later PR.

Another thing I will point out is that this is an API-breaking change based on the `trunk-minor` branch, which is not normally something we would allow.

I think @joaander should have the final say on whether this type of PR is okay to merge.

joaander:
@tommy-waltmann It is not API breaking because the MPCD code is not yet in a tagged release version. @mphoward is proposing the quick fix in now so that the future PR is not API breaking.

tommy-waltmann (approved):

syjlee (approved):

PR #1810, opened 2024-06-11
Enhancement/md pair gaussian value check

josephburkhart:
## Description

I have modified the `TypeParameterDict` for `md.pair.Gaussian`, `md.pair.ExpandedGaussian`, and `md.pair.LJGaussian` so that `sigma` must be a positive real number. Correspondingly, I have updated docstrings and pytests to reflect the new requirement.

## Motivation and context

This change ensures that users cannot accidentally or mistakenly set `sigma` equal to 0 for any of the Gaussian-type pair potentials, since doing so would cause divide-by-zero errors during simulation execution.

Resolves #1806

## How has this been tested?

I have added new invalid parameter dictionaries to the constructor function `_invalid_params()` in `hoomd/md/pytest/test_potential.py`.

I have built the documentation using Sphinx and confirmed that the changes I made to the docstrings display properly.

## Change log

```
Added a positive-number check to parameters for Gaussian-type pair potentials.
```

## Checklist:

- [x] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/CONTRIBUTING.rst).
- [x] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/ContributorAgreement.md).
- [x] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

PR #1817, opened 2024-06-13
Add expanded gaussian pair for hpmc

josephburkhart:
## Description

I have created a new expanded gaussian pair potential for HPMC. The implementation of this potential closely follows the implementation of the lennard-jones pair potential for HPMC, which has led to some minor code duplication that may warrant later refactoring. The mathematical definition of the new pair potential should be equivalent to that of the expanded gaussian in the MD module.  As part of this change, the following files were created:

- `hoomd/hpmc/PairPotentialExpandedGaussian.h` - templated off of `PairPotentialLennardJones.h`
- `hoomd/hpmc/PairPotentialExpandedGaussian.cc` - templated off of `PairPotentialLennardJones.cc`
- `hoomd/hpmc/pair/expanded_gaussian.py` - templated off of `lennard_jones.py`
- `hoomd/hpmc/pair/pytest/test_pair_expanded_gaussian.py` - templated off of `test_pair_lennard_jones.py`

## Motivation and context

Adding this pair potential will allow users to stochastically simulate systems with interactions that are defined by gaussian models, which previously has only been possible in MD. This may provide benefits, for example, in very large systems.

## How has this been tested?

- [x] HOOMD compiled locally without errors
- [x] ctest completed locally without errors
- [x] pytest completed locally without errors 
- [x] sphinx documentation built locally without errors and appears correct
- [x] an example simulation appeared visually correct (tested with both repulsive and attractive parameters)

## Change log
```
Added:

* ``hpmc.pair.ExpandedGaussian`` computes the expanded Gaussian pair potential in HPMC
  (`#1817 <https://github.com/glotzerlab/hoomd-blue/pull/1817>`__).
```

## Checklist:


josephburkhart:
pre-commit.ci autofix

josephburkhart:
pre-commit.ci autofix

josephburkhart:
In an offline conversation, @SchoeniPhlippsn and I were discussing some revisions to the pre-computed values in `PairPotentialExpandedGaussian.h`, namely, removing the lines that pre-compute `r_cut_squared` and `r_on_squared`, since I just take the square root of them in `PairPotentialExpandedGaussian.cc`. 

I pointed out that in `PairPotentialExpandedGaussian.cc`, there are some calculations that could actually be moved to the header file - many of the [lines ](https://github.com/glotzerlab/hoomd-blue/blob/65645aa54770ba0c5c062a6bd879dc2a96966dbb/hoomd/hpmc/PairPotentialExpandedGaussian.cc#L35-L53) that calculate shifting/smoothing functions when the mode is `xplore` or `shift` do not actually depend on the current value of `r` at all, so they could just be pre-computed in the header.

I left the code this way because it's how it's structured in the HPMC LJ Pair Potential, and I wanted to keep things consistent. But moving the computation to the header file would make things more efficient. What do others think?

joaander:
Yes, you can precompute additional constant terms and store them in the `ParamType` data structure. This adds memory per type pair, but saves a few cycles per interaction. On the CPU this is probably a performance win.

And this is a good observation on `r_cut_squared`. It may be best to keep this one consistent with LJ - if we move to a generic base class then they will need to be consistent. Using the pre-computed value avoids the need to compute `sqrt(r_cut_squared)` for every particle pair interaction when `mode == 'shift'`.

hoomd/hpmc/pytest/test_pair_expanded_gaussian.py
```diff
@@ -0,0 +1,222 @@
...
+    simulation.run(0)
+
+    assert expanded_gauss_1.energy == pytest.approx(expected=eg(
+        1.5, 1.0, 1.0, 1.0),
+                                                    rel=1e-5)
+    assert expanded_gauss_2.energy == pytest.approx(expected=eg(
+        1.5, 2.0, 1.0, 1.0),
+                                                    rel=1e-5)
+    assert simulation.operations.integrator.pair_energy == pytest.approx(
+        expected=eg(1.5, 1.0, 1.0, 1.0) + eg(1.5, 2.0, 1.0, 1.0), rel=1e-5)
```
tcmoore3: You could save the expected results into variables. That would help with readability, especially the last line, which becomes
```python
assert simulation.operations.integrator.pair_energy == pytest.approx(expected_1 + expected_2)
```

tcmoore3 (approved):
This looks good, nice work.

Regarding whether to precompute quantities or not, I have no strong opinions, so my vote would be to keep as consistent as possible with the other HPMC pair potentials.

joaander:
We can make an issue and plan to refactor the precomputing in a future PR.

On the GPU, memory is expensive and floating point operations are cheap. This is why the HOOMD MD code does not precompute these quantities. That behavior was copied into HPMC. However, these HPMC potentials are CPU only where the balance is different. I expect a small (but measurable) performance gain by precomputing these on the CPU (especially when sqrt is required).

This is another reason to wait a bit on refactoring common isotropic pair methods into a HPMC base class. We have not yet recognized all of the common elements (e.g. precomputation) that could be combined.

hoomd/hpmc/pytest/test_pair_expanded_gaussian.py
josephburkhart: Apologies, I just noticed this feedback. I've just implemented your suggested changes and pushed to this branch.

joaander (approved):
Thanks! Merge in `trunk-minor` and this is done.

hoomd/hpmc/pair/expanded_gaussian.py
```diff
@@ -0,0 +1,106 @@
...
+        if default_r_cut is None:
+            default_r_cut = float
+        else:
+            default_r_cut = float(default_r_cut)
+
+        params = hoomd.data.typeparam.TypeParameter(
+            'params', 'particle_types',
+            hoomd.data.parameterdicts.TypeParameterDict(
+                epsilon=float,
+                sigma=float,
```
joaander:
```suggestion
                sigma=positive_real,
```

hoomd/hpmc/pair/expanded_gaussian.py
```diff
@@ -0,0 +1,106 @@
...
+
+    simulation = hoomd.util.make_example_simulation()
+    sphere = hoomd.hpmc.integrate.Sphere()
+    sphere.shape['A'] = dict(diameter=0.0)
+    simulation.operations.integrator = sphere
+"""
+
+import hoomd
+
+from .pair import Pair
```
joaander, line 16:
```suggestion
from .pair import Pair
from hoomd.data.typeconverter import positive_real
```

joaander (approved):
One more small tweak: we should ensure that sigma is positive as in #1810.

PR #1822, opened 2024-06-19
Sphere geometry and basic rejection sampling framework for MPCD virtual particles

mphoward:
## Description

This PR adds support for generating MPCD virtual particles for a geometry using rejection sampling. Particles are drawn everywhere in space, but only the ones outside the geometry are kept. This can be inefficient but isn't usually a bottleneck for most simulations (main limitation is the extra memory). We plan to refine this approach in the future.

While working on this, I realized that the way the virtual particle fillers were getting seeded had some variation between classes and CPU vs. GPU. This PR fixes that.

To be able to test the filler, I needed a geometry that doesn't have its own filler already. I added a spherical confinement that we had available from another project. 

## Motivation and context

This type of sampling is required for some geometries where it's impossible to mathematically define the volume to sample virtual particles in (see #1737 for more discussion).

## How has this been tested?

I added a C++ test checking the statistics of drawn particles are roughly right. Python tests were also added for `mpcd.geometry.Sphere`, including for the filler.

## Change log

```
* `mpcd.geometry.Sphere` for spherical confinement.
```

## Checklist:


hoomd/mpcd/pytest/test_stream.py
```diff
@@ -517,3 +523,216 @@ def test_check_mpcd_particles(self, simulation_factory, snap):
...
+            np.testing.assert_array_almost_equal(
+                snap.mpcd.position[2],
+                0.99 * np.array([-1., -2., np.sqrt(11)]))
+            np.testing.assert_array_almost_equal(
+                snap.mpcd.velocity[2],
+                0.25 * np.array([-1., -2., np.sqrt(11)]))
+            np.testing.assert_array_almost_equal(snap.mpcd.position[3],
+                                                 [2., -2., -np.sqrt(8)])
+            np.testing.assert_array_almost_equal(snap.mpcd.velocity[3],
+                                                 [0.8, -0.4, -0.5])
```
tommy-waltmann: Can this set of checks be placed in its own function? It appears to be checking the same things many times throughout different tests, but with different values each time.

It may even be useful to do the repeated checks in a loop.

tommy-waltmann (changes requested):
I just have one suggestion for cleaning up the tests.

hoomd/mpcd/pytest/test_stream.py
mphoward: I think that `np.testing.assert_array_almost_equal` is more or less that function, since I would need to pass lists of positions and velocities. I can transpose the checks so that there are only two calls to `np.testing.assert_array_almost_equal` if you want to reduce the number of those calls, but this pattern of checking each particle individually is used in other spots in this test file. I think the original reason for writing it this way was that it was easier to figure out which particle was causing the error, likely when it was using `unittest` rather than `pytest`.

mphoward: I rearranged these units to reduce the number of assertions, so this should be addressed. For debugging these geometries, having explicit lists of coordinates that can be referred to on each step is helpful, so I don't think any further abstraction should be done.

tommy-waltmann (approved):
LGTM

Issue #1852, opened 2024-07-19
Walls don't work with external plugins 


astatt:
### Description

The [wall potential python class](https://github.com/glotzerlab/hoomd-blue/blob/bf35be69c9229745a0efed6f083afbe63a1d76c6/hoomd/md/external/wall.py#L174) isn't working with plugins because it is missing the string that let's you redefine where the module is implemented, see [pair python class](https://github.com/glotzerlab/hoomd-blue/blob/trunk-patch/hoomd/md/pair/pair.py) for reference. This is the same issue as documented in [Issue #1591](https://github.com/glotzerlab/hoomd-blue/issues/1591). 

### Script

```python
class WallPotential(force.Force):
    r"""Base class wall force.

    Warning:
        `WallPotential` should not be used directly.  It is a base class that
        provides features and documentation common to all standard wall
        potentials.
    """

    def __init__(self, walls):
        self._walls = None
        self.walls = hoomd.wall._WallsMetaList(walls, _to_md_cpp_wall)

    def _attach_hook(self):
        if isinstance(self._simulation.device, hoomd.device.CPU):
            cls = getattr(_md, self._cpp_class_name)
        else:
            cls = getattr(_md, self._cpp_class_name + "GPU")
        self._cpp_obj = cls(self._simulation.state._cpp_sys_def)
        self._walls._sync({
            hoomd.wall.Sphere:
                _ArrayViewWrapper(
                    _WallArrayViewFactory(self._cpp_obj, hoomd.wall.Sphere)),
            hoomd.wall.Cylinder:
                _ArrayViewWrapper(
                    _WallArrayViewFactory(self._cpp_obj, hoomd.wall.Cylinder)),
            hoomd.wall.Plane:
                _ArrayViewWrapper(
                    _WallArrayViewFactory(self._cpp_obj, hoomd.wall.Plane)),
        })


changed to 

class WallPotential(force.Force):
    r"""Base class wall force.

    Warning:
        `WallPotential` should not be used directly.  It is a base class that
        provides features and documentation common to all standard wall
        potentials.
    """

    # Module where the C++ class is defined. Reassign this when developing an
    # external plugin.
    _ext_module = _md

    def __init__(self, walls):
        self._walls = None
        self.walls = hoomd.wall._WallsMetaList(walls, _to_md_cpp_wall)

    def _attach_hook(self):
        if isinstance(self._simulation.device, hoomd.device.CPU):
[...]

PR #1853, opened 2024-07-19
Add _ext_module variable to base python classes

astatt:
## Description

Without the variable `_ext_module` some of the base python classes cannot be easily extended in external plugins, because it will not correctly find it, i.e when wanting to use or import "plugin_name.wall.PotentialName" it will instead look for "hoomd.wall.PotentialName". The simple inclusion of "_ext_module = _md" that can be overwritten by the plugin fixes this issue.

## Motivation and context

Needed for convenient development of external plugins. 
Resolves Issue #1852.

## How has this been tested?

This has been tested locally on the CPU. Because it only touches *.py files, no other tests are necessary. 

## Change log

```
Fix: External walls, external fields, and constrains can now be implemented via external plugins 
```

## Checklist:

### DIFF
--- a/hoomd/md/pair/aniso.py
+++ b/hoomd/md/pair/aniso.py
@@ -47,6 +47,10 @@ class AnisotropicPair(Pair):
         mode (`str`, optional) : the energy shifting mode, defaults to "none".
     """
 
+    # Module where the C++ class is defined. Reassign this when developing an
+    # external plugin.
+    _ext_module = _md
+
     _accepted_modes = ("none", "shift")
 
     def __init__(self, nlist, default_r_cut=None, mode="none"):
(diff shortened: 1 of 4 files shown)


hoomd/md/pair/aniso.py
```diff
@@ -47,6 +47,10 @@ class AnisotropicPair(Pair):
         mode (`str`, optional) : the energy shifting mode, defaults to "none".
     """
 
+    # Module where the C++ class is defined. Reassign this when developing an
+    # external plugin.
+    _ext_module = _md
```
mphoward: We have a plugin class that successfully inherits this class and overrides `_ext_module`:

https://github.com/mphowardlab/azplugins/blob/0c520e05afe59ef07ab3405fccaff4d90524fc45/src/pair.py#L371

I think this change is not needed because `_ext_module` is inherited from `Pair`.

astatt: Ah! Great. I'll remove it.

joaander (approved):
Thanks!

PR #1870, opened 2024-08-15
Enable dumping only part of the buffer for burst writer.

DomFijan:
## Description
Adds 3 options to dump function of burst writer: start, stop and clear buffer. Start and stop determine which subset of frames will be dumped, while clear buffer flag clears the buffer  or not. Default options should give same results as the current implementation.

## Motivation and context
This would be very useful to have for interactivity with dupin where often we just want to dump a  subset of the deque buffer of dump writer.

## How has this been tested?
Tests testing new functionality have been added.

## Change log

```
Changed:

* Add option to dump a subset of frames in burst writer
  (`#1870 <https://github.com/glotzerlab/hoomd-blue/pull/1870>`__)
```

## Checklist:

### DIFF
--- a/hoomd/GSDDequeWriter.cc
+++ b/hoomd/GSDDequeWriter.cc
@@ -42,7 +42,7 @@ GSDDequeWriter::GSDDequeWriter(std::shared_ptr<SystemDefinition> sysdef,
         else
             {
             analyze(timestep);
-            dump(0, 0, true);
+            dump(0, -1, true);
             }
         }
     }
@@ -69,6 +69,10 @@ void GSDDequeWriter::dump(long int start, long int end, bool empty_buffer)
     if (start < 0 || start >= buffer_length)
         {
         throw std::runtime_error("Burst.dump's start index is out of range.");
+        }
+    if (end < 0)
+        {
+        end = buffer_length;
         }
     for (auto i = end - 1; i >= start; --i)
         {
--- a/hoomd/GSDDequeWriter.cc
+++ b/hoomd/GSDDequeWriter.cc
@@ -14,11 +14,13 @@ GSDDequeWriter::GSDDequeWriter(std::shared_ptr<SystemDefinition> sysdef,
                                int queue_size,
                                std::string mode,
                                bool write_at_init,
+                               bool clear_whole_buffer_after_dump,
                                uint64_t timestep)
     : GSDDumpWriter(sysdef, trigger, fname, group, mode), m_queue_size(queue_size)
     {
     setLogWriter(logger);
     bool file_empty = true;
+    m_clear_whole_buffer_after_dump = true;
 #ifdef ENABLE_MPI
     if (m_sysdef->isDomainDecomposed())
         {
@@ -42,9 +44,10 @@ GSDDequeWriter::GSDDequeWriter(std::shared_ptr<SystemDefinition> sysdef,
         else
             {
             analyze(timestep);
-            dump(0, -1, true);
+            dump(0, -1);
             }
         }
+    m_clear_whole_buffer_after_dump = clear_whole_buffer_after_dump;
     }
 
 void GSDDequeWriter::analyze(uint64_t timestep)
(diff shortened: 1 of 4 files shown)


hoomd/md/pytest/test_burst_writer.py
```diff
@@ -242,3 +248,36 @@ def test_write_burst_log(sim, tmp_path):
...
+    if sim.device.communicator.rank == 0:
+        with gsd.hoomd.open(name=filename, mode='r') as traj:
+            assert len(traj) == 4
+
+    sim.run(4)
+    burst_writer.dump()
+    burst_writer.flush()
+    if sim.device.communicator.rank == 0:
+        with gsd.hoomd.open(name=filename, mode='r') as traj:
+            assert len(traj) == 6 if empty_buffer else 8
```
janbridley:
```suggestion
            assert len(traj) == (6 if empty_buffer else 8)
```

Prevents the unintended `assert 8`

janbridley (approved):

hoomd/write/gsd_burst.py
```diff
@@ -114,20 +114,28 @@ def _attach_hook(self):
...
-        """Write all currently stored frames to the file and empties the buffer.
+    def dump(self, start=0, end=-1, clear_entire_buffer=True):
+        """Write stored frames in range to the file and empties the buffer.
 
         This method alllows for custom writing of frames at user specified
         conditions.
 
+        Args:
+            start (int): The first frame to write. Defaults to 0.
+            end (int): The last frame to write, must be positive integer.
```
janbridley:
```suggestion
            end (int): The last frame to write.
```
Default is not a positive int

hoomd/GSDDequeWriter.cc
```diff
@@ -59,14 +59,42 @@ void GSDDequeWriter::analyze(uint64_t timestep)
...
+    if (end < 0)
+        {
+        iterator_end = buffer_length - start;
+        iterator_start = 0;
+        }
+    else
+        {
+        iterator_end = buffer_length - start;
+        iterator_start = buffer_length - end;
+        }
```
janbridley, line 86: So if `end < 0`, we write out the range `0, length-start`? Why is the correct range not `start, length+end`?

hoomd/GSDDequeWriter.cc
```diff
@@ -59,14 +59,42 @@ void GSDDequeWriter::analyze(uint64_t timestep)
...
+    for (auto i = iterator_end - 1; i >= iterator_start; --i)
         {
         write(m_frame_queue[i], m_log_queue[i]);
         }
```
janbridley, line 90: Does this not write the trajectory in reverse? I could be misunderstanding how write works

joaander: See line 54 which is `m_log_queue.push_front(getLogData());`. Why this is `push_front` I do not know. But frames are added at the front of the queue and removed from the end.

hoomd/GSDDequeWriter.h
```diff
@@ -31,7 +31,7 @@ class PYBIND11_EXPORT GSDDequeWriter : public GSDDumpWriter
 
     void analyze(uint64_t timestep) override;
 
-    void dump();
+    void dump(long int start, long int end, bool clear_entire_buffer);
```
joaander: Let's move `clear_entire_buffer` to a class level attribute (with get/set methods in C++ and a `param_dict` entry in Python). This way, users only need to set the parameter once and leave it.

I think the default could be either `True` or `False`. For backward compatibility, `dump()` with no arguments writes and clears all frames with either setting. We should set the default to be what we think is the most common use-case.

joaander (changes requested):

hoomd/GSDDequeWriter.cc
DomFijan: Anything beyond buffer_length would cause out of bounds accessing problems. 
The reason why this is weird is because the dequeue here is in reverse order. So to dump the frames in their time progression we have to do this weird index swapping stuff.

hoomd/GSDDequeWriter.cc
DomFijan: It does not. There are tests that specifically test if correct frames are dumped out. The reason why this works is because the dequeue is for some reason ordered in reverse (unsure why)...

janbridley (approved):

joaander (approved):
Thanks!

PR #1897, opened 2024-09-26
Fix segfaults for MPCD with large particle counts under MPI

### LINKED ISSUE #1895: Segmentation fault for MPI operations on large data sizes
### Description

MPCD segfaults for large system sizes:

* https://github.com/glotzerlab/hoomd-blue/issues/774#issuecomment-1979411890
* https://github.com/glotzerlab/hoomd-blue/discussions/1887

Our working guess is that this is caused by HOOMD's MPI helper methods like `scatter_v` and `gather_v` accidentally overflowing the MPI count (signed int) when data gets serialized to bytes.

I propose to register `MPI_Datatype`s for common data types we use these methods on that don't need to be serialized (like `Scalar3`). I would register these types in the `MPIConfiguration` and provide getters to access them. Then, callers can either invoke the MPI operations they want directly, or we can provide helper methods using these types (likely in the `MPIConfiguration` class as well).

We should also, at minimum, add a check in the MPI helper methods and throw an exception if the serialized data is expected to overflow a signed int.

The script below was user reported to cause segfaults on 48 CPUs.

### Script

```python
import hoomd
import numpy

device = hoomd.device.CPU()
simulation = hoomd.Simulation(device=device, seed=1)

snapshot = hoomd.Snapshot()
L = 128
density = 50
kT = 1

if snapshot.communicator.rank == 0:
    rng = numpy.random.default_rng(seed=42)
    snapshot.configuration.box = [L,L,L,0,0,0]
    snapshot.mpcd.types = ['A']
    snapshot.mpcd.N = int(density * L * L * L)
    snapshot.mpcd.position[:] = rng.uniform(low=-0.5*L,high=0.5*L,size=(snapshot.mpcd.N,3))

    velocity = rng.normal(0.0, numpy.sqrt(kT), (snapshot.mpcd.N, 3))
    velocity -= numpy.mean(velocity, axis=0)
    snapshot.mpcd.velocity[:] = velocity

simulation.create_state_from_snapshot(snapshot)

integrator = hoomd.mpcd.Integrator(dt=0.02)
integrator.collision_method = hoomd.mpcd.collide.StochasticRotationDynamics(
    period=1, angle=130, kT=kT
)

integrator.streaming_method = hoomd.mpcd.stream.Bulk(
    period=integrator.collision_method.period
)

integrator.mpcd_particle_sorter = hoomd.mpcd.tune.ParticleSorter(trigger=20)
simulation.operations.integrator = integrator

simulation.run(100)
device.notice(f'{simulation.tps}')
```

### Input files

_No response_

### Output

```shell
Segmentation fault
```

### Expected output

_No response_

### Platform

CPU, GPU, Linux

### Installation method

Compiled from source

### HOOMD-blue version

4.8.2

### Python version

3.12

mphoward:
## Description

This PR refactors parts of the MPCD code that relied on generic serialized MPI routines to instead use custom MPI datatypes. This increases the amount of data that can be sent because each object is larger. This is an attempt to fix segfaults reported when either initializing from or taking a snapshot with a large number of MPCD particles.

I replaced most uses of `bcast` with `MPI_Bcast` while I was at it since this is a relatively simple MPI call.

I also added exceptions when the serialized methods exceed the byte count that can be stored in a signed int.

## Motivation and context

Large numbers of particles seem to lead to overflow of the serialized MPI methods.

Resolves #1895

## How has this been tested?

Existings tests pass. I will confirm with reporting user (or test on a cluster myself) that this fixes the segfaults they were observing.

## Change log

```
* Fix segmentation faults initializing MPCD simulations with large numbers (~100+ million) of MPCD particles under MPI.
```

## Checklist:


mphoward:
Not stale—haven't had time to finish this off.

mphoward:
@joaander I will rebase this onto `trunk-major` and finish this week.

mphoward:
@joaander this should be ready to take a look at! I updated the regular particle communicator to use these new data structures, but I didn't touch the use of `bcast`, `scatter_v`, etc. outside MPCD. I did add code to throw an exception if the serialization buffer sizes will overflow an int.

hoomd/mpcd/ParticleData.cc
```diff
@@ -449,89 +463,118 @@ void mpcd::ParticleData::takeSnapshot(mpcd::ParticleDataSnapshot& snapshot,
...
+                    rank_displacements.data(),
+                    m_mpi_pdata_element,
+                    root,
+                    mpi_comm);
 
         if (rank == root)
             {
+            /*
+             * First sort the particles into tag order. This is probably a slow step, but could be
+             * sped up by sorting on each rank first then merging the sorted segments here.
```
mphoward, line 511: I can make this optimization in a subsequent PR, as well as packing the buffers on the GPU. I left it off for now in order to get this finalized.

joaander: OK, no problem. You might find the `GatherTagOrder` class helpful in implementing the optimization. I wrote it to solve a similar problem when writing GSD files and gathering forces.
https://github.com/glotzerlab/hoomd-blue/blob/f725ae3b796da9d1128c0388fcf36111bcdd781c/hoomd/HOOMDMPI.h#L511-L531

joaander (approved):
Thanks!

mphoward:
FYI: In running with one of the other branches that didn’t have this fix, I found that some MPI libraries internally check and report the error we were unintentionally triggering (count too large). This supports that this is the right fix.

hoomd/mpcd/ParticleData.cc
mphoward: Thanks! I remembered that you had written this and wanted to try it, but didn’t have time now. I will keep this mind when I come back to it later. Gathering these snapshots can be very slow.

joaander:
> FYI: In running with one of the other branches that didn’t have this fix, I found that some MPI libraries internally check and report the error we were unintentionally triggering (count too large). This supports that this is the right fix.

It is strange that the MPI API hasn't addressed this defect more aggressively. It has been 25 years since 64 bit processors have become commonplace.

mphoward:
It is indeed strange and frustrating. It should be addressed as part of the MPI-4 standard, but that will probably take a while to percolate through the implementations.

https://www.mpi-forum.org/docs/mpi-4.1/mpi41-report/node499.htm#Node499

PR #1955, opened 2024-11-26
Fix hpmc potential energy calculation

Charlottez112:
Fix for calculating potential energy correctly when there are multiple potentials in `mc.pair_potentials` with different `r_cut`


joaander (approved):
Thanks!

PR #1970, opened 2024-12-15
Wang frenkel

MartinGirard:
## Description

This draft PR would add the Wang-Frenkel potential to hoomd (equation (2) in Wang & Frenkel, https://pubs.rsc.org/en/content/articlelanding/2020/cp/c9cp05445f). 

## Motivation and context

This is largely motivated by coarse-grained IDP models, specifically by MPIPI (https://www.nature.com/articles/s43588-021-00155-3), which uses this type of potential.

## How has this been tested?

This is work in progress, unit tests will be added to hoomd. At the time of writing this draft PR, I would want some feedback on the implementation. Specifically, in eq (2) of Wang et al, the cutoff radius appears in the potential. I have chosen to change "r_c" there by an adjustable parameter.

As the potential requires calculation of arbitrary powers, which are integers, this has required addition of pow(float, int) in hoomdmath. 

## Checklist:


joaander:
Thanks. I'm busy right now but will review this when I get a chance. I would be helpful if you could address the failing CI checks.

MartinGirard:
> Thanks. I'm busy right now but will review this when I get a chance. I would be helpful if you could address the failing CI checks.

Failing test is the pre-commit.ci.

hoomd/HOOMDMath.h
```diff
@@ -349,6 +349,50 @@ inline HOSTDEVICE double pow(double x, double y)
     return ::exp(y * log(x));
     }
 
+//! Compute the pow of x,y in double precision when y is an integer using squaring
+inline HOSTDEVICE double pow(double x, unsigned int y){
```
joaander: Add a `hoomd/test/test_math.cc` file using UPP11 to unit test these new math functions.

hoomd/HOOMDMath.h
```diff
@@ -349,6 +349,50 @@ inline HOSTDEVICE double pow(double x, double y)
...
+        double result = 1.0;
+        for(;;){
+                if(y & 1)
+                    result *= x;
+                y >>= 1;
+                if(!y)
```
joaander: Even with all these branches, this method is faster than `::expf(y * logf(x))`? Did you benchmark on both the CPU and GPU? I would like to see performance metrics showing that the method is worth it.

hoomd/md/EvaluatorPairWangFrenkel.h
```diff
@@ -0,0 +1,213 @@
...
+#ifndef __HIPCC__
+#include <string>
+#endif
+
+#include "hoomd/HOOMDMath.h"
+
+/*! \file EvaluatorPairWangFrenkel.h
+    \brief Defines the pair evaluator class for Wang-Frenkel potentials
+    \details As the prototypical example of a MD pair potential, this also serves as the primary
+   documentation and base reference for the implementation of pair evaluators.
```
joaander:
```suggestion
```

hoomd/md/pair/pair.py
```diff
@@ -2020,3 +2020,78 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0.0, mode="none"):
...
+                 \left(\frac{2 \nu + 1}
+                 {2\nu \left(\left(\frac{R}{\sigma}\right)^{2\mu} - 1\right)}\right)^{2\nu + 1}
+
+    The potential was introduced in `Xipeng Wang et al. 2020`_.
+
+    .. _Xipeng Wang et al. 2020:
+       https://dx.doi.org/10.1039/c9cp05445f
+
+    Note:
+        In the original paper, the pair parameter R is explicitly equal to the cutoff.
```
joaander:
```suggestion
    Warning:
        Set `R` and `r_cut` to the same value so the potential goes to
        0 smoothly at the cutoff.
```

joaander (changes requested):
Thanks!

Regarding pre-commit: Yes -- you need to run `pre-commit`, commit the automatic changes, and make any manual adjustments needed to satisfy all the lints. See https://hoomd-blue.readthedocs.io/en/v5.0.1/style.html

Overall, this PR is very clean and complete. Aside from pre-commit, I have a few questions and requests around the new pow math functions. I will be happy to merge after these are addressed and the needed unit test is added.

hoomd/HOOMDMath.h
MartinGirard: I briefly tested the best-case scenario on an A100, which is of monodisperse values for the exponent (µ=v=2). The integer version produces 4100 TPS for integers (vs 4300 for 12-6 LJ), and 2600 with FP exponents. Other values of exponents I tried do not give measurable differences in performance. 

I will have to inspect the PTX in more details, but I'm assuming nvcc generated branchless code, eg `result = y&1 ? x : 1.0` and a select op. Cost would be proportional to the log2 of the largest exponent in the warp. Exponents are generally small, say 1 ~ 4. Largest value in the force-field of interest (MPIPI) is 11.

joaander: Thanks!

MartinGirard: I looked into the PTX, nvcc indeed yielded branchless code with FSEL for the the multiplication. The performance difference is even higher when compute bound; I tested on my old 2060, and got 225 TPS (int) versus 100 TPS (float).

MartinGirard:
Should be more or less set now. There is a last detail, if users input negative mu or nu, the cast to uint in the evaluator fails (the exponent should always be positive for this potential to make sense anyway). The user receives a cryptic message:

RuntimeError: Unable to cast Python instance of type <class 'int'> to C++ type '?' (#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details)

Python doesn't have built-in types for unsigned integers. I can check this on the C++ side, but that yields lengthy messages. Do we have a neater way on the python side?

joaander:
> Do we have a neater way on the python side?

Yes, use a validator. We have one for positive reals, but not positive ints so you will need to write one:
https://github.com/glotzerlab/hoomd-blue/blob/769ec5922f672a123c804d10d9e89208e9b89faa/hoomd/data/typeconverter.py#L87-L95

Here is an example of how it gets used:
https://github.com/glotzerlab/hoomd-blue/blob/769ec5922f672a123c804d10d9e89208e9b89faa/hoomd/md/pair/pair.py#L328-L332

Using these will give a somewhat better formatted message indicating which parameter keys caused the error message.

joaander (approved):
Thanks!

PR #1984, opened 2025-01-14
Cylinder-Wall Spheropolyhedron Support

seoulfood:
## Description

I added in spheropolyhedron support for cylindrical walls. I changed one user seen python file and the ExternalWall.h file.

## Motivation and context

It's useful to be able to scan through convex polyhedra, convex spheropolyhedra, and spheres within cylindrical walls.

## How has this been tested?

NEED TO ADD

Documentation builds and displays correctly.

## Checklist:

- [ x ] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/CONTRIBUTING.rst).
- [ x ] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/trunk-minor/ContributorAgreement.md).
- [ x ] My name is on the list of contributors (`sphinx-doc/credits.rst`) in the pull request source branch.

### DIFF
--- a/hoomd/hpmc/ExternalFieldWall.h
+++ b/hoomd/hpmc/ExternalFieldWall.h
@@ -481,6 +481,86 @@ inline bool test_confined(const CylinderWall& wall,
     return accept;
     }
 
+// Cylindrical Walls and Convex Spheropolyhedra
+DEVICE inline bool test_confined(const CylinderWall& wall,
+                                 const ShapeSpheropolyhedron& shape,
+                                 const vec3<Scalar>& position,
+                                 const vec3<Scalar>& box_origin,
+                                 const BoxDim& box)
+    {
+    bool accept = true;
+    Scalar3 t = vec_to_scalar3(position - box_origin);
+    t.x = t.x - wall.origin.x;
+    t.y = t.y - wall.origin.y;
+    t.z = t.z - wall.origin.z;
+    vec3<Scalar> shifted_pos(box.minImage(t));
+
+    vec3<Scalar> dist_vec
+        = cross(shifted_pos,
+                wall.orientation); // find the component of the shifted position that is
+                                   // perpendicular to the normalized orientation vector
+    Scalar max_dist = sqrt(dot(dist_vec, dist_vec));
+    if (wall.inside)
+        {
+        max_dist += (shape.getCircumsphereDiameter() / Scalar(2.0));        
+        }
+    else
+        {
+        max_dist -= (shape.getCircumsphereDiameter() / Scalar(2.0));
+        if (max_dist < 0)
+            {
+            max_dist = 0;
+            }
+
+        }
+
+    bool check_verts
+        = wall.inside ? (wall.rsq <= max_dist * max_dist)
+                      : (wall.rsq >= max_dist * max_dist); // condition to check vertices, dependent
+                                                           // on inside or outside container
+
+    if (check_verts)
+        {
+        if (shape.verts.N)
+            {
+            if (wall.inside)
+                {
+                for (unsigned int v = 0; v < shape.verts.N && accept; v++)
+                    {
+                    vec3<Scalar> pos(shape.verts.x[v], shape.verts.y[v], shape.verts.z[v]);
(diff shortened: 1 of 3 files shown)


hoomd/hpmc/ExternalFieldWall.h
```diff
@@ -481,6 +481,86 @@ inline bool test_confined(const CylinderWall& wall,
...
+    t.x = t.x - wall.origin.x;
+    t.y = t.y - wall.origin.y;
+    t.z = t.z - wall.origin.z;
+    vec3<Scalar> shifted_pos(box.minImage(t));
+
+    vec3<Scalar> dist_vec
+        = cross(shifted_pos,
+                wall.orientation); // find the component of the shifted position that is
+                                   // perpendicular to the normalized orientation vector
+    Scalar max_dist = sqrt(dot(dist_vec, dist_vec));
```
joaander, line 502: This value is only ever used squared.

Remove the sqrt and rename to max_dist_squared. Remove the multiplications that square it.

hoomd/hpmc/ExternalFieldWall.h
```diff
@@ -481,6 +481,86 @@ inline bool test_confined(const CylinderWall& wall,
...
+    Scalar3 t = vec_to_scalar3(position - box_origin);
+    t.x = t.x - wall.origin.x;
+    t.y = t.y - wall.origin.y;
+    t.z = t.z - wall.origin.z;
+    vec3<Scalar> shifted_pos(box.minImage(t));
+
+    vec3<Scalar> dist_vec
+        = cross(shifted_pos,
+                wall.orientation); // find the component of the shifted position that is
+                                   // perpendicular to the normalized orientation vector
```
joaander, line 501: Comments are easier to read at the same level as the code. You can also use English language rules when writing comments.

```suggestion
    // Find the component of the shifted position that is
    // perpendicular to the normalized orientation vector.
    vec3<Scalar> dist_vec
        = cross(shifted_pos,
                wall.orientation); 
                                   
```

hoomd/hpmc/ExternalFieldWall.h
```diff
@@ -481,6 +481,86 @@ inline bool test_confined(const CylinderWall& wall,
...
+    Scalar max_dist = sqrt(dot(dist_vec, dist_vec));
+    if (wall.inside)
```
joaander, line 503: A comment would be helpful to understand what the code is doing here.
```suggestion
    // Check if the shape's bounding sphere intersects the wall.
    if (wall.inside)
```

Also, consider rewriting this so that it evaluates the `<=` or `>` inside the if. The branch is repeated again with a ternary operator on the next line which is very confusing. Better yet, combine the content of the `if check_verts -> if shape.verts.N -> ` branches here as well. Bringing all the "inside" and "outside" code together will make it much easier to understand.

hoomd/hpmc/ExternalFieldWall.h
```diff
@@ -481,6 +481,86 @@ inline bool test_confined(const CylinderWall& wall,
...
+                unsigned int err = 0;
+                ShapeSpheropolyhedron wall_shape(q, *wall.verts);
+                ShapeSpheropolyhedron part_shape(shape.orientation, shape.verts);
+
+                accept = !test_overlap(shifted_pos, wall_shape, part_shape, err);
+                }
+            }
+        // Edge case; pure sphere. In this case, check_verts implies that the
+        // sphere will be outside. 
+        // Note from gabs: I don't fully understand this edge case
```
joaander, line 555: When `shape.verts.N` is zero, it is assume to have a single vertex at 0,0,0 - making it a sphere. In that case the bounding sphere check above is the only check that is necessary. I agree that this is really confusing. This case should be handled FIRST, not last. It would be more readable moved up above and handled with its own `return false` branch.

joaander (changes requested):
Thanks! I have many suggestions to improve the code.

ALSO, you need to add unit tests to check that the intersection test is working as intended. `test_external_wall.py` tests the walls.

PR #2022, opened 2025-03-13
Correctly account for neighbors in HPMC when updaters move particles.

joaander:
## Description

Specifically fix an issue where using `RemoveDrift` would cause overlapping hard particles. More generally, ensure that `IntegratorHPMCMono` properly accounts for all potential neighbors after updaters move particles outside of the integrator's control.

## Motivation and context

A user reported overlapping particles in a Frenkel-Ladd simulation that used `RemoveDrift`.

In serial: `IntegratorHPMCMono` moved particles and set the aabb invalid flag. If a logger then examined the number of overlaps, the AABB is rebuilt and the invalid flag cleared. An updater could then move particles and the following call to integrate would NOT rebuild the AABB tree - instead using the stale tree and therefore missing neighbors. 

In parallel: An updater might move a particle outside the local domain. `IntegratorHPMCMono` assumes that all particles are local. Any updater that moves particles needs to call `communicate` to ensure this and `RemoveDrift` did not do so.

## How has this been tested?

Cases that trigger this bug are extremely rare. I have tested these fixes with the minimal example script provided by the user. Short runs now behave correctly in serial and parallel. We will test this branch with full production runs before merging.

I will also run the hoomd-validation test suite before merging.

## Checklist:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,9 @@ Change Log
   (`#2020 <https://github.com/glotzerlab/hoomd-blue/pull/2020>`__).
 * MD integrators no longer integrate the z degree of freedom in 2D simulation boxes
   (`#2021 <https://github.com/glotzerlab/hoomd-blue/pull/2021>`__).
+* HPMC integrators no longer miss pairwise interactions when updaters, such as
+  ``RemoveDrift`` move particles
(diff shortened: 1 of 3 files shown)


joaander:
This branch passes hoomd-validation tests.

joaander:
Production Frenkel-Ladd simulations behave correctly with this branch.

Issue #2001, opened 2025-02-20
Support rigid bodies in MPCD


mzbush:
### Description

Rigid bodies are currently incompatible with MPCD collision methods because:

1. The collision method needs to read the velocities of constituent particles, but they are not currently set.
2. The collision method updates the velocity of constituent particles in a way that is inconsistent with rigid motion.

We should add support for rigid bodies to the collision methods.

### Proposed solution

1. ForceComposite should set the velocity of constituent particles based on the linear and angular velocity of the rigid body when it sets their positions and orientations.

    - HOOMD's angular momentum quaternion can be used to compute the angular velocity in the body frame.

    - The body's angular velocity can be used to get constituent particle velocities in the body frame relative to the center-of-mass translation.

    - These velocities can then be rotated to the space frame and have the center-of-mass translational velocity added.

    - Question: should this always be done, or should it be opt in?

2. The CollisionMethod should store the net change in space-frame linear and angular momentum for each rigid body as a result of collisions on constituent particles.

    - The changes can be computed by the base class taking the difference in velocity of constituent particles before/after collision or by recording the change during each collision. The first option is more generic, but the second option is probably faster because it has less read/write transactions.

    -  In MPI simulations, we will need to reduce these contributions onto the rank that owns the central particle, then broadcast the result.

    - These changes will be used to update the velocity and angular momentum of the central particle for the body, then reset the constituent velocities.

joaander:
Point 1 is briefly discussed in #278. Note that implementing it with MPI will require communicating the ghost angular momentum so that all ghost constituents have the needed information to compute their velocity locally. For consistency with the rest of HOOMD-blue, the constituent velocities should always be set. I think there is a flags mechanism to avoid communicating angular momentum when it is not needed.

mphoward:
This is a good point: DPD (#278) has a similar issue as MPCD, in that the momentum transferred to constituent particles needs to be passed to the central particle to enforce the rigid constraint. We should consider this when we design our solution for MPCD so that maybe the same code can also be used for DPD.

Our plan is to set the velocities of the constituent particles first (making sure it works CPU, GPU & MPI), then we'll figure out how to address the momentum transfer part.

joaander:
[...]

PR #2024, opened 2025-03-17
Set constituent particle velocities for rigid bodies

### LINKED ISSUE #278: DPD should work with rigid bodies
**[Original report](https://bitbucket.org/glotzer/hoomd-blue/issue/278) by Jens Glaser (Bitbucket: [jens_glaser](https://bitbucket.org/jens_glaser), GitHub: [jglaser](https://github.com/jglaser)).**

From my recollection, the DPD force compute operates on all particles
and computes friction forces. The integrator only operates on the central particles.
Therefore computed temperature values are wrong.

The underlying reason is likely the constituent particles have no velocities. If these
were computed from the angular velocity, it should work.

A user reported this and the reproducer script is attached.

mzbush:
## Description

The velocity of constituent particles in rigid bodies is set and updated when the constituent particle positions and orientations are updated. 

A second call to the MPI communicator to update ghost particles is done after the 2nd Verlet step to ensure the constituent particle velocities are updated correctly after the integration step. 

## Motivation and context

Resolves #278 and is a step required for #2001

## How has this been tested?

The pytests test that the velocity calculated is correct in the case of a constant set angular momentum for two rigid bodies. One where the inertia tensor is 1 on the diagonal and another where the inertia tensor is 0 in x and neither 0 nor 1 in the y and x directions. Angular momenta values are tested for each axial direction, and for a zero and nonzero linear velocity. A second test checks a constant torque propagated for one time step for both a zero and nonzero linear velocity.

## Checklist:


CHANGELOG.rst
```diff
@@ -4,6 +4,14 @@ Change Log
 5.x
 ---
 
+5.2.0 (not released yet)
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+*Added*
+
+* Set constituent particle velocities for rigid bodies
+  (`#1970 <https://github.com/glotzerlab/hoomd-blue/pull/2024>`__).
```
mphoward:
```suggestion
  (`#2024 <https://github.com/glotzerlab/hoomd-blue/pull/2024>`__).
```

mphoward (approved):
Thanks @mzbush! This looks good to me, other than the one small fix to the PR number in the changelog.

@joaander this is ready for review by you or others. I worked with Michaela on this PR in a fork, so I've been over the code a few times myself and I think her unittests have good coverage.

I wanted to record two inefficiencies @mzbush noted while working on this:

* Constituent positions and velocities are updated in the same method, so you can't update only the velocities if you know the positions haven't changed.
* Relatedly, an additional full ghost update is triggered after the second Verlet step, even though only certain fields may be updated during that step.

For generality, we decided not to optimize for these in case some future integration method might break the assumption that the second Verlet step only modifies velocities.

joaander:
Thanks for working this! I won't have a chance to review this in detail until after March 31.

Does the additional ghost communication always occur? Or only when rigid bodies are present?

mphoward:
> Thanks for working this! I won't have a chance to review this in detail until after March 31.

No problem, we will be working on the MPCD implementation of the rigid bodies in the meantime.

> Does the additional ghost communication always occur? Or only when rigid bodies are present?

Right now, yes, it always occurs. However, it would be easy to only trigger this update when rigid bodies are present if we want to assume an update is otherwise not needed!

joaander:
> > Does the additional ghost communication always occur? Or only when rigid bodies are present?
> 
> Right now, yes, it always occurs. However, it would be easy to only trigger this update when rigid bodies are present if we want to assume an update is otherwise not needed!

OK, I'll benchmark this branch as part of the review to see if the extra effort is needed.

mphoward:
The effort is minimal on our side (add `m_rigid_bodies &&` to the check on whether we have a domain decomposition before doing the extra ghost update), but a benchmark would still be helpful to decide if it's worth continuing to assume the second integration step never modifies any ghost properties.

mphoward:
Not stale

hoomd/Communicator.cc
```diff
@@ -2724,6 +2797,32 @@ void Communicator::beginUpdateGhosts(uint64_t timestep)
...
+                                                  access_location::host,
+                                                  access_mode::overwrite);
+            ArrayHandle<unsigned int> h_copy_ghosts(m_copy_ghosts[dir],
+                                                    access_location::host,
+                                                    access_mode::read);
+            ArrayHandle<unsigned int> h_rtag(m_pdata->getRTags(),
+                                             access_location::host,
+                                             access_mode::read);
+
+            // copy velocity of ghost particles
```
joaander:
```suggestion
            // copy angmom of ghost particles
```
Should this be angmom?

hoomd/Communicator.cc
```diff
@@ -2724,6 +2797,32 @@ void Communicator::beginUpdateGhosts(uint64_t timestep)
...
+            for (unsigned int ghost_idx = 0; ghost_idx < m_num_copy_ghosts[dir]; ghost_idx++)
+                {
+                unsigned int idx = h_rtag.data[h_copy_ghosts.data[ghost_idx]];
+
+                assert(idx < m_pdata->getN() + m_pdata->getNGhosts());
+
+                // copy velocity into send buffer
```
joaander:
```suggestion
                // copy angmom into send buffer
```
And this too?

hoomd/md/ForceComposite.cc
```diff
@@ -799,7 +804,7 @@ void ForceComposite::computeForces(uint64_t timestep)
                 h_torque.data[central_idx].z += delta_torque.z;
 
                 /* from previous rigid body implementation: Access Torque elements from a single
-                   particle. Right now I will am assuming that the particle and rigid body reference
+                   particle. Right now I will be assuming that the particle and rigid body reference
```
joaander: I think the whole "Right now..." sentence can be removed. The particle and body reference are the same. If this code was incorrect, then validation tests would fail.

joaander (commented):
Thanks!

I have a few minor suggestions. I will also check MPI benchmarks to see how the 2nd ghost communication impacts performance.

hoomd/md/ForceComposite.cc
```diff
@@ -630,6 +630,11 @@ CommFlags ForceComposite::getRequestedCommFlags(uint64_t timestep)
     // request orientations
     flags[comm_flag::orientation] = 1;
 
+    // request velocity, angular momentum, and inertia
+    flags[comm_flag::velocity] = 1;
+    flags[comm_flag::angmom] = 1;
+    flags[comm_flag::inertia] = 1;
```
joaander, line 636: I think these requests should be in PotentialPairDPDThermo. That is currently the only place that uses `comm_flag::velocity` that would be compatible with rigid bodies.

`ForceDistanceConstraint` also sets the velocity flag, but that isn't compatible with rigid bodies.

joaander (commented):
With 64,000 particles on 16 ranks this branch adds a 10% performance hit on the CPU (using hoomd-benchmarks):
```
                  5.1.1  mpcd-rigid-body
MDPairLJ     151.327200       142.909117
MDPairOPP     43.387061        43.345766
MDPairTable  130.033553       113.557941
MDPairWCA    632.891359       526.533988
```

With 16,000 particles on 16 ranks, the hit is higher - especially for the cheap to compute potentials like WCA:
```
                   5.1.1  mpcd-rigid-body
MDPairLJ      539.949773       482.350209
MDPairOPP     157.449643       158.363591
MDPairTable   475.856353       404.518839
MDPairWCA    1965.876524      1267.190448
```

I think it would be ideal to skip the second ghost communication when it isn't necessary. Comment that code clearly to explain what is going on.

hoomd/md/ForceComposite.cc
mphoward: If these flags are not set by `ForceComposite`, `updateRigidBodies` will need to check if this information is being communicated to decide if it can set the constituent velocities. This will lead to diverging behavior in serial and MPI simulations, since in serial `updateRigidBodies` will always set the constituent velocities correctly, while in MPI simulations it will depend on which other objects have been attached to the communicator.

If we wanted to implement this behavior, I would suggest a subscriber model for `ForceComposite` to request the constituent velocities be updated, then `ForceComposite` would only request this information if that was true. This seems like more work than it's worth, though.

joaander: Understood.

joaander:
The numbers are slightly different and more scattered than above (possibly because of a different load on the cluster node I got today), but back to back runs show there is no longer an obvious performance degradation.

64,000 particles:
```
                  5.1.1  mpcd-rigid-body-e0ff36e3
MDPairLJ     145.263652                145.642383
MDPairOPP     42.794153                 42.641170
MDPairTable  126.072872                116.429795
MDPairWCA    603.838532                605.912932
```

16,000 particles:
```
                   5.1.1  mpcd-rigid-body-e0ff36e3
MDPairLJ      493.554553                487.831982
MDPairOPP     147.933232                148.339457
MDPairTable   443.633694                428.346691
MDPairWCA    1713.408675               1926.993684
```

joaander (approved):
Thanks!

PR #2093, opened 2025-07-03
Mesh dynamic bonding

SchoeniPhlippsn:
## Description

Adds a mesh dynamical bonding updater that allows flipping edges within a mesh.

## Motivation and context

This addition allows modelling fluidized membranes. 

## How has this been tested?

added pytests

## Checklist:


hoomd/ForceCompute.h
```diff
@@ -75,6 +75,42 @@ class PYBIND11_EXPORT ForceCompute : public Compute
     //! Computes the forces
     virtual void compute(uint64_t timestep);
 
+    //! Computes the parameters for meshes
+    virtual void precomputeParameter() { };
```
joaander: Why are these methods in the base class `ForceCompute`? I think they should be defined in a subclass. If one doesn't exist, you could add a class that can be the parent of all mesh force computes.

hoomd/data/syncedlist.py
```diff
@@ -192,7 +192,10 @@ def _register_item(self, value):
             return
         else:
             if value._attached:
-                raise RuntimeError(f"Cannot place {value} into two simulations.")
+                if not value._two_attach:
+                    raise RuntimeError(f"Cannot place {value} into two simulations.")
+                else:
+                    value._two_attach = False
```
joaander: Why do we need this change?

I don't understand how an object can be attached to two simulations at once. They can only store one `m_sysdef`.

hoomd/md/BendingRigidityMeshForceCompute.h
```diff
@@ -89,6 +89,33 @@ class PYBIND11_EXPORT BendingRigidityMeshForceCompute : public ForceCompute
 
     //! Actually compute the forces
     virtual void computeForces(uint64_t timestep);
+
+    virtual Scalar calcEnergy(unsigned int idx_a,
+                              unsigned int idx_b,
+                              unsigned int idx_c,
+                              unsigned int idx_d,
+                              unsigned int type_id);
```
joaander, line 94:
```suggestion
    Scalar calcEnergy(unsigned int idx_a,
                              unsigned int idx_b,
                              unsigned int idx_c,
                              unsigned int idx_d,
                              unsigned int type_id) override;
```
... and repeated changes in similar examples.

In modern C++, classes that override methods should use the override keyword.

hoomd/md/MeshDynamicBondUpdater.cc
```diff
@@ -0,0 +1,411 @@
...
+
+        // Initialize the RNG
+        RandomGenerator rng(hoomd::Seed(RNGIdentifier::MeshDynamicBondUpdater, timestep, seed),
+                            hoomd::Counter(i));
+
+        // compute the random force
+        UniformDistribution<Scalar> uniform(0, Scalar(1));
+
+        Scalar rand_number = uniform(rng);
+        Scalar part_func = exp(-m_inv_T * energyDifference);
```
joaander: m_inv_kT perhaps?

Readers would expect T to be in Kelvin or some other temperature unit. Naming it kT makes it clear that this is a thermal energy.

hoomd/md/PotentialBond.h
```diff
@@ -66,6 +66,12 @@ template<class evaluator, class Bonds> class PotentialBond : public ForceCompute
 
     //! Actually compute the forces
     virtual void computeForces(uint64_t timestep);
+
+    virtual Scalar energyDiff(unsigned int idx_a,
```
joaander: Same as for `ForceCompute`. The mesh specific virtual methods should be in a mesh-specific base class.

joaander (commented):
Sorry for the delay. I have some questions about some of the changes and some suggestions for improvement.

hoomd/ForceCompute.h
```diff
@@ -75,6 +75,11 @@ class PYBIND11_EXPORT ForceCompute : public Compute
     //! Computes the forces
     virtual void compute(uint64_t timestep);
 
+    virtual bool checkSurrounding()
+        {
+        return false;
+        }
+
```
joaander: I think this method can be moved to `MeshForceCompute` as well. It is only used by `MeshDynamicBondUpdater` which operates on a list of `MeshForceCompute` objects.

hoomd/data/syncedlist.py
```diff
@@ -192,7 +192,10 @@ def _register_item(self, value):
             return
         else:
             if value._attached:
-                raise RuntimeError(f"Cannot place {value} into two simulations.")
+                if not value._allow_if_updater:
+                    raise RuntimeError(f"Cannot place {value} into list twice.")
+                else:
+                    value._allow_if_updater = False
```
joaander: I can see that a workaround is needed here because now a `MeshForceCompute` might be added to two synced lists: `integrator.forces` and `updater.forces`. I think we can avoid the workaround by fixing the original error condition. 

The message states "Cannot place {value} into two simulations", but the condition around it is not validating the two simulations part. `value._attach` in `if self._synced` above actually does do that. So it seems that this whole `else` block is not necessary. Try removing it (along with the other `_allow_if_updater` code you added) and see if that works.

hoomd/md/update.py
```diff
@@ -357,8 +361,122 @@ def _setattr_param(self, attr, value):
...
+
+    """
+
+    def __init__(self, trigger, mesh, kT, forces=[]):
+        # initialize base class
+        super().__init__(trigger)
+
+        for force in forces:
+            force._allow_if_updater = True
+
```
joaander:
```suggestion
```

This seems fragile. What if the user adds another force to `updater.forces` after initializing `updater`?

I don't think this workaround is necessary anyway. See my other comment where `_allow_if_updater` is used.

hoomd/operation.py
```diff
@@ -547,6 +547,9 @@ class see its documentation.
         + inspect.cleandoc(AutotunedObject._doc_inherited)
     )
 
+    def __init__(self):
+        self._allow_if_updater = False
+
```
joaander:
```suggestion
```
I don't think this is necessary. See my comment where `_allow_if_updater` is used.

joaander (changes requested):
Thanks! I have a few small suggestions for improvement. Let's finish this up and merge it soon.

joaander (approved):
Thanks!

Issue #2120, opened 2025-09-11
Implement Lees-Edwards boundary conditions for shear flow


wkdarko:
### Description

Simulations of bulk shear flow are currently not supported in HOOMD-Blue. Even though box deformations are supported, the necessary periodic boundary conditions to simulate the bulk behavior of shear rheology are not available. We would like to implement Lees-Edwards boundary conditions to support these types of simulations.

### Proposed solution

We will implement the Lagrangian-Rhomboid version of Lees-Edwards boundary conditions. The following steps will be taken:

- modify the simulation box object to include deformation of the lattice vectors
- modify the `wrap`, `minImage` and `shift` methods of `BoxDim` to account for velocity wrapping alongside positions, as particles cross periodic boundaries
- modify integration methods or equations of motion that use PBCs, to include particle velocities
- define a box deformation updater that will trigger at special points in the run loop to set the deformation rate on the box; this object will also have methods for resetting the tilt factors if necessary

mphoward:
We are working on this now and will have a PR implementing the BoxDim modifications ready soon. We have confirmed these changes do not significantly affect performance in benchmarks.

joaander:
@SchoeniPhlippsn has also worked on code for Lees-Edwards boundary conditions.

mphoward:
Great! We have completed the first 3 points above and should have a PR open today or Monday. We implemented a full deformation matrix a la LAMMPS `fix deform`, and I think we propagated the changes everywhere needed in the code but a careful review would be helpful.

@SchoeniPhlippsn do you have a branch you can share with any relevant code? It could be very helpful for finishing the last point to get the deformations set on the box.

@astatt also has a hacked implementation of Lees–Edwards BCs as well as a SLLOD integration method. We are planning to work with her to get the integrator moved into HOOMD once the boundary conditions are taken care of.

joaander:
How do your changes to `wrap` work with HPMC? Velocity is unused in most HPMC code paths.

mphoward:
We overloaded versions of `wrap`, `minImage`, and `shift` that take a velocity, then we reviewed all the code by grep for places they were called. We made no changes in HPMC because we assumed velocity would not be used, so it has all the old methods without velocity still. In MD and MPCD, we checked whether the calls we found needed to also wrap velocity one-by-one. For example, we use the new `minImage` with velocity for DPD potential, but we use the old `minImage` without velocity in the other pair potentials.

[...]

PR #2122, opened 2025-09-13
Lees-Edwards boundary conditions for shear flow

wkdarko:
## Description
This PR implements the Lees-Edwards boundary conditions for performing shear flow simulations.

## Motivation and context
The implementation will allow users to simulate bulk rheology under full periodic boundary conditions.

## How has this been tested?
Unit tests on box deformation methods were added.

## Checklist:


mphoward:
pre-commit.ci autofix

joaander:
Now that 5.4 is out, this is next on my list to review. I'll look at the proposed changes soon.

mphoward:
Thanks! I’ve also asked Kwabena to merge up trunk-minor to keep this current.

hoomd/Communicator.cc
```diff
@@ -2403,10 +2407,11 @@ void Communicator::exchangeGhosts()
             for (unsigned int idx = start_idx; idx < start_idx + m_num_recv_ghosts[dir]; idx++)
                 {
                 Scalar4& pos = h_pos.data[idx];
+                Scalar4& vel = h_vel.data[idx];
 
                 // wrap particles received across a global boundary
                 int3& img = h_image.data[idx];
-                shifted_box.wrap(pos, img);
+                shifted_box.wrap(pos, vel, img);
```
mphoward: Should this branch to the old vs. new method based on whether velocities are being communicated in the comm flags?

hoomd/Communicator.cc
```diff
@@ -2966,15 +2971,19 @@ void Communicator::beginUpdateGhosts(uint64_t timestep)
...
             const BoxDim shifted_box = getShiftedBox();
             for (unsigned int idx = start_idx; idx < start_idx + m_num_recv_ghosts[dir]; idx++)
                 {
                 Scalar4& pos = h_pos.data[idx];
+                Scalar4& vel = h_vel.data[idx];
 
                 // wrap particles received across a global boundary
                 int3 img = make_int3(0, 0, 0);
-                shifted_box.wrap(pos, img);
+                shifted_box.wrap(pos, vel, img);
```
mphoward: Same question here.

hoomd/CommunicatorGPU.cu
```diff
@@ -860,14 +862,24 @@ __global__ void gpu_pack_wrap_kernel(unsigned int n_out,
         }
 
     box.setPeriodic(periodic);
+    Scalar4 postype = d_postype[idx];
     int3 img = make_int3(0, 0, 0);
-    if (d_img)
+    if (out_img && d_img)
```
mphoward: I'm pretty sure the read arrays are always valid, not null, based on the assertions below, right?
```suggestion
    if (out_img)
```

hoomd/CommunicatorGPU.cu
```diff
@@ -860,14 +862,24 @@ __global__ void gpu_pack_wrap_kernel(unsigned int n_out,
...
+        {
         img = d_img[idx];
-    Scalar4 postype = d_postype[idx];
-    box.wrap(postype, img, wrap);
+        }
+    if (out_vel && d_vel)
```
mphoward:
```suggestion
    if (out_vel)
```

hoomd/CommunicatorGPU.cu
```diff
@@ -944,27 +965,15 @@ void gpu_exchange_ghosts_pack(unsigned int n_out,
                            n_out,
                            d_ghost_idx_adj,
                            d_pos,
-                           d_img,
+                           send_vel ? d_vel : 0,
+                           send_image ? d_img : 0,
```
mphoward: `d_img` can be left alone, just add `d_vel`
```suggestion
                           d_vel,
                           d_img,
```

mphoward (changes requested):
Thanks @wkdarko! This looks good to me based on our internal rounds of review. I had a couple small requests below related to the fixes you pushed to the communicators, then I think it should be good to go from my side.

joaander (commented):
Thanks! The code is clean and the changes are minimal.

I am concerned about the large number of `minImage`, `wrap`, and `shift` methods. Some return results, others modify references in place. To someone reading this code a year from now, it will not be clear what method is called. Could you add a suffix to the new methods to ameliorate this problem?

This PR is also missing a user interface to set the deformation rate and documentation of the feature in `features.rst`. We need to clearly document what classes in HOOMD-blue interoperate with the deformation rate and what ones do not. At this point, it is not clear to me whether that can be done in `features.rst` or if it needs to be mentioned in each class's documentation. Ideally, classes that don't support box deformations will raise an error when the deformation rate is non-zero.

Longer term (after this PR), you should add a tutorial to https://github.com/glotzerlab/hoomd-examples that shows users how to perform Lees-Edwards simulations.

mphoward:
> I am concerned about the large number of `minImage`, `wrap`, and `shift` methods. Some return results, others modify references in place. To someone reading this code a year from now, it will not be clear what method is called. Could you add a suffix to the new methods to ameliorate this problem?

The methods we added have the same argument / return types as the existing methods, so we thought overloading to include a velocity argument would be cleaner than appending a suffix. Would it be OK if we address this by expanding the doxygen documentation for the new methods? We can explain that they are used in the same way as the position-only methods but when the velocity also needs to be manipulated. Otherwise, we will need to go back through all the files in the diff manually to rename.

> This PR is also missing a user interface to set the deformation rate and documentation of the feature in `features.rst`. We need to clearly document what classes in HOOMD-blue interoperate with the deformation rate and what ones do not. At this point, it is not clear to me whether that can be done in `features.rst` or if it needs to be mentioned in each class's documentation. Ideally, classes that don't support box deformations will raise an error when the deformation rate is non-zero.

Yes, we are working on the user interface to do this in a separate PR! We were worried this one was getting too bulky. If you are OK with it, I think this PR could be merged as a temporarily undocumented feature because it doesn't change any behavior of existing code without an interface to set the deformation rates. (They are all zero by default.)

@wkdarko left a comment on #2120 to ask about how it would be best to implement the deformation, and we would appreciate your feedback there before we proceed too far!

Because of how we have modified the `BoxDim`, all the MD integration methods and pair forces should work with the deformation. I'm not sure if there are any other MD features that would break... the thermo compute will return unphysical values, but that is expected for NEMD. I know that some things in MPCD will break (and we will fix them), so we will error out and document that in the follow up PR until we have them fixed. We plan to do some thorough validation and comparison with LAMMPS once we have the user interface created.

> Longer term (after this PR), you should add a tutorial to https://github.com/glotzerlab/hoomd-examples that shows users how to perform Lees-Edwards simulations.

Yes, definitely will do!

hoomd/md/PotentialPairDPDThermo.h
```diff
@@ -184,7 +184,7 @@ template<class evaluator> void PotentialPairDPDThermo<evaluator>::computeForces(
             assert(typej < this->m_pdata->getNTypes());
 
             // apply periodic boundary conditions
-            dx = box.minImage(dx);
+            box.minImage(dx, dv);
```
joaander, line 187: @mphoward `minImage` is the method I was referring to that returns a value in two out of four overloads, but not in the other two. I understand that C++ lacks a natural syntax for multiple return value unpacking, so I don't have a better suggestion.

joaander:
> Yes, we are working on the user interface to do this in a separate PR! We were worried this one was getting too bulky. If you are OK with it, I think this PR could be merged as a temporarily undocumented feature because it doesn't change any behavior of existing code without an interface to set the deformation rates. (They are all zero by default.)

Yes, I would prefer everything in one PR. Reviewing the changes in batches is helpful.

> @wkdarko left a comment on #2120 to ask about how it would be best to implement the deformation, and we would appreciate your feedback there before we proceed too far!

Understood, I will move this discussion there.

hoomd/md/PotentialPairDPDThermo.h
mphoward: Oh, I see what you mean. If you have a strong preference toward something like `minImageWithVelocity(dx, dv)` we can do that.

joaander: No, I don't have a strong preference. Perhaps it is best to leave as is and expand the documentation as you suggest.

mphoward: I'll work with @wkdarko to expand this documentation to see if we can resolve it that way.

mphoward:
@wkdarko you'll need to update your fork and reopen this PR onto `trunk` per #2183.

joaander:
> @wkdarko you'll need to update your fork and reopen this PR onto `trunk` per #2183.

I recreated `trunk-minor` temporarily so that I could reopen this PR and change the base.

mphoward:
FYI: @wkdarko is iterating the box deformation implementation in a fork with me:

https://github.com/wkdarko/hoomd-blue/pull/7

I think it's getting close, so we will bring it into this PR soon

wkdarko:
Hello @mphoward and @joaander , the implementation of box deformation is ready for review on this PR. Could you please take a look when you have a chance? Thanks!

mphoward (approved):
Thank you @wkdarko!

@joaander I reviewed this PR incrementally with Kwabena as he worked on it. During initial modification of `BoxDim`, we also ran `hoomd-benchmarks` to check for regressions. In addition to the unit tests, we've also run validation tests against LAMMPS `fix deform`, and the results are comparable. Hence, I think this PR is ready to be merged once you approve!

joaander:
Thanks! I will review this pull request soon.

hoomd/md/LeesEdwardsBoxDeformer.cc
```diff
@@ -0,0 +1,174 @@
...
+                                      access_location::host,
+                                      access_mode::readwrite);
+
+            const unsigned int N = m_pdata->getN();
+            Scalar Ly = new_box.getL().y;
+
+            for (unsigned int i = 0; i < N; i++)
+                {
+                h_pos.data[i].x -= Scalar(flip) * Ly;
+                h_image.data[i].x -= flip;
```
joaander: Please explain why the signs are both negative here. How does this lead to continuity of motion in the unwrapped coordinates?

joaander (changes requested):
Thanks!

Please rename the module from `deformer` to `deform` for consistency with the other module names.

hoomd/md/LeesEdwardsBoxDeformer.cc
mphoward: Good catch, I think this is an error (and there may be more) that was not covered by a unit test. I talked with Kwabena, and he is going to rewrite this code so that the flipped box is calculated first, then flipped_box.shift() is used to move the particles one or more images back into the flipped box. He will also expand the unit tests to check the unwrapped coordinates.

joaander: Is it even possible to maintain continuity as the box vectors change?

mphoward: I think we can make the unwrapped coordinate be continuous using the procedure described above, which would be desirable for calculating things like MSD in shear, but the specific particle we are technically tracking probably changes on a flip because you pick up a different (identical) image.

hoomd/md/pytest/test_deform.py
```diff
@@ -0,0 +1,225 @@
...
+        sim = simulation_factory(snap)
+        deformer = md.deform.LeesEdwardsBoxDeformer(
+            shear_rate=shear_rate, max_tilt=max_xy
+        )
+        sim.operations.integrator = md.Integrator(dt=dt, deformer=deformer)
+
+        sim.run(1)
+
+        snap_after = sim.state.get_snapshot()
+        if snap_after.communicator.rank == 0:
```
mphoward, line 213: The math in these tests is complicated. Please rewrite to do two things:

1. Confirm that the _wrapped_ coordinates lie inside the flipped box.
2. Confirm that the _unwrapped_ coordinates do not change as a result of the flip.

I don't think HOOMD has a convenience method for unwrapping, but it is not too bad to write yourself: get the lattice vectors of the box, then add the image multiples of these to the wrapped coordinates.

mphoward (changes requested):
Please update the unit tests to more directly check the quantities we need to make sure are handled correctly during a flip. Otherwise, I think this addresses @joaander's comments!

hoomd/md/pytest/test_deform.py
wkdarko: Thanks @mphoward
The unit tests have been revised accordingly.

mphoward (approved):
Thank you @wkdarko!

@joaander we fixed the issue you spotted with the unwrapped coordinates by solving the system of equations that ensures they are unchanged by the flip. We also updated the unit tests so that it explicitly checks wrapped coordinates are in the flipped box but unwrapped coordinates do not change. I think this is good to go from my side!

joaander (approved):
Thanks!

PR #2136, opened 2025-09-30
Make MPCD collision methods compatible with rigid bodies

### LINKED ISSUE #2001: Support rigid bodies in MPCD

mzbush:
## Description

This PR implements a means of momentum transfer from constituent particles to the rigid body during the collision step. This method requires that the constituent particles be thermalized prior to the collision step, where all constituent particles of rigid bodies have a random velocity vector added to their current velocity such that the net gain of these random velocities to the linear and angular velocity of the rigid body is zero. This is to ensure that rigid bodies reach the set temperature in the MPCD fluid.

This implementation is currently incompatible with MPI and only works for 1 CPU or 1 GPU.

## Motivation and context

Resolves #2001

## How has this been tested?

Unit tests have been added to ensure momentum conservation and error checking.

## Checklist:


mphoward (approved):
Looks great, thank you!

@joaander we have done thorough physical validation of this approach (manuscript being finalized), and we get excellent agreement with state of the art. I would recommend to squash merge this one to due to the commit history.

mphoward:
@joaander could you resolve the conflict in the changelog entry based on your release plans? We have continued to use this branch and not found any problems, so we'd like to get it merged soon if possible.

joaander (approved):
Thanks! And sorry for the delay.

mphoward:
No problem! I fixed a typo in the CHANGELOG link I noticed.

PR #2184, opened 2025-11-24
Update for frictional contacts

khofmann6:
## Description

- I updated the Documentation to align with the style general style.  

- The calculation of the frictional force has been modified. It is now based exclusively on the z-component of the moment of inertia. This change introduces support for 2D simulations, which can be configured by using a moment of inertia of [0, 0, I]. For 3D simulations, the system requires isotropic spheres, specified by an identical moment of inertia in all dimensions (e.g., [I, I, I]).


joaander (approved):
Thanks!

PR #2197, opened 2025-12-09
Fix Inf error in table writing

josephburkhart:
## Description

Miniscule add-on to #2189 - addresses an additional failure case for Inf values.

## Motivation and context

See #2189.

## How has this been tested?

Same as #2189.

## Checklist:


josephburkhart:
Note that my solution [does not distinquish](https://docs.python.org/3/library/math.html#math.isinf) between + and - inf values.

joaander (approved):
Thanks!

PR #2206, opened 2025-12-18
Fix mpcd particle positions

mzbush:
## Description

Fix bug in mpcd pytest with rigid bodies in which the mpcd particles are not placed in the same cell as the constituent particles.

## Motivation and context

The mpcd pytests that check momentum conservation when colliding with rigid body constituent particles require that mpcd particles be placed in the same location as the constituent particles to ensure a collision occurs. When testing periodic boundaries by placing the rigid body in a corner of the simulation box, the mpcd particles were not placed in the right location.

## How has this been tested?

The positions in the test have been checked manually to ensure they are now in the right place.

## Checklist:

### DIFF
--- a/hoomd/mpcd/pytest/test_collide.py
+++ b/hoomd/mpcd/pytest/test_collide.py
@@ -225,7 +225,7 @@ class TestCollisionMethod:
             initial_snap.particles.velocity[:] = [velo_rigid]
             initial_snap.particles.angmom[:] = [angmom_rigid]
 
-            # place the mpcd particles on top of constituents
+            # place the mpcd particles on top of constituents, accounting for PBCs
             positions = np.add(def_rigid["positions"], pos_rigid)
             positions[positions < -11 * 0.5] = positions[positions < -11 * 0.5] + 11
             positions[positions > 11 * 0.5] = positions[positions > 11 * 0.5] - 11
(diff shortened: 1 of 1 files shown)


hoomd/mpcd/pytest/test_collide.py
```diff
@@ -216,7 +216,7 @@ def test_rigid_collide(
 
         # create simulation
         initial_snap = one_particle_snapshot_factory(
-            particle_types=["A", "B"], position=pos_rigid, L=21
+            particle_types=["A", "B"], position=pos_rigid, L=11
```
mphoward: Please store the value of L in a variable you can use here and also on line 230–231 so there are not "magic" numbers in the test.

hoomd/mpcd/pytest/test_collide.py
```diff
@@ -226,9 +226,12 @@ def test_rigid_collide(
             initial_snap.particles.angmom[:] = [angmom_rigid]
 
             # place the mpcd particles on top of constituents
```
mphoward: Please document this code change in this comment so it is clear why it is needed
```suggestion
            # place the mpcd particles on top of constituents, accounting for PBCs
```

mphoward (changes requested):
Thanks! This is a needed and straightforward fix. Please implement the minor changes below to make the test clearer in future.

mphoward (approved):
Looks good to me, thanks!

PR #2215, opened 2026-01-14
Logging correct Helfrich energy

SchoeniPhlippsn:
## Description

The Helfrich energy has been assigned to each vertex for each outgoing mesh bond, making it roughly 6 times larger. 

## Motivation and context

Calculating energy correctly for logging purposes. 

## How has this been tested?

Running simulations. 

## Checklist:


hoomd/md/HelfrichMeshForceCompute.cc
```diff
@@ -360,8 +360,8 @@ void HelfrichMeshForceCompute::computeForces(uint64_t timestep)
             h_force.data[idx_a].x += Fa.x;
             h_force.data[idx_a].y += Fa.y;
             h_force.data[idx_a].z += Fa.z;
-            h_force.data[idx_a].w += h_params.data[meshbond_type] * 0.5
-                                     * dot(sigma_dash_a, sigma_dash_a) * inv_sigma_a;
+            h_force.data[idx_a].w = h_params.data[meshbond_type] * 0.5
```
joaander, line 363: How does changing `+=` to `=` not introduce new problems? What if `idx_a` or `idx_b` participate in multiple mesh bonds? With `=`, only the energy due to the last bond evaluated will be set.

SchoeniPhlippsn: This is true. But `sigma_dash_a` and `inv_sigma_a` are both precalculated quantities. That means they will be the same across all bonds and I apply the same value each time.

joaander: Thanks for the explanation.

joaander (approved):

PR #2249, opened 2026-03-09
GSD Double Precision Parameter

### LINKED ISSUE #2194: Precision loss when writing GSD files
### Description

When using charge normalization for elementary_charges, floating point errors result in total charges that don't quite cancel out, something close to 1e-17. I think the primary cause for this is the charge factor, [0.0848385920 e](https://hoomd-blue.readthedocs.io/en/latest/units.html), which is pretty much unavoidable. 

This small of a rounding issue doesn't seem to cause a problem, unless the snapshot is written to a gsd file, and then loaded back into a simulation using `sim.create_state_from_gsd`. The total charge jumps from 1e-16 to 1e-7, and this can cause charge issues in the simulation. I especially see this when I run a system with >5000 molecules, which should be charge neutral, but end up with a net charge. 

I've tried using the [Decimal](https://docs.python.org/3/library/decimal.html) to improve the float precision to set to 64 bit precision during the initial conversion from the elementary_charge units to the dimensionless units, but saw no change in behavior.
```python
from decimal import Decimal, getcontext

getcontext().prec = 64  # Set desired precision
```

Maybe there should be a way to charge balance the hoomd state once something is loaded in? Even just shifting every charge by the factor of `total_charge/sim.state.N_particles` might counteract issues like this?

### Script

```python
import hoomd 
import gsd.hoomd
import numpy as np

print(f"Hoomd version: {hoomd.version.version}")
print(f"GSD version: {gsd.version.version}")

# NOTE: first charges fails as well, but has more particles so simplified to just 4 particles.
# charges = [-0.18,  -0.06,  -0.18,  0.51,  -0.43,  -0.33,  0.16,  0.06,  0.06,  0.06,  0.06,  0.06,  0.06,  0.06,  0.03,  0.03,  0.03]
charges = [-0.1, 0.1/3, 0.1/3, 0.1/3]
print(f"Initial float sum for charges with single molecule: {sum(charges)}")
charge_factor = 0.0848385920 # from https://hoomd-blue.readthedocs.io/en/latest/units.html
print(f"Initial normalized sum for charges with single molecule: {sum(charges)/charge_factor}")

snapshot = gsd.hoomd.Frame()
box = [10, 10, 10, 0, 0, 0] # Lx, Ly, Lz, xy, xz, yz
n_mols = 100
snapshot.particles.N = len(charges)*n_mols # replicates highlight charge issue
positions = (np.random.rand(snapshot.particles.N , 3) - 0.5) * 10
snapshot.particles.position = positions
all_charges = np.array(charges*n_mols) /charge_factor
print(f"Final charges plugged in: {sum(all_charges)}")
snapshot.particles.charge = all_charges
print(f"Ending total charge {sum(snapshot.particles.charge)}")

# NOTE: fill in meaningless data
# Define particle types
# Assign type 'A' (typeid 0) to all particles
snapshot.particles.typeid = [0] * snapshot.particles.N 

# save to gsd
device = hoomd.device.CPU()
sim = hoomd.Simulation(device=device, seed=1)
hoomd.write.GSD.write(
    state=sim.state, filename="checkcharge.gsd", mode="wb"
)
del sim

# reload gsd
device = hoomd.device.CPU()
sim = hoomd.Simulation(device=device, seed=1)
sim.create_state_from_gsd("checkcharge.gsd")
snap_loaded = sim.state.get_snapshot()
print(f"Reloaded charge: {sum(snap_loaded.particles.charge)}")
```

### Input files

_No response_

### Output

```shell
Hoomd version: 6.0.0
GSD version: 4.2.0
Initial float sum for charges with single molecule: -6.938893903907228e-18
Initial normalized sum for charges with single molecule: -8.178935718201485e-17
Final charges plugged in: -1.1102230246251565e-16
Ending total charge -1.1102230246251565e-16
Reloaded charge: -8.940696716308594e-06
```

### Expected output

```
Hoomd version: 6.0.0
GSD version: 4.2.0
Initial float sum for charges with single molecule: -6.938893903907228e-18
Initial normalized sum for charges with single molecule: -8.178935718201485e-17
Final charges plugged in: -1.1102230246251565e-16
Ending total charge -1.1102230246251565e-16
Reloaded charge: -1.1102230246251565e-16
```

### Platform

macOS, CPU

### Installation method

Conda-forge package

### HOOMD-blue version

6.0.0

### Python version

3.12.8

CalCraven:
## Description

This should allow for a passable parameter for utilizing gsd methods in hoomd, including writing, reading, and creating the state, to leverage the ability for gsd's stored in single or double precision.
See this [PR for details](https://github.com/glotzerlab/gsd/pull/495)

## Motivation and context

Required for highly precise tracking of structures, including tightly packed boxes, large numbers of decimals in charge that must be close to neutrality, etc.

Resolves #2194

## How has this been tested?

Tested by writing charges as double and single and testing the dtype

## Checklist:


CalCraven:
Apologies @joaander for letting this sit so long. Not sure this is the greatest solution but tried to just get something out for review.

hoomd/write/gsd_burst.py
joaander, line 46:
```suggestion
            to ``True``.
        precision (str): Write precision for floating-point data. One of
        ``'single'`` or ``'double'``. Defaults to ``'single'``.
```

hoomd/GSDDumpWriter.cc
```diff
@@ -272,7 +277,6 @@ void GSDDumpWriter::initFileIO()
             ostringstream o;
             o << "HOOMD-blue " << HOOMD_VERSION;
```
joaander, line 281:
```suggestion
        m_exec_conf->msg->notice(3) << "GSD: create or overwrite gsd file " << m_fname << endl;
```
Let's keep the notice messages unchanged.

hoomd/GSDDumpWriter.cc
```diff
@@ -294,7 +298,7 @@ void GSDDumpWriter::initFileIO()
             populateNonDefault();
 
             // open the file in append mode
-            m_exec_conf->msg->notice(3) << "GSD: open gsd file " << m_fname << endl;
+            m_exec_conf->msg->notice(1) << "GSD: open gsd file " << m_fname << endl;
```
joaander:
```suggestion
            m_exec_conf->msg->notice(3) << "GSD: open gsd file " << m_fname << endl;
```

hoomd/GSDDumpWriter.cc
```diff
@@ -403,6 +407,9 @@ void GSDDumpWriter::write(GSDDumpWriter::GSDFrame& frame, pybind11::dict log_dat
         }
     else
 #endif
+    {
+    writeLogQuantities(log_data);
+    }
```
joaander:
```suggestion
```

hoomd/GSDDumpWriter.h
```diff
@@ -259,6 +275,24 @@ class PYBIND11_EXPORT GSDDumpWriter : public Analyzer
...
+    std::vector<double> m_position_double;
+    std::vector<float> m_position_float;
+    std::vector<double> m_orientation_double;
+    std::vector<float> m_orientation_float;
+    std::vector<double> m_velocity_double;
+    std::vector<float> m_velocity_float;
+    std::vector<double> m_angmom_double;
+    std::vector<float> m_angmom_float;
+    std::vector<double> m_diameter_double;
+    std::vector<float> m_diameter_float;
```
joaander:
```suggestion
    /// Temporary buffers for single precision writes
    std::vector<float> m_float_buffer;
```
We can reuse the same buffer for all data chunks.

hoomd/GSDDumpWriter.h
```diff
@@ -271,6 +305,21 @@ class PYBIND11_EXPORT GSDDumpWriter : public Analyzer
...
+                               std::vector<double>& double_buffer,
+                               std::vector<float>& float_buffer);
+
+    /// Helper to write float/double chunks to vec3 with precision handling
+    void writeVec3FloatDoubleChunk(const std::string& name,
+                                   const std::vector<vec3<double>>& data,
+                                   uint32_t N,
+                                   std::vector<double>& double_buffer,
+                                   std::vector<float>& float_buffer);
+
```
joaander, line 303:
```suggestion
    /// Helper to write float/double chunks with precision handling
    void writeFloatDoubleChunk(const std::string& name,
                               const std::vector<double>& data,
                               uint32_t N,
                               uint32_t M);

    /// Helper to write float/double chunks to vec3 with precision handling
    void writeVec3FloatDoubleChunk(const std::string& name,
                                   const std::vector<vec3<double>>& data,
                                   uint32_t N);

```
No need to pass buffer as an argument, these methods can use the `m_float_buffer` and `m_double_buffer` member variables.

hoomd/GSDDumpWriter.cc
```diff
@@ -1270,7 +1454,7 @@ void GSDDumpWriter::populateLocalFrame(GSDDumpWriter::GSDFrame& frame, uint64_t
 
         for (unsigned int index : m_index)
             {
-            float diameter = static_cast<float>(h_diameter.data[index]);
+            double diameter = static_cast<double>(h_diameter.data[index]);
 
             if (diameter != 1.0f)
```
joaander:
```suggestion
            if (diameter != Scalar(1.0))
```

hoomd/GSDDumpWriter.cc
```diff
@@ -1333,7 +1517,7 @@ void GSDDumpWriter::populateLocalFrame(GSDDumpWriter::GSDFrame& frame, uint64_t
 
         for (unsigned int index : m_index)
             {
-            quat<float> angmom = quat<float>(h_angmom.data[index]);
+            quat<double> angmom = quat<double>(h_angmom.data[index]);
 
             if (angmom.s != 0.0f || angmom.v.x != 0.0f || angmom.v.y != 0.0f || angmom.v.z != 0.0f)
```
joaander:
```suggestion
            if (angmom.s != Scalar(0.0) || angmom.v.x != Scalar(0.0) || angmom.v.y != Scalar(0.0) || angmom.v.z != Scalar(0.0))
```

hoomd/GSDDumpWriter.cc
```diff
@@ -519,6 +547,95 @@ void GSDDumpWriter::writeFrameHeader(const GSDDumpWriter::GSDFrame& frame)
         }
     }
 
+/// Helper to write float/double chunks with precision handling
+void GSDDumpWriter::writeFloatDoubleChunk(const std::string& name,
+    const std::vector<double>& data,
+    uint32_t N,
+    uint32_t M,  // number of components (1 for scalar, 3 for vec3, etc)
+    std::vector<double>& double_buffer,
+    std::vector<float>& float_buffer)
```
joaander:
```suggestion
    uint32_t M)
```

hoomd/GSDDumpWriter.cc
```diff
@@ -519,6 +547,95 @@ void GSDDumpWriter::writeFrameHeader(const GSDDumpWriter::GSDFrame& frame)
...
+    {
+        double_buffer[i] = data[i];
+    }
+    retval = gsd_write_chunk(&m_handle,
+            name.c_str(),
+            GSD_TYPE_DOUBLE,
+            N,
+            M,
+            0,
+            (void*)double_buffer.data());
```
joaander:
```suggestion
    retval = gsd_write_chunk(&m_handle,
            name.c_str(),
            GSD_TYPE_DOUBLE,
            N,
            M,
            0,
            (void*)data.data());
```
No need to copy the double precision data before writing.

hoomd/GSDDumpWriter.cc
```diff
@@ -519,6 +547,95 @@ void GSDDumpWriter::writeFrameHeader(const GSDDumpWriter::GSDFrame& frame)
...
+    double_buffer[i * 3 + 1] = data[i].y;
+    double_buffer[i * 3 + 2] = data[i].z;
+    }
+    retval = gsd_write_chunk(&m_handle,
+        name.c_str(),
+        GSD_TYPE_DOUBLE,
+        N,
+        3,
+        0,
+        (void*)double_buffer.data());
```
joaander:
```suggestion
    uint32_t N)
    {
    int retval;
    if (m_precision == "double")
    {
    retval = gsd_write_chunk(&m_handle,
        name.c_str(),
        GSD_TYPE_DOUBLE,
        N,
        3,
        0,
        (void*)data.data());
```
No need to copy the data before writing.

hoomd/GSDDumpWriter.cc
```diff
@@ -519,6 +547,95 @@ void GSDDumpWriter::writeFrameHeader(const GSDDumpWriter::GSDFrame& frame)
...
+    {
+    float_buffer[i] = static_cast<float>(data[i]);
+    }
+    retval = gsd_write_chunk(&m_handle,
+            name.c_str(),
+            GSD_TYPE_FLOAT,
+            N,
+            M,
+            0,
+            (void*)float_buffer.data());
```
joaander:
```suggestion
    m_float_buffer[i] = static_cast<float>(data[i]);
    }
    retval = gsd_write_chunk(&m_handle,
            name.c_str(),
            GSD_TYPE_FLOAT,
            N,
            M,
            0,
            (void*)m_float_buffer.data());
```

hoomd/GSDDumpWriter.cc
```diff
@@ -519,6 +547,95 @@ void GSDDumpWriter::writeFrameHeader(const GSDDumpWriter::GSDFrame& frame)
...
+    float_buffer[i * 3 + 1] = static_cast<float>(data[i].y);
+    float_buffer[i * 3 + 2] = static_cast<float>(data[i].z);
+    }
+    retval = gsd_write_chunk(&m_handle,
+        name.c_str(),
+        GSD_TYPE_FLOAT,
+        N,
+        3,
+        0,
+        (void*)float_buffer.data());
```
joaander:
```suggestion
    m_float_buffer.resize(data.size() * 3);
    for (size_t i = 0; i < data.size(); i++)
    {
    float_buffer[i * 3 + 0] = static_cast<float>(data[i].x);
    float_buffer[i * 3 + 1] = static_cast<float>(data[i].y);
    float_buffer[i * 3 + 2] = static_cast<float>(data[i].z);
    }
    retval = gsd_write_chunk(&m_handle,
        name.c_str(),
        GSD_TYPE_FLOAT,
        N,
        3,
        0,
        (void*)m_float_buffer.data());
```

hoomd/GSDDumpWriter.cc
```diff
@@ -532,7 +649,7 @@ void GSDDumpWriter::writeAttributes(const GSDDumpWriter::GSDFrame& frame)
         writeTypeMapping("particles/types", frame.particle_data.type_mapping);
         }
 
-    if (frame.particle_data.type.size() != 0)
+    if (m_dynamic[gsd_flag::particles_types] || frame.particle_data.type.size() != 0)
```
joaander:
```suggestion
    if (frame.particle_data.type.size() != 0)
```
This was actually correct the way it was. To boost efficiency in MPI simulations, the typeid data is not even collected unless it is dynamic (see `gatherGlobalFrame`).

hoomd/GSDDumpWriter.cc
```diff
@@ -1040,7 +1225,6 @@ void GSDDumpWriter::populateNonDefault()
     int retval;
 
     // open the file in read only mode
```
joaander, line 1153:
```suggestion
    // open the file in read only mode
    m_exec_conf->msg->notice(3) << "GSD: check frame 0 in gsd file " << m_fname << endl;
```

hoomd/GSDReader.cc
joaander, line 1: The user doesn't choose the precision in `GSDReader`. It needs to check the data type on each chunk in the file before it reads. I would suggest making the snapshot double precision. You can read directly into the snapshot if the chunk is double. If not, read into a float buffer and then cast up to double when copying into the snapshot.

joaander (changes requested):
Thanks! No worries on the time, these things always take a long time to get right. I've made some comments to simplify the code in `GSDDumpWriter`. `GSDReader` needs some more work so it can handle files where the user doesn't know the precision of the file or the file may have some single and some double precision chunks.

hoomd/write/gsd_burst.py
CalCraven: Addressed in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Yeah, had a bunch of debug messages of my own in here, so these got removed alongside those messages unwittingly.
Fixed in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Fixed in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Removed in 00e5df497

hoomd/GSDDumpWriter.h
CalCraven: Yeah that's way more nasty. 00e5df497

hoomd/GSDDumpWriter.h
CalCraven: Removed in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Updated in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Updated in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Removed in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: yeah that was dumb.
00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Fixed in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Fixed in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Fixed in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Noted, yeah I was just playing around with some errors I didn't quite understand. Thanks for the explanation!
Reverted in 00e5df497

hoomd/GSDDumpWriter.cc
CalCraven: Fixed in 00e5df497

hoomd/GSDReader.cc
CalCraven: I pushed b8eab2eda, which should drop that argument for the GSDReader, so now it should hopefully be identified for all chunks. My only question is, do we want everything to be read to the snapshot as a double? Will that cause any memory issues?

joaander: The data needed to represent a snapshot is a small fraction of the total memory used by HOOMD-blue.

joaander (commented):
Thanks!

I pushed revisions directly to the branch to improve the robustness of the reader and test the reader with double precision files. CI tests will fail until I release the new version of GSD.

CalCraven:
Sounds good. Following [this initial action plan](https://github.com/glotzerlab/hoomd-blue/issues/2194#issuecomment-3617996251), I've opened an issue with Ovito.
[stuko/ovito#237](https://gitlab.com/stuko/ovito/-/work_items/237)

joaander:
> Sounds good. Following [this initial action plan](https://github.com/glotzerlab/hoomd-blue/issues/2194#issuecomment-3617996251), I've opened an issue with Ovito. [stuko/ovito#237](https://gitlab.com/stuko/ovito/-/work_items/237)

That is great, thanks!

joaander (approved):
Thanks!

Issue #2013, opened 2025-03-03
Refactor MPCD cell property calculations and communication


mphoward:
### Description

This will require multiple steps to implement. I think the sequence of tasks will be:

- [ ] Remove cell list data structures.
- [ ] Calculate cell properties directly from particles not cell list.
- [ ] Realign MPCD domain decomposition to snap with edges of cells.
- [ ] Add forward and backward communication of particles to `CollisionMethod`.
- [ ] Set a configurable buffer width for communicating MPCD particles, rather than use coverage box.
- [ ] Sort particles using Hilbert or Z-order curve rather than cell list. Make a standard updater.
- [ ] Consider removing / refactoring cell list to live in collision method.
- [ ] Benchmark with both single and double particle data, and consider using a mixed-precision data model for MPCD particles.

### Motivation and context

The supplementary material of https://doi.org/10.1016/j.cpc.2024.109494 (a similar benchmark / issue as discussed in #1887) shows that HOOMD underperforms their implementation on both CPUs and GPUs. On the CPU, HOOMD has better strong scaling but lower performance for most core counts. On the GPU, there is also a performance difference, but the strong scaling is now abysmal. Based on the implementation presented in Section 3, I think the main likely differences are:

- Mixed precision particle data (they use float positions & double velocities). I reproduced their benchmark numbers using a fully double build of HOOMD. The impact of this could be assessed by rerunning with MPCD mixed precision.
- Their cell property calculation is done using atomic operations, which bypasses the cell list. This should lead to fewer memory transactions, even if they are more scattered, and MPCD is typically quite memory bound.
- They communicate MPCD particles to the rank owning the collision cell, then backward communicate the effect of the collision. This pattern may have less communication overhead than HOOMD's strategy to communicate overlapping cells because it is point-to-point. It also means that particles can drift farther before migration.

There are other benefits to refactoring as proposed above:

- Lower memory demands, especially for large systems or those with boundaries.
- Easier to calculate more advanced cell statistics because particles are wrapped when migrated (#1883).
- Should fix round-off errors with domain decomposition for skewed boxes (#2009).

mphoward:
@joaander we are going to have a large number of code changes coming in that are needed to address this issue. These changes are going to help us do some other things in future too:

* Angular momentum conservation (#1883)
* Lees-Edwards BCs for MPCD (#2120)
* Future developments in cell list scheme for MPCD and use of mixed precision

[...]

PR #2335, opened 2026-09-11
Refactor MPCD cell list

mzbush:
## Description

Revamps how the CellList for mpcd is calculated so that the properties are computed while particles are being binned into cells so that a structure that stores a list of particles per cell is no longer needed. Removes CellThermoCompute and does property calculation in the CellList

## Motivation and context

Reduces memory usage during the collision step and reduces memory transactions. Is a prerequisite step to improving parallel scaling and adding additional cell statistics such as those needed for angular momentum conversion.

Is a step toward resolving #2013

## How has this been tested?

New unit tests have been created to ensure cell property calculations are correct

## Checklist:

- [X] I have reviewed the [**Contributor Guidelines**](https://github.com/glotzerlab/hoomd-blue/blob/trunk/CONTRIBUTING.rst).
- [X] I agree with the terms of the [**HOOMD-blue Contributor Agreement**](https://github.com/glotzerlab/hoomd-blue/blob/trunk/ContributorAgreement.md).


hoomd/mpcd/CellList.h
mphoward:
```suggestion
    void compute(uint64_t timestep) override;
```

mphoward (approved):
The serial unit tests all pass except for an error related to the Andersen thermostat, which is expected. I am merging this into the development branch now so that we can move onto adding MPI support.

